PR #4, opened 2018-03-08
use efi instead of linux in acrn.conf

miguelinux:
acrn.efi is an EFI executable image and not a linux kernel image.
This commit changes linux to efi in the boot-loader configuration.

For more reference please review:
https://www.freedesktop.org/wiki/Software/systemd/systemd-boot/

Signed-off-by: Miguel Bernal Marin <miguel.bernal.marin@linux.intel.com>


miguelinux:
@jren1 I updated this PR to match current master branch

PR #8, opened 2018-03-13
apicv: cancel event injection if vcpu is scheduled out

fyin1:
And re-inject the event after vcpu is scheduled in.

### DIFF
--- a/common/schedule.c
+++ b/common/schedule.c
@@ -147,6 +147,9 @@ static void context_switch_out(struct vcpu *vcpu)
 	if (vcpu == NULL)
 		return;
 
+	/* cancel event(int, gp, nmi and exception) injection */
+	cancel_event_injection(vcpu);
+
 	atomic_store_rel_32(&vcpu->running, 0);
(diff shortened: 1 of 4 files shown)


common/schedule.c
```diff
@@ -147,6 +147,9 @@ static void context_switch_out(struct vcpu *vcpu)
 	if (vcpu == NULL)
 		return;
 
+	/* cancel event(int, gp, nmi and exception) injection */
```
yakuizhao: Maybe the vcpu in context_switch_out is not the current vmcs structure. In such case the exec_vmwrite won't work as expected.
It will be better that the exec_vmptrld is called for the safety.

fyin1: Hi Yakui,
Thanks for the comments.
Now, we only have CPU partition feature (no CPU sharing yet). So from performance perspective, we tries best to avoid loading VMCS. There is comments in line 168 explaining it and what should be done if we want to support CPU sharing.

lijinxia:
verified+1

PR #13, opened 2018-03-13
add data structure in vm for enable trusty feature

mingqiangchi:
add key info structure
add sworld_eptp in vm structure, and rename ept->nworld_eptp
add secure world control structure

Change-Id:
Tracked-On:220921


lijinxia:
verified +1

PR #15, opened 2018-03-14
align the coding style when access VMEXIT_QUAL

yonghuah:
- align coding style when operating on reading
   VMX_EXIT_QUALIFICATION
 - avoid redunant operatons when read the
   value of VMEXIT_QUAL.


yakuizhao:
It will be better that the commit can be split into two commits.(One thing in one patch)
One is to avoid the redundant operation
The second is to align the coding style.

yonghuah:
duplicated.

PR #18, opened 2018-03-14
vtd: fix coherency issue of vtd tables

binbinwu1:
flush cacheline after modifying vtd table entries


binbinwu1:
Will fix the issue by setting the attribute of tables to write through, so close the pull request.

PR #19, opened 2018-03-14
Remove ASSERT in lib functions

lyan3:
Replace ASSERT in lib functions with error message print and return a
value indicating error to allow the caller of lib functions to handle
the error.

### DIFF
--- a/lib/mem_mgt.c
+++ b/lib/mem_mgt.c
@@ -266,7 +266,8 @@ void *malloc(unsigned int num_bytes)
 	}
 
 	/* Check if memory allocation is successful */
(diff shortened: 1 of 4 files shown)


yakuizhao:
It seems that the ASSERT can help to dump some infos if HV_DEBUG is enabled. This is helpful.
But it will cause that the system enters the state of "pause" if the HV_DEBUG is not enabled.

In fact the condition check in ASSERT is not expected.
Is it ok to rework the implementation of assert when the HV_DEBUG is not enabled?  Maybe like what you mentioned, it can be checked by caller instead of pause.

lyan3:
it's not the latest patch set.

PR #20, opened 2018-03-14
Remove ASSERT in lib functions

lyan3:
Replace ASSERT in lib functions with error message print and return a
value indicating error to allow the caller of lib functions to handle
the error.

Change-Id: If166484238dc0734041adfdbb19a5b374c044e33

### DIFF
--- a/lib/mem_mgt.c
+++ b/lib/mem_mgt.c
@@ -266,7 +266,8 @@ void *malloc(unsigned int num_bytes)
 	}
 
 	/* Check if memory allocation is successful */
-	ASSERT(memory != NULL, "");
+	if (memory == NULL)
+		pr_err("%s: failed to alloc 0x%x Bytes", __func__, num_bytes);
 
 	/* Return memory pointer to caller */
 	return memory;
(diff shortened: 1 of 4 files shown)


lyan3:
Comments from yakuizhao: 
It seems that the ASSERT can help to dump some infos if HV_DEBUG is enabled. This is helpful.
But it will cause that the system enters the state of "pause" if the HV_DEBUG is not enabled.
In fact the condition check in ASSERT is not expected.
Is it ok to rework the implementation of assert when the HV_DEBUG is not enabled? Maybe like what you mentioned, it can be checked by caller instead of pause.

lyan3:
@yakuizhao  for lib functions, I think ASSERT should not be allowed. And for ASSERT in other parts of code, how about we use another MACRO like "CHECK" to check the condition, dump the context and return a value if condition is false? ASSERT(with pause) will be needed somewhere when HV_DEBUG is not enabled, what do you think?

yakuizhao:
@lyan3 , Your new suggestion make sense. it looks good that the "ASSERT" is forbidden in lib function.

lijinxia:
verified +1

PR #21, opened 2018-03-15
Simple vmx

lifeix:


lijinxia:
verified +1

lifeix:
@dongyaozu @lyan3 Yes. But IMO, we need ASSERT in some condition if this condition would cause the hypervisor to crash or can't boot.  Anyway, Like would fix this later if it necessary.

dongyaozu:
Leave ASSERT there is OK,  but we need careful review of each ASSERT to make sure a UOS won't be able to attack SOS thru ASSERT

PR #23, opened 2018-03-15
add create secure world ept API

mingqiangchi:
This patch is prepared for enabling secure world feature.
this api will create new eptp for secure world, whose PDPT
entries are copied form normal world,the PML4/PDPT for secure
world are separated from Normal World, PD/PT are shared in the
Secure World's EPT and Normal World's EPT.Secure world can
access Normal World's memory, but Normal World can not access
Secure World's memory
This function implemented：
-- Unmap specific memory from guest ept mapping
-- Copy PDPT from Normal world to Secure world
-- Map specific memory for Secure world
-- Unmap specific memory from SOS ept mapping


lijinxia:
verified +1

PR #25, opened 2018-03-16
Jmp to vm_exit directly if vm-entry fails in vmresume

yakuizhao:
…ails in vmresume

It is possible that the vm-entry fails in vmresume instr under some scenarios.
It will pass to next instruction following vmresume. In such case it will call
the vmlaunch again.


lijinxia:
verified +1

PR #26, opened 2018-03-16
Fix typo

mlouielu:
Should this be `UOSloader`? Since I didn't see anywhere refer `AOSloader` in `trusty.txt` or [`acrn-documentation`](https://projectacrn.github.io/introduction/index.html#boot-sequence) anymore.

Signed-off-by: Louie Lu <git@louie.lu>

### DIFF
--- a/Documentation/Trusty.txt
+++ b/Documentation/Trusty.txt
@@ -68,7 +68,7 @@ API
 Trusty Boot flow
 ****************
 Per design, AOSloader will trigger boot of Trusty. So the boot flow will be:
-    AOSloader --> ACRN --> Trusty --> ACRN --> AOSloader
+    UOSloader --> ACRN --> Trusty --> ACRN --> UOSloader
 
 Detail:
 1. UOSloader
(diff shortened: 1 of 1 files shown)


jren1:
Hi Bing, Could you please comment this patch?

gvancuts:
In case we want to merge this, please fix the line above too (as it also mentions `AOSloader`. It would actually be even better if the term was described a bit more in details in the document. One can _guess_ what it means but spelling it out would make it clearer.

Finally, is there any reason why we shouldn't move this to https://github.com/projectacrn/acrn-documentation? This looks like a very good and useful document to have and I would much prefer to have it with the rest of the document. It would have to be converted to ReST as well but I'm pretty sure @dbkinder could help there if needed.

anthonyzxu:
is below flow easier to understand?
ACRN --> UOSloader --> Trust --> UOSloader --> Android

bzhu5:
AOSLoader -> Android OS Loader (specific for Android Only).
UOSLoader -> UOS Loader.

so i think it is fine to use UOSLoader.

Documentation/Trusty.txt
```diff
@@ -68,7 +68,7 @@ API
 Trusty Boot flow
 ****************
 Per design, AOSloader will trigger boot of Trusty. So the boot flow will be:
```
bzhu5: in this line #70, there is yet another AOSloader

mlouielu:
Addressed @bzhu5 issue and add the full name at UOSloader first appearance.

lijinxia:
verified +1

PR #28, opened 2018-03-16
Remove unnecessary variable in map_mem

mingqiangchi:
the variable 'table_present' is redundant
in function(map_mem_region)

### DIFF
--- a/arch/x86/mmu.c
+++ b/arch/x86/mmu.c
@@ -166,7 +166,6 @@ static uint32_t map_mem_region(void *vaddr, void *paddr,
 		int ept_entry, enum mem_map_request_type request_type)
 {
 	uint64_t table_entry;
-	uint64_t table_present;
 	uint32_t table_offset;
 	uint32_t mapped_size;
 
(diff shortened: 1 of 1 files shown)


arch/x86/mmu.c
```diff
@@ -265,7 +252,7 @@ static uint32_t map_mem_region(void *vaddr, void *paddr,
 		}
 		case PAGING_REQUEST_TYPE_UNMAP:
 		{
-			if (tmp & table_present) {
+			if (entry) {
```
yakuizhao: The check is only for PAGING_REQUEST_TYPE_UNMAP.
Can we always clear the entry to zero for like what is done for PAGING_REQUEST_TYPE_MAP?
In such case the MEM_READ64 can be reduced.

mingqiangchi: I will add 'invalidate TLB' in these 3 cases in another patch, then it is necessary to check the entry here

lijinxia:
verified +1

PR #29, opened 2018-03-19
Add page mask MACRO

mingqiangchi:
Now used IA32E_REF_MASK when calculate the address,
it will clear bit63, should use CPU_PAGE_MASK

### DIFF
--- a/arch/x86/mmu.c
+++ b/arch/x86/mmu.c
@@ -640,7 +640,7 @@ void obtain_last_page_table_entry(struct map_params *map_params,
 	}
 
 	/* Obtain page table entry from PD table*/
-	table_addr = (void *)(table_entry&IA32E_REF_MASK);
+	table_addr = (void *)(table_entry & IA32E_REF_MASK);
 	table_entry = get_table_entry(map_params, addr,
 			table_addr, IA32E_PD);
 	table_present = check_page_table_present(map_params, table_entry);
@@ -667,7 +667,7 @@ void obtain_last_page_table_entry(struct map_params *map_params,
 	}
 
 	/* Obtain page table entry from PT table*/
-	table_addr = (void *)(table_entry&IA32E_REF_MASK);
+	table_addr = (void *)(table_entry & IA32E_REF_MASK);
 	table_entry = get_table_entry(map_params, addr,
 			table_addr, IA32E_PT);
 	table_present = check_page_table_present(map_params, table_entry);
--- a/include/arch/x86/cpu.h
+++ b/include/arch/x86/cpu.h
@@ -41,6 +41,7 @@
 /* Define page size */
 #define CPU_PAGE_SHIFT          12
 #define CPU_PAGE_SIZE           0x1000
+#define CPU_PAGE_MASK           0xFFFFFFFFFFFFF000
 
 /* Define CPU stack alignment */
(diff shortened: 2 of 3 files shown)


mingqiangchi:
'IA32E_REF_MASK'  has other use, if it is a ept entry, need to clear bit63

arch/x86/ept.c
```diff
@@ -83,7 +83,7 @@ static void *find_next_table(uint32_t table_offset,
 	}
 
 	/* Get address of the sub-table */
-	sub_table_addr = (void *)(table_entry & IA32E_REF_MASK);
+	sub_table_addr = (void *)(table_entry & CPU_PAGE_MASK);
```
yakuizhao: This is for the EPT usage.
It will be better that the IA32E_REF_MASK is still used.(The bit63 should be cleared).

arch/x86/guest/guest.c
```diff
@@ -181,7 +181,7 @@ uint64_t gva2gpa(struct vm *vm, uint64_t cr3, uint64_t gva)
 	addr = cr3;
 
 	for (level = 3; level >= 0; level--) {
-		addr = addr & IA32E_REF_MASK;
+		addr = addr & CPU_PAGE_MASK;
```
yakuizhao: This make senses as it is used to walk the page_table for guest.

arch/x86/mmu.c
```diff
@@ -442,7 +442,7 @@ static void  *walk_paging_struct(void *addr, void *table_base,
 				    (uint64_t) sub_table_addr | table_present);
 		} else {
 			/* Get address of the sub-table */
-			sub_table_addr = (void *)(table_entry & IA32E_REF_MASK);
+			sub_table_addr = (void *)(table_entry & CPU_PAGE_MASK);
```
yakuizhao: This file is still used for EPT. (Of course it is also used in paging_init of hypervisor).
In such case the IA32E_REF_MASK is better.

yakuizhao:
This changes will affect the ept and gva2gpa.
For the gva2gpa, the  change of Bit63 mask makes sense.
But for the EPT usage, it is unnecessary to change IA32E_REF_MASK to CPU_PAGE_MASK.

So I think that we can limit the new PAGE_Macro to gva2gpa.
Thanks

yakuizhao:
The updated patch looks good. And the change is limited to gva2gpa.

And this patch also includes the change of coding style in mmu.c.  It will be better that this can be split into two patches. One is the change of page macro in gva2gpa. And the second is the coding style in mmu.c.

Thanks

arch/x86/guest/guest.c
JasonChenCJ: Why we use CPU_PAGE_MASK here? What will happen if XD set? Should we still keep IA32_REF_MASK?

yakuizhao: @JasonChenCJ ,  This is only used to obtain the corresponding address in gva2hva. 
The paging mechanism can use the CPU_PAGE_MASK macro to return the address for next level.

mingqiangchi: for cr3, here we can use CPU_PAGE_MASK, but next to ‘addr = entry;‘, then here should use IA323_REF_MASK

lijinxia:
verified +1

PR #31, opened 2018-03-20
ptdev: fix bug when update ptdev entry

JasonChenCJ:
info->phys_pin need be used by ptdev_build_native_rte when updating entry

TODO: currently ptdev entry is virtual based, the better solution should
be physical based.


lijinxia:
verified +1

PR #32, opened 2018-03-20
vm load: fix bug in loading kernel

zgenx:
According to the explaination for pref_address
in Documentation/x86/boot.txt, a relocating bootloader
should attempt to load kernel at pref_address if possible.
But due to a non-relocatable kernel will unconditionally
move itself and to run at perf address, no need to copy
kernel to perf_address by bootloader.


lijinxia:
verified +1

PR #30, opened 2018-03-19
Makefile: fix a couple of wrong variable assignments

gvancuts:
Fix a couple of wrong assignments to D_SRCS and C_SRCS.

Signed-off-by: Geoffroy Van Cutsem <geoffroy.vancutsem@intel.com>

PR #35, opened 2018-03-20
Makefile: fix build issue when RELEASE=1

gvancuts:
Fix a build issue when using RELEASE=1. This was introduced
by mistake in a previous commit and this one partially reverts
the change and also addresses another issue that had not been
noticed before.


gvancuts:
This partially reverts a change introduced in #30 and also addresses a pre-existing issue. With this PR, all combination of `RELEASE` and `PLATFORM` can be successfully compiled. I have not tested those on real hardware though.

| `RELEASE` | `PLATFORM` | Result |
| :-: | :-: | :-: |
| 0 | 0 | OK |
| 0 | 1 | OK |
| 1 | 0 | OK |
| 1 | 1 | OK |

dbkinder:
The commit message should say what the issue actually is that’s being fixed.

gvancuts:
Force-pushed with an updated commit.

anthonyzxu:
@gvancuts 

all c files under debug should be in D_SRC,
please check my comment in 
https://github.com/projectacrn/acrn-hypervisor/pull/30#issuecomment-374797288

JasonChenCJ:
we should only leave printf.c built for release version, and do not build in any others, Please check the SOS kernel cmdline, and make sure under release version, what's the real HW device ttyS0 point to, if this will conflict with others,  the simple way is remove "pci_devices_ignore=(0:18:2)" and "console=ttyS0" from cmdline.

gvancuts:
Got it @anthonyzxu 

It looks like this PR would be superceded by #45 , is that correct? If true, I'll simply close this one.

anthonyzxu:
That's correct, please close this one.
And thank you for bringing this issue up!

PR #37, opened 2018-03-20
makefile: adjust linker path in Makefile

dbkinder:
As mentioned in https://github.com/projectacrn/acrn-documentation/pull/38 , 
on some operating systems, the ``gnu-efi`` package installs the
linker under a ``gnuefi`` folder in ``${LIBDIR}``. This is the
case in Fedora for example. Check if the gnuefi folder is there
and use it if it is, in the path to the linker.  This PR fixes the Makefile rather
than documenting how to edit it yourself.

### DIFF
--- a/bsp/uefi/efi/Makefile
+++ b/bsp/uefi/efi/Makefile
@@ -55,8 +55,15 @@ endif
 INCDIR := /usr/include
 
 # gnuefi sometimes installs these under a gnuefi/ directory, and sometimes not
-CRT0 := $(LIBDIR)/crt0-efi-$(ARCH).o
-LDSCRIPT := $(LIBDIR)/elf_$(ARCH)_efi.lds
+ifneq ("$(wildcard $(LIBDIR)/gnuefi/crt0-efi-$(ARCH).o)","")
+	CRT0 := $(LIBDIR)/gnuefi/crt0-efi-$(ARCH).o
+	LDSCRIPT := $(LIBDIR)/gnuefi/elf_$(ARCH)_efi.lds
+else
+	CRT0 := $(LIBDIR)/crt0-efi-$(ARCH).o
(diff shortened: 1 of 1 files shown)


bsp/uefi/efi/Makefile
```diff
@@ -55,8 +55,15 @@ endif
 INCDIR := /usr/include
 
 # gnuefi sometimes installs these under a gnuefi/ directory, and sometimes not
-CRT0 := $(LIBDIR)/crt0-efi-$(ARCH).o
-LDSCRIPT := $(LIBDIR)/elf_$(ARCH)_efi.lds
+ifneq ("$(wildcard $(LIBDIR)/gnuefi/crt0-efi-$(ARCH).o)","")
```
gvancuts: Nitpick, you could just make this `ifneq ($(wildcard $(LIBDIR)/gnuefi/*),)` (it makes it a little more readable).

dbkinder: Ha.  OK.  I thought it was best to check for an expected file (for that $ARCH rather than just having the folder there.

gvancuts:
Verified that compilation passes for both `make` and `make PLATFORM=uefi` on the following Linux distributions:
* CentOS 7
* Fedora 26 and 27
* Ubuntu 14.04 and 16.04
* Debian 8
* Clearlinux

Couldn't test with `RELEASE=1` as this is still broken (regardless of the Linux distribution)

jren1:
Did you verify on at least one UEFI platform, such as APL NUC?

gvancuts:
Hi @jren1 , I have just tried an `acrn.efi` binary built with that patch (built on a Fedora 27 dev machine). It went straight to the SOS as before. The network is functional as before. Let me know if there is any specific test you want me to try out.

lijinxia:
verified +1

PR #38, opened 2018-03-20
Add minimal .travis.yml file

shrmrf:
This can be extended for more fine-grained control of the build process. I created a pull request (instead of an issue) because issues are not enabled in the `acrn-*` repositories.

Feedback on CI builds is welcome. I'm a noob at this.


dbkinder:
Let's get issues enabled!

gvancuts:
+1 on getting issues enabled! :-)

shrmrf:
@gvancuts thanks :) 
idk if this issue will be merged but I think (now) the minimal `.travis.yml` should also consist of the `make clean` step. (I'll add it and update this pull request)

I also figured that the device model does not get built with `gcc >= 7` (the fix is only one flag `-format-truncation=0` ) .. would be really nice to have issues running

shrmrf:
Updated `.travis.yml` 

Even if you don't go with travis for CI, I think there must be an automated build status and config file (maybe even to prove that builds work)

gvancuts:
Hi @shrmrf , thanks a lot for your contribution and suggestion. I more than 100% agree that we need some sort of CI in place. I don't know if the engineering team has already thought about it. I gave this some (short) thought over the last couple of days as I was battling with build issues myself. In my opinion, we should test all combinations of `RELEASE={1|0}` and `PLATFORM={sbl|uefi}`.

Ideally we would also verify that these can be built in different operating systems. I have been testing with these ones (in my little corner) so far:
* Ubuntu 14.04 and 16.04
* Debian 8
* CentOS 7
* Fedora 26 and 26
* Clearlinux

The little bit that I looked at Travis CI, I could not find a way to test in different host environment (it seems to provide a Ubuntu host and also a MacOS if you specify it). I've heard of Shippable too and I was going to investigate whether it is a little more flexible in that space. All we'd really need here is the ability to run containers. This is how I test these different OSs in my office, I have a container for each of those that comes pre-installed with all build tools and dependencies and I can test all build combinations that I want that way.

gvancuts:
> I also figured that the device model does not get built with gcc >= 7 (the fix is only one flag -format-truncation=0 ) .. would be really nice to have issues running

Is this what is described (and tracked) here: https://github.com/projectacrn/acrn-devicemodel/pull/14?

shrmrf:
@gvancuts I believe so, yes.

gvancuts:
> idk if this issue will be merged but I think (now) the minimal .travis.yml should also consist of the make clean step. (I'll add it and update this pull request)

Yes, I've noticed that building for a different platform will fail if you don't do a 'make clean' in between. The 'make clean' step isn't always cleaning everything at the moment as what it cleans (or not) is affected by the PLATFORM assignment... we ought to fix that at some point.

shrmrf:
@gvancuts I believe that we can provide a `Dockerfile` with each repository. This way, you won't have to have a complex build matrix and can target a couple of OS. Travis can use docker to do the builds as well.

I did email Jack internally to ask him to enable issues. Let's see...

gvancuts:
@shrmrf , can you send me a pointer to the Travis doc that explains how to run Docker containers to do the testing? All I found so far was this page: https://docs.travis-ci.com/user/reference/overview/. What they describe is that they can use a container to run your test but as far as I can see, it does not let you choose what container it uses (and based on the doc, they only have Ubuntu 14.04). I really would like to see us being able to validate the compilation on different OS (at a minimum: Clearlinux, Ubuntu 16.04 and Fedora 27)... I don't see how to do that with Travis (but I really don't know much about Travis yet).

I don't think merging this will automatically get Travis going, so someone with sufficient access rights will need to link this repo with Travis.

fuzhongl:
Verified+1

shrmrf:
Hello @gvancuts sure... there are various ways:-

- Using [Docker Compose](https://docs.travis-ci.com/user/docker/#Using-Docker-Compose) 
- Using [dockerhub](https://docs.travis-ci.com/user/docker/#Using-a-Docker-Image-from-a-Repository-in-a-Build)

But I think [building the image](https://docs.travis-ci.com/user/docker/#Building-a-Docker-Image-from-a-Dockerfile) might be correct in these cases. Docker compose can also be used if you want to test against more environments

shrmrf:
Should I create an example Dockerfile and create a pull-request?

gvancuts:
I think it would be interesting if you could create an example Dockerfile. I was playing a bit with Shippable today with a Docker container based on Clearlinux but I'm stuck. Although it looks like you can use pretty much any container, there are still a number of operations being done in the background and one of them fails for me (see https://github.com/Shippable/support/issues/4223).

@Rraghuveer , are you the right person to activate Travis?

shrmrf:
Created [new pull request](https://github.com/projectacrn/acrn-hypervisor/pull/52)

PR #42, opened 2018-03-21
Move s[n]printf to lib for release

lifeix:
@anthonyzxu


lifeix:
Seems OK

lib/sprintf.c
```diff
@@ -0,0 +1,725 @@
...
+#endif
+
+#define PRINT_STRING_MAX_LEN 4096
+
+/** Command for the emit function: copy string to output. */
+#define PRINT_CMD_COPY			0x00000000
+
+/** Command for the emit function: fill output with first character. */
+#define PRINT_CMD_FILL			0x00000001
+
```
JasonChenCJ: PRINT_CMD_COPY & PRINT_CMD_FILL is dup with header file

debug/printf.c
```diff
@@ -34,622 +34,6 @@
 #include <hv_arch.h>
```
JasonChenCJ: why dont we move the whole printf.c into lib directly?

anthonyzxu: some functions are used only for debug, so put them under debug/ , and reduce the code size.

lijinxia:
verified +1

PR #43, opened 2018-03-21
mmu: add invalidate TLB after update MMU entry

mingqiangchi:
This patch add invalidate operation after update MMU entry including
map/unmap/modify memory.
if it is the first mapping, no need to invalidate,
unmap and modify, need to invalidate TLB.


mingqiangchi:
Close it, will submit a new patch to implement  TLB shootdown.

PR #44, opened 2018-03-21
Add handle EOI case when APIC-write VM exit

yonghuah:
-If "virtual-interrupt delivery" VM-execution control is 0,
  Processor will causes an APIC-write VM exit if page offset
  is 0xB0 (EOI), SDM Vol3, Chapter 29.4.3


yonghuah:
duplicated with PR#46

PR #46, opened 2018-03-21
Vapic enhance

yonghuah:
these patches  are to detect and enable only APICv features which are actually supported by the processor, instead of turning on all features by default. and as prerequisites, apic_access vexit handler and EOI handler in apic-write handler are need.


yonghuah:
Below scenarios tested and works (sos & uos an boot normally)
   APICv   VIRT-REG   vINTR-Delivery
    ON  	     ON          ON
    ON  	     ON          OFF
    ON  	     OFF         ON
    ON  	     OFF         OFF
    OFF  	     n/a         n/a

PR #47, opened 2018-03-22
hv: rename VM exit handlers in vlapic

yonghuah:
to keep aligned with the vmexit reasons' name defined in
SPEC.

### DIFF
--- a/arch/x86/guest/vlapic.c
+++ b/arch/x86/guest/vlapic.c
@@ -2337,7 +2337,7 @@ apicv_inject_pir(struct vlapic *vlapic)
 	}
 }
 
-int apic_access_exit_handler(struct vcpu *vcpu)
+int apic_access_vmexit_handler(struct vcpu *vcpu)
 {
 	bool ret;
 	int access_type, offset;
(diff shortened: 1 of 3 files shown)


arch/x86/guest/vlapic.c
```diff
@@ -2366,7 +2366,7 @@ int apic_access_exit_handler(struct vcpu *vcpu)
 	return 0;
 }
 
-int apicv_virtualized_eoi_exit_handler(struct vcpu *vcpu)
+int virtualized_eoi_vmexit_handler(struct vcpu *vcpu)
 {
```
dongyaozu: Should this be eoi_vmexit_handler? or apicv_eoi_vmexit_handler?

dongyaozu: then veoi_vmexit_handler or eoi_vmexit_handler is fine.  We didn't use "virtualized" in other APIs.

yonghuah: Got it, will update.

lijinxia:
verified+1

PR #48, opened 2018-03-22
ioreq: only clear vcpu->req when io request needed

JasonChenCJ:


yakuizhao:
Nice cleanup.
It will be better that the commit log can be described more clearly.(Why and the purpose).

PR #49, opened 2018-03-23
hv: microcode: Enable microcode update from SOS.

fyin1:
microcode update from UOS is disabled.
microcode version checking is available for both SOS and UOS.

There are two TODOs of this patch:
1. This patch only update the uCode on pCPUs SOS owned. For the
   pCPUs not owned by SOS, the uCode is not updated. To handle
   this gap, we will have SOS own all pCPUs at boot time. So
   all pCPUs could have uCode updated. This will be handled
   in the patch to enable SOS own all pCPUs at boot time.

2. gva2gpa now doesn't check possible page table walk failure.
   Will add the failure check in gva2gpa in different patch.

Reviewed-by: Anthony Xu (anthony.xu@intel.com)
Acked-by: Tian, Kevin <kevin.tian@intel.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,6 +113,7 @@ C_SRCS += arch/x86/guest/vpic.c
 C_SRCS += arch/x86/guest/vmsr.c
 C_SRCS += arch/x86/guest/vioapic.c
 C_SRCS += arch/x86/guest/instr_emul.c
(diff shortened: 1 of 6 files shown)


fuzhongl:
Verified+1

PR #53, opened 2018-03-26
debug: make ASSERT empty under release version

JasonChenCJ:
ASSERT is only for debug purpose, for release version, it should try
error handling instead of deadloop there.

v1:
- change the ASSERT under release version to empty code

TODO: revise all ASSERT usage

Acked-by: Eddie Dong (Eddie.dong@intel.com)


lijinxia:
verified +1

PR #54, opened 2018-03-26
Fix compiler warnings when building for UEFI

rustylynch:
Straight forward compiler warning fix patch when building with PLATFORM=uefi

The most controversial bits are where to place the function definitions for the uefi related functions.  Instead of adding a new header I just placed them in a CONFIG_EFI_STUB protected section of apicreg.h.  Not sure if there is a better place.


lijinxia:
verified+1

PR #61, opened 2018-03-29
Trusty enabling patch series

YadongQi:
This patch series enables trusty feature base on ACRN hypervisor.

There are four main parts of this patch series.
1. HC_WORLD_SWITCH hypercall
   This hypercall is used to simulate Secure Monitor Call(SMC) instruction to
   switch vCPU world state between Normal World and Secure World.
2. HC_INITIALIZE_TRUSTY hypercall
   This hypercall is used by UOS_Loader to request hypervisor to init Secure World
   environment and launch Trusty-OS.
3. Add key_info for Trusty
   key_info is a structure which contains many sensetive info such like seed,
   rpmp_key, platform, manufacturing state, etc. This structure must be provided
   when Trusty OS boot up, otherwise Trusty OS will fail to boot. Per previsous
   design, hypervisor will got key_info from bootloader and derive vkey_info for
   Trusty. Currently, bootloader did not pass key_info to hypervisor, so for
   trusty bring-up, a dummy key_info is used temporarily.
4. Remove pio emulation assertion for Normal World
   Currently, the UOS serial log is printed through IO(0x3f8). Secure World also
   print serial log by this IO port. So the assert should removed for Trusty OS
   booting up.
5. Rename TIME_MS_DELTA to CYCLES_PERMS
   TIME_MS_DELTA is not clear enough, so rename it to CYCLES_PER_MS.

For more trusty detail design, please read Documentation/Trusty.txt for reference.

### DIFF
--- a/Documentation/Trusty.txt
+++ b/Documentation/Trusty.txt
@@ -49,8 +49,8 @@ Note: Trusty OS is running in Secure World in the architecture above.
 **************************
 Trusty specific Hypercalls
 **************************
-1. HC_LAUNCH_TRUSTY
-   ->This Hypercall is used by UOSloader (User OS Bootloader) to request ACRN to launch Trusty.
+1. HC_INITIALIZE_TRUSTY
+   ->This Hypercall is used by UOS_Loader to request ACRN to initialize Trusty.
    ->The Trusty memory region range, entry point must be specified.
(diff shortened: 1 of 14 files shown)


lijinxia:
verified +1

PR #62, opened 2018-03-29
hv: Save secure world memory info into vm instead of vm0

mingqiangchi:
A bugfix for saving secure world memory info.
Maybe there are multiple UOS, each VM has its own secure
world and normal world, should save memory info into individual VM.


TianhuaSun:
verified +1

PR #63, opened 2018-03-29
Fix struct assignment: no memcpy with -nostdlib.

busykai:
This fixes link error for struct assignment introduced in 4af2f04bd2f.

Use `memcpy_s` instead of struct assignment.

The error seen on current master:
```
gcc <...> -o src/acrn-hypervisor/build/acrn.out -static -nostdlib
<...>
src/acrn-hypervisor/build/arch/x86/trusty.o: In function `setup_trusty_info':
src/acrn-hypervisor/arch/x86/trusty.c:297: undefined reference to `memcpy'
```

### DIFF
--- a/arch/x86/trusty.c
+++ b/arch/x86/trusty.c
@@ -294,7 +294,8 @@ static bool setup_trusty_info(struct vcpu *vcpu,
 	/* TODO: prepare vkey_info */
 
 	/* copy key_info to the first page of trusty memory */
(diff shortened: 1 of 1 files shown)


busykai:
ok. on a side note, favoring the latest over the first might be not the most encouraging open source practice.

anthonyzxu:
Agreed, I should go over all the PRs first. I didn't expect two PRs fixing the same issue. lesson learnt!

PR #64, opened 2018-03-29
Fix compile error caused by missing memcpy function

rustylynch:
On some toolchain configurations direct struct assignments will
default to a memcpy operation which is not present in this
environment, so explicitly use the internal memcpy_s function.

Signed-of-by: Rusty Lynch <rusty.lynch@intel.com>

### DIFF
--- a/arch/x86/trusty.c
+++ b/arch/x86/trusty.c
@@ -34,6 +34,7 @@
 #include <acrn_hv_defs.h>
 #include <hv_debug.h>
 #include <hkdf.h>
+#include "rtl.h"
 
 _Static_assert(NR_WORLD == 2, "Only 2 Worlds supported!");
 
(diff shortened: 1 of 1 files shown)


busykai:
I have submitted #63 for this as well, an hour ago. :)

arch/x86/trusty.c
```diff
@@ -294,7 +295,8 @@ static bool setup_trusty_info(struct vcpu *vcpu,
 	/* TODO: prepare vkey_info */
 
 	/* copy key_info to the first page of trusty memory */
-	mem->first_page.key_info = g_key_info;
+	memcpy_s(&mem->first_page.key_info, sizeof(g_key_info),
```
busykai: technically 2nd argument needs to be the size of the destination, not the source.

TianhuaSun:
verified +1

PR #65, opened 2018-03-30
Change the acrn boot flow

zhenggen:
The following is the change list of this patch list.

a. Bug fix on delivering RSDP from efi stub to hypervisor
   With current code, memcpy rsdp to 0x500 may overwrite uefi
   code/data region. So just deliver the pointer of RSDP,
   which is already saved in uefi system table, to hypervisor.

b. Change the acrn boot flow on uefi platform
   With current code, the acrn.efi is inserted between
   cl bootloader.efi and bzImage.efi that destroyed the chain
   relationship of cl bootloader and cl bzImage.efi.
   And the following is current boot flow:
   UEFI -> cl bootloader.efi -> acrn.efi -> bzImage.efi
   Resume above chain relationship, and make uefi vm return to
   efi stub context once launched, then continue to call the
   UEFI API(LoadImage/StartImage) to launch cl bootloader or
   other bootloaders. So the boot flow will change to be as below:
   UEFI -> acrn.efi -> cl bootloader.efi -> bzImage.efi

c. Change the efi_ctx passing method
   Save the pointer of efi_ctx in mi_drivers_addr field of
   multiboot structure and pass to hypervisor, not by
   saving in register RDX(the third default parameter in
   64bit call function). And with this method, we can be
   compatible with the original 32bit boot parameters passing
   method and no need to large the array size of boot_regs
   in hypervisor.

d. Cleanup the mass code in efi stub
   Move the header content from boot.c to boot.h.
   Remove uncessary functions and create new function named
   construct_mbi to cleanup the mass code for construcing the
   multiboot information.

e. Update documentation
   Add a new document named ACRN_UEFI.txt to direct user how
   to use the acrn.efi.


TianhuaSun:
verified +1

avahidi:
With commit f216b574 ( 5725c78 )  the hypervisor no longer boots on the upsquared board. The call to LoadImage from efi_main() never returns.

Any suggestions on how to find the cause of this issue?

gvancuts:
Hi @avahidi , just as an FYI, I highlighted your question on the mailing list as it's probably the best place to continue that conversation with the engineering team (there was a thread related to the UP Squared board already going on there) - see https://lists.projectacrn.org/g/acrn-dev/message/1208

PR #67, opened 2018-04-02
treewide: fix declarations of functions with empty parameter lists

junjiemao1:
According to C99:

    The empty list in a function declarator that is not part of a definition of
    that function specifies that no information about the number or types of the
    parameters is supplied.

This means gcc is happy with the following code, which is undesirable.

    void foo();    /* declaration with an empty parameter list */

    void bar() {
        foo();     /* OK */
        foo(1);    /* OK */
        foo(1, 2); /* OK */
    }

This patch fixes declarations of functions with empty parameter lists by adding
an unnamed parameter of type void, which is the standard way to specify that a
function has no parameters. The following coccinelle script is used.

    @@
    type T;
    identifier f;
    @@

    -T f();
    +T f(void);

New compilation errors are fixed accordingly.


lijinxia:
verified +1

PR #68, opened 2018-04-02
hv: pass-through xsave feature to guests

yonghuah:
enable Xsave feature and pass-through it to guests
  - enable host xsave before expose it to guests.
  - add validation for the value to be set to 'xcr0' before call xsetbv
    when handling xsetbv vmexit.
  - tested in SOS guest, created two threads to do different
    FP calculations,test code runs in user land of sos.


arch/x86/cpuid.c
```diff
@@ -329,6 +329,18 @@ void guest_cpuid(struct vcpu *vcpu,
...
 
+		/*no xsave support for guest if it is not enabled on host*/
+		if (!(*ecx & CPUID_ECX_OSXSAVE))
+			*ecx &= ~CPUID_ECX_XSAVE;
+
+		*ecx &= ~CPUID_ECX_OSXSAVE;
+		if (*ecx & CPUID_ECX_XSAVE) {
+			uint64_t cr4;
+			/*read guest CR4*/
+			cr4 = exec_vmread(VMX_GUEST_CR4);
```
dongyaozu: Donot know why we need to check guest CR4 here. This is equal to vCR4, right?

yonghuah: Hi Eddie,  
it's to check guest CR4 (vCR4),  We can't return CPUID.ECX.OSXSAVE = 1 if guest did not set vCR4.OSXSAVE.

yonghuah:
@dongyaozu , VMexit will occur guest write CR4. in handling VMX_EXIT_REASON_CR_ACCESS, function cr_access_handler() will write guest cr4  by call write_cr4();

lijinxia:
verified +1

PR #69, opened 2018-04-03
Create sworld

mingqiangchi:


arch/x86/trusty.c
```diff
@@ -96,56 +96,68 @@ _Static_assert(sizeof(struct trusty_startup_param)
...
-				+ (i * IA32E_COMM_ENTRY_SIZE));
-		pml4e = MEM_READ64(vm->arch_vm.sworld_eptp);
-		MEM_WRITE64((pml4e & IA32E_REF_MASK)
-				+ (i * IA32E_COMM_ENTRY_SIZE),
-				entry);
-	}
+	vm->arch_vm.sworld_eptp = alloc_paging_struct();
+
+	/* The trusty memory is remapped to guest physical address
+	 * of 511G to 511GB + 16MB
```
bzhu5: how about using corresponding macros for 511GB and 16MB? (16MB will potentially be changed in future)

lijinxia:
verified+1

arch/x86/trusty.c
mingqiangchi: here the map is according to the input parameters(rebased_gpa and size), i will update the comment

lijinxia:
verified +1

PR #71, opened 2018-04-03
hv: Use the maximum physical address number from CPUID

junjiemao1:
Currently we assume that the lower 52 bits of physical addresses are always
effective. However this may not be the case on some platforms. Per SDM:

    If CPUID.80000008H:EAX[7:0] is supported, the maximum physical address
    number supported should come from this field.

This patch series reads the maximum physical address number at runtime from
CPUID and calculate the actual mask before any physical address operations.

v1 -> v2:
   * Panic on platforms where CPUID.80000008H is not supported.


lijinxia:
@junjiemao1 Met error when apply the patches:
error: patch failed: arch/x86/cpu.c:367
error: arch/x86/cpu.c: patch does not apply

junjiemao1:
I have rebased the patches. Would you make another try, please?

lijinxia:
verified +1

PR #74, opened 2018-04-04
hv: shell: add cpuid command

lifeix:
Add cpuid shell command for QA cpuid test


lifeix:
@dongyaozu  yes. The QA want to compare it with the value in Guest OS and find out which is emulated.

jren1:
verified+1

PR #76, opened 2018-04-05
refine functions in mmu.c

JasonChenCJ:


JasonChenCJ:
@dongyaozu, which patch you are talking about, is "replace ASSERT with panic in fetch_page_table_offset"?

JasonChenCJ:
will update with a new PR

PR #77, opened 2018-04-05
replace malloc and memset with calloc

yechunliang:
malloc: allocate a block of memory, the contents of the block are undefined.
calloc: allocate a block of memory for an array of num elements and initializes all its bits to zero.

Signed-off-by: yechunliang <yechunliangcn@163.com>


jren1:
verified+1

PR #80, opened 2018-04-06
Travis CI: enable compilation testing on multiple distros

gvancuts:
These two commit shamelessly built upon the work done by @shrmrf and implement the following:
* Build 3 Docker containers based on the following Linux distributions (inc. all build tools and dependencies):
   * Clearlinux
   * Ubuntu 16.04
   * Fedora 26
* For each of those, build the `acrn-hypervisor` code for all combinations of the following variables
   * PLATFORM={sbl|uefi}
   * RELEASE={0|1}

*Note:* additional Dockerfile for other distributions are available too but not used for now.

This would allow the project to use the Travis CI system and make sure that every Pull Request can at least be built (no functional testing is done).


gvancuts:
For a quick look at the details of what the Travis CI system does, check:
[Build Status](https://travis-ci.org/gvancuts/acrn-hypervisor)

shrmrf:
@gvancuts and @jren1 can you maybe also add the badge to the readme file? So it can look something like this for example: https://github.com/llogiq/mutagen/blob/master/README.md

Colors are always nice.

gvancuts:
> @gvancuts and @jren1 can you maybe also add the badge to the readme file? So it can look something like this for example: https://github.com/llogiq/mutagen/blob/master/README.md

I was planning to do that but after the PR has been merged **and** Travis CI enabled. Until that happens, the badge will use an invalid URL so there is not much point.

> Colors are always nice

Completely agree! :-)

jren1:
I just pushed one patch to my acrn-hypervisor, then the Travis CI system sent one email to me and reminded me that my patches has passed the build. It is nice! I will love to merge this patch to allow the future patches to be built in more Linux distributions by Travis CI!

NanlinXie:
I also have tried to setup [my Travis CI system](https://travis-ci.org/NanlinXie/acrn-hypervisor) and run build successfully, it is cool! I may considering whether we could do some more on it for testing.

shrmrf:
I think we should enable:
-  Github issues on this repo (instead of our forks)
  - Issues might be the only way to get feedback from the community
  - That, and something like [Gitter](https://gitter.im/) or [Slack](https://slack.com/) (Gitter is better because of its signup using Github account)
- Travis CI on this repo
  - Add Travis Build Badges to the `README.md` file

PR #81, opened 2018-04-06
UEFI: get the guest entry address explicitly

jren1:
Previously with the help from gcc, __builtin_return_address(0) is
used as the guest entry address, but it can't work well with gcc 7.3
and leads to the guest unable to return to UEFI correctly.
Let's get it explicitly.


lijinxia:
verified +1

PR #82, opened 2018-04-07
HV: Guest Pstate control support

jsun26intel:
To enable guest pstate control, guest must have intel_pstate driver or
acpi-cpufreq driver enabled in kernel. Especialy for acpi-cpufreq control,
some ACPI objects/methods (i.e. _PCT/_PPC/_PSS) must be included in UOS
ACPI DSDT or SSDT table.

To avoid complex work on parsing native ACPI table or evaluate object via
kernel acpica lib, here we use a more simplified method:

	1. Hard code the Px data in HV;
	2. Host load the Px data and store them to boot_cpu_data during
	   boot;
	3. Each VM has a Px data copy when create VM;
	4. Add a VHM service in kernel that could interact with HV and DM
	   on data transtion;
	5. Before launching UOS, DM query HV the supported Px data;
	6. DM generate the DSDT table with these Px data;
	7. HV need to trap all UOS Px request to make sure it is valid or
	   not;
	
change log:
	v4: (1) px data should be per-cpu of target vm, but v3 actually
		query it from vm0, fixed it to query target vcpu of target
		vm;
	    (2) move vm0 check to validate_pstate();
	    (3) add px data of APL NUC;

	v3: (1) change dynamic allocated px data to fixed MAX_PSTATE size
		px data array struct in vm;
	    (2) fix missed perf_ctl msr read trap;

	v2: (1) fix cpu model name char buffer bug;
	    (2) load host px data to boot_cpu_data struct and store to each
	        vm while create vm instead of store in global variable;
	    (3) vhm hypercall only support transit one specific pstate data
	        instead of all cpu state data;

	v1: initial version;


lijinxia:
I met with a compile error:
arch/x86/cpu_state_tbl.c:84:2: error: ‘for’ loop initial declarations are only allowed in C99 mode
for (int i = 0; i < count; i++) {

jsun26intel:
Hi Jinxia, I've pushed again by force update, please check, thanks!

jsun26intel:
Hi Eddie,
I've changed the code from 
for (int i = 0; i < ...;  i++)
to
int i;
for (i = 0; i < ...; i++)

please approve, thanks a lot!
BTW, Jason said he has a forced update also, please check. Thanks!

lijinxia:
verified +1

PR #86, opened 2018-04-09
trusty: initialize tsc_offset to zero

YadongQi:
Per trusty design, trusty requires a monotonic increasing
secure tick(TSC) at run time. This secure tick will used
to mitigate password/pin force attack, control key expiration,
etc.

Currently, the TSC_OFFSET is enabled. And guest will got
(host_tsc + tsc_offset) when execute rdtsc/rdtscp/rdmsr to
aquire tsc value. The host_tsc is always keeping increasing
during the runtime.

So initialize tsc_offset of trusty to 0 will ensure the
secure tick feature.


lijinxia:
verified +1

PR #94, opened 2018-04-11
remove 'noxsave' in acrn.conf

yonghuah:
- xsave is enabled for guests


lijinxia:
verified +1

PR #95, opened 2018-04-11
Add IO requrest 'req_buf' check before reference

yonghuah:
This address maybe invalid if a hostile address was set
in hypercall 'HC_SET_IOREQ_BUFFER'.

In the perspective of security, this address should be
validated before using.

### DIFF
--- a/arch/x86/ept.c
+++ b/arch/x86/ept.c
@@ -344,8 +344,9 @@ int dm_emulate_mmio_post(struct vcpu *vcpu)
 {
 	int ret = 0;
 	int cur = vcpu->vcpu_id;
-	struct vhm_request_buffer *req_buf =
-		(void *)HPA2HVA(vcpu->vm->sw.req_buf);
+	struct vhm_request_buffer *req_buf;
+
+	req_buf = (struct vhm_request_buffer *)(vcpu->vm->sw.req_buf);
 
 	vcpu->req.reqs.mmio_request.value =
 		req_buf->req_queue[cur].reqs.mmio_request.value;
--- a/arch/x86/io.c
+++ b/arch/x86/io.c
@@ -39,11 +39,12 @@ int dm_emulate_pio_post(struct vcpu *vcpu)
 {
 	int cur = vcpu->vcpu_id;
 	int cur_context = vcpu->arch_vcpu.cur_context;
-	struct vhm_request_buffer *req_buf =
-		(void *)HPA2HVA(vcpu->vm->sw.req_buf);
+	struct vhm_request_buffer *req_buf = NULL;
 	uint32_t mask =
 		0xFFFFFFFFul >> (32 - 8 * vcpu->req.reqs.pio_request.size);
 	uint64_t *rax;
+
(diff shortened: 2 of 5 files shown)


arch/x86/ept.c
```diff
@@ -344,8 +344,12 @@ int dm_emulate_mmio_post(struct vcpu *vcpu)
...
 	int ret = 0;
 	int cur = vcpu->vcpu_id;
-	struct vhm_request_buffer *req_buf =
-		(void *)HPA2HVA(vcpu->vm->sw.req_buf);
+	struct vhm_request_buffer *req_buf = NULL;
+
+	if (vcpu->vm->sw.req_buf == 0)
+		return -1;
+
+	req_buf = (void *)HPA2HVA(vcpu->vm->sw.req_buf);
```
dongyaozu: GPA2HVA ?   And we need to check if the GPA is valid... And check HVA is valid

yonghuah:
@dongyaozu @anthonyzxu 
Thanks for comments, will update and save it as HVA

arch/x86/ept.c
```diff
@@ -344,8 +344,12 @@ int dm_emulate_mmio_post(struct vcpu *vcpu)
 {
 	int ret = 0;
 	int cur = vcpu->vcpu_id;
-	struct vhm_request_buffer *req_buf =
-		(void *)HPA2HVA(vcpu->vm->sw.req_buf);
+	struct vhm_request_buffer *req_buf = NULL;
+
+	if (vcpu->vm->sw.req_buf == 0)
+		return -1;
```
JasonChenCJ: could we use errno instead of -1 here

arch/x86/io.c
```diff
@@ -39,12 +39,16 @@ int dm_emulate_pio_post(struct vcpu *vcpu)
...
 	int cur_context = vcpu->arch_vcpu.cur_context;
-	struct vhm_request_buffer *req_buf =
-		(void *)HPA2HVA(vcpu->vm->sw.req_buf);
+	struct vhm_request_buffer *req_buf = NULL;
 	uint32_t mask =
 		0xFFFFFFFFul >> (32 - 8 * vcpu->req.reqs.pio_request.size);
 	uint64_t *rax;
 
+	if (vcpu->vm->sw.req_buf == 0)
+		return -1;
```
JasonChenCJ: ditto

common/hypercall.c
```diff
@@ -783,6 +782,8 @@ int acrn_insert_request_wait(struct vcpu *vcpu, struct vhm_request *req)
 	if (!vcpu || !req || vcpu->vm->sw.req_buf == 0)
 		return -1;
```
JasonChenCJ: ditto, could you help change it together?

yonghuah:
have updated according to comments, thanks.

yonghuah:
@dongyaozu  Have removed 2 'req_buf' checks.
  but not add this kind of check in prepare_vcpu( ) as we discussed.   as by check the device model code, it will not continue .. if  calling 'IC_SET_IOREQ_BUFFER' returns error. make sense?

lijinxia:
verified +1

PR #96, opened 2018-04-11
Add range check for 'vm0_gpa' before creating EPT

yonghuah:
Input('vm0_gpa') should be validated when handling
hypercall 'HC_VM_SET_MEMMAP to create EPT for UOS
in 'hcall_set_vm_memmap()' function.
at least 'hpa' range (from 'vm0_gpa') can't overlap
hypervisor memory region.


lijinxia:
verified +1

PR #98, opened 2018-04-11
hv: Remove hpet from acrn.conf

mingqiangchi:
Remove clocksource=hpet from SOS kernel cmdline, as ACRN is providing
tsc and hpet will not be supported in the future


lijinxia:
verified +1

PR #101, opened 2018-04-12
ptdev: bug fix on operating list

zhenggen:
Before using a node of list, initialize it.

### DIFF
--- a/arch/x86/assign.c
+++ b/arch/x86/assign.c
@@ -166,6 +166,10 @@ alloc_entry(struct vm *vm, enum ptdev_intr_type type)
 	ASSERT(entry, "alloc memory failed");
 	entry->type = type;
 	entry->vm = vm;
+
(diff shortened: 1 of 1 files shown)


arch/x86/assign.c
```diff
@@ -166,6 +166,10 @@ alloc_entry(struct vm *vm, enum ptdev_intr_type type)
 	ASSERT(entry, "alloc memory failed");
 	entry->type = type;
 	entry->vm = vm;
+
+	INIT_LIST_HEAD(&entry->softirq_node);
+	INIT_LIST_HEAD(&entry->entry_node);
```
dongyaozu: entry->entry_node is added in the next code, do we need to INIT here ?

zhenggen: Per the standard operating, it better and safe to call INIT_LIST_HEAD before use list head.

lijinxia:
verified +1

PR #103, opened 2018-04-12
VMX: bug fix on operating vmx

zhenggen:
Switch all the referenced virtual address to physical address
include ept mapping, vmcs field, vmxon, vmclear, and vmptrld.

Reviewed-by: Chen, Jason Cl <jason.cj.chen@intel.com>
Reviewed-by: Yakui, Zhao <yakui.zhao@intel.com>

### DIFF
--- a/arch/x86/vmx.c
+++ b/arch/x86/vmx.c
@@ -98,17 +98,18 @@ int exec_vmxon_instr(void)
 	uint64_t tmp64;
 	uint32_t tmp32;
 	int ret_val = -EINVAL;
-	void *vmxon_region;
+	void *vmxon_region_va;
+	uint64_t vmxon_region_pa;
 
 	/* Allocate page aligned memory for VMXON region */
-	vmxon_region = alloc_page();
+	vmxon_region_va = alloc_page();
 
-	if (vmxon_region != 0) {
+	if (vmxon_region_va != 0) {
 		/* Initialize vmxon page with revision id from IA32 VMX BASIC
 		 * MSR
 		 */
 		tmp32 = msr_read(MSR_IA32_VMX_BASIC);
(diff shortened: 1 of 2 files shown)


arch/x86/vmx.c
```diff
@@ -1314,11 +1317,12 @@ int init_vmcs(struct vcpu *vcpu)
 	memcpy_s((void *) vcpu->arch_vcpu.vmcs, 4, &vmx_rev_id, 4);
 
 	/* Execute VMCLEAR on current VMCS */
-	status = exec_vmclear((void *)&vcpu->arch_vcpu.vmcs);
+	vmcs_pa = HVA2HPA((uint64_t)vcpu->arch_vcpu.vmcs);
```
dongyaozu: Why uint64_t here?  HVA2HPA takes va as parameter, it is good enough

yakuizhao: @dongyaozu , the function of exec_vmclear is define as: exec_vmclear(void *)
So one uint64_t variable is defined and assigned by HVA2HPA. Then exec_vmclear(&var) is called.

dongyaozu: But, it doesn't need the uint64_t inside (uint64_t)vcpu->arch_vcpu.vmcs)..

yakuizhao: @dongyaozu You are right. The uint64_t in HVA2HPA((uint64_t)vcpu->arch_vcpu.vmcs) is redundant. 
It seems that the one uint64_t is added in the macro of HVA2HPA.
So  Gen can remove it.

zhenggen: I will remove the redundant uint64_t.

lijinxia:
verified +1

PR #104, opened 2018-04-12
MMU: bug fix on operating va <=> pa convertion

zhenggen:
1. Before referencing to physical address of devs include lapic, ioapic,
and uart, switch to virtual address.
2. Before referencing to phisical address of acpi tables, switch it to
virtual address.
3. Before pasing cmdline, switch the phisical address of param to virtaul
address.
4. Use a phisical address of pml4 to write CR3.


include/lib/types.h
```diff
@@ -58,8 +58,7 @@ typedef unsigned int uint32_t;
 typedef unsigned long uint64_t;
 typedef signed long int64_t;
 typedef unsigned int size_t;
-typedef unsigned long mmio_addr_t;
-typedef unsigned long vaddr_t;
+typedef void* vaddr_t;
```
yakuizhao: Can we also define one mmio_addr_t like vaddr_t?
typedef void * mmio_addr_t

include/arch/x86/io.h
```diff
@@ -180,7 +180,7 @@ int dm_emulate_pio_post(struct vcpu *vcpu);
  *  @param value The 32 bit value to write.
  *  @param addr The memory address to write to.
  */
-static inline void mmio_write_long(uint32_t value, mmio_addr_t addr)
+static inline void mmio_write_long(uint32_t value, vaddr_t addr)
```
yakuizhao: Can we also redefine the mmio_addr_t type like vaddr_t so that the mmio_addr_t is used in io.h?
typedef void * mmio_addr_t

zhenggen:
I will split this patch and apply paddr_t type for physical address.

PR #107, opened 2018-04-12
spinlock: bug fix in spinlock code

zhenggen:
Using ebx will truncate the high 32bit part of 64bit virtual address.
So using rbx instead of ebx.

Reviewed-by: Yakui, Zhao <yakui.zhao@intel.com>


lijinxia:
verified +1

PR #108, opened 2018-04-12
[acrn-dev] [PATCH] VMX: change PAT register default value

feijiang1:
This is workaround patch to resolve Disti performance issue.
In kernel 4.14, PAT is skipped to initialize if MTRR is not enabled,
while graphics driver need set WC to GGTT memory to accelerate memcpy,
if PAT is not initialized, default PAT register will treat UC- as
uncacheable, which will impact gfx performance. Change PAT default
register value to treat UC- as WC to workaroud this problem.

Revert me when PAT/MTRR strong correlation is removed in kernel.


lijinxia:
verified +1

PR #111, opened 2018-04-16
cpu cpuid/msr feature/capability detect API refine

lifeix:
use cpu_has_cap for cpuid related feature/capability detect
use cpu_has_vmx_ept_cap/cpu_has_vmx_vpid_cap for vmx feature/capability detect


lijinxia:
verified +1

PR #112, opened 2018-04-16
rename vmexit handlers

yonghuah:
this patch is to align the suffix of the handlers' name.


lijinxia:
verified +1

PR #113, opened 2018-04-16
revise type of 'exit_reason' and 'inst_len' in vcpu_arch

yonghuah:
SDM 24.9.1 Volume3:
 - 'Exit reason' field in VMCS is 32 bits.

 SDM 24.9.4 in Volume3
 - 'VM-exit instruction length' field
   in VMCS is 32 bits.

 This patch is to redefine the data types of above fields
 in 'struct vcpu_arch' and udpate the code using these
 two fields.


lijinxia:
verified +1

PR #114, opened 2018-04-17
Bug fix on address convertion

zhenggen:
a, Change the address arguments type of mmio access api from uint64_t
   to void*.
b, before referencing to physical address of devs such as lapic, ioapic,
   vtd, and uart, switch to virtual address.
c, before referencing to phisical address of acpi tables, switch it to
   virtual address.
d, before pasing commandline from boot param, switch the phisical address
   of param to virtaul address.


lijinxia:
verified +1

PR #116, opened 2018-04-17
hv: save early hv logmsg before sos booted

lyan3:
Add percpu earlylog sbuf within hypervisor memory space to accommodate log
before sos booting, and copy the earlylog to SOS allocated sbuf at the first
do_logmsg after SOS sets the sbuf during booting.

With the patch, earlylog of hypervisor could be captured with acrnlog
service if SOS booted. And it may help debugging hv issue if SOS failed to
boot, with "logdump" command.

Reviewed-by: Zhao Yakui <yakui.zhao@intel.com>


lijinxia:
verified +1

PR #119, opened 2018-04-18
enable rdtscp instruction for guest OS all vCPUs

mcao6:
before just AP can run "rdtscp" intruction, if run it on BSP,
it will cause "illegal instruction"; now align BSP & AP.
also remove duplicated code.


lijinxia:
verified +1

PR #121, opened 2018-04-19
hv: bug fix on synchronizing with APs

zhenggen:
Using eax will truncate the high 32bit part of 64bit virtual address.
And the type of sync is unsigned long, so using rbx instead of ebx.

Reviewed-by: Zhao Yakui <yakui.zhao@intel.com>

### DIFF
--- a/arch/x86/cpu.c
+++ b/arch/x86/cpu.c
@@ -732,11 +732,11 @@ static void print_hv_banner(void)
 
 static void pcpu_sync_sleep(unsigned long *sync, int mask_bit)
 {
-	int wake_sync = (1 << mask_bit);
+	uint64_t wake_sync = (1UL << mask_bit);
 
 	if (get_monitor_cap()) {
 		/* Wait for the event to be set using monitor/mwait */
-		asm volatile ("1: cmpl      %%ebx,(%%eax)\n"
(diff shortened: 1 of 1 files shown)


arch/x86/cpu.c
```diff
@@ -732,11 +732,11 @@ static void print_hv_banner(void)
 
 static void pcpu_sync_sleep(unsigned long *sync, int mask_bit)
 {
-	int wake_sync = (1 << mask_bit);
+	uint64_t wake_sync = (1UL << mask_bit);
```
dongyaozu: Whether we need a 64 bits comparison or 32 bits comparison is another story.  Do we need 64 bits?
@yakuizhao @zhenggen

yakuizhao: @dongyaozu , Based on the caller of pcpu_sync_sleep, I think that either of 32-bits and 64-bits are OK.
The pcpu_sync is used as the input of pcpu_sync_sleep.(It is unsigned long type). And only 0 is passed as mask_bit.
After using uint64_t, it can assure that wake_sync and sync are based on the same data type. (If so, %%rbx should be used).
Of course based on the mask_bit input, it is no problem that 32-bits are used. (If so, %%ebx should be used). 

In the updated patch Gen uses the uint64_t type and %%rbx data type. I think that it is also ok.

PR #123, opened 2018-04-20
hv: add hardware_detect_support to detect which hardware we support

lifeix:
Now just add some basic feature/capability detect (not all). Vapic
didn't add here for if we must support vapic then the code which
for vapic not supported must remove, like mmio apic r/w.


lijinxia:
verified +1

PR #125, opened 2018-04-23
[v2] update 'I/O instruction' VM-exit handler for security

yonghuah:
- add port range check to avoid handling requests
  that port addresses across boundary.

- update function 'dm_emulate_pio_pre()', it is not
  necessary to parse 'exit_qual' one more time.

- v1- v2 update: remove some unnecessary changes and rename
  local viarable


yonghuah:
will review it in maillist

PR #126, opened 2018-04-24
UEFI: bug fix on redundant waking up APs

zhenggen:
Due to redundant waking up APs When rebooting UOS, the crash occurs
and fail to reboot UOS.


lijinxia:
verified+1

PR #132, opened 2018-04-26
ptdev: change remapping entry from virtual to physical based

JasonChenCJ:
currently, pass-thru devices are managed by per-vm's remapping entries
which is virtual based:
- MSI entry is identified by virt_bdf+msix_index
- INTx entry is identified by virt_pin+vpin_src
it works but it's not a good design for physical resource management, for
example a physical IOAPIC pin could belong to different vm's INTx entries,
the Device Model then must make sure there is no resource conflict from
application's level.

This patch change the design from virtual to physical based:
- MSI entry is identified by phys_bdf+msix_index
- INTx entry is identified by phys_pin
The physical resource is directly managed in hypervisor, a miss adding
entry will be found by hypervisor and return error message with failure.

Acked-by: Xu, Anthony <anthony.xu@intel.com>

### DIFF
--- a/arch/x86/assign.c
+++ b/arch/x86/assign.c
@@ -35,14 +35,17 @@
 #include <hv_debug.h>
 
 #define ACTIVE_FLAG 0x1 /* any non zero should be okay */
-#define INTX_ID_IOAPIC	(0 << 23)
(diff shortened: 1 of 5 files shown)


lijinxia:
verified+1

PR #134, opened 2018-04-28
HV: Guest Cstate control support

jsun26intel:
Like guest pstate control, guest must have intel_idle driver or acpi-idle
driver enabled in kernel to enable guest cstate control. Especialy for
acpi-idle subsystem, idle space info is needed in ACPI FACP or DSDT table.

Acrn Cx enabling is similar with Px enabling implementation:

	1. Hard code the host Cx data in HV;
	2. Host load the host Cx data and store them to boot_cpu_data
	   during boot;
	3. Each VM has a Cx data copy saved in its pminfo when create VM;
	4. Add a VHM service in kernel that could interact with HV and DM
	   on data transtion. HV would expose native cx data to VHM/DM
	   without validation, because guest OS would check it before
	   enter idle;
	5. Before launching UOS, DM query HV the supported Cx data;
	6. DM generate the ACPI CST objects for UOS with these Cx data;
	7. For CPU partioned HV, we don't need to trap on idle operation,
	   when CPU need to be shared in the future, we will enable the
	   idle trap then;

change log:

	v4: 1) Change term of cx_entry_cnt to cx_cnt to keep consistence
	       with px_cnt;
	    2) Disable idle vmexit handler for now;
	    3) Add cx_data of APL NUC;

	v3:
	    1) Change term of Cx to Cx entry to make concept clear;
	    2) Initialize vm cx data entry array size from MAX_CX_ENTRY
	       to MAX_CSTATE, i.e. treat cx_data[0] as a dummy place holder
	       representing C0, then cx_data[idx] representing indexed Cx
	       entry;
	    3) VHM hypercall transit native cx data, so that guest could
	       execute idle instruction natively without vm exit which
	       impact performance;
	    4) Remove idle lantency check in trap, just favor decisions
	       from guest OS.
	    
	v2: VHM hypercall should transit virtualized cx data, not original
	    host cx data from VM pminfo;

	v1: initial version;

### DIFF
--- a/arch/x86/cpu_state_tbl.c
+++ b/arch/x86/cpu_state_tbl.c
@@ -73,6 +73,13 @@ struct cpu_px_data px_j3455[] = {
 	{0x320, 0, 0xA, 0xA, 0x0800, 0x0800}  /* P8 */
 };
 
+/* The table includes cpu cx info of Intel J3455 SoC */
+struct cpu_cx_data cx_j3455[] = {
+	{{SPACE_FFixedHW, 0x1, 0x2, 0x1, 0x01}, 0x1, 0x1, 0x3E8}, /* C1 */
+	{{SPACE_FFixedHW, 0x1, 0x2, 0x1, 0x21}, 0x2, 0x32, 0x0A}, /* C2 */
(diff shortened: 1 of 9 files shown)


lijinxia:
verified+1

PR #140, opened 2018-05-03
security: fix issues reported by Klocwork

yonghuah:
- NULL pointer reference risk
 - buffer overflow risk


yonghuah:
No, no because format, but reference a NULL pointer, e.g.
if (!entry) {  /*entry is NULL */
 pr_err(...entry->msi.msix_entry_index,  /* We can't use "entry->...", as entry is NULL in this case.*/
-				entry->vm->attr.id); /*entry is NULL  in this case */
}

@dongyaozu

lijinxia:
verified+1

PR #141, opened 2018-05-03
hv: cpuid: don't reference the crossed array

lifeix:
We should consider the boundary condition although we didn't
access it.

@yonghuah


lijinxia:
verified+1

PR #143, opened 2018-05-05
Add canonical `Dockerfile` and `docker-compose.yml`

shrmrf:
To get users up and running quickly, I propose adding a canonical `docker-compose` command.

## Prerequisites 
[`docker-compose`](https://docs.docker.com/compose/install/) and `docker` set up.

## Usage

```bash
docker-compose run [--rm] acrn-make <ARGS>
<ARGS>: arguments to the make command
```

```bash
docker-compose run --rm acrn-make 
```
Defaults to running `make all` on the local directory.

```bash
docker-compose run --rm acrn-make clean
```
Runs `make clean` on the local directory.

### `acrn-make` alias
`--rm` flag ensures that there are no zombie containers that aren't cleaned up by the `docker` daemon. A useful alias could be defined:-
```bash
alias acrn-make='docker-compose run --rm acrn-make'
```

Then `acrn-make <ARGS>` applies simply.


NanlinXie:

gvancuts:
I like what you've done here! This is taking the original idea we had (and implemented via Dockerfiles in `./travis-dockerfiles`), making it even more simple for a first-time user to use.

A few comments after playing with this tonight:
* It works just fine on MacOS (Docker 18.03.0-ce and compose 1.20.1)
* I checked that we could also pass variables such as `PLATFORM=uefi` and it worked :)
* The `--rm` trick does not work for me (did it work for you?)
* We should change the Dockerfile to use Clearlinux as the base layer since this is the primary OS we use for building the hypervisor)
* Last but not least, if we go down this path, we should also provide a similar mechanism for the acrn-devicemodel

shrmrf:
@gvancuts comments inline:-

>I checked that we could also pass variables such as PLATFORM=uefi and it worked :) 

Cool.. the idea was to provide a tool that can be used just like `make` with all dependencies built-in.

>The `--rm` trick does not work for me (did it work for you?)

Yes, it does. Are you using a Mac? I tested it on Windows and Linux. For Windows you need to define an environment variable in the `.env` file.. something might be required for Macs as well. The `.env` file is useful for passing `http_proxy` variables as well for corporate settings.

>We should change the Dockerfile to use Clearlinux as the base layer since this is the primary OS we use for building the hypervisor)

The only reason I chose Ubuntu was that it's understandable by the lay developers and the dependencies are easy to read in the `Dockerfile`. On clearlinux, one will have to go into the bundle definition. But if you think that Clearlinux is better, we can change the pull request, sure.

>Last but not least, if we go down this path, we should also provide a similar mechanism for the acrn-devicemodel

Sure!!

shrmrf:
Got some time to test with `http_proxy` behind a corporate proxy.

shrmrf:
@gvancuts where can I check the build logs? Any idea on why the build failed??

gvancuts:
I think @NanlinXie may be able to provide more insights.

NanlinXie:
We manually check this PR pass.  Previous failed is because our NUC deployment is upgraded due to patch "Make the Service OS bootloader configurable" .

NanlinXie:

gvancuts:
This Pull Request needs to be rebased now that we've combined all three repositories into one. I would suggest that we wait for [this patch](https://lists.projectacrn.org/g/acrn-dev/message/3050) to land before rebasing as it will make things easier.

shrmrf:
So now only three `make` commands exist? (`make`, `make hv`, `make dm` - right?)

I will change the default to `make` we don't need the `CMD` line then in the `Dockerfile`. (I only had it there because of muscle memory anyways)

gvancuts:
> So now only three make commands exist? (`make`, `make hv`, `make dm` - right?)

Not quite yet but that will be the case when the patch I mentionned earlier is merged (the review needs to be finalized on the mailing list and then the patch sent as a PR). You could still do the job manually by entering the various folders and calling make in there, but this path will make it easier. I suspect we'll also have a third invocation of `make` shortly, to build the tools as these were just moved to a top-level folder (see https://github.com/projectacrn/acrn-hypervisor/commit/26e7b260b1b8887e022512695aac747846505171). It would make sense to me to have the following in the end:
* `make`: build *all* targets (i.e. `hv`, `dm and `tools`)
* `make hv`: build the hypervisor
* `make dm`: build the devicemodel
* `make tools`: build the tools

> I will change the default to make we don't need the CMD line then in the Dockerfile. (I only had it there because of muscle memory anyways)

:-)

shrmrf:
Got to know through a contact today about the reasons why Intel wants to do everything with Clearlinux. I think it all makes sense. I'll also modify my PR to use a dockerfile based on Clearlinux instead of Ubuntu. 

I think Travis builds (in parallel to the Jenkins stuff) makes a lot of sense so everything is tested in the open as well and clearly visible on the project README.

gvancuts:
> Got to know through a contact today about the reasons why Intel wants to do everything with Clearlinux. I think it all makes sense. I'll also modify my PR to use a dockerfile based on Clearlinux instead of Ubuntu.

Thanks!

> I think Travis builds (in parallel to the Jenkins stuff) makes a lot of sense so everything is tested in the open as well and clearly visible on the project README.

I agree it is highly desirable to have system that's completely visible in the open. We had just enabled anonymous access to Jenkins so anyone could check the status of the testing. I have not, however, looked at exactly what tests are being performed and I would also support using Travis if it is complementary to what Jenkins does today.

NanlinXie:
@gvancuts Currently our CI includes build and NUC deploy test, we will perform SOS/UOS boot up test for each PR. The log could be checked through Jenkins console output. 
@shrmrf We have also investigated Travis, it is easy to create build test but not including deploy test on NUC. Thus currently we choose Jenkins to implement it.  But we would also support to do more on Travis if Travis could do more things.

PR #147, opened 2018-05-07
rename 'req_buf' field in 'struct vm_sw_info'

yonghuah:
- rename it to 'io_shared_page' to keep consistent
   with ACRN HDL foils.

 - update related code that reference this data structure.


lijinxia:
verified+1

PR #149, opened 2018-05-07
HV: series of patches for making AP trampoline code relocatable

ZideChen0:


lijinxia:
Verified -1
SOS boot failed on MRB

PR #156, opened 2018-05-08
HV: correct the handling of MU_MEM_ATTR_READ MMU flag

ZideChen0:


ZideChen0:
I have updated the commit message. Besides to that, combined with PR #157, it will raise page fault if writing to read-only pages, which is useful for debugging.

lijinxia:
verified +1

PR #157, opened 2018-05-08
HV: Enable CR0.WP

ZideChen0:
Page fault could be raised if writing to read-only pages. This is
useful for debugging.


lijinxia:
verified +1

PR #160, opened 2018-05-10
Vlapic timer: refine vlapic timer after revisit hv timer

lifeix:


NanlinXie:

lijinxia:
verified+1

PR #162, opened 2018-05-11
cleanup port i/o vmexit handler

yonghuah:
- print error in case i/o range spans devices

- remove test for  'io_read' & 'io_write', as they
   never be NULL as default handlers are installed
   if no handler is set in register function.

- remove ASSERT if errors happen to avoid hv crash
  in error case.


NanlinXie:

PR #163, opened 2018-05-11
cleanup: check the validity of I/O handlers to be installed

yonghuah:
- emulated I/O device driver must install valid handlers
  for both IO read & write

### DIFF
--- a/arch/x86/io.c
+++ b/arch/x86/io.c
@@ -222,21 +222,6 @@ static void deny_guest_io_access(struct vm *vm, uint32_t address, uint32_t nbyte
 	}
 }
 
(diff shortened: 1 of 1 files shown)


bzhu5:
why you removed the default r/w IO handler in this patch? You didn't explain this in your commit message.

yonghuah:
default I/O handler will operate on physical i/o port,  which may not be expected by emulated device driver,   driver should register its own i/o handler if it  really need.  if so , NULL i/o handlers are not allowed for device drivers. @bzhu5 @dongyaozu

NanlinXie:

PR #173, opened 2018-05-14
Move ACRN tools code directory one level higher

lyan3:
The tools directory is moved out of ./devicemodle, to be in parallel with
hypervisor, devicemodel and doc.


NanlinXie:

PR #174, opened 2018-05-14
Optimize patch

yakuizhao:
This is one patch set that tries to enable the -O2 option.
The patch[1..5] are used to fix the issues when trying
to enable -O2 option.
The patch6 is used to enable the -O2 option.


NanlinXie:

lijinxia:
verified+1

PR #180, opened 2018-05-15
remove potential using default i/o handler

yonghuah:
- device driver should register valid i/o handlers
  in any cases, avoid referencing to default handler

- remove i/o handler test code as they shall
  never be NULL.


NanlinXie:

lijinxia:
verified+1

PR #181, opened 2018-05-15
add a makefile under project root directory

mcao6:
it can be used as following:
  make ==> to make hypervisor & devicemodel & tools
also modify tools makefile to support out-dir setting.

Reviewed-by: Jack Ren <jack.ren@intel.com>


NanlinXie:

lijinxia:
verified +1

PR #192, opened 2018-05-15
HV: automatically generate acrn (hypervisor) version from git

gvancuts:
This patch changes how the acrn hypervisor version is set.
Previously, a "version.h" was constructed that included a number
of macros that were used to build up the version number. With
this patch, the version is directly taken from git (using
'git describe').

The most recent tag is used as the baseline for the version and
more information is automatically added (such as number of commits
on top of that tag, whether the working tree is dirty, etc.). For
more information on the exact format, see 'man git describe'

### DIFF
--- a/hypervisor/Makefile
+++ b/hypervisor/Makefile
@@ -2,9 +2,8 @@
 # ACRN Hypervisor
 #
 
-MAJOR_VERSION=0
-MINOR_VERSION=1
-RC_VERSION=4
+HV_VERSION := $(shell git --no-pager describe --tags --always --dirty)
+HV_BUILD_TIME := $(shell date -Iseconds)
 
 API_MAJOR_VERSION=1
 API_MINOR_VERSION=0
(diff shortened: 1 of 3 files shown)


dongyaozu:
@jren1 @gvancuts  I think we need a way to manage version/build well. So far, we defined something, but not well maintained. We have VERSION, MAJOR_VERSION/MINOR_VERSION, API_MAJOR_VERSION/API_MINOR_VERSION, HV_MAJOR_VERSION/HV_MINOR_VERSION, HV_API_MAJOR_VERSION/HV_API_MINOR_VERSION, HV_BUILD_VERSION and many other versions.  We need to think of an overall solution for this.  Can you two propose?
Thanks!

dongyaozu:
@gvancuts @jren1  so, what is the conclusion? Do we need this patch? if not, can we drop it?

gvancuts:
Hi @dongyaozu , I think we need something but not this patch since @jren1 is not happy with it. I need to follow-up on your request to come up with a proposal. I have some ideas but still need to put them down in an email and share with @jren1 and the community (was thinking of sending that as a "Request For Comment" email on the `acrn-dev` mailing list.

I'll close this PR in the meantime (and #193 as well).

PR #193, opened 2018-05-15
DM: automatically generate acrn-dm version from git

gvancuts:
This patch changes how the 'acrn-dm' version is set. Previously,
a "version.h" was constructed that included a number of macros
that were used to build up the version number. With this patch,
the version is directly taken from git (using 'git describe').

The most recent tag is used as the baseline for the version and
more information is automatically added (such as number of commits
on top of that tag, whether the working tree is dirty, etc.). For
more information on the exact format, see 'man git describe'

### DIFF
--- a/devicemodel/Makefile
+++ b/devicemodel/Makefile
@@ -1,11 +1,11 @@
 #
 # ACRN-DM
 #
-MAJOR_VERSION=0
-MINOR_VERSION=1
-RC_VERSION=4
 BASEDIR := $(shell pwd)
 DM_OBJDIR ?= $(CURDIR)/build
+
+DM_VERSION := $(shell git --no-pager describe --tags --always --dirty)
+DM_BUILD_TIME := $(shell date -Iseconds)
 
 CC ?= gcc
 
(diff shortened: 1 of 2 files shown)


gvancuts:
Hi @jren1 . As I explained on the mailing list, I disagree with you :) See the long explanation [here](https://lists.projectacrn.org/g/acrn-dev/message/3287) and more below (I'm still hoping to convince you ;)

The current scheme does not tell us too much about the version a customer may be using. The commit ID will likely not be something we know anything about and all we are left is the version info that is currently set in the Makefile. That version could easily be changed by the customer pseudo-randomly (and their versioning scheme could easily conflict with ours at some point). Since the commit ID is (most likely) meaningless to us (because it only exists in the customer tree), all we know (assuming the version hasn't been tampered with) is the version they are running. That version in fact maps to *many* commits in our tree. As an example, if a customer is running version 0.1-rc2, he could be on any of *72* commits.

With the versioning scheme I am proposing, we know *for a fact* what the closest tag (hence version) is. `git describe` will also tell us how many commits are included since that tag (a piece of information that's not available with the current scheme). The situation regarding the actual commit the code is based off is the same between the current scheme and my proposal (i.e. if it matches one of the commit ID in our tree, it's fine, if not, then we have no clue of what it includes regardless).

You have a good point though that *if* they do not use `git`, then we loose the version number altogether... let me think about that some more, perhaps that drawback outweighs the benefits...

jren1:
Hi Geoffroy, as I said in the mail list, the git commit has been part of the versioning:

For example:
ACRN Hypervisor
HV version 0.1-rc4-20180503-b2a5dab build by root, start time 678656us
API version 1.0

b2a5dab is the git commit ID already.

gvancuts:
> Hi Geoffroy, as I said in the mail list, the git commit has been part of the versioning:

Yes, no argument there :)

My point was that both mechanisms are identical from that point of view.

gvancuts:
As per the closing comment in PR #192 , we will come up with a proposal and discuss it on the mailing list. Closing this PR in the meantime.

PR #198, opened 2018-05-16
HV: refine copy_from_vm/copy_to_vm

YadongQi:
If the target is an array, then only the first element
will be copied.
So replace structure assignment with memcpy_s().


NanlinXie:

PR #201, opened 2018-05-16
Uefi boot

JasonChenCJ:


ttzeng:
@lijinxia, how did you verify the HugeTLB function? and, on which platform? Since NUC6CAYH is our [supported hardware platform](https://projectacrn.github.io/latest/hardware.html), I suppose the validation had been done on the NUC6CAYH NUC. If that is the case, would you share the configuration you used for the verification? Thanks.

JasonChenCJ:
@ttzeng , you can setup test based on https://github.com/projectacrn/acrn-hypervisor/blob/master/doc/primer/index.rst, there is a chapter "UEFI boot" to introduce how to boot with UEFI which including hugetlb configuration.

ttzeng:
@JasonChenCJ I already did, the CMA approach works fine, but after changing to the HugeTLB approach (removed the cma kernel command line), I can't get neither 1G nor 2M huge TLB work on my APL NUC. For 1G page reservation, I had tried hugepages=1, 2, 4, neither setting can boot the UOS. For 2M page reservation, I only tried nr_hugepages=512, 1024, neither work. So I am just wondering any BKC you or @lijinxia could share? Thanks.

JasonChenCJ:
@ttzeng , you nee update your SOS kernel as well.

ttzeng:
@JasonChenCJ , what's the minimum version of SOS kernel is required? I am using 39-31 SOS kernel.

PR #204, opened 2018-05-16
improve Makefiles from new strucutre

miguelinux:
Use the -C parameter from make command to load the makefiles inside subdirecotories instead of changing directory using "cd" command.

Also with the new structure of acrn-hyperviso the install target from the make file is missing.
The install target was added to the makefiles from the tools subdirectory. 

The install target is an essential target to the Linux Distributions.


Makefile
```diff
@@ -13,23 +13,18 @@ TOOLS_OUT := $(ROOT_OUT)/tools
 all: hypervisor devicemodel tools
 
 hypervisor:
-	cd $(T)/hypervisor; \
-	make HV_OBJDIR=$(HV_OUT) PLATFORM=$(PLAT) RELEASE=$(RELEASE) clean; \
-	make HV_OBJDIR=$(HV_OUT) PLATFORM=$(PLAT) RELEASE=$(RELEASE)
+	make -C $(T)/hypervisor HV_OBJDIR=$(HV_OUT) PLATFORM=$(PLAT) RELEASE=$(RELEASE) clean; \
```
anthonyzxu: good idea to use "-C".  can we remove "; \" at the end of the line?

jren1: can't remove ';' because there is another line following '\' after that

anthonyzxu: from below , seems "; \" are not needed.


+	make -C $(T)/tools/acrnlog OUT_DIR=$(TOOLS_OUT) install
+	make -C $(T)/tools/acrn-manager OUT_DIR=$(TOOLS_OUT) install
+	make -C $(T)/tools/acrntrace OUT_DIR=$(TOOLS_OUT) install

Makefile
```diff
@@ -13,27 +13,34 @@ TOOLS_OUT := $(ROOT_OUT)/tools
...
+.PHONY: install
+install: hypervisor-install devicemodel-install tools-intall
 
+hypervisor-install:
+	make -C $(T)/hypervisor HV_OBJDIR=$(HV_OUT) PLATFORM=$(PLAT) RELEASE=$(RELEASE) install
+
+devicemodel-install:
+	make -C $(T)/devicemodel DM_OBJDIR=$(DM_OUT) install
+
+tools-intall:
```
anthonyzxu: tools-intall  -> tools-install

Makefile
jren1: Then need to remove both ";" and "\" together. 
Currently after this patch:
hypervisor:
	make -C $(T)/hypervisor HV_OBJDIR=$(HV_OUT) PLATFORM=$(PLAT) RELEASE=$(RELEASE) clean; \
	make -C $(T)/hypervisor HV_OBJDIR=$(HV_OUT) PLATFORM=$(PLAT) RELEASE=$(RELEASE)

Makefile
jren1: Good catch!

Makefile
anthonyzxu: ";\" are not needed with "make -C".
";\" are needed with "cd " because they have to be in the same statement.

please remove all unnecessary ";\" and ";"

Thanks,
Anthony

anthonyzxu: ";\\" are not needed with "make -C".
";\\" are needed with "cd " because they have to be in the same statement.

please remove all unnecessary ";\\" and ";"

anthonyzxu: finally find out a way input '\\',

miguelinux:
I just added a commit which fix the typos and styles

PR #207, opened 2018-05-17
 HV: refine EPT violation VM-exit handler

yonghuah:
- refine EPT violation vmexit handler
- add check for mmio access that spans devices


lijinxia:
verified+1

PR #208, opened 2018-05-17
doc: Getting Started Guide for UP2 board

kiendinh:
Signed-off-by: Kien Dinh <kien.t.dinh@intel.com>


NanlinXie:

ttzeng:
@kiendinh you might also add a link to the [top most page](https://github.com/projectacrn/acrn-hypervisor/blob/master/doc/index.rst), or host the doc in [`howtos/tech/`](https://github.com/ttzeng/acrn-hypervisor/tree/master/doc/howtos/tech) folder to be collected in How-Tos section group.

jren1:
@shiqingg Shiqing, please verify the steps in UP2

kiendinh:
@ttzeng Thank you for the suggestion. I've added the link to the top most page.

shiqingg:
Verified+1

dongyaozu:
@dbkinder  how is your side check? Can we merge now?

dbkinder:
Don't merge yet please.  I'll work on the doc tomorrow.

ailun258:
Verified +1
@kiendinh @dbkinder @gvancuts 
the steps in the doc is working, but please don't merge it at this moment, we can make it better to follow up by the community.
3 proposal points:
1) We would change the title to "how to boot up the ACRN on UP2 board", and keep Getting started guide for mainstream Hardware platform only that official supported as the priority 1. 
2) Remove service-so and kernel-pk installation because it may confuse people - why after installed them, people still need to build source and deploy files again. 
3) guide people to build hypervisor, device model and SOS kernel and deploy them manually

ailun258:
what i am suggesting to build SOS kernel for UP2 manually because there are some issues with kernel-pk

ailun258:
also i noticed in this doc, it mentioned to modify source code to enable serial port of the UP2 board,
can we merge this code into master branch and make it common or adaptable? @jren1  so people don't need to modify source code manually again, and if the latest source code can be integrated  into service-os bundle,  people may don't need to build source and deploy again manually after installing the service-os bundle

dbkinder:
@ailun258 I too noticed that, and was hoping there was some configuration options with the hypervisor for handling this kind of stuff (something like Kconfig) where the values for #ifdefs and values are configured outside of the code.  Having folks modify code is bad form.

kiendinh:
@ailun258 Thank you for the feedbacks.
1. Can you explain more about the causal relationship between the title and the priority? It should be up to the user to decide which board to start with, while the priority is something internal.
2. service-os package contains necessary systemd service to start the device model. As long as Clear Linux is the only choice for SOS, why not use its utility as much as possible? 
3. It's beyond the scope of this commit. Building the kernel should refer to an article of Clear Linux.

kiendinh:
@dbkinder, @ailun258, @jren1 Agreed. It should be easier if there is a compile option for different types of boards. It will be obvious choice when ACRN supports other boards from different architectures. They have something similar in Zephyr, say
$ make BOARD=up2

And it will switch to the configurations of the specific board.

gvancuts:
A patch (from @jren1 ) landed (about 5 hours ago) that allows the `uart` parameter to be passed to the hypervisor via a cmdline option - see #246.

We don't have a tag yet that includes the relevant commit (i.e. https://github.com/projectacrn/acrn-hypervisor/pull/246/commits/6c09a0e093ffde8f88398f08a5cff82102345ccc ) but this should be the case at the latest tomorrow. The `service-os` maintainers (@miguelinux and @bryteise ) can then update the bundle with that version. At that point, there is no need anymore to modify the hypervisor code, just need to adjust this line: https://github.com/projectacrn/acrn-hypervisor/pull/208/files#diff-79b7e33150037d7f2fd6d44103a49c05R120 (and remove the `bootloaderx64.efi` renaming operation which is not needed anymore).

gvancuts:
Why do we need to build the Service OS kernel manually? What is broken in the one shipping in the `kernel-pk` bundle?

gvancuts:
> @dbkinder, @ailun258, @jren1 Agreed. It should be easier if there is a compile option for different types of boards. It will be obvious choice when ACRN supports other boards from different architectures. They have something similar in Zephyr, say
> $ make BOARD=up2
> And it will switch to the configurations of the specific board.

There is something like this being proposed as we speak - see https://lists.projectacrn.org/g/acrn-dev/message/3840

dongyaozu:
@gvancuts  yes, we plan to do so. It will be very soon.  If there is no major other issues, let us merge it. We will have a big reshuffle later on when we switch to board based configuration.

jren1:
@kiendinh @dbkinder Can you rebase and refine this patch?

kiendinh:
@jren1 I just modified the PR to replace the code modification for UART with cmdline support.

gvancuts:
@dbkinder (as the owner of the documentation) has requested changes to this, we shouldn't merge documentation before this is addressed. Please do not do that anymore.

FYI, a more comprehensive set of guidelines has just been published to help people write `reST` documents: https://projectacrn.github.io/latest/contribute/doc_guidelines.html. It should make it easier for everyone to understand how to write docs so that it can be easily published on the project website.

PR #214, opened 2018-05-17
Makefile: compliant with existing conventions

ttzeng:
We have some conventions in existing code and documents, these patches are to follow those conventions:
(1) Current documents presume build the code with 'make PLATFORM=uefi'
(2) Both HV and DM invoke compiler with $(CC) for flexibility, the new tools code should follow the same practice

Signed-off-by: Tonny Tzeng tonny.tzeng@intel.com

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 
 # global helper variables
 T := $(CURDIR)
-PLAT := uefi
-RELEASE := 0
+PLATFORM ?= uefi
+RELEASE ?= 0
 
 ROOT_OUT := $(T)/build
 HV_OUT := $(ROOT_OUT)/hypervisor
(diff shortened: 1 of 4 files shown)


gvancuts:
This looks good to me! And it triggered a question in my mind (that I had already asked during the patch review on the mailing list): shouldn't we re-use the same compiler options than when compiling the devicemodel code?

I'm assuming there are specific reasons why we use the current set of [compiler options](https://github.com/projectacrn/acrn-hypervisor/blob/master/devicemodel/Makefile#L10-L43) and it would be good to understand if they should be applicable to these tools too (and if not, why).

@lyan3 @mcao6 , can you comment?

mcao6:
The tools related need "Yan, Like" to confirm.

lyan3:
it looks good to me. 
For the compiler options, the configuration tool for acrn project is under discussion. we will follow up to the same options and tool. @gvancuts

lijinxia:
Verified+1

PR #220, opened 2018-05-18
DM: add spanning devices check for i/o access

yonghuah:
- i/o access whose handler is not registered should
  not be allowed.


lijinxia:
Verified+1

PR #229, opened 2018-05-21
exception: only trap #MC

JasonChenCJ:
HV only print out message for #MC then inject it back to guest.


dongyaozu:
@JasonChenCJ why check fails?

JasonChenCJ:
@dongyaozu seemed it's Jekins issue, @lijinxia could you help check?

lijinxia:
Verified +1

PR #230, opened 2018-05-21
DM: fix incorrectly check buffer boundaries issues

yonghuah:
- fix issues report by Klocwork that function 'strncpy'
  may incorrectly check buffer boundaries and may
  overflow buffer 'basl_stemplate' & 'basl_temaplate'


yonghuah:
drop this solution.

PR #237, opened 2018-05-22
Revert "dm: free entries in pci_businfo[] when deinit"

ssqre:
This commit is reported to cause UOS reboot fail becasue the
pci_businfo[] only be allocated when calling pci_parse_slot
in dm initialization while UOS reboot will not allocate again.
So we can't free it here.

This reverts commit 7aaff68798a4ba3564e4d913678eb84c610dad6c.


lijinxia:
Verified +1

PR #238, opened 2018-05-22
DM:fix the possible buffer overflow issue using 'strncpy'

yonghuah:
function 'strncpy' may incorrectly check buffer boundaries
and may overflow buffers.


lijinxia:
Verified +1

PR #239, opened 2018-05-22
hv:Use exception vector MACRO instead of hardcode

mingqiangchi:
Now use hardcode when inject GP/NMI to guest,
replace it with MACRO.


lijinxia:
Verified +1

PR #240, opened 2018-05-22
DM: initialize 'create_vm' before reference it in 'vm_opn'

yonghuah:


lijinxia:
verified +1

PR #242, opened 2018-05-22
DM:fix suspicious dereference of pointer in 'pci_emul_deinit()

yonghuah:
suspicious dereference of pointer 'fi->fi_devi'
  by passing it to function 'pci_emul_free_bars()'
  before  NULL check.


lijinxia:
Verified +1

PR #243, opened 2018-05-22
doc: move DM memory setup to GSG

dbkinder:
DM memory allocation mechanism is better placed in the Getting Started
material, so move it there from the primer.


doc/getting_started/index.rst
```diff
@@ -311,6 +311,42 @@ Set up Reference UOS
...
+   - For 2M huge page reservation, after the SOS starts up, run the
+     command::
+
+        echo reserved_pg_num > /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages
+
+   .. note::
+      You can use 2M reserving method to do reservation for 1G page size, but it
+      may fail.  For an EFI platform, you may skip 1G page reservation
+      by using a 2M page, but make sure your huge page reservation size is
+      large enough for your usage.
```
gvancuts: I do not have any prior knowledge on CMA vs. HugeTLB and I'm a little confused as to which I should use after reading this.

There is a large gap in terms of the amount of memory being reserved (2M vs. 1G). Can we add a little explanation (or a link to more info) on how to determine what to choose?

What would typically cause the 1G reservation (using the 2M method) to fail?

Would it make sense to use the 1G reservation method to reserve 2M?

JasonChenCJ: @gvancuts , hugetlb is new mechanism which should be preferred as CMA (old one) has exist problems like 
- fail to reserve
- share with other clients
but currently we still keep two solutions, so user can still choose.  As 1G page need big block of continuous memory, so if after system run for a while, it's difficult to find such big memory block if the physical ram size is not so big for the platform. Normally we are easy to reserve 2M page, so we can use normal way (echo pg# > nr_hugepages).

gvancuts: Hi @JasonChenCJ , can you confirm if my understanding is correct? If so, I would like to add some text to our GSG myself. Thanks

PR #244, opened 2018-05-22
HV:initialize variables before reference in vmx.c

yonghuah:
- to avoid complains from code static scan tool


lijinxia:
Verified +1

PR #245, opened 2018-05-22
HV: refine 'decode_instruction() function

yonghuah:
update:
   1. remove 'struct mem_io *'from input arguments
   2. return 'opsize' instead of status.
   3. rename 'vmm_decode_instruction()'


lijinxia:
Verified +1

PR #254, opened 2018-05-23
DM: bugfix - use of freed memory in 'monitor_close()'

yonghuah:
-memory was dereferenced after being freed:
  object 'client' was derefenced after being freed
  by call 'vmm_client_free_res()'

 -Root cause: MACRO 'LIST_FOREACH()' dereference
  'client' for next list node after 'client'
  was freed.

### DIFF
--- a/devicemodel/core/monitor.c
+++ b/devicemodel/core/monitor.c
@@ -453,7 +453,7 @@ int monitor_init(struct vmctx *ctx)
 
 void monitor_close(void)
 {
-	struct vmm_client *client;
+	struct vmm_client *client, *pclient;
 	if (!monitor_thread)
 		return;
 	shutdown(monitor_fd, SHUT_RDWR);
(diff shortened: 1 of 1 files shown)


devicemodel/core/monitor.c
```diff
@@ -465,7 +465,9 @@ void monitor_close(void)
 	/* client buf-mem and fd may be still in use by msg-handler */
 	/* which is handled by mevent */
 	pthread_mutex_lock(&client_mutex);
-	LIST_FOREACH(client, &client_head, list) {
+	list_foreach_safe(client, &client_head, list, pclient) {
+		num_client--;
```
dongyaozu: num_client will be -- in                 vmm_client_free_res(client);   Do we need this one?

yonghuah:
here , "vmm_client_free_res() " is called, not "vmm_client_free()".  monitor_close() is called mainly in exit process and free resource. have dropped the changes for 'num_client' and updated.  thanks.    
 @dongyaozu

lijinxia:
Verified +1

PR #255, opened 2018-05-23
add platform related config and launch script

mcao6:


mcao6:
Let me check with Jack.

gvancuts:
@miguelinux , heads-up that this patch will introduce new files to be added to the `acrn-hypervisor` package.

gvancuts:
@jren1 , what else is going to land in this new `platform/` folder?

If nothing else, please move this under `devicemodel\samples` where we already have a similar set of scripts.

PR #257, opened 2018-05-23
Getting Started Guide: add build instructions for the ACRN tools

gvancuts:
This adds basic information about the new ACRN tools that have
been integrated in the the acrn-hypervisor repo (under tools/).

It also adds the build dependencies for the different development
environment that we support.

Signed-off-by: Geoffroy Van Cutsem <geoffroy.vancutsem@intel.com>

PR #225, opened 2018-05-19
Fix the logic that sets the top-level build directory

gvancuts:
The 'ROOT_OUT' variable does not allow to set the folder in
which all output files (objects, binaries, etc.) are built
and placed.

Overriding it from the command-line leads to 'build' folders
of the name set that are created in the tree eventually leading
to a failure during the build. This patch changes how the
output directory is constructed so that *all* output files
are put in there. It introduces a new variable called
'BUILD_DIR' to that effect.

It also incidentally fix the fact that 'make clean' only removes
that output directory which means it does not properly clean up
the tree if one overrides the 'ROOT_OUT' variable manually today.

To change where the build artefacts are placed, set 'BUILD_DIR'
to whatever you want, e.g.:
   $ make BUILD_DIR=build-test
Will create a top-level folder called 'build-test' where everything
will be put. 'make clean' will remove it (if called with the same
argument)

Fix #223 

Signed-off-by: Geoffroy Van Cutsem <geoffroy.vancutsem@intel.com>

PR #260, opened 2018-05-23
acrn: allow to specify the build output folder

jren1:
Usages: make O=<any folder>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,8 @@ T := $(CURDIR)
 PLATFORM ?= uefi
 RELEASE ?= 0
 
-BUILD_DIR := build
-ROOT_OUT := $(T)/$(BUILD_DIR)
+O ?= build
+ROOT_OUT := $(shell mkdir -p $(O);cd $(O);pwd)
 HV_OUT := $(ROOT_OUT)/hypervisor
 DM_OUT := $(ROOT_OUT)/devicemodel
 TOOLS_OUT := $(ROOT_OUT)/tools


gvancuts:
Hi @jren1 , what is the purpose of this patch (besides renaming the `BUILD_DIR` variable to `O` - which I find less intuitive, but I can live with that)?

Is it fixing any issue you've observed?

gvancuts:
Just to give you more context, the `BUILD_DIR` variable was introduced (by me) in PR #225 specifically to allow people to set the build directory (the logic was not in place). PR #257 (finally) adds a little text to explain how to use this (see [this line](https://github.com/projectacrn/acrn-hypervisor/pull/257/files#diff-f25f8087f145213406fc67b2c7c42cb1R397).

gvancuts:
> @gvancuts The clear Linux team complainted that they tried to build both SBL and UEFI platform, but they are over written by each other:
> Command as below (Please try):-
> make PLATFORM=uefi HV_OBJDIR=uefi RELEASE=0
> make PLATFORM=sbl HV_OBJDIR=sbl RELEASE=0
> The output still is inside build and the sbl will overwrite the uefi.
> This patch can solve that issue by simply specifying O=sbl o O=uefi in make command line.
> That patch is waiting for Clear Linux to verify whether it can solve their issue.

This is exactly what my PR (#225) was addressing, is there any reason why they cannot use `make BUILD_DIR=build-uefi` or `make BUILD_DIR=build-sbl` today?

jren1:
Got your context. but we can update the document to tell the people to use O= to put the objects files to anywhere and don't need to modify the Makefile, right?

gvancuts:
> Got your context. but we can update the document to tell the people to use O= to put the objects files to anywhere and don't need to modify the Makefile, right?

They don't need to modify the Makefile today. All they need to do is set the BUILD_DIR variable to whatever they want, e.g.:
   $ make BUILD_DIR=build-uefi PLATFORM=uefi RELEASE=0
   $ make BUILD_DIR=build-sbl PLATFORM=sbl RELEASE=0
And the corresponding build artefacts will be placed in respectively 'build-uefi' and 'build-sbl'.

What I mean is that the feature you are proposing with your patch is already there so unless it's somehow broken, I don't see why we need it. Does that clarify?

jren1:
But that still limit the output folder under acrn-hypervisor folder. For our internal E2E build, the output folder is even not in acrn-hypervisor folder.

jren1:
@miguelinux please help verify whether this patch can allow you to build both sbl and uefi platform in autospec.

jren1:
from @miguelinux ：
”
Looks like it could work.

I will do a PR to add other changes, please merge your patch.
“

PR #263, opened 2018-05-24
DM: using 'strncpy' coding style cleanup

yonghuah:
- check buffer boundaries to avoid buffer overflow

### DIFF
--- a/devicemodel/core/sw_load_bzimage.c
+++ b/devicemodel/core/sw_load_bzimage.c
@@ -131,8 +131,7 @@ acrn_parse_kernel(char *arg)
 	size_t len = strlen(arg);
 
 	if (len < STR_LEN) {
-		strncpy(kernel_path, arg, len);
-		kernel_path[len] = '\0';
+		strncpy(kernel_path, arg, len + 1);
 		if (check_image(kernel_path) != 0){
 			fprintf(stderr, "SW_LOAD: check_image failed for '%s'\n",
 				kernel_path);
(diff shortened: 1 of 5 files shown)


devicemodel/core/sw_load_bzimage.c
```diff
@@ -131,7 +131,7 @@ acrn_parse_kernel(char *arg)
 	size_t len = strlen(arg);
 
 	if (len < STR_LEN) {
-		strncpy(kernel_path, arg, len);
+		strncpy(kernel_path, arg, STR_LEN);
```
dongyaozu: That means you want to copy more bytes, do you really want? I think it is STR_LEN-len.

devicemodel/core/sw_load_bzimage.c
```diff
@@ -151,7 +151,7 @@ acrn_parse_ramdisk(char *arg)
 	size_t len = strlen(arg);
 
 	if (len < STR_LEN) {
-		strncpy(ramdisk_path, arg, len);
+		strncpy(ramdisk_path, arg, STR_LEN);
```
dongyaozu: same here and others.

devicemodel/core/sw_load_bzimage.c
yonghuah: Copy 'len' bytes only if 'arg' is properly null-terminated and copies at most 'STR_LEN' if 'arg' is not null - terminated to avoid buffer overrun.

char *strncpy(char *dest, const char *src, size_t n);
       The  strncpy()  function copies at most n bytes of src
       are copied.  If  the  length of src is less than n, strncpy() writes additional null
       bytes to dest to ensure that a total of n bytes are written.

Thanks.
@dongyaozu

dongyaozu: Yes, but len=strlen(arg_, so it is already NULL terminated.  we probably should use "n+1".  We don't want to copy too many bytes here.

yonghuah: agree with you,  "len = strlen() "excludes the terminating null bytes ('\0'), so strncpy( .. , len) will not copy  '\0' to dest.  
 Will update, thanks.
@dongyaozu

lijinxia:
Verified+1

PR #265, opened 2018-05-24
fix "warning:range expression in switch statements are not standard"

425686558:
Range expression in switch statement is in gcc extension standard(gcc
manual 6.28),not in c99 standard.
GCC manual 6.28 reference link below:
(https://gcc.gnu.org/onlinedocs/gcc-8.1.0/gcc/Case-Ranges.html#Case-Ranges)


NanlinXie:

lijinxia:
ok to test

lijinxia:
OK to verify

PR #266, opened 2018-05-24
fix "warning:ISO C forbids 'return' with expression,in function retur…

425686558:
…ning void"

returning void function must return void.


NanlinXie:

425686558:
invalid request,please ignore

PR #270, opened 2018-05-24
Configure install locations in a single place.

busykai:
Also, fix default install destinations for some files (most notably
binaries should be located in /usr/lib/acrn).

This is needed to correctly package acrn.

### DIFF
--- a/Defs.mk
+++ b/Defs.mk
@@ -1,0 +1,6 @@
+# Copyright (c) 2018 Intel Corporation. All rights reserved.
+# See LICENSE for terms.
+
+INSTALL_BINDIR := /usr/bin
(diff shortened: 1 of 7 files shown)


NanlinXie:

busykai:
i just wanted to clarify: i do not insist that you need to accept this patch as is, but it at least eliminates the hardcoded locations and moves the binaries from `/usr/share/acrn` to `/usr/lib/acrn` where they belong.

miguelinux:
+1

jren1:
Since we merged Miguel's patch, we can drop this patch?

busykai:
@jren1 you still have the install paths hardcoded all over the place. i assume you want to continue like that by dropping this patch, correct?

PR #271, opened 2018-05-24
DM: validate port range explicitly in inout.c

yonghuah:
- check the port address range explicitly
  to avoid buffer overflow.


lijinxia:
Verified+1

PR #272, opened 2018-05-24
Move APL MRB sample files (launch scripts) into a new folder

gvancuts:
Move the new scripts to be used on the APL MRB platform under
a new folder inside 'devicemodel/samples' where similar scripts
are already placed.

There are now two subfolders under 'devicemodel/samples':
* apl-mrb: for scripts, config files to be used with the APL MRB
* uefi: scripts and helper files for UEFI-based platforms (e.g.
  APL NUC)


gvancuts:
Had an offline conversation with @jren1 . More files (that are not related to the Service OS or devicemodel) will eventually land in `platform/apl-mrb` so I agreed to keep things as they are. Closing this.

PR #273, opened 2018-05-24
Platform:Change boot_args to assign one plane of pipe B and pipe C to SOS

yakuizhao:
After i915.avail_planes_per_pipe(0x0000F) is added to SOS, it will cause
that the pipe B/C is initialized without plane. In such case this pipe
can't be used by UOS, with error: primary plane disabled.
The bootarg is changed so that the pipe B/C on SOS can be initialized with
at least one plane.

Signed-off-by: Jiang, Fei <fei.jiang@intel.com>
Signed-off-by: Zhao, Yakui<yakui.zhao@intel.com>


miguelinux:
@jren1 

I did some changes at #274  which include these files
please take at look at #274

yakuizhao:
The patch is updated after the PR #274 is merged.

jren1:
Hi Yakui,
    After merging Miguel's patch, the boot_args files have been to devicemodel/samples/sbl-mrb.
   Can you rebase your patch?

PR #274, opened 2018-05-24
Move config and scripts

miguelinux:
This patch series do some changes in the file structure.

1. Move the platform apl-mrb samples to devicemodel samples directory.
2. Add the install target to the missing samples files and re-organize the
    samples directory structure to have nuc and apl-mrb samples.
3. install acrn.efi to /usr/lib
4. install acrn.32.out to /usr/lib/acrn/acrn.sbl


dbkinder:
@gvancuts @jren1 this patch broke the documentation build (it had a reference to include the sources of bridge.sh and launch_uos.sh scripts in the documentation).     We would have caught this if a doc build was part of the CI acceptance.

miguelinux:
mmm ... yes I only updated the path from the files not the references

PR #279, opened 2018-05-25
Ioc mediator：Support lifecycle emulation

liuyuan-GitHub:
This patch set implements IOC mediator lifecycle work flow that including
suspend, shutdown and resume. IOC mediator needs notify suspend/shutdown
event for PM DM once received corresponding heartbeat messages from UOS.
And VM Manager needs forward S3/S5 wakeup reason to IOC DM to resume/boot UOS.

Liu Yuan (3):
  IOC mediator: Separate wakeup reason and heartbeat
  IOC mediator: Implement state transfer framework
  IOC mediator: Implement state transfer operations

v2:
  1) Replace the term of FSM with IOC state.
  2) Refine the patches order.
  3) Embed ioc_dev in cbc_pkt to simplify data exchange.

### DIFF
--- a/devicemodel/hw/platform/ioc.c
+++ b/devicemodel/hw/platform/ioc.c
@@ -768,13 +768,32 @@ cbc_request_dequeue(struct ioc_dev *ioc, enum cbc_queue_type qtype)
 }
 
 /*
+ * Send a cbc_request to TX handler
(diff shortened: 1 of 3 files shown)


lijinxia:
Verified+1

PR #287, opened 2018-05-28
Dev/dm pt reset check

gzhai7:


lijinxia:
OK to verify

lijinxia:
Verified +1

PR #262, opened 2018-05-24
fix "obsolete use of designated initializer without '='"

425686558:
According in C99 manual 6.7.8,'=' is required.

Signed-off-by: huihuang.shi <huihuang.shi@intel.com>

PR #288, opened 2018-05-28
refomatted GET_DATA_SIZE macro

425686558:
The macro GET_DATA_SIZE's code style is not correct,
it is merged by accident,the Pull Request #262
<fix "obsolete use of designated initializer without '='"> 
have two commid id,the second commit changed the ucode.c
by mistake.

Now fixed.


lijinxia:
Verified+1

PR #289, opened 2018-05-28
hv: handle the case of empty hypervisor cmdline

jren1:
Fix a bug:
When the hypervisor cmdline is empty, the hypervisor will be
unable to boot up.


lijinxia:
Verified+1

PR #291, opened 2018-05-28
Fix unnamed struct/union fields

junjiemao1:
According to the syntax defined in C99, each struct/union field must have an
identifier. This patchset removes unnamed struct/union fields using the
following approach.

* The unnamed fields are refactored if the structure can be easily expressed in
  a C99-compatible way without semantic differences.

* Otherwise a proper name is given to each previously unnamed field.

v1 -> v2:

    * Separate changes according to the approach used for the fixes
    * Name updates:
      - intr_info -> ptdev_intr_info
      - fields_u(8|32|64) -> fields_(8|32|64)

Junjie Mao (2):
  treewide: remove unnecessary unnamed structs/unions
  HV: treewide: give names to unnamed structs/unions


NanlinXie:

lijinxia:
OK to verify

lijinxia:
Verified+1

PR #292, opened 2018-05-28
HV: debug: stop using ## __VA_ARGS__

junjiemao1:
It is an extension of GCC CPP to:

* allow omitting a variable macro argument entirely, and
* use ## __VA_ARGS__ to remove the the comma before ## __VA_ARGS__ when
  __VA_ARGS__ is empty.

The only use of ## _VA_ARGS__ is to define the pr_xxx() macros, with the first
argument being the format string and the rest the to-be-formatted arguments. The
format string is explicitly spelled out because another macro pr_fmt() is used
to add to the format string a prefix which is customizable by defining what
pr_fmt() expands to.

For C99 compliance, this patch changes the pr_xxx() macros in the following
pattern.

    - #define pr_fatal(fmt, ...)				\
    -     do_logmsg(LOG_FATAL, pr_fmt(fmt), ## __VA_ARGS__);	\
    + #define pr_fatal(...)					\
    +     do_logmsg(LOG_FATAL, pr_prefix __VA_ARGS__);		\

Reference:

* https://gcc.gnu.org/onlinedocs/gcc/Variadic-Macros.html#Variadic-Macros

### DIFF
--- a/hypervisor/arch/x86/guest/vioapic.c
+++ b/hypervisor/arch/x86/guest/vioapic.c
@@ -28,7 +28,7 @@
  * $FreeBSD$
  */
 
-#define pr_fmt(fmt)	"vioapic: " fmt
+#define pr_prefix	"vioapic: "
 
 #include <hypervisor.h>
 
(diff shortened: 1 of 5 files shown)


NanlinXie:

lijinxia:
OK to verify

lijinxia:
Verified+1

PR #300, opened 2018-05-29
fix "ISO C99 does not support '_Static_assert'"

HuiHuangShi:
_Static_assert is supported in C11 standard.
Please see N1570(C11 mannual) 6.4.1.
acked-by kevin

### DIFF
--- a/hypervisor/include/debug/assert.h
+++ b/hypervisor/include/debug/assert.h
@@ -30,6 +30,13 @@
 
 #ifndef ASSERT_H
 #define ASSERT_H
+#define __STDC_99__	199901L
+
+/* Force a compilation error if condition is false */
+#if __STDC_VERSION__ == __STDC_99__
+#define _Static_assert(expr, error)	\
+	extern char Static_Error[(expr) ? 1 : -1]
+#endif
 
 #ifdef HV_DEBUG
 void __assert(uint32_t line, const char *file, char *txt);


425686558:
This change ignore the second parameter of the macro, and according the 'expr' to expand,
if the 'expr' is not 0,it will compile successful, otherwise it will faile.

junjiemao1:
_Static_assert(expr, note) basically:

1. leads to compilation failure if expr statically evaluates to false,
2. has no effect at runtime if the code compiles, and
3. can be used both inside or outside a function.

The definition of _Static_assert introduced in this patch meets the requirements above because:

1. C99 requires that if the array size "is a constant expression, it shall have a value greater than zero." (Section 6.7.5.2 Array declarators),
2 & 3. Declarations is allowed inside or outside a function, and has no effects if the declared variable is not used.

There're some differences between the macro in this patch and the _Static_assert keyword, though, including:

1. When the expression is not a constant at compile time, _Static_assert must fail. C99 does allow array sizes to be a non-constant expression (though the behavior is undefined) though gcc reports compile errors in such cases. A better option may be defining a structure with a bit-field of negative size, since C99 explicityly requires bit-field widths to be constants.
2. Confusion arises if the declared variable (i.e. Static_Error) is used. But it is unlikely to happen because our current naming convention forbids using capitalized names.

dongyaozu:
@junjiemao1 if we want to ASSERT at runtime, we can simply use ASSERT. If we want to report the error at compile time, we can use something else...  This patch assumes we defined Static_Error  + complain on the minus index.   Not sure if this is the best solution.  @HuiHuangShi You need to get more support from others...

junjiemao1:
@dongyaozu In the current code we use "_Static_assert" for compile-time assertion, but this is a c11-only feature. For c99-compatible code using a negative size of array or bit-field is a common practice As an example, in the definition of _Static_assert in glibc a bit-field of negative width is used.

For Static_error, this patch does not assume Static_Error is defined anywhere for the trick to work.

dongyaozu:
Let us completely remove _Static_assert.  We can use ASSERT(), which is NULL in release version...

junjiemao1:
I see. @HuiHuangShi please close this PR and cook another patch replacing all _Static_assert with ASSERT? Thanks.

PR #301, opened 2018-05-29
ASSERT cleanup for SOS/UOS creation

lifeix:
v2:
1. fix cover letter comments grammar error
2. datail patch descriptions.

v1:

ASSERT will hang the whole system in debug version while doing nothing
in release version. However, in most of the case, we should not hang
the whole system: like invalid input parameter or creating an UOS. And
in release version, we shouldn't do nothing when something bad happened.

So ASSERT cleanup is necessary. This serial is just a beginning.
It will panic the system when creating SOS failed while not when create
UOS failed. It just returns error number to the caller when something
nonfatal error heppened otherwise panic when something fatal error happened.

In this serial, failure to create SOS is an fatal error.

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -367,6 +367,7 @@ static void get_cpu_name(void)
 
 void bsp_boot_init(void)
 {
+	int ret;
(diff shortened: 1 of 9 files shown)


lijinxia:
Verified+1

PR #302, opened 2018-05-29
DM sample: update sos/uos boot arg to let Android UOS see two displays

feijiang1:
Android HWC has problem when only see one display, then this is
workaround patch to make Android HDMI2 display normal.

V2: only change AaaG UOS kernel boot arg


lijinxia:
Verified+1

PR #306, opened 2018-05-29
Control register handling cleanup

binbinwu1:
Remove unused API init_cpu.
Control register interface cleanup.


lijinxia:
Verified+1

PR #311, opened 2018-05-30
Virtio 1.0 cap pci cfg

chejianj:
The VIRTIO_PCI_CAP_PCI_CFG capability creates an alternative access method to the common configuration, notification, ISR and device-specific configuration regions.

The capability is immediately followed by an additional field like so:
struct virtio_pci_cfg_cap {
        struct virtio_pci_cap cap;
            uint8_t pci_cfg_data[4]; /* Data for BAR access. */ }

To access a device region, the driver writes into the capability structure (ie. within the PCI configuration space) as follows:

- The driver sets the BAR to access by writing to cap.bar
- The driver sets the size of the access by writing 1, 2 or 4 to cap.length
- The driver sets the offset within the BAR by writing to cap.offset

At that point, pci_cfg_data will provide a window of size cap.length into the given cap.bar at offset cap.offset.

v1 --> v2:
- add check of the input p_capoff in pci_emul_find_capability
- revise the commit message of 2/3 to clarify the current issue
- add check of the return value of virtio_find_capability
- use *rv instead of reading from config space again in
  virtio_pci_modern_cfgread
- use val instead of reading from config space again in
  virtio_pci_modern_cfgwrite

### DIFF
--- a/devicemodel/include/virtio.h
+++ b/devicemodel/include/virtio.h
@@ -518,6 +518,7 @@ struct virtio_base {
 	uint8_t config_generation;	/**< configuration generation */
 	uint32_t device_feature_select;	/**< current selected device feature */
 	uint32_t driver_feature_select;	/**< current selected guest feature */
+	int cfg_coff;			/**< PCI cfg access capability offset */
 };
 
 #define	VIRTIO_BASE_LOCK(vb)					\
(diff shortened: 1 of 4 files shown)


devicemodel/include/virtio.h
```diff
@@ -873,6 +874,44 @@ void virtio_dev_error(struct virtio_base *base);
...
+ * Handle virtio PCI configuration space writes. Only the specific registers
+ * that need speical operation are handled in this callback. For others just
+ * fallback to pci core. This interface is only valid for virtio modern.
+ *
+ * @param ctx Pointer to struct vmctx representing VM context.
+ * @param vcpu VCPU ID.
+ * @param dev Pointer to struct pci_vdev which emulates a PCI device.
+ * @param coff Register offset in bytes within PCI configuration space.
+ * @param bytes Access range in bytes.
+ * @param value The value to write.
```
dbkinder: Please verify doxygen comments during review -- param name is "val" not "value"

PR #313, opened 2018-05-30
Remove current vcpu

JasonChenCJ:


NanlinXie:

dongyaozu:
@JasonChenCJ it is time to rebase now...  VPID patch is in.

hypervisor/arch/x86/trusty.c
```diff
@@ -133,9 +135,13 @@ static void create_secure_world_ept(struct vm *vm, uint64_t gpa_orig,
 	vm->sworld_control.sworld_memory.base_hpa = hpa;
 	vm->sworld_control.sworld_memory.length = size;
 
-	invept(vm->current_vcpu);
-	invept(vm0->current_vcpu);
+	foreach_vcpu(i, vm, vcpu) {
```
dongyaozu: why we need to flush EPT? with current vpid solution, we don't need to .. Not?

JasonChenCJ: @dongyaozu , there is EPT mapping change here for both UOS normal world and SOS.

dongyaozu: ok, SOS removed some mappings, so we need flush.
for UOS, we uses different EPTP, so they won't conflict with each other, why we need to flush?

JasonChenCJ: For UOS, we are also removing trusty memory from normal world. We are just using different EPTP for normal world and trusty world.

lijinxia:
Verified+1

PR #315, opened 2018-05-30
DM: Disable audio NHLT table by default

gzhai7:
NHLT table is used by some audio driver for topology data, but current
default audio driver doens't depend on it. Disable it by default due
to possible boot failure with different SOS firmware.


lijinxia:
Verified+1

PR #318, opened 2018-05-30
Support DRD capability for APL xHCI.

YangLiang3:
On Intel ApolloLake platform, there has one couple USB2/USB3 ports
shared one couple USB2/USB3 PHY with xDCI(dwc3) controller to support
dual role mode. There has one hardware logic inside xHCI controller
to support dynamic USB PHY switch between xHCI and xDCI. And two
xHCI mmio registers are used for controling this logic, which are exposed
as xHCI extended capability for software.
This patch set implements USB dual role switch for Intel ApolloLake platform.
DRD which means dual role device is used for switching xHCI and xDCI between
PHY USB2&USB3.

### DIFF
--- a/devicemodel/hw/pci/xhci.c
+++ b/devicemodel/hw/pci/xhci.c
@@ -30,11 +30,58 @@
  *  devices:
  *    tablet             USB tablet mouse
  */
+
+/*
+ * xHCI DRD control flow digram.
+ *  +---------------------------+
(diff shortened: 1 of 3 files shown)


lijinxia:
Verified+1

gvancuts:
Is this true for *all* APL platforms or really more specific to one (e.g. APL NUC vs. UP2)?

PR #325, opened 2018-05-31
ASSERT cleanup for hypercall

lifeix:
1. hypercall fail would not panic hypervisor.
2. hypervisor don't deal with hypercall fail.
3. hypervisor return the errno to SOS when hypercall failed.


lijinxia:
Verified+1

PR #330, opened 2018-05-31
ioc: add cbc_lifecycle service

alekdu:
The cbc_lifecycle service will listen on /dev/cbc-lifecycle tty port to
receive "wakeup reason" information from ioc controller as well as send
heartbeat to ioc to set different power cycle method.

The RTC funtionality is not implemented by current ioc firmware yet,
hence the service leaves it as an empty handler now.

### DIFF
--- a/misc/cbc_lifecycle/Makefile
+++ b/misc/cbc_lifecycle/Makefile
@@ -1,0 +1,12 @@
+
+OUT_DIR ?= .
+
+$(OUT_DIR)/cbc_lifecycle: cbc_lifecycle.c ../../tools/acrn-manager/acrn_mngr.h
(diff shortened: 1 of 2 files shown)


NanlinXie:

lijinxia:
OK to verify

lijinxia:
Verified+1

gvancuts:
@lijinxia how did this pass our validation?

I understood from @jren1 that the Jenkins CI system uses a specific build directory as output (e.g. `make O=/path/to/outdir`). But, as I pointed out during the review on the mailing list, you cannot use `O=` anymore as `cbc_lifecycle` will fail to compile (because it does not find `acrn-mngr` since the path is wrongly hardcoded [in the Makefile](https://github.com/projectacrn/acrn-hypervisor/blob/master/misc/cbc_lifecycle/Makefile#L5)

Can you comment?

PR #333, opened 2018-05-31
Documentation: small addition to the 'acrnlog' tool documentation

gvancuts:
Make it clear that some commands mentionned in the 'acrnlog' tool
documentation are meant to be run in the hypervisor shell and *not*
the Service OS shell (as are the other commands throughout the rest
of that document).

### DIFF
--- a/tools/acrnlog/README.rst
+++ b/tools/acrnlog/README.rst
@@ -42,12 +42,14 @@ steps:
 
       # acrnlog -n 8 -s 4 &
 
-You can use ``get_loglevel`` and ``set_loglevel`` to query and change
-the hypervisor loglevel.
+You can use ``get_loglevel`` and ``set_loglevel`` commands
+in the hypervisor shell (not the Service OS shell)
+to query and change the hypervisor log level.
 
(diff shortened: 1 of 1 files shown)


tools/acrnlog/README.rst
```diff
@@ -57,6 +57,11 @@ output to the console. For example:
    ACRN:\>get_loglevel
    console_loglevel: 2, mem_loglevel: 5
 
+.. note::
```
dbkinder: Rather than as an afterthought, it would be better to add this information earlier in the document.  Delete this addition and instead add it above on line 45:

    You can use ``get_loglevel`` and ``set_loglevel`` commands
    in the hypervisor shell (not the Service OS shell)
    to query and change the hypervisor log level.

and on line 50, just to emphasize it again, change the last sentence to:

    For example, in the hypervisor shell you can use these commands:

When we have a document to list all the shell commands we
can add a reference to that here too.

gvancuts: Just force-pushed a new version. Thanks

PR #341, opened 2018-05-31
DM: Turn on WIFI/BT passthrough for Android UOS

gzhai7:
This patch turn on WIFI/BT in launch_UOS.sh based on WIFI BDF on SBL,
which is different from ABL. Still need WIFI/BT driver in Android to
make WIFI/BT passthrough work.


lijinxia:
Verified+1

PR #343, opened 2018-06-01
fix "warning: ISO C forbids zero-size array"

HuiHuangShi:
ISO C not allowed array type whith the zero limit.
Change the array type to flexible array type.

Signed-off-by: huihuang shi <huihuang.shi@intel.com>


NanlinXie:

dongyaozu:
I saw several patches to fix this kind of similar issues. Can we check all and make one patch for all of them?

PR #347, opened 2018-06-01
fix "ISO C99 does not support '_Static_assert'"

HuiHuangShi:
Acked-by:Eddie Dong
_Static_assert is supported in C11 standard.
Please see N1570(C11 mannual) 6.4.1.
replace _Static_assert with ASSERT.

Signed-off-by: huihuang shi <huihuang.shi@intel.com>


lijinxia:
OK to verify

lijinxia:
Verified +1

PR #348, opened 2018-06-01
Enable IOC mediator by default on MRB

liuyuan-GitHub:
Update MRB launch_uos.sh script to enable IOC mediator by default.
And change the PTY path from the /tmp/ioc_$vm_name to /run/acrn/ioc_$vm_name for aligning with acrnctl.
Set one default boot wakeup reason to avoid acrn-dm boot failed with invalid wakeup reason parameter.

v2):
   1. Avoid enable invalid wakeup reason bit.
   2. Add restriction comments in check_dir function-call.


lijinxia:
Verified+1

PR #350, opened 2018-06-01
Support different vcpu modes

binbinwu1:
Note: this patch series need to work with new vSBL (start from real mode) on SBL platform!

In current code, hv only support starting vcpu from real mode or 64bit mode.
And sos/uos bsp can only start from 64bit mode. gva2gpa translation only works
on long mode.

This patch series adds support to start vcpu from protected mode. And has the
following vcpu start mode settings:
- For sbl platform:
This patch starts sos bsp from protected mode by default.
CONFIG_START_VM0_BSP_64BIT is defined to allow start sos bsp
from 64bit mode. If a config CONFIG_START_VM0_BSP_64BIT
defined in config file, then sos bsp will start from 64bit mode.
This patch start uos bsp from real mode, which needs the integration
of virtual bootloader (vsbl).
- For uefi platform:
This patch sets sos bsp vcpu mode according to the uefi context.
This patch starts uos bsp from protected mode, because vsbl is not ready
to publish for uefi platform yet. After vsbl is ready, can change to
start uos bsp from real mode.

This patch adds support to do gva2gpa translation in different paing modes.

This patch series assumes that VMX supports "unrestricted guest" feature to
support guest real mode and unpaged protected mode.
If hardware doesn't support "unrestricted guest", hardware capability check fails.

### DIFF
--- a/devicemodel/core/sw_load_bzimage.c
+++ b/devicemodel/core/sw_load_bzimage.c
@@ -314,7 +314,7 @@ acrn_sw_load_bzimage(struct vmctx *ctx)
 		if (setup_size <= 0)
 			return -1;
 		*kernel_entry_addr = (uint64_t)
-			(KERNEL_LOAD_OFF(ctx) + setup_size + 0x200);
+			(KERNEL_LOAD_OFF(ctx) + setup_size);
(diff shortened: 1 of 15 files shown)


lijinxia:
Verified +1

PR #358, opened 2018-06-04
add cbc_lifecycle service systemd support

alekdu:
Now cbc_attach service installed by default and thus we can enable
cbc_lifecycle service too.


NanlinXie:

dongyaozu:
Hold a while, since it may conflict with 355.

PR #367, opened 2018-06-06
tools: acrn-crashlog: change source of config file

chengangc:
This patch is to change the source of config file according
to the doc of telemetrics-client:

There is a config file installed at
``/usr/share/defaults/telemetrics/telemetrics.conf``
To make modifications, copy that file to
``/etc/telemetrics/telemetrics.conf``
and modify the /etc version.

Otherwise, the modification may be lost after ``swupd update``.

Reviewed-by: Zhang Di <di.zhang@intel.com>
Reviewed-by: Jin Zhi <Zhi.jin@intel.com>
Reviewed-by: Liu Xinwu <xinwu.liu@intel.com>


lijinxia:
Verified+1

PR #368, opened 2018-06-06
Tools: add Makefile for all tools

lyan3:
Add a Makefile under ./tools to avoid changing top-level Makefile frequently


NanlinXie:

PR #377, opened 2018-06-07
Refine guest control register initialization

binbinwu1:
This patch series ajust the order of control registers for more clear  in logic, but no change in guest init stage.
And in current code, on uefi platform, vmcs will be overwritten according to uefi context, using the exec_vmwrite directly. This patch series uses vmx_write_cr<#> interface to init control registers, which can be more clear, and some check logic can be reused.
Add comment to point out CR4 value should be inited before CR0, due to CR4 value will be checked when set CR0.


NanlinXie:

lijinxia:
Verified+1

PR #378, opened 2018-06-07
dm: fix the error code issues in passthrough.c

shiqingg:
With current implementation:
vm_init_vdevs only handles the negative error code, while passthru_init
returns positive error code when error occurs.
This causes unexpected dm crash since the real error is not being
handled properly.

What this patch does:
Change the error code to be negative value in passthru_init because it
is common in Linux kernel to return negative value when error occurs.

v2 -> v3
* add more comments about the reason to convert the return value

v1 -> v2:
* add a wrapper API to convert the error returned from pci_system_init
  to the ERROR we defined in DM
* use the defined errno as the return value rather than -1

### DIFF
--- a/devicemodel/hw/pci/passthrough.c
+++ b/devicemodel/hw/pci/passthrough.c
@@ -865,6 +865,31 @@ cfginit(struct vmctx *ctx, struct passthru_dev *ptdev, int bus,
 }
 
 /*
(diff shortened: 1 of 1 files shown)


NanlinXie:

lijinxia:
Verified+1

PR #385, opened 2018-06-08
DM USB: Fix potential crash issues.

xiaoguangwu:
This patch fix some potential crash issues, like wild pointers access, buffer overflow and etc.


NanlinXie:

lijinxia:
Verified+1

PR #391, opened 2018-06-08
apl-mrb: Add in weston.service

YinThong:
This service will start weston once power up.

Signed-off-by: Choong Yin Thong <yin.thong.choong@intel.com>

### DIFF
--- a/devicemodel/Makefile
+++ b/devicemodel/Makefile
@@ -178,6 +178,7 @@ install-samples-nuc: $(SAMPLES_NUC)
 
 install-samples-mrb: $(SAMPLES_MRB)
 	install -D -t $(DESTDIR)/usr/share/acrn/samples/apl-mrb $^
+	install -p -D -m 0644 $(DESTDIR)/usr/share/acrn/samples/apl-mrb/weston.service $(DESTDIR)/usr/lib/systemd/system
 
 install-bios: $(BIOS_BIN)
 	install -d $(DESTDIR)/usr/share/acrn/bios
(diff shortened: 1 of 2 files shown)


NanlinXie:

dongyaozu:
With current working style, You need to get it reviewed in mailinglist, and get ACK from maintainer. And then send PR request with those review info.    

Very simple patch can go PR directly...

dongyaozu:
@jren1 @gvancuts  can you review?

dongyaozu:
Thanks YinThong for the patches:) Let us wait some time for Geoff to comment

gvancuts:
> In the future, please follow the process:
> 1. send the patches to acrn-dev@lists.acrnproject.org for review
> 2. get the ack-by from maintainer and then send the git pull request here.

We should really update our documentation as today there is no indication there that it's the process one should follow: https://projectacrn.github.io/latest/contribute/contribute_guidelines.html#contribution-workflow

(On a personal note: I realize this is a common methodology in various open-source projects but I find it very cumbersome for me to work with this as it interferes with the corporate IT set-up. I wouldn’t want that to become a barrier for other community members to join and contribute).

gvancuts:
A few questions on this:
* Does that enable the `weston.service` (or does that make it available but still disabled by default)?
* Should we have something similar for `devicemodel/samples/nuc`?
   * I also believe this will unconditionally install this file on *any* platform (since it's part of the `make install` step). I don't think this is the intended behaviour?
* What bothers me a bit with this is that it assumes that `weston` is installed but that dependency is not called out anywhere. So if this patch gets merged as-is, someone would install the `service-os` bundle on a new Clear Linux installation and have a `systemd` service file installed that will not work until he installs an extra bundle and that won't happen automatically nor is documented anywhere.

@miguelinux , could you also review this patch?

YinThong:
@miguelinux will create proper step to be integrated on OS side.

PR #393, opened 2018-06-08
hv:replace unsigned long long with uint64_t

mingqiangchi:
unsigned long long--> uint64_t
long long --> int64_t

### DIFF
--- a/hypervisor/lib/sprintf.c
+++ b/hypervisor/lib/sprintf.c
@@ -143,7 +143,7 @@ static const char *get_flags(const char *s, int *flags)
 }
 
 static const char *get_length_modifier(const char *s,
-			int *flags, unsigned long long *mask)
+			int *flags, uint64_t *mask)
 {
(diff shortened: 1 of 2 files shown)


NanlinXie:

hypervisor/lib/sprintf.c
```diff
@@ -515,19 +515,19 @@ int do_print(const char *fmt, struct print_param *param,
 						(param->vars.flags &
 						PRINT_FLAG_LONG_LONG) ?
 						__builtin_va_arg(args,
```
dongyaozu: For this one, it relies on int/long/long long... I am not sure if we should change.. Can you check with Jason? Same for others in __builtin_va_arg.

lijinxia:
OK to verify

mingqiangchi:
have checked with Jason, will not change  __builtin_va_arg

lijinxia:
Verified+1

PR #398, opened 2018-06-09
hv: patch series for trampoline code relocation

ZideChen0:


NanlinXie:

lijinxia:
Verified+1

gvancuts:
Out of curiosity, is there any specific reason why the functions in the code are called `trampline*` instead of `trampoline*`?

PR #400, opened 2018-06-11
nuc-samples: use huge page as default launch_uos cmdline

JasonChenCJ:
we should use hugetlb as default uos memory allocation solution

### DIFF
--- a/devicemodel/samples/nuc/launch_uos.sh
+++ b/devicemodel/samples/nuc/launch_uos.sh
@@ -16,7 +16,10 @@ fi
 #for memsize setting
 mem_size=1000M
 
-acrn-dm -A -m $mem_size -c $2 -s 0:0,hostbridge -s 1:0,lpc -l com1,stdio \
+# make sure there is enough 2M hugepages in the pool
+echo 1024 > /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages
+
+acrn-dm -T -A -m $mem_size -c $2 -s 0:0,hostbridge -s 1:0,lpc -l com1,stdio \
   -s 5,virtio-console,@pty:pty_port \
   -s 6,virtio-hyper_dmabuf \
   -s 3,virtio-blk,/root/clear-21260-kvm.img \


NanlinXie:

devicemodel/samples/nuc/launch_uos.sh
```diff
@@ -16,7 +16,10 @@ fi
 #for memsize setting
 mem_size=1000M
 
-acrn-dm -A -m $mem_size -c $2 -s 0:0,hostbridge -s 1:0,lpc -l com1,stdio \
+# make sure there is enough 2M hugepages in the pool
+echo 1024 > /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages
```
gvancuts: I'm trying to understand this one better, with `mem_size=1000M`, are you allocating more memory than you really need (twice as much)?

gvancuts: Another thought on this, I think it would be better to add `hugepagesz=1G hugepages=2` to https://github.com/projectacrn/acrn-hypervisor/blob/master/hypervisor/bsp/uefi/clearlinux/acrn.conf instead? If I recall correctly, you mentionned earlier that trying to allocate those on a running system is more likely to fail.

My recommendation would therefore be to modify this patch to do a couple of things:
1. Add the parameter to `acrn.conf` (as mentionned above in my comment)
2. Modify this script to check if there is enough hugepage available (by checking `/sys/kernel/mm/hugepages-1048576kB/nr_hugepages` and `/sys/kernel/mm/hugepages-2048kB/nr_hugepages`) and *only if needed* try to allocate them. If that fails, we should also exit the script with an error message indicating why it failed.

JasonChenCJ: @gvancuts , here just is pre-allocation, so we do it with 1024*2M, later we will move this into DM. For 1G reservation, actually we dont know what's the user's platform memory size, we can do a hint by add hugepagesz=1G hugepages=2 to acrn.conf, or we can let user to modify it according to the HOWTO.
What do you prefer?

gvancuts: Hi @JasonChenCJ , I'm not completely following you. Here is what I think we should do, let me know if this makes sense to you:
* Booting with `hugepagesz=1G hugepages=2` is the same as doing `echo 1024 > /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages` when the system is already running (main difference being that the latter is more likely to fail based on how long the system has been running for and the amount of memory it has). Is that correct?
* If my assumption above is correct, I would recommend that we add `hugepagesz=1G hugepages=2` to `acrn.conf` and just check in the script if there is enough huge pages already reserved before attempting to start the VM.

Does that make sense?

JasonChenCJ: hi, @gvancuts , there is some difference, yes, you can do 1G page reservation through cmdline, but we need take care of UOS memory size which is not 1G align. For example, 1000M need 500 2M page -- it means reserved 1G page cannot be used.

gvancuts: OK, I'm still finding all this quite confusing (and I realize it's also most likely because of a lack of knowledge in this area on my part). Let's keep you patch as is then but I think we **really** need a proper document that explains how to manage the system memory overall. I.e. what memory to reserve, when, using what mechanism and what should be taken into consideration. Can you develop such guide?

PR #401, opened 2018-06-11
irq: convert irq/vector numbers to unsigned

junjiemao1:
Currently irq and vector numbers are used inconsistently.

    * Sometimes vector or irq ids is used in bit operations, indicating
      that they should be unsigned (which is required by MISRA C).

    * At the same time we use -1 to indicate an unknown irq (in
      common_register_handler()) or unavailable irq (in
      alloc_irq()). Also (irq < 0) or (vector < 0) are used for error
      checking. These indicate that irq or vector ids should be signed.

This patch converts irq and vector numbers to unsigned 32-bit integers, and
replace the previous -1 with IRQ_INVALID or VECTOR_INVALID. The branch
conditions are updated accordingly.


NanlinXie:

lijinxia:
Verified+1

PR #403, opened 2018-06-11
doc: use graphviz for intro boot-flow diagram

dbkinder:
Trying out the graphviz capability rather than using static images that
we can't easily modify.


NanlinXie:

PR #407, opened 2018-06-12
hv: replace vlapic_init by vlapic_reset in vcpu_reset

lyan3:
This change is to fix a guest vm hang issue at vm reset, especially easy to
be seen when it's a watchdog timeout reset.
vlapic_init create and init vlapic.vlapic_timer without deleting the
timer from cpu_times list, which breaks the list, results in a timer remains
with callback points to an invalid location.


NanlinXie:

lijinxia:
Verified +1

PR #408, opened 2018-06-12
HV: add bsp acpi info support

jsun26intel:
On some occations HV operates relying on host acpi info, eg. S3/S5/RTC...,
etc. Add ACPI parser code in HV would increase code size and complexity,
so a simple approach is use a c file to store all needed data in bsp.

Fow now we can store the data by hardcoding but we will provide a user
space tool shortly that run on target first to collect the data and then
generate the C file automatically.

The limitation is if ACPI data get changed due to BIOS update, we have to
regenerate the file and rebuild the HV. But this case is very rare.

Changelog:

	v3:	Remove Kconfig for different ApolloLake SKU,
		since all APL platform on SBL should use same
		ACPI table except _PSS.

	v2:	Add Kconfig support

	v1:	Initial version

### DIFF
--- a/hypervisor/arch/x86/guest/pm.c
+++ b/hypervisor/arch/x86/guest/pm.c
@@ -103,3 +103,21 @@ void vm_setup_cpu_state(struct vm *vm)
 	vm_setup_cpu_cx(vm);
 	init_cx_port(vm);
 }
+
+/* This function is for power management Sx state implementation,
+ * VM need to load the Sx state data to implement S3/S5.
+ */
+int vm_load_pm_s_state(struct vm *vm)
+{
+	if ((boot_cpu_data.x86 == host_acpi_info.x86_family)
+		&& (boot_cpu_data.x86_model == host_acpi_info.x86_model)) {
+		vm->pm.sx_state_data = (struct pm_s_state_data *)
+						&host_acpi_info.pm_s_state;
+		pr_info("System S3/S5 is supported.");
+		return 0;
+	} else {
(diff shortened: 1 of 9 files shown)


NanlinXie:

hypervisor/include/arch/x86/guest/vm.h
```diff
@@ -63,6 +63,7 @@ struct vm_pm_info {
 	struct cpu_px_data	px_data[MAX_PSTATE];
 	uint8_t			cx_cnt;		/* count of all Cx entries */
 	struct cpu_cx_data	cx_data[MAX_CSTATE];
+	struct pm_s_state_data	*s_state_data;
```
dongyaozu: rename s_state_data to sx_state_data; and add a comment to mention this is for S3/S5 xxxx ....

hypervisor/arch/x86/guest/pm.c
```diff
@@ -103,3 +103,18 @@ void vm_setup_cpu_state(struct vm *vm)
 	vm_setup_cpu_cx(vm);
 	init_cx_port(vm);
 }
+
+int vm_load_pm_s_state(struct vm *vm)
```
dongyaozu: add a comment it is for power management S3/S5 state load. xxxxxx..

jsun26intel:
Hi Eddie,
I've pushed again by force update, please review.
BR,
Victor

lijinxia:
Verified+1

PR #409, opened 2018-06-12
doc: add some docs about graphviz for drawings

dbkinder:
Add some practical examples and pointers to additional information about
using graphviz.

### DIFF
--- a/doc/howtos/process/graphviz.rst
+++ b/doc/howtos/process/graphviz.rst
@@ -1,0 +1,118 @@
+.. graphviz-examples:
+
+Drawings using graphviz
+#######################
+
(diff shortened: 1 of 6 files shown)


NanlinXie:

gvancuts:
Unless I've missed something, this does not publish it though (not until it's added as part of a `toc` somewhere), am I correct? Any reason why you don't want yet to include it as part of the documentation?

dbkinder:
@gvancuts The content in the howtos folder are published with the tech/* and process/* TOC entries.

gvancuts:
Thanks for the clarification!

PR #415, opened 2018-06-13
Handle softirq race condition

zhenggen:
HV: bug fix in operating softirq

a. handle a race condition in current softirq code.
b. add volatile declaration to pointer parameter in current bit operation code.


NanlinXie:

lijinxia:
Verified+1

PR #417, opened 2018-06-13
script: workaround to limit the plane numbers in UOS

minhe1:
This patch is to limit the plane numbers in UOS as a workaround for
some known issues.


NanlinXie:

PR #419, opened 2018-06-13
hv: fix android boot fail on 4GB memory board

mingqiangchi:
Config remained 1GB page to 1 on 4GB memory


NanlinXie:

PR #422, opened 2018-06-14
bugfix: make remove_intx_remapping static

yechun1:
remove_intx_remapping is not global function.

Unlike global functions in C, access to static functions is restricted to the file where they are declared. Another reason for making functions static can be reuse of the same function name in other files

Signed-off-by: Chris Ye <chris.ye@intel.com>


NanlinXie:

dongyaozu:
can you rebase ? We made a reshuffle for this file ..

lijinxia:
Verified+1

PR #426, opened 2018-06-14
hv:refine create/destroy secure world ept api

mingqiangchi:
refine create/destroy secure world ept to
handle discontinuous hpa


NanlinXie:

Issue #395, opened 2018-06-08
Errors when building the documentation (`make html` on Fedora 27)


gvancuts:
I followed [these instructions](https://projectacrn.github.io/latest/howtos/process/docbuild.html#installing-the-documentation-tools) to install the dependencies to build the documentation but I am encountering the following error(s) when running `make html`. I am on Fedora 27.

```
$ make html
make: [Makefile:42: html] Error 1 (ignored)

New errors/warnings found, please fix them:
==============================================

warning: Tag `CLANG_ASSISTED_PARSING' at line 1048 of file `-' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature enabled.
warning: Tag `CLANG_OPTIONS' at line 1056 of file `-' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature enabled.
Traceback (most recent call last):
  File "/usr/local/bin/sphinx-build", line 7, in <module>
    from sphinx.cmd.build import main
ModuleNotFoundError: No module named 'sphinx.cmd'
```

gvancuts:
The following patch gets me passed the initial warning (although I don't believe the warning is fatal):
```
diff --git a/doc/acrn.doxyfile b/doc/acrn.doxyfile
index 96b54ea..2c44681 100644
--- a/doc/acrn.doxyfile
+++ b/doc/acrn.doxyfile
@@ -1045,7 +1045,7 @@ VERBATIM_HEADERS       = YES
 # generated with the -Duse-libclang=ON option for CMake.
 # The default value is: NO.
 
-CLANG_ASSISTED_PARSING = NO
+#CLANG_ASSISTED_PARSING = NO
 
 # If clang assisted parsing is enabled you can provide the compiler with command
 # line options that you would normally use when invoking the compiler. Note that
@@ -1053,7 +1053,7 @@ CLANG_ASSISTED_PARSING = NO
 # specified with INPUT and INCLUDE_PATH.
 # This tag requires that the tag CLANG_ASSISTED_PARSING is set to YES.
 
-CLANG_OPTIONS          =
+#CLANG_OPTIONS          =
 
 #---------------------------------------------------------------------------
 # Configuration options related to the alphabetical class index
```

But I'm still left with this:
```
$ make html
[...]

PR #427, opened 2018-06-14
doc: update doc building tools versions

dbkinder:
The latest versions of Sphinx and Breathe work well together so we're
updating the requirements.txt installation information (and
documentation) to use the latest versions of these tools.

Fixes: #395


NanlinXie:

PR #411, opened 2018-06-13
Support menuconfig and drop duplicated command line options

junjiemao1:
This patch set adds the following features to the make system.

    1. 'make menuconfig' now shows a curses-based TUI for modifying
        configurations.

    2. 'PLATFORM=' or 'RELEASE=' are no longer needed when calling 'make' if a
        configuration exists.

The getting started guide is updated accordingly.

v1 -> v2:

    * Fix 'make oldconfig' overwritting the modified .config with out-dated
      values in config.mk.
    * Rename minimalconfig to savedefconfig.
    * Add link to EPEL in the getting started guide.
    * Split the step using a defconfig from those generating a defconfig in the
      getting started guide.

PR #428, opened 2018-06-14
doc: add hypervisor kconfig option reference

dbkinder:
Modify the Zephyr project script for creating documentation from Kconfig
files to generate the hypervisor option configuration docs.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -14,9 +14,15 @@ SPHINXPROJ    = "Project ACRN"
 SOURCEDIR     = .
 BUILDDIR      ?= _build
 
+# document publication assumes the folder structure is setup
+# with the acrn-hypervisor and projectacrn.github.io repos as
+# sibling folders and make is run inside the acrn-hypervisor/docs
+# folder.
+
+ACRN_BASE     = "$(CURDIR)/../.."
 DOC_TAG      ?= development
 RELEASE      ?= latest
-PUBLISHDIR    = ../../projectacrn.github.io/$(RELEASE)
+PUBLISHDIR    = $(ACRN_BASE)/projectacrn.github.io/$(RELEASE)
 
 # Put it first so that "make" without argument is like "make help".
 help:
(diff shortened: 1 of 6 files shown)


NanlinXie:

dbkinder:
@junjiemao1 This should get us the hypervisor config option documentation (mention in #411)

dbkinder:
It's also worth noting the Zephyr project uses its own version of kconfiglib and the config documentation generation script relies on this version so it included here too.

junjiemao1:
@ulfalizer welcome :)

@dbkinder I've tried with the vanilla Kconfiglib from Pypi. It also works and there is no differences among the generated files.

doc/Makefile
```diff
@@ -14,9 +14,10 @@ SPHINXPROJ    = "Project ACRN"
 SOURCEDIR     = .
 BUILDDIR      = _build
 
+ACRN_BASE     = "`pwd`/../.."
```
gvancuts: Could we use `ACRN_BASE = $(CURDIR)/../..` here instead? (or `$(shell pwd)/../..`)

dbkinder:
I've incorporated your comments @ulfalizer @gvancuts @junjiemao1 Thanks!

dbkinder:
@ulfalizer looks like genrest.py (when run in the CI environment) doesn't like non-ascii characters:

```
make[1]: Entering directory '/workspace/acrn-hypervisor-premerge-build-nuc/acrn-hypervisor/doc'
Traceback (most recent call last):
  File "scripts/genrest.py", line 377, in <module>
    write_kconfig_rst()
  File "scripts/genrest.py", line 97, in write_kconfig_rst
    write_sym_rst(sym, out_dir)
  File "scripts/genrest.py", line 123, in write_sym_rst
    kconfig_definition_rst(sym))
  File "scripts/genrest.py", line 373, in write_if_updated
    f.write(s)
UnicodeEncodeError: 'ascii' codec can't encode character '\u2192' in position 335: ordinal not in range(128)
make[1]: Leaving directory '/workspace/acrn-hypervisor-premerge-build-nuc/acrn-hypervisor/doc'
make[1]: *** [Makefile:48: kconfig] Error 1
make: *** [Makefile:41: doc] Error 2
```

ulfalizer:
@dbkinder 
Only meant safe in the font sense there. ;)

Yeah, should probably just force UTF-8 in `genrest.py`. Kconfiglib already defaults to UTF-8.

Might be a good idea to switch to some UTF-8 locale in the CI environment too (`C.UTF-8` is usually available). The root cause of the problem is that that the environment has `LC_CTYPE` set to `C`, which tells Python to use ASCII as the encoding. Other tools besides Python will check that setting as well.

Related Python PEP (it's been enough of a pain that they decided to add a workaround for Python 3.7): https://www.python.org/dev/peps/pep-0538/

dbkinder:
@ulfalizer thanks for the info, should we tweak genrest.py in Zephyr as well, just in case?

ulfalizer:
@dbkinder 
Can't hurt. Want to make a PR?

PR #431, opened 2018-06-14
doc: add graphviz installation instructions

dbkinder:


gvancuts:
Just a suggestion (low-priority): add it too to the bullet list of SW components (and their versions) just above.

PR #433, opened 2018-06-15
HV: enable SMEP in hypervisor

yonghuah:
- this patch is to enable SMEP in hypervisor, SMEP protects
   guests' memory from supervisor-mode instruction fetches,
   in other words, hypervisor which operating in supervisor
   mode can't fetch instructions from (guests' memory)
   linear addresses that are accessible in user mode.


NanlinXie:

PR #437, opened 2018-06-15
DM sample: force enabling HDMI1 and HDMI2 connectors

feijiang1:
When initial_modeset is enabled, will check EDID in early stage,
while sometime didn't get EDID in time through I2C, and will cause
CRTC not used correctly in some corner cases. Then we need force
enabling two HDMI connectors to attach CRTC in initial modeset stage.

Reviewed-by: Zhao, Yakui <yakui.zhao@intel.com>
Reviewed-by: Min He <min.he@intel.com>


NanlinXie:

TianhuaSun:
Verified+1

PR #441, opened 2018-06-16
dm: change the android guest selinux policy as permissive

jren1:


NanlinXie:

PR #446, opened 2018-06-19
fix parted of "missing for discarded return value"

HuiHuangShi:
MISRA C required that return value should be used, missing for it should
add "(void)" prefix before the function call.
Some function can be declared without return value to avoid this problem.


lijinxia:
OK to verify

lijinxia:
Verified+1

PR #453, opened 2018-06-19
hv: fix error use of list_del for delete timer

lifeix:
We should use list_del_init here. Otherwise it would misorder the
timer list if we would delete timer more than once if there're more
than one timer on it.

Change-Id: I2d18248173ced1ea1b723352327d2c3d0efc4900

### DIFF
--- a/hypervisor/arch/x86/timer.c
+++ b/hypervisor/arch/x86/timer.c
@@ -96,7 +96,7 @@ int add_timer(struct timer *timer)
 void del_timer(struct timer *timer)
 {
 	if (timer && !list_empty(&timer->node))
-		list_del(&timer->node);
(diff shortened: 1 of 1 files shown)


NanlinXie:

lifeix:
@dongyaozu 
When we shutdown the vm, we would call 
reset_vcpu -> vlapic_reset -> vlapic_reset_timer -> del_timer 
and
destroy_vcpu->vlapic_free -> del_timer.  
vlapic_reset will reset the timer and delete the timer of course. while vlapic_free should delete the timer then free the vlapic. Now we would call vlapic_reset before vlapic_free, would we  follow the rule later ?  If yes, I will remove del_timer in vlapic_free. Thx

PR #458, opened 2018-06-19
HV: convert hexadecimals to unsigned

junjiemao1:
Per MISRA C, operands to bit-wise operations should have unsigned
types. However, C99 prioritizes to use signed integers for hexadecimal constants
without the 'u' suffixes, leading to tons of bit operations on signed integers.

This patch series add the 'u' suffixes to the constants which are used in bit
operations, and add the intended width of these integers when applicable
(i.e. the target value is at least 32-bit wide) to avoid functional differences
due to signed vs. unsigned extensions.


NanlinXie:

hypervisor/include/arch/x86/cpu.h
```diff
@@ -41,80 +41,80 @@
 /* Define page size */
 #define CPU_PAGE_SHIFT          12
 #define CPU_PAGE_SIZE           0x1000
-#define CPU_PAGE_MASK           0xFFFFFFFFFFFFF000
+#define CPU_PAGE_MASK           0xFFFFFFFFFFFFF000ULL
```
dongyaozu: Why this is ULL ? It is uint64_t, right?

hypervisor/include/arch/x86/cpu.h
```diff
@@ -310,7 +310,7 @@ void start_cpus();
 /* This macro writes the stack pointer. */
 #define CPU_SP_WRITE(stack_ptr)                             \
 {                                                           \
-	uint64_t rsp = (uint64_t)stack_ptr & ~(CPU_STACK_ALIGN - 1);  \
+	uint64_t rsp = (uint64_t)stack_ptr & ~(CPU_STACK_ALIGN - 1U);  \
```
dongyaozu: This should be 1UL.. Not?

hypervisor/arch/x86/gdt.c
```diff
@@ -11,14 +11,14 @@ static void set_tss_desc(union tss_64_descriptor *desc,
 {
 	uint32_t u1, u2, u3;
 
-	u1 = ((uint64_t)tss << 16) & 0xFFFFFFFF;
-	u2 = (uint64_t)tss & 0xFF000000;
-	u3 = ((uint64_t)tss & 0x00FF0000) >> 16;
+	u1 = ((uint64_t)tss << 16) & 0xFFFFFFFFULL;
```
dongyaozu: I am lost on the usage of UL/ULL. I will stop here to get it clarified

junjiemao1: Per C99, L (long) is at least 32-bit while LL (long long) is at least 64-bit. The actual bits of each integer type is implementation-defined. Thus I use ULL for 64-bit constants whenever possible.

hypervisor/include/arch/x86/cpu.h
junjiemao1: My bad. CPU_STACK_ALIGN and the '1' here should all be uint64_t.

lijinxia:
Verified+1

PR #459, opened 2018-06-19
 tools: acrn-crashlog: fix potential issues for acrnprobe

liuxinw:
The aim of this patch set is to fix potential issues, which are reported by
static analysis tool, for acrnprobe.

For the convenience of review, break it into two patches.


NanlinXie:

lijinxia:
Verified+1

PR #467, opened 2018-06-20
hv: add sanity check for the input vcpu_id

zhijianli88:
Previously, HV assigns vm cpuid accroding to vm->hw.created_vcpus, which is
not always same with the input vcpu_id. And the DM uses the input vcpu_id
directly later. This might cause arcn has 2 different vcpuid pointing to one vcpu

### DIFF
--- a/hypervisor/common/hypercall.c
+++ b/hypervisor/common/hypercall.c
@@ -218,6 +218,12 @@ int64_t hcall_create_vcpu(struct vm *vm, uint64_t vmid, uint64_t param)
 		return -1;
 	}
 
+	if ((uint32_t)vm->hw.created_vcpus != cv.vcpu_id) {
+		pr_err("vcpu %d:%d is invalid(already assigned ?)\n",
+		        vm->hw.created_vcpus, cv.vcpu_id);
+		return -EINVAL;
+	}
+
 	pcpu_id = allocate_pcpu();
 	if (-1 == pcpu_id) {
 		pr_err("%s: No physical available\n", __func__);


NanlinXie:

dongyaozu:
@gvancuts  can you review the 2nd patch?

hypervisor/bsp/uefi/efi/Makefile
```diff
@@ -43,8 +43,8 @@ HOST  = $(shell $(CC) -dumpmachine | sed "s/\(-\).*$$//")
 ARCH := $(shell $(CC) -dumpmachine | sed "s/\(-\).*$$//")
 
 ifeq ($(ARCH),x86_64)
-	LIBDIR := $(shell if [ -d /usr/lib64 ]; then echo /usr/lib64; \
-			else if [ -d /usr/lib ]; then echo /usr/lib; fi ; fi;)
+	LIBDIR := $(shell if [ -e /usr/lib64/elf_x86_64_efi.lds ]; then echo /usr/lib64; \
```
gvancuts: This is broken on Fedora as this file is put in a `gnuefi/` subdir (i.e. under `/usr/lib64/gnuefi/`).

dongyaozu: Thanks Geoff!  If @zhijianli88 can split this 2 patches to 2 PRs, I can merge the 1st one now.

zhijianli88: at my ubuntu16.04, gnu-efi is installed into /usr/lib instead of /usr/lib64
```
~/workspace$ dpkg -L gnu-efi | grep elf_x86_64_efi.lds
/usr/lib/elf_x86_64_efi.lds
```
but the previous Makefile try to link /usr/lib64/elf_x86_64_efi.lds becuase my host has /usr/lib64 as well which is provided by libc6-amd64:i386
```
~/workspace$ dpkg -S /usr/lib64
libc6-amd64:i386: /usr/lib64
```

the previous logic always assume LIBDIR is /usr/lib64 if it exists

zhijianli88: @gvancuts you are right, it might  break Fedora. let's consider more...

gvancuts:
> at my ubuntu16.04, gnu-efi is installed into /usr/lib instead of /usr/lib64
> ~/workspace$ dpkg -L gnu-efi | grep elf_x86_64_efi.lds /usr/lib/elf_x86_64_efi.lds
> but the previous Makefile try to link /usr/lib64/elf_x86_64_efi.lds becuase my host has /usr/lib64 as well which is provided by libc6-amd64:i386
> ~/workspace$ dpkg -S /usr/lib64
> libc6-amd64:i386: /usr/lib64
> the previous logic always assume LIBDIR is /usr/lib64 if it exists

Thanks for the explanation. I now understand how the logic is flawed in the Makefile. You're right, this is a problem that we should fix. We should probably just look for the file using 'find' (it's already used in various other Makefiles) and then extract the path where it's at (and maybe assign it to a variable called GNUEFIDIR).

zhijianli88:
@dongyaozu  I have removed the 2nd patch.

dongyaozu:
@zhijianli88 seems checks have failed... Make sure you are rebased?

zhijianli88:
I noticed that, the log said "could not created VCPU0"
http://darcy-optiplex.sh.intel.com:8080/job/acrn-hypervisor-premerge-deployment-nuc/414/console

i'm looking into it.

BTW, where is the hv log, i can only find dm log.

zhijianli88:
@dongyaozu  i added some debug info this patch, which is not good to be merged into upstream.

lijinxia:
UOS boot failed！
vcpu 0 is invalid(already assigned ?)

zhijianli88:
any way to access the hv log from internal intel(i can access intel internal network). i have no hw platform on hand, so i cannot debug it locally.

zhijianli88:
It's possible caused by dm and kernel use different struct acrn_create_vcpu
dm side
···
struct acrn_create_vcpu {
	/** the virtual CPU ID for the VCPU created */
	uint16_t vcpu_id;

	/** the physical CPU ID for the VCPU created */
	uint16_t pcpu_id;
} __aligned(8);
···

SOS kernel side
```
struct acrn_create_vcpu {
	/** the virtual CPU ID for the VCPU created */
	uint32_t vcpu_id;

	/** the physical CPU ID for the VCPU created */
	uint32_t pcpu_id;
} __attribute__((aligned(8)));
```
Noticed some one is fixing on this, will try again later.

PR #472, opened 2018-06-20
irq_count accounting

sainath14:
When guest frees vector, counts needs to be reset. "int" command from hv shell ends up incrementing per cpu irq_count. Fixed it.


NanlinXie:

lijinxia:
Verified+1

PR #477, opened 2018-06-21
HV: remove unused API lookp_entry_by_id()

zhijianli88:

### DIFF
--- a/hypervisor/arch/x86/assign.c
+++ b/hypervisor/arch/x86/assign.c
@@ -68,17 +68,6 @@ _lookup_entry_by_id(uint32_t id)
 	return NULL;
 }
 
-static inline struct ptdev_remapping_info *
-lookp_entry_by_id(uint32_t id)
-{
-	struct ptdev_remapping_info *entry;
-
-	spinlock_obtain(&ptdev_lock);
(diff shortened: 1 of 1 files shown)


NanlinXie:

hypervisor/arch/x86/assign.c
```diff
@@ -69,7 +69,7 @@ _lookup_entry_by_id(uint32_t id)
 }
 
 static inline struct ptdev_remapping_info *
-lookp_entry_by_id(uint32_t id)
+lookup_entry_by_id(uint32_t id)
```
dongyaozu: Good Catch @zhijianli88 
But, why this is an orphan API ?  @gzhai7

edwinzhai: Sorry. It's a mistake when spliting the big patches. This orphan API will be removed by the 2nd patches.  I'll continue to split the 2nd into small ones to address this.

Leave it as is or delete it. I'll handle it in next small patches.

dongyaozu: In this case, @zhijianli88 please delete the API.

lijinxia:
Verified+1

PR #478, opened 2018-06-22
hv mmu & ept revisit serial

lifeix:
hv mmu & ept revisit serial 2
v1-v2:
1. remove how to support cma document.
2. remove sample script about how to use hugetlb by default 
3. remove git commit changed-id

v1:
Now the hv mmu & ept map/unmap/modify logic is not clear. Map may map 
mapped page tables; unmap may unmap non-mapped or unmapped page tables;
modify may modify gfn to mfn mapping besides page table attributes or
modify non-mapped page table.
We want to change this. Let's set some rules first:
1. map can only map unmapped page tables
2. unmap can only unmap mapped page tables;
3. modify can only modify mapped page tables' attributes.

Besides, we want to move configure page table attributes outsides internal
implement logic.

This patch is the serial two 1st half.
After this serial, we can achieve:
1. emove cma support in the dm;

Note:
We should remove the "-T" option in script for launch UOS.

Next will
1. remove cma support in the dm vhm
2. remove MAP_MMIO in sos vhm then remove MAP_MMIO in hv
3. add MAP_MODIFY in hv then add MAP_MODIFY in sos vhm
4. apply rules list above in hv

hv mmu & ept revisit serial 1

v1-v2:
1. minor fix when mask EPT cache type.

v1:
Now the hv mmu & ept map/unmap/modify logic is not clear. Map may map 
mapped page tables; unmap may unmap non-mapped or unmapped page tables;
modify may modify gfn to mfn mapping besides page table attributes or
modify non-mapped page table.
We want to change this. Let's set some rules first:
1. map can only map unmapped page tables
2. unmap can only unmap mapped page tables;
3. modify can only modify mapped page tables' attributes.

Besides, we want to move configure page table attributes outsides internal
implement logic.

This patch is the serial one.
After this serial, we can achieve:
1. We would not unmap unmap non-mapped or unmapped page tables;
2. set page tables attributes at frist before call internal implement function.
Next will
1. remove cma support in the dm & sos vhm
2. remove MAP_MMIO in sos vhm then remove MAP_MMIO in hv
3. add MAP_MODIFY in hv then add MAP_MODIFY in sos vhm
4. apply rules list above in hv

### DIFF
--- a/devicemodel/core/main.c
+++ b/devicemodel/core/main.c
@@ -72,7 +72,6 @@ char *guest_uuid_str;
 char *vsbl_file_name;
 uint8_t trusty_enabled;
 bool stdio_in_use;
-bool hugetlb;
 
 static int guest_vmexit_on_hlt, guest_vmexit_on_pause;
 static int virtio_msix = 1;
@@ -146,7 +145,6 @@ usage(int code)
 		"       -U: uuid\n"
 		"       -w: ignore unimplemented MSRs\n"
 		"       -W: force virtio to use single-vector MSI\n"
-		"       -T: use hugetlb for memory allocation\n"
 		"       -x: local apic is in x2APIC mode\n"
 		"       -Y: disable MPtable generation\n"
 		"       -k: kernel image path\n"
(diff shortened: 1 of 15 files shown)


NanlinXie:

lifeix:
Close this PR and split it into two PR (HV and DM) later.

PR #486, opened 2018-06-22
hv: cpu_context is not only used by guest.

fyin1:
It could be also used by host as well. So we remove GUEST from
MACRO name

Acked-by: Eddie Dong <Eddie.dong@intel.com>


NanlinXie:

lijinxia:
Verified+1

PR #487, opened 2018-06-22
Dockerfiles: add Python3 and kconfiglib build dependency

gvancuts:
This adds Python3 and the 'kconfiglib' tool to all Dockerfiles.
This is required in order to build (and/or configure) the ACRN
hypervisor.

It also adds 'findutils' to all Fedora Dockerfiles as both 'find'
and 'xargs' are used in some Makefiles and provided by this package.


NanlinXie:

PR #489, opened 2018-06-23
creating a guide on how to use Ubuntu as sos

ailun258:
this guide is for people who are interested in booting up Ubuntu system as the Service OS on top of ACRN hypervisor,

### DIFF
--- a/doc/howtos/using_ubuntu_as_sos.rst
+++ b/doc/howtos/using_ubuntu_as_sos.rst
@@ -1,0 +1,184 @@
+Using Ubuntu as the Service OS with the ACRN hypervisor
+
+
+Based on the ACRN open source project and getting started guide. This documentation is to guide people how to boot Ubuntu system as the 
+Service OS with the ACRN hypervisor. By default, in the getting started guide documentation of the ACRN open source project, Clear Linux
+is being used as the Service OS and User OS on top of the ACRN hypervisor. 
+Here we are trying to boot Ubuntu system as the Service OS and launch Clear Linux as User OS with the ACRN hypervisor on official 
+supported NUC, playing steps as below:
+
+Step #1,  Installing a Ubuntu system on the NUC with standard installing steps. 
+For example with Ubuntu 16.04.4 LTS distribution.
+Download it here: https://www.ubuntu.com/download/desktop
+Burn Ubuntu 16.04.4 image to a USB key
+Install Ubuntu system with the USB key 
+After complete installation and rebooting, you could boot Ubuntu from grub menu.
+
+Step #2,  setup proxy with you prefered ways, command line tool or system settings tool.  If you don’t need to setup proxy, you could 
+skip this step
+
+Step #3,  installing ssh service with below commands and allow “root” login over SSH optional, 
+People could just ssh into the Service OS and use 'sudo' once logged in (which would be preferable from a security standpoint), then 
+restart SSH service. Of cause, this step is not must 
+if you know how to play, just skip this step.
+# sudo apt-get install openssh-server
+# sudo service ssh status
+In /etc/ssh/sshd_config 
+Change PermitRootLogin from prohibit-password to yes
+# sudo service ssh restart
+
+Step #4, additional packages needed to install, please check the related sections of getting started guide in the ACRN open source
+project, if you don’t have an chance to setup additional packages flowing up that guid to build source. You can reference below commands
+to install directly here. 
+# sudo apt-get update 
+# sudo apt-get install git
+# sudo apt-get install libpciaccess-dev
+# sudo apt-get install gnu-efi
+# sudo apt-get install uuid-dev
+# sudo apt-get install libssl-dev
+# sudo apt-get install vim
+# sudo apt-get install libsystemd-dev
+# sudo apt-get install libevent-dev
+# sudo apt-get install libxml2-dev
+
+Step #5, Checkout source code from open source project repositories
+# git clone https://github.com/projectacrn/acrn-hypervisor
+Note: if you encounter issues with “git clone”, you can download ZIP package directly. It’s so easy.
+
+Step # 6, building the ACRN hypervisor and device model source code on Ubuntu system
+Building device model and hypervisor from source trees on NUC you just setup. You can reference the section of Getting started guide 
+of the ACRN open source project, or following up below commands to build directly here. 
+# cd acrn-hypervisor
+# make PLATFORM=uefi
+The build results are found in the build directory.
+For example:
+# ls acrn-hypervisor/build
+devicemodel
+hypervisor
+tools
+
+Steps #7,  Preparing Service OS kernel 
+You can download latest Service OS kernel from below links:
+https://download.clearlinux.org/releases/current/clear/x86_64/os/Packages/
+
+for example:
+https://download.clearlinux.org/releases/current/clear/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
+or
+https://cdn.download.clearlinux.org/current/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
+
+# mkdir ~/kernel-build
+# cd ~/kernel-build
+# wget  https://download.clearlinux.org/releases/current/clear/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
+# sudo apt-get install rpm2cpio
+# rpm2cpio linux-pk414-sos-4.14.41-39.x86_64.rpm | cpio -idmv 
+# sudo cp -r ~/kernel-build/usr/lib/modules/4.14.41-39.pk414-sos/ /lib/modules/
+
+Step #8, Deploying the hypervisor
+Mount the EFI partition and verify you have the following files, then add the ACRN hypervisor and Clear Linux kernel to the EFI 
+Partition, 
+# umount /boot/efi 
+# lsblk
+# mount /dev/sda1 /mnt
+# ls /mnt/EFI/ubuntu
+  fw  fwupx64.efi  grub.cfg  grubx64.efi  MokManager.efi  shimx64.efi
+
+Copy acrn.efi  to /mnt/EFI/acrn/ 
+# sudo mkdir /mnt/EFI/acrn/
+# sudo cp acrn-hypervisor/build/acrn.efi /mnt/EFI/acrn	
+
+Step #9, Configure the EFI firmware to boot the ACRN hypervisor by default
+# sudo -s
+# efibootmgr -c -l "\EFI\acrn\acrn.efi" -d /dev/sda -p 1 -L "ACRN Hypervisor                                                                                                    NUC" -u "bootloader=\EFI\ubuntu\grubx64.efi"
+Then you can see the result of boot order. Also you can check the boot order with below command:
+# efibootmgr -v            	
+or setting boot order with this command to make sure the acrn.efi is the first boot option:
+# efibootmgr -o x, x       	
+Note: by default, the “ACRN Hypervisor” you just added should be the first one to boot. But please make sure it is, or select it 
+manually after press “F10” when system booting up.   
+
+Step #10, Deploy the SOS kernel
+# sudo cp ~/kernel-build/usr/lib/kernel/org.clearlinux.pk414-sos.4.14.41-39 /boot/acrn/
+
+Step #11, Customize the boot of the grub entry 
+Use grub custom file (/etc/grub.d/40_custom) to create a new grub entry
+Add the following to the end of grub file
+menuentry 'ACRN ubuntu SOS' {
+        recordfail
+        load_video
+        insmod gzio
+        insmod part_gpt
+        insmod ext4
+        linux /boot/acrn/org.clearlinux.pk414-sos.4.14.41-39 pci_devices_ignore=(0:18:1) maxcpus=1 console=tty0 console=ttyS0 
+        i915.nuclear_pageflip=1 root=PARTUUID=<ID of rootfs partition> rw rootwait ignore_loglevel no_timer_check consoleblank=0 
+        i915.tsd_init=7 i915.tsd_delay=2000 i915.avail_planes_per_pipe=0x00000F i915.domain_plane_owners=0x011111110000 
+        i915.enable_guc_loading=0 i915.enable_guc_submission=0 i915.enable_preemption=1 i915.context_priority_mode=2 
+        i915.enable_gvt=1 hvlog=2M@0x1FE00000
+}
+Note: here you have to use PARTUUID, UUID doesn’t work.  
+
+Step #12,  Update grub
+# sudo update-grub
+At this moment, you need to modify /boot/grub/grub.cfg file manually to enable the timeout so that the system has opportunity to show 
+you grub menu to select. Or you need to set the “ACRN ubuntu SOS” as the default boot option. Without enabling  timeout, the system 
+will not show grub menu for people to select and boot into incorrect kernel.
+
+Enable timeout:
+ if [ x$feature_timeout_style = xy ] ; then
+ -  set timeout_style=hidden
+ + #set timeout_style=hidden
+ -  set timeout=0
+ + set timeout = 10 
+ # Fallback hidden-timeout code in case the timeout_style feature is
+  # unavailable.
+
+Step #13, Check the boot process 
+Reboot system to see if the acrn.efi is loaded successfully, and if the Menu of Grub being shown to select “ACRN ubuntu SOS”.  
+after boot successfully, the system will start to Ubuntu Desktop successfully.  to check if the system is running on virtualization 
+evn and the ACNR hypervisor is running. 
+For example:
+   
+NUC6CAYH:~$ dmesg | grep ACRN
+[    0.000000] Hypervisor detected: ACRN
+[    0.862942] ACRN HVLog: acrn_hvlog_init
+
+This output means the boot flow is correct and the SOS system with ACRN is running well.
+ 
+
+Step #14, Deploy Clear Linux as the UOS 
+Download clear linux UOS from download.clearlinux.org
+#cd /root
+# wget https://download.clearlinux.org/releases/22780/clear/clear-22780-kvm.img.xz
+# unxz clear-22780-kvm.img.xz
+Copy  launch_uos.sh from source code
+# cp ~/Download/acrn-hypervisor/devicemodel/samples/nuc/launch_uos.sh /root
+Download standard PK kernel with below command:
+# wget https://download.clearlinux.org/releases/22780/clear/x86_64/os/Packages/linux-pk414-standard-4.14.47-44.x86_64.rpm
+# rpm2cpio linux-pk414-standard-4.14.47-44.x86_64.rpm | cpio -idmv
+
+Use the following steps to update clearlinux kernel:
+# sudo losetup -f -P --show /root/clear-22789-kvm.img  
+# sudo mount /dev/loop0p3 /mnt
+# sudo cp -r /root/usr/lib/modules/4.14.47-44.pk414-standard /mnt/lib/modules/
+# sudo cp -r /root/usr/lib/kernel /lib/modules/
+# sudo umount /mnt
+# sync
+
+You may Need to a permission issue
+# sudo chmod 777 /dev/acrn_vhm
+You also need to install a package.
+# sudo apt-get instal iasl
+# sudo cp /usr/bin/iasl /usr/sbin/iasl
+
+Step 15, Deploy device module - acrn-dm
+# cp ~/download/acrn-hypervisor//build/devicemodel/acrn-dm /usr/bin
+
+Step 16, modify launch_uos.sh 
+-s 3,virtio-blk,/root/clear-22780-kvm.img
+-k /lib/modules/kernel/org.clearlinux.pk414-standard.4.14.47-44
+
+Step 17, Run UOS
+	# cd /root
+	# ./launch_uos.sh
+
+Congratulation !!! 
+You are now see UOS booting up…..   


NanlinXie:

ailun258:
@dbkinder 
any progress?

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+for example:
+https://download.clearlinux.org/releases/current/clear/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
+or
+https://cdn.download.clearlinux.org/current/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
+
+# mkdir ~/kernel-build
+# cd ~/kernel-build
+# wget  https://download.clearlinux.org/releases/current/clear/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
+# sudo apt-get install rpm2cpio
+# rpm2cpio linux-pk414-sos-4.14.41-39.x86_64.rpm | cpio -idmv
```
dbkinder: All these ``linux-pk414-*`` file names are referencing files that don't exist.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+Step #11, Customize the boot of the grub entry 
+Use grub custom file (/etc/grub.d/40_custom) to create a new grub entry
+Add the following to the end of grub file
+menuentry 'ACRN ubuntu SOS' {
+        recordfail
+        load_video
+        insmod gzio
+        insmod part_gpt
+        insmod ext4
+        linux /boot/acrn/org.clearlinux.pk414-sos.4.14.41-39 pci_devices_ignore=(0:18:1) maxcpus=1 console=tty0 console=ttyS0
```
dbkinder: version number in this filename is wrong too... :(

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+}
+Note: here you have to use PARTUUID, UUID doesn’t work.  
+
+Step #12,  Update grub
+# sudo update-grub
+At this moment, you need to modify /boot/grub/grub.cfg file manually to enable the timeout so that the system has opportunity to show 
+you grub menu to select. Or you need to set the “ACRN ubuntu SOS” as the default boot option. Without enabling  timeout, the system 
+will not show grub menu for people to select and boot into incorrect kernel.
+
+Enable timeout:
```
dbkinder: This notation might be confusing (with the leading ``+`` and ``-`` signs).  Better would be to just tell them what it should look like after the changes are made.

dbkinder:
Also, PR #523 is doing a document reorganization so there's no longer a howtos section.  It will probably be easiest to close this PR and I'll open a new one with the reST, spelling, and grammar edits.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+https://download.clearlinux.org/releases/current/clear/x86_64/os/Packages/
+
+for example:
+https://download.clearlinux.org/releases/current/clear/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
+or
+https://cdn.download.clearlinux.org/current/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
+
+# mkdir ~/kernel-build
+# cd ~/kernel-build
+# wget  https://download.clearlinux.org/releases/current/clear/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
```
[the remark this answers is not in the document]
gvancuts: It's going to be a moving target if we use the `current` URL. I would suggest we point at a known good kernel as a reference but still recommend that people use the latest and greatest if possible (by looking inside this `current` URL).

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+supported NUC, playing steps as below:
+
+Step #1,  Installing a Ubuntu system on the NUC with standard installing steps. 
+For example with Ubuntu 16.04.4 LTS distribution.
+Download it here: https://www.ubuntu.com/download/desktop
+Burn Ubuntu 16.04.4 image to a USB key
+Install Ubuntu system with the USB key 
+After complete installation and rebooting, you could boot Ubuntu from grub menu.
+
+Step #2,  setup proxy with you prefered ways, command line tool or system settings tool.  If you don’t need to setup proxy, you could
```
gvancuts: I would leave this out completely. Setting up the proxy is what I'd consider part of the usual installation steps. We could at most have a note that invites them to make sure they have internet access as we'll need to download some stuff later.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+For example with Ubuntu 16.04.4 LTS distribution.
+Download it here: https://www.ubuntu.com/download/desktop
+Burn Ubuntu 16.04.4 image to a USB key
+Install Ubuntu system with the USB key 
+After complete installation and rebooting, you could boot Ubuntu from grub menu.
+
+Step #2,  setup proxy with you prefered ways, command line tool or system settings tool.  If you don’t need to setup proxy, you could 
+skip this step
+
+Step #3,  installing ssh service with below commands and allow “root” login over SSH optional,
```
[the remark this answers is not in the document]
gvancuts: Let's stick to explaining the standard SSH access (and let people use sudo in that session if/when they need it).

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+restart SSH service. Of cause, this step is not must 
+if you know how to play, just skip this step.
+# sudo apt-get install openssh-server
+# sudo service ssh status
+In /etc/ssh/sshd_config 
+Change PermitRootLogin from prohibit-password to yes
+# sudo service ssh restart
+
+Step #4, additional packages needed to install, please check the related sections of getting started guide in the ACRN open source
+project, if you don’t have an chance to setup additional packages flowing up that guid to build source. You can reference below commands
```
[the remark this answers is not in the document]
gvancuts: Can we add a link to that particular section in the Getting Started Guide?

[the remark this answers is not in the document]
gvancuts: I agree we should not repeat these steps in here... as a matter of fact, the list below is already outdated (missing Python3 and kconfiglib which are required to build the hypervisor).

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# sudo apt-get install uuid-dev
+# sudo apt-get install libssl-dev
+# sudo apt-get install vim
+# sudo apt-get install libsystemd-dev
+# sudo apt-get install libevent-dev
+# sudo apt-get install libxml2-dev
+
+Step #5, Checkout source code from open source project repositories
+# git clone https://github.com/projectacrn/acrn-hypervisor
+Note: if you encounter issues with “git clone”, you can download ZIP package directly. It’s so easy.
```
gvancuts: I would like to see a specific version (tag, commit ID, ...) that is known to work.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
+Using Ubuntu as the Service OS with the ACRN hypervisor
+
+
+Based on the ACRN open source project and getting started guide. This documentation is to guide people how to boot Ubuntu system as the 
+Service OS with the ACRN hypervisor. By default, in the getting started guide documentation of the ACRN open source project, Clear Linux
+is being used as the Service OS and User OS on top of the ACRN hypervisor. 
+Here we are trying to boot Ubuntu system as the Service OS and launch Clear Linux as User OS with the ACRN hypervisor on official 
+supported NUC, playing steps as below:
+
+Step #1,  Installing a Ubuntu system on the NUC with standard installing steps.
```
gvancuts: I believe it's still possible to install Ubuntu on a standard Bios system. IIRC, the boot mechanism that will be used is the same as the one that was used to boot of the installation medium. If this is true, we should add a note telling folks that they **must** install using UEFI.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+
+# mkdir ~/kernel-build
+# cd ~/kernel-build
+# wget  https://download.clearlinux.org/releases/current/clear/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
+# sudo apt-get install rpm2cpio
+# rpm2cpio linux-pk414-sos-4.14.41-39.x86_64.rpm | cpio -idmv 
+# sudo cp -r ~/kernel-build/usr/lib/modules/4.14.41-39.pk414-sos/ /lib/modules/
+
+Step #8, Deploying the hypervisor
+Mount the EFI partition and verify you have the following files, then add the ACRN hypervisor and Clear Linux kernel to the EFI
```
gvancuts: We should refer to this kernel as the Service OS kernel and not the Clear Linux kernel.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# cd ~/kernel-build
+# wget  https://download.clearlinux.org/releases/current/clear/x86_64/os/Packages/linux-pk414-sos-4.14.41-39.x86_64.rpm
+# sudo apt-get install rpm2cpio
+# rpm2cpio linux-pk414-sos-4.14.41-39.x86_64.rpm | cpio -idmv 
+# sudo cp -r ~/kernel-build/usr/lib/modules/4.14.41-39.pk414-sos/ /lib/modules/
+
+Step #8, Deploying the hypervisor
+Mount the EFI partition and verify you have the following files, then add the ACRN hypervisor and Clear Linux kernel to the EFI 
+Partition, 
+# umount /boot/efi
```
gvancuts: I think I asked this before but I don't remember the answer :-( ... why do we need to unmount the partition first to remount it right away (somewhere else)?

doc/howtos/using_ubuntu_as_sos.rst
gvancuts: We should also recommend that people do **not** have `secure boot` on.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# sudo cp acrn-hypervisor/build/acrn.efi /mnt/EFI/acrn	
+
+Step #9, Configure the EFI firmware to boot the ACRN hypervisor by default
+# sudo -s
+# efibootmgr -c -l "\EFI\acrn\acrn.efi" -d /dev/sda -p 1 -L "ACRN Hypervisor                                                                                                    NUC" -u "bootloader=\EFI\ubuntu\grubx64.efi"
+Then you can see the result of boot order. Also you can check the boot order with below command:
+# efibootmgr -v            	
+or setting boot order with this command to make sure the acrn.efi is the first boot option:
+# efibootmgr -o x, x       	
+Note: by default, the “ACRN Hypervisor” you just added should be the first one to boot. But please make sure it is, or select it
```
gvancuts: We've just added `ACRN Hypervisor NUC` (and not `ACRN Hypervisor`)

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
+Using Ubuntu as the Service OS with the ACRN hypervisor
+
+
+Based on the ACRN open source project and getting started guide. This documentation is to guide people how to boot Ubuntu system as the 
+Service OS with the ACRN hypervisor. By default, in the getting started guide documentation of the ACRN open source project, Clear Linux
+is being used as the Service OS and User OS on top of the ACRN hypervisor. 
+Here we are trying to boot Ubuntu system as the Service OS and launch Clear Linux as User OS with the ACRN hypervisor on official 
+supported NUC, playing steps as below:
```
gvancuts: Can we be more specific and indicate the model (using a link). I would like that we add a Baby Canyon (KBL NUC) to the list soon and at that point, the "official supported NUC" becomes ambiguous.

The instructions below would likely need to be tweaked a little for a different NUC.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# efibootmgr -c -l "\EFI\acrn\acrn.efi" -d /dev/sda -p 1 -L "ACRN Hypervisor                                                                                                    NUC" -u "bootloader=\EFI\ubuntu\grubx64.efi"
+Then you can see the result of boot order. Also you can check the boot order with below command:
+# efibootmgr -v            	
+or setting boot order with this command to make sure the acrn.efi is the first boot option:
+# efibootmgr -o x, x       	
+Note: by default, the “ACRN Hypervisor” you just added should be the first one to boot. But please make sure it is, or select it 
+manually after press “F10” when system booting up.   
+
+Step #10, Deploy the SOS kernel
+# sudo cp ~/kernel-build/usr/lib/kernel/org.clearlinux.pk414-sos.4.14.41-39 /boot/acrn/
```
gvancuts: This is the first time you use `/boot/acrn` so I expect it does not exist. Should it be created?

I'm also wondering if the target folder shouldn't really be somewhere under `/mnt/`, where the ESP is mounted at the moment (and the kernel needs to be somewhere on the ESP for Grub to be able to load it).

doc/howtos/using_ubuntu_as_sos.rst
gvancuts: Enabling the time-out and changing the default boot target should always be done by editing `/etc/default/grub` (and **then** running `sudo update-grub`). If something (bug or something else) is forcing us to edit the `/boot/grub.cfg` directly, we should explain this here.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+ + #set timeout_style=hidden
+ -  set timeout=0
+ + set timeout = 10 
+ # Fallback hidden-timeout code in case the timeout_style feature is
+  # unavailable.
+
+Step #13, Check the boot process 
+Reboot system to see if the acrn.efi is loaded successfully, and if the Menu of Grub being shown to select “ACRN ubuntu SOS”.  
+after boot successfully, the system will start to Ubuntu Desktop successfully.  to check if the system is running on virtualization 
+evn and the ACNR hypervisor is running.
```
gvancuts: evn?

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+This output means the boot flow is correct and the SOS system with ACRN is running well.
+ 
+
+Step #14, Deploy Clear Linux as the UOS 
+Download clear linux UOS from download.clearlinux.org
+#cd /root
+# wget https://download.clearlinux.org/releases/22780/clear/clear-22780-kvm.img.xz
+# unxz clear-22780-kvm.img.xz
+Copy  launch_uos.sh from source code
+# cp ~/Download/acrn-hypervisor/devicemodel/samples/nuc/launch_uos.sh /root
```
gvancuts: This assumes that the `git clone` operation above was run from within the `~/Download` folder which may not be true (I never clone anything under that folder for example).

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+Download standard PK kernel with below command:
```
gvancuts: I would expand `PK` by `Production Kernel (PK)`

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# wget https://download.clearlinux.org/releases/22780/clear/x86_64/os/Packages/linux-pk414-standard-4.14.47-44.x86_64.rpm
+# rpm2cpio linux-pk414-standard-4.14.47-44.x86_64.rpm | cpio -idmv
+
+Use the following steps to update clearlinux kernel:
```
gvancuts: We need to be more verbose here and refer to the kernel as the Linux PK kernel (standard). What we're doing is updating a Clear Linux image by **manually** installing a kernel that will make it work as a User OS.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# sudo losetup -f -P --show /root/clear-22789-kvm.img  
+# sudo mount /dev/loop0p3 /mnt
```
gvancuts: I realize we've done a reboot but I see a risk that someone does not do the reboot, proceeds with the instructions on how to set-up a User OS and here they're using a mountpoint that is already in use.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# sudo cp -r /root/usr/lib/modules/4.14.47-44.pk414-standard /mnt/lib/modules/
+# sudo cp -r /root/usr/lib/kernel /lib/modules/
```
gvancuts: I don't understand what this step does. You are copying something from the Linux PK standard kernel on the Service OS rootfs, is that really what you mean to do?

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+NUC6CAYH:~$ dmesg | grep ACRN
+[    0.000000] Hypervisor detected: ACRN
+[    0.862942] ACRN HVLog: acrn_hvlog_init
+
+This output means the boot flow is correct and the SOS system with ACRN is running well.
+ 
+
+Step #14, Deploy Clear Linux as the UOS 
+Download clear linux UOS from download.clearlinux.org
+#cd /root
```
gvancuts: This will fail if you're not root so we should add a preceding `sudo -s` instruction and get rid of all the `sudo` below, until the end of the doc (since we are now `root`).

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# sudo mount /dev/loop0p3 /mnt
+# sudo cp -r /root/usr/lib/modules/4.14.47-44.pk414-standard /mnt/lib/modules/
+# sudo cp -r /root/usr/lib/kernel /lib/modules/
+# sudo umount /mnt
+# sync
+
+You may Need to a permission issue
+# sudo chmod 777 /dev/acrn_vhm
+You also need to install a package.
+# sudo apt-get instal iasl
```
gvancuts: s/instal/install (and no `sudo`)

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# sudo cp /usr/bin/iasl /usr/sbin/iasl
+
+Step 15, Deploy device module - acrn-dm
```
gvancuts: We should move this step to be right after step 6 (where we build the repo), and use `make install` (which *should* work and install not only the devicemodel but also the tools).

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# cp ~/download/acrn-hypervisor//build/devicemodel/acrn-dm /usr/bin
+
+Step 16, modify launch_uos.sh 
+-s 3,virtio-blk,/root/clear-22780-kvm.img
+-k /lib/modules/kernel/org.clearlinux.pk414-standard.4.14.47-44
```
gvancuts: I think this is the reason why you copied stuff from the linux PK standard package on the Service OS rootfs. I would actually prefer if we kept the rootfs 'clean' and make the `launch_uos.sh` script pick the kernel from `/root/...` directly (so we can also save the copy operation above)

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
+Using Ubuntu as the Service OS with the ACRN hypervisor
```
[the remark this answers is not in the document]
ailun258: i created it and followed up it to boot up system , and i was able to have complete system setup, let's check again to make sure there is no technical issues.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+
+Based on the ACRN open source project and getting started guide. This documentation is to guide people how to boot Ubuntu system as the 
+Service OS with the ACRN hypervisor. By default, in the getting started guide documentation of the ACRN open source project, Clear Linux
+is being used as the Service OS and User OS on top of the ACRN hypervisor. 
+Here we are trying to boot Ubuntu system as the Service OS and launch Clear Linux as User OS with the ACRN hypervisor on official 
+supported NUC, playing steps as below:
+
+Step #1,  Installing a Ubuntu system on the NUC with standard installing steps. 
+For example with Ubuntu 16.04.4 LTS distribution.
+Download it here: https://www.ubuntu.com/download/desktop
```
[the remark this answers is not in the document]
ailun258: 16.04 and 18.04 worked either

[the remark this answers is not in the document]
ailun258: should change 16.04.4 to   16.04

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+Burn Ubuntu 16.04.4 image to a USB key
```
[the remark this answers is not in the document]
ailun258: good,  i agree

doc/howtos/using_ubuntu_as_sos.rst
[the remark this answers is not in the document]
ailun258: i have different thoughts with you,  i prefer to have one-stop information in one page.

doc/howtos/using_ubuntu_as_sos.rst
```diff
@@ -0,0 +1,184 @@
...
+# make PLATFORM=uefi
+The build results are found in the build directory.
+For example:
+# ls acrn-hypervisor/build
+devicemodel
+hypervisor
+tools
+
+Steps #7,  Preparing Service OS kernel 
+You can download latest Service OS kernel from below links:
```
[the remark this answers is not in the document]
ailun258: yes, you may feel confusing, but for me, it's not, 
we may need to explain more here like you mentioned, 
we are Using Ubuntu system as the Service OS, but we need to customize SOS kernel to adapt hypervisor,   2 options. 1) build Ubuntu kernel with source and hypervisor patches,  i don't recommend this way for community new user.  so i  didn't mention this.
2) replace Ubuntu kernel with Clear based Service OS kernel.  this is so easy to replace,  we are using this way

doc/howtos/using_ubuntu_as_sos.rst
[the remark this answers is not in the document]
dbkinder: While "one stop information" is a nice goal, duplicating information creates a problem to keep them in sync, and as @gvancuts noted, the information is already out of sync :(

doc/howtos/using_ubuntu_as_sos.rst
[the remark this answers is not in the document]
ailun258: yes, this is going to be a moving target,  lets use 4.14.41-39 as the example only

doc/howtos/using_ubuntu_as_sos.rst
ailun258: yes, if you are using exact version,  let's using linux-pk414-sos-4.14*

doc/howtos/using_ubuntu_as_sos.rst
ailun258: no solution so far, we don't need to modify it manually until  having good tool for that

ailun258:
@dbkinder  can you please submit a .rst version?   let's move to new version to modify

doc/howtos/using_ubuntu_as_sos.rst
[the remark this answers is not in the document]
ailun258: no,  at this moment,  we need to have some little changes because release model is not very good. when the release model trend stable.  we don't need to do any modification manually

doc/howtos/using_ubuntu_as_sos.rst
ailun258: yes, you got it,
it's hard to implement what said now because clear team provides kvm,img for user OS, if there is a UOS.img provided from Clear team, that would be easy for us to save the copy operation

doc/howtos/using_ubuntu_as_sos.rst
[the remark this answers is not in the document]
gvancuts: > no, at this moment, we need to have some little changes because release model is not very good. when the release model trend stable. we don't need to do any modification manually

I don't understand this comment @ailun258 , can you elaborate some more?

doc/howtos/using_ubuntu_as_sos.rst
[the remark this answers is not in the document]
gvancuts: No disagreement that the steps and methodology followed is the correct one. We just need to explain some more what we're doing here (and why) because people not familiar with ACRN will be confused about this.

doc/howtos/using_ubuntu_as_sos.rst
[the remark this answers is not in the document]
gvancuts: > yes, this is going to be a moving target, lets use 4.14.41-39 as the example only

Can you update the URL so they all point to a location where that kernel can be downloaded from?

doc/howtos/using_ubuntu_as_sos.rst
gvancuts: Keep the exact name and fix the URL instead.

doc/howtos/using_ubuntu_as_sos.rst
gvancuts: > no solution so far, we don't need to modify it manually until having good tool for that

Are you saying it's not possible to make these modifications in `/etc/default/grub` directly and then run `sudo update-grub`?

doc/howtos/using_ubuntu_as_sos.rst
gvancuts: I don't understand what is hard about this. You could just adjust the `-k` parameter in the launch script to point at the kernel in `/root/usr/lib/kernel`? Any reason why this wouldn't work?

gvancuts:
@ailun258 , I have done an initial conversion to ReST as per your request and took the opportunity to also address some of the comments that were made during the review. The version I have sent you (via PR https://github.com/ailun258/acrn-hypervisor/pull/1) will still need a bit of work but I would suggest we continue from there.

jren1:
@gvancuts  I just read https://github.com/gvancuts/acrn-hypervisor/blob/0f27f711b3accb0fe8eeef750411d15b27092f36/doc/howtos/tech/using_ubuntu_as_sos.rst, it looks pretty good

dbkinder:
@gvancuts @ailun258 , at this point it seems that closing this PR and using Geoffroy's version is a better way to go.  There are a few fixes needed in Geoffroy's version, but it would be best to have Geoffroy submit a new PR and we pick up there (or Ailin needs to update his PR with Geoffroy's version).

dbkinder:
OK, tutorials works for me too :)

dbkinder:
@ailun258 can you accept @gvancuts edits so we can continue editing his version rather than yours.

dbkinder:
Picking up the doc and editing with #595

PR #492, opened 2018-06-25
hv: irq: fix type for vector in ioapic setup

lyan3:
Fix the type for vector in ioapic setup, which is a potential problem:
- return VECTOR_INVALID instead of false in irq_desc_alloc_vector()
  when irq is out of range;
- change variable type from int to uint32_t for vector, and correct
  the returned value check.


NanlinXie:
Verify +1

PR #494, opened 2018-06-25
guest.h: Condition operator fix of foreach_vcpu loop

yuchuyang:
The for loop statement should work in both & and &&, but it seems
make more sense to have && when executing a condtion.


NanlinXie:

lijinxia:
OK to verify

lijinxia:
Verified+1

PR #497, opened 2018-06-26
HV: refine 'init_mtrr()'

yonghuah:
to avoid reading operations on MTRR registers if
no MTRR feature support on current platform.


NanlinXie:

hypervisor/arch/x86/mtrr.c
```diff
@@ -92,6 +92,9 @@ void init_mtrr(struct vcpu *vcpu)
 	union mtrr_cap_reg cap = {0};
 	int i;
 
+	if (!cpu_has_cap(X86_FEATURE_MTRR))
```
dongyaozu: @yonghuah the general policy so far used is to check the minimal hardware capability at beginning of AVRN. MTRR is one of the minimal capability. Can you check if we do this kind of minimal hardware cap check at beginning?  If yes, we don't need this change

yonghuah: @dongyaozu  not check MTRR cap when cpu boot /hardware_detect_support( ),  do we need move it to hardware_detect_support()?

dongyaozu:
That is better for all those basic features we must rely on.

yonghuah:
@dongyaozu  Have updated, thanks.

PR #501, opened 2018-06-26
HV: treewide: convert suffix ULL to UL

junjiemao1:
It is already assumed that ''long'' has 8-bytes, and thus there is no need to use ULL to indicate a 8-byte unsigned constant.

This patch changes all ULL suffixes found in the hypervisor to UL. Also remove a trailing empty line in trusty.h.


NanlinXie:

PR #504, opened 2018-06-27
fix "function return type inconsistent"

HuiHuangShi:
MISRA C required function return type should be consistented.


NanlinXie:

PR #506, opened 2018-06-27
DM: virtio_rnd: fix rnd->fd and vbs_k->fd leak

zhijianli88:
Previously, either rnd->fd or vbs_k->fd isn't be closed in some cases.
this patch will close them in time.

V2: fix vbs_k->fd leak as well


NanlinXie:

PR #507, opened 2018-06-27
refactor host pirq file structures

lyan3:
- rename related source files
- add a header for common data struct and APIs


NanlinXie:

PR #509, opened 2018-06-27
HV: Remove 'register' prefix for data type

yonghuah:
- it is unnecessary.


NanlinXie:

PR #514, opened 2018-06-27
hv:Change parameter to udelay to uint64_t

rarindam:
Modified the parameter from int to uint64_t.
Reasong being, the delay should be greater than 0.
uint64_t gives greater range.


NanlinXie:

dongyaozu:
There are discussions in the mailinglist regarding the udelay: we want to use uint32_t to avoid the overflow in multiply operations and use uint32_t for tsc_hz.  I am waiting for the owner to send a new patch.

jren1:
@rarindam Please rebase the patches

PR #517, opened 2018-06-28
DM/samples: remove clocksource assignment

shuox:
HPET reading is much slower than TSC. Using it will hurt SOS performance
a lot, and then the whole system performance.

Remove the strict assignment from cmdline. Then SOS kernel will pickup
TSC as default clocksource.


lijinxia:
OK to verify

lijinxia:
Verified+1

PR #521, opened 2018-06-28
HV:common:transfer local variable type

HuiHuangShi:
The local variable type should be transfer to non-basic type,
chaned it to length-prefix(uint32_t,int32_t ...) type.

Char *type or char array type which used to pointer a string
will be keeped.

V1->V2 add extra comments.


NanlinXie:

lijinxia:
Verified+1

PR #522, opened 2018-06-28
Dev/fix start failure

gzhai7:
All patches and the fix of UOS start failure got acked already. This is just re-sending after reverting.


NanlinXie:

lijinxia:
Verified+1

PR #527, opened 2018-06-29
HV: vpic: cleanup integral types

junjiemao1:
This patch series cleans up the integral types in vpic by changing icw_num and rd_cmd_reg to uint8_t, converting port addresses and width to be unsigned in alignment with the general port I/O interfaces, and fixing the uses of boolean variables.


NanlinXie:

lijinxia:
Verified+1

PR #529, opened 2018-06-29
build: Using id tool to get builder username

shuox:
Some build environment might has no USER pre-defined. So use id tool to
get builder username instead of USER environemnt.

Also add a version cleanup for tools to keep them updated.


NanlinXie:

lijinxia:
Verified+1

PR #535, opened 2018-06-29
HV: vpic: using uint8_t to represent PIC pins

junjiemao1:
This patch series change the type of pic pins to be uint8_t and clean up the
uses of integral types, constants and variables by explicitly spelling out
narrowing conversions and unsigned constants related to this type change.

v1 -> v2:

    * Rename the number of pins per PIC to NR_VPIC_PINS_PER_CHIP, and the number
      of pins of 2 cascading PICs to NR_VPIC_PINS_TOTAL.
    * Using "%hhu" instead of "%d" in format string when a pin is expected.


NanlinXie:

lijinxia:
Verified+1

PR #537, opened 2018-06-29
misc: fix cbc_attach blocks APL NUC boot for 20 seconds

liuyuan-GitHub:
This patch resolved systemd be blocking for 20 seconds due to waiting the
"READY=1" from cbc_attach. And cbc_attach will retry 60 sec at the worst case
which can be reproduced on non-IOC boards.

This is one temporary solution. The final solution need modify SBL to export IOC
ACPI device through ACPI table, and also Linux IOC CBC driver need trigger
uevent which should be the trigger of launching cbc_attach service.

This patch changes the service type to simple instead of notify due to systemd is
needn't wait cbc related services. And change Restart to no due to cbc_attach
already has retry mechanism.

These changes are also reasonable for the board integrated IOC component.
So they should be keep even the long term final solution implemented.

Reviewed-by: Alex Du <alek.du@intel.com>

### DIFF
--- a/misc/cbc_attach/Makefile
+++ b/misc/cbc_attach/Makefile
@@ -1,7 +1,7 @@
 OUT_DIR ?= .
 
 all:
-	gcc -o $(OUT_DIR)/cbc_attach cbc_attach.c -lsystemd
+	gcc -o $(OUT_DIR)/cbc_attach cbc_attach.c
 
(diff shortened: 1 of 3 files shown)


NanlinXie:

lijinxia:
Verified+1

PR #541, opened 2018-06-29
HV: Remove vm->attr.name

KaigeFu:
We define attr.name in struct vm and named as ("vm_%d", attr.id). attr.name only
be used in debug tool vm_list. It does't deserve to do so in OS created flow
(aka function create_vm). It's better to handle this in vm_list tool.

Acked-by: Eddie Dong <eddie.dong@inte.com>


NanlinXie:

lijinxia:
Verified+1

PR #550, opened 2018-06-30
HV: bug fix update RTE when interrupt is unmasked

abdulrehman-MentG:
Current code incorrectly assumes that RTL cannot be updated while
the interrupt is masked.

Signed-off-by: Abdul Rehman <abdulrehmanee010@gmail.com>

### DIFF
--- a/hypervisor/arch/x86/guest/vioapic.c
+++ b/hypervisor/arch/x86/guest/vioapic.c
@@ -406,12 +406,13 @@ vioapic_write(struct vioapic *vioapic, uint32_t addr, uint32_t data)
 
 		/* remap for active: interrupt mask -> unmask
 		 * remap for deactive: interrupt mask & vector set to 0
+		 * remap for trigger mode change
+		 * remap for polarity change
 		 */
-		data64 = vioapic->rtbl[pin].reg;
-		if ((((data64 & IOAPIC_RTE_INTMASK) == IOAPIC_RTE_INTMCLR)
-		  && ((last & IOAPIC_RTE_INTMASK) == IOAPIC_RTE_INTMSET))
-		  || (((data64 & IOAPIC_RTE_INTMASK) == IOAPIC_RTE_INTMSET)
-		  && ((vioapic->rtbl[pin].reg & IOAPIC_RTE_INTVEC) == 0))) {
+		if ( (changed & IOAPIC_RTE_INTMASK) != 0UL ||
+		     (changed & IOAPIC_RTE_TRGRMOD) != 0UL ||
+		     (changed & IOAPIC_RTE_INTPOL ) != 0UL ) {
+
 			/* VM enable intr */
 			struct ptdev_intx_info intx;
 


NanlinXie:

hypervisor/arch/x86/guest/vioapic.c
```diff
@@ -393,14 +393,10 @@ vioapic_write(struct vioapic *vioapic, uint32_t addr, uint32_t data)
 			vioapic_send_intr(vioapic, pin);
 		}
 
-		/* remap for active: interrupt mask -> unmask
-		 * remap for deactive: interrupt mask & vector set to 0
+		/*
+		 * remap for any update to RTL
```
dongyaozu: Why it is called RTL ?

hypervisor/arch/x86/guest/vioapic.c
```diff
@@ -393,14 +393,10 @@ vioapic_write(struct vioapic *vioapic, uint32_t addr, uint32_t data)
...
 		 */
-		data64 = vioapic->rtbl[pin].reg;
-		if ((((data64 & IOAPIC_RTE_INTMASK) == IOAPIC_RTE_INTMCLR)
-		  && ((last & IOAPIC_RTE_INTMASK) == IOAPIC_RTE_INTMSET))
-		  || (((data64 & IOAPIC_RTE_INTMASK) == IOAPIC_RTE_INTMSET)
-		  && ((vioapic->rtbl[pin].reg & IOAPIC_RTE_INTVEC) == 0))) {
+		if ( changed != 0U ) {
```
dongyaozu: @fyin1 can you have a review?

fyin1: From the log posted from Abdul, the RTE is changed from edge triggered to low level triggered. This is unusual usage. We need to dig more to find out why this operation happens with this device first.

hypervisor/arch/x86/guest/vioapic.c
abdulrehman-MentG: It's a mistake it should be RTE (redir table entry)

dongyaozu:
WHat does RTL mean? shouldn't it be RTE?

dongyaozu:
Can you fix all the RTL to RTE? There are many RTL ...

hypervisor/arch/x86/guest/vioapic.c
```diff
@@ -393,14 +393,10 @@ vioapic_write(struct vioapic *vioapic, uint32_t addr, uint32_t data)
...
-		/* remap for active: interrupt mask -> unmask
-		 * remap for deactive: interrupt mask & vector set to 0
+		/*
+		 * remap for any update to RTE
 		 */
-		data64 = vioapic->rtbl[pin].reg;
-		if ((((data64 & IOAPIC_RTE_INTMASK) == IOAPIC_RTE_INTMCLR)
-		  && ((last & IOAPIC_RTE_INTMASK) == IOAPIC_RTE_INTMSET))
-		  || (((data64 & IOAPIC_RTE_INTMASK) == IOAPIC_RTE_INTMSET)
-		  && ((vioapic->rtbl[pin].reg & IOAPIC_RTE_INTVEC) == 0))) {
```
JasonChenCJ: we want to keep the filter,  and would not like to trigger ptdev remapping for every change. Could we think about what field need be reflect to native, and just add them into check filter?
Like here we can add  IOAPIC_RTE_TRGRMOD, IOAPIC_RTE_INTPOL, IOAPIC_RTE_DESTMOD, IOAPIC_RTE_DELMOD and IOAPIC_RTE_DEST.

abdulrehman-MentG: Yes we can add the fields. For this specific case being discussed, only IOAPIC_RTE_TRGRMOD and IOAPIC_RTE_INTPOL would suffice. But my original question remains what is the reason behind original filter of only mask change?
Related mailing list: https://lists.projectacrn.org/g/acrn-users/topic/question_regarding/22866607?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,22866607

JasonChenCJ: @abdulrehman-MentG, usually, we should not change RTE setting during unmask, and for performance consideration, we don't want ptdev remapping run every time vioapic RTE got update. That's why we only take care mask change.

abdulrehman-MentG:
I have restored the filter with IOAPIC_RTE_TRGRMOD and IOAPIC_RTE_INTPOL updates allowed.

dongyaozu:
@abdulrehman-MentG  Why you made 3 patches/commit? Didn't rebase ?

abdulrehman-MentG:
@dongyaozu I have made a single patch and rebased. Thanks!

lijinxia:
Verified+1

PR #552, opened 2018-07-02
DM: Remove 'strictio' from UOS bootargs' options

yonghuah:
- UOS will boot fail if 'strictio' is enabled ('-e' option), in this
   case (with '-e'), device model will block all PIO accesses whose
   handlers were not registered, after that, device model program will
   exit, hence UOS boot fail.

   actually, such kind of accesses exist, e.g. UOS would program
   PIT registers (port address: 0x43) if hpet is disabled.

 - For debug, we can trap unexpected PIO access in 'default_inout()'

### DIFF
--- a/devicemodel/core/inout.c
+++ b/devicemodel/core/inout.c
@@ -88,13 +88,12 @@ register_default_iohandler(int start, int size)
 }
 
 int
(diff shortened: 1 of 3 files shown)


NanlinXie:

lijinxia:
Verified+1

PR #555, opened 2018-07-02
HV: treewide: drop debug-only helpers in release build

junjiemao1:
There are multiple helpers for collecting info to be printed on the ACRN uart
console. They are unreachable code in release builds in which the uart console
is removed.

To define a precise boundary for safety-related activities, this patch wraps the
declarations and definitions to these helpers with "#ifdef HV_DEBUG" so that
these unreachable APIs will be dropped in release builds.

v1 -> v2:

    * Fix coding style: no empty lines between #ifdef and the wrapped code.
    * Also drop get_rte_info() in ioapic.c, which is solely used by
      get_ioapic_info().


NanlinXie:

lijinxia:
Verified+1

PR #556, opened 2018-07-02
HV: Remove unnecessary vm0 check in vm0 specific func

KaigeFu:
Function prepare_vm0_memmap_and_e820 and init_vm0_boot_info are specific for vm0.
There is no need to check is_vm0 again in those functions.

This patch remove the unnecssary checks.

v1 -> v2:
   - Add pre-condition comment before the function as Junjie's suggestion.


NanlinXie:

lijinxia:
Verified+1

PR #557, opened 2018-07-02
misc: disable cbc_lifecycle service always restart by systemd

liuyuan-GitHub:
Due to cbc_lifecycle service already has its own retry mechanism,so change the
Restart type to "no".

This change is also indirectly avoid cbc_lifecycle always restart for the
non-IOC boards. Actually, there has one defect for current IOC CBC architecture
due to lack IOC hardware detection mechanism. The related guys are already aware
it and WIP.

Reviewed-by: Alex Du <alek.du@intel.com>


NanlinXie:

PR #558, opened 2018-07-02
Update return value type as unint16 for bit operations

wuxyintel:
Update return value type as unint16 for bit operations
Update the type of bit index, vcpu id, softirq id, 
event id as uint16_t to reduce type conversion, the 
following steps are doing:
(1) Update return value type for fls and clz;
(2) Update return value type for fls64 and clz64;
(3) Update return value type for ffs64 and ffz64.
For each step, for the related caller,the type of
variables which store return value is updated as
uint16_t and return value checking is updated.

V1-->V2:
        INVALID_BIT_INDEX instead of INVALID_NUMBER;
        INVALID_CPU_ID instead of INVALID_PCPU_ID or INVALID_VCPU_ID;
        "%hu" is used to print vcpu id (uint16_t);
        Add 'U/UL' for constant value as needed.
V2-->V3:
       Clean up extra updates, such as add "%hu";
       For bit operations, it return INVALID_BIT_INDEX
       directly when input parameter value is zero;
       Add 'U/UL' for constant value as needed.
V3-->V4:
        Clean up some comments.

Xiangyang Wu (3):
  HV:treewide:Update return type for bit operations fls and clz 
  HV:treewide:Update return type for function fls64 and clz64
  HV:treewide:Update return type of function ffs64 and ffz64

### DIFF
--- a/hypervisor/arch/x86/guest/guest.c
+++ b/hypervisor/arch/x86/guest/guest.c
@@ -75,11 +75,11 @@ inline struct vcpu *get_primary_vcpu(struct vm *vm)
 
 inline uint64_t vcpumask2pcpumask(struct vm *vm, uint64_t vdmask)
 {
-	int vcpu_id;
+	uint16_t vcpu_id;
 	uint64_t dmask = 0;
 	struct vcpu *vcpu;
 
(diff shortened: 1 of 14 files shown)


NanlinXie:

lijinxia:
Verified+1

PR #562, opened 2018-07-03
HV: timer: convert tsc_hz to tsc_khz

junjiemao1:
Representing TSC freqeuency in Hz generally needs a uint64_t with range checks
upon use to avoid integer overflow. This patch series uses KHz instead to reduce
the TSC frequency to a uint32_t.


NanlinXie:

lijinxia:
Verified+1

PR #564, opened 2018-07-03
cbc: fix the install issue when do make misc-install only

alekdu:
The install -D is misleading people since it only creates the leading
directories.

The problem was hidden due to we always call make install and the
destination folder was already created in previous install script.


NanlinXie:

lijinxia:
Verified+1

PR #565, opened 2018-07-03
tools: acrn-crashlog: replace ldconfig with pkg-config in Makefile

chengangc:
This patch is to fix the dependency issue with autospec.

Using ldconfig in the autospec build environment is not
going to work as the packages do not have the cache
generate after a rpm install. This patch replaces ldconfig
with pkg-config to check the existance of the libraries.

Reviewed-by: Jin Zhi <zhi.jin@intel.com>
Reviewed-by: Liu Xinwu <xinwu.liu@intel.com>
Acked-by: Zhang Di <di.zhang@intel.com>

### DIFF
--- a/tools/acrn-crashlog/Makefile
+++ b/tools/acrn-crashlog/Makefile
@@ -20,16 +20,27 @@ export BUILDDIR
 export CC
 export RM
 
-EXTRA_LIBS = -lsystemd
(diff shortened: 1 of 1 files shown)


NanlinXie:

lijinxia:
Verified+1

PR #568, opened 2018-07-03
HV: Don't pass retval to vmm_emulate_instruction()

KaigeFu:
We pass retval to vmm_emulate_instruction and assign the return value to retval
at the same time. The retval will be passed to mmio_read/write finally as memarg
and the functions don't use the parameter actually. Apparently, we misused the
retval.

This patch fix it by passing 'NULL' to vmm_emulate_instruction.


NanlinXie:

lijinxia:
Verified+1

PR #574, opened 2018-07-04
HV:treewide:Update cpu_id type as uint_16 and rename related cpu_id

wuxyintel:
There are extra type conversion in the HV since cpu_id
type is uint32_t and the return value type of get_cpu_id
is uint16_t. BTW, the name of cpu_id is not clear enough
to express its usage.

So the following updates are made in this patch:
Update cpu_id type as unit_16 to reduce type casting;
Update related temporary variables type;
Update related print argument;
Change the input parameter name of interrupt_init as
cpu_id to keep align with function implement;
Rename related cpu_id as pcpu_id as needed.


NanlinXie:

lijinxia:
Verified+1

PR #576, opened 2018-07-04
HV:CPU:Update logical_id type as uint_16 and rename logical_id

wuxyintel:
There are extra type conversion in the HV since logical_id
type is uint32_t and the input argument of bitmap operations
is uint16_t. BTW, the name of logical_id is not clear enough
to express its usage.

So the following updates are made in this patch:
Update logical_id type as unit_16 to reduce type casting;
Update related print argument;
Rename related logical_id as pcpu_id as needed.

Note: logical_id in the interrupt_init definition have been
updated in cpu_id cleanup patch.


NanlinXie:

lijinxia:
Verified+1

PR #580, opened 2018-07-04
doc: add support for kerneldoc API tools

dbkinder:
Some ACRN kernel components are using the API documentation methods of
the Linux kernel.  While they use Sphix for generating
their documentation, they don't use doxygen to collect the API
information as we do for the rest of the project.  Instead they use their own tools called "kerneldoc".
This PR incorporates those tools into our documentation build process.

There is a prescribed directory structure for this to work: that the
acrn-hypervisor and acrn-kernel repos are cloned to sibling folders,
e.g.:

    projectacrn
       acrn-hypervisor
       acrn-kernel

so that documentation references from acrn_hypervisor/doc can access the
source code in ../../acrn-kernel to do the kerneldoc processing.  A full
display of the kerneldoc API material for a source file in the
acrn-kernel tree can be done using a sphinx extention directive:

    .. kernel-doc:: /tools/virtio/linux/scatterlist.h

where the assumed root of these file references is ../../acrn-kernel.

The format for kerneldoc comments is documented in
https://www.kernel.org/doc/html/latest/doc-guide/kernel-doc.html and
references to kerneldoc API material in .rst files is documented in
https://www.kernel.org/doc/html/latest/doc-guide/kernel-doc.html#including-kernel-doc-comments

Without options, the kernel-doc directive includes all documentation
comments from the source file. With options, you can display subsets of
these comments.

The intention is to limit use of kerneldoc comments to the acrn-kernel
repo and not use them elsewhere within the ACRN project (where doxygen
comments are expected.)


NanlinXie:

gvancuts:
This looks good to me!! Thanks @dbkinder 

A couple of notes from my basic testing:
- The leading `/` in the file path is not required (`kernel-doc` directive)
- I used a couple of files in `include/linux/vhm/` when playing with this but was getting warnings... are we already using `kernel-doc` there or is it in fact standard doxygen stuff?
- I think it would be good to add a paragraph in https://projectacrn.github.io/latest/developer-guides/doc_guidelines.html on how to use both these directives (`kernel-doc` and `doxygengroup`, etc.) This can most definitely wait until later, just like I wouldn't cover the topics extensively, just put an intro and links to online references for more details

None of my notes should block merging this so :+1:

dbkinder:
Hmmm... I wasn't expecting to see the hypervisor include files in the acrn-kernel repo?  are they duplicated from the acrn-hypervisor repo?  In any case kerneldocs processing only works on files with kerneldocs comments, and I see the file acrn-kernel/include/linux/vhm/acrn_common.h is a copy of what's in the acrn-hypervisor repo and has doxygen comments.  How odd?  Is this intended?

I thought we were getting the GVT-g interfaces from acrn-kernel but now I'm confused.  What are the hypervisor includes doing over in acrn-kernel too?

PR #583, opened 2018-07-05
HV:treewide:string assigned to const object

wuxyintel:
In the hypervisor, some strings are assigned to non const
object, this violates MISRA C:2012.

Update the type of the object as const type since it always
points to string.


NanlinXie:

lijinxia:
Verified+1

PR #585, opened 2018-07-05
DM: Add write sync in fwrite

yang8621:
An immediate reset or power down will cause a loss of write content.
The cause is the data write to disk is at cache within a short
time window before it's synced to storage media.
An explicit fsync() forces to sync the data to storage to prevent
the data loss of such immediate reset.

Signed-off-by: Huang Yang <yang.huang@intel.com>
Signed-off-by: duminx <minx.du@intel.com>


NanlinXie:

lijinxia:
Verified+1

PR #586, opened 2018-07-05
hv:fix "missing for discarded return value" for memset

mingqiangchi:
No need to check the return value for memset
code like this:
int a(void) {
	return 0;
}
int b(void){
	a();
}
fix as follow:
int a(void) {
	return 0;
}
int b(void){
	(void)a();
}

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -368,7 +368,7 @@ void bsp_boot_init(void)
 	start_tsc = rdtsc();
 
 	/* Clear BSS */
-	memset(_ld_bss_start, 0, _ld_bss_end - _ld_bss_start);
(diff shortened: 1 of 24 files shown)


NanlinXie:

dongyaozu:
@mingqiangchi @junjiemao1   Nothing wrong with the patch, but I am wondering if we should just use void for the return value of memset, or we need to add "void" in every caller side of memset.

junjiemao1:
We have met quite some cases where pre-condition is significant now. We plan to systematically revisit these constraints after MISRA C cleanup, when there'll be some component-level interface specifications ready.

lijinxia:
@mingqiangchi Could you please rebase and fix the conflicts?
Conflicting files
hypervisor/arch/x86/guest/instr_emul.c

lijinxia:
Verified+1

PR #590, opened 2018-07-05
hv:fix "missing for discarded return value" for memcpy_s and strcpy_s

mingqiangchi:
It will print error information inside memcpy_s/strcpy_s/strncpy_s if
the parameter is invalid, then no need to check
the return value of memcpy_s/strcpy_s/strncpy_s for the caller
code like this:
int a(void) {
return 0;
}
int b(void){
a();
}
fix as follow:
int a(void) {
return 0;
}
int b(void){
(void)a();
}


NanlinXie:

lijinxia:
Verified+1

PR #596, opened 2018-07-05
HV: uart16550.c: check the denominator before use

yuchuyang:
To prevent the value is devided by zero, checks the denominator
before the calculation. Adding the if statement to check before use.
If the baud_rate is equal to zero, using default baud_rate.


NanlinXie:

lijinxia:
Verified+1

PR #598, opened 2018-07-06
hv: fix 'Recursion in procedure calls found'

shiqingg:
Here is how the recursion might happen:

                     when there is something wrong
                     |
sbuf_put -> memcpy_s -> pr_err -> do_logmsg
   |                                 |
   -----------------------------------

Replace 'pr_err' with 'ASSERT' in 'memcpy_s' to break this kind of
recursion.

### DIFF
--- a/hypervisor/include/debug/assert.h
+++ b/hypervisor/include/debug/assert.h
@@ -12,7 +12,6 @@ void __assert(uint32_t line, const char *file, const char *txt);
 
 #define ASSERT(x, ...) \
 	if (!(x)) {\
-		pr_fatal(__VA_ARGS__);\
 		__assert(__LINE__, __FILE__, "fatal error");\
 	}
 #else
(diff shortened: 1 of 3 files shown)


hypervisor/include/debug/assert.h
```diff
@@ -12,7 +12,7 @@ void __assert(uint32_t line, const char *file, const char *txt);
 
 #define ASSERT(x, ...) \
 	if (!(x)) {\
-		pr_fatal(__VA_ARGS__);\
+		printf(__VA_ARGS__);\
```
dongyaozu: Remove this printf, it is not necessary.

shiqingg:
Hi Eddie, I have removed the pr_fatal in ASSERT in the latest version.

Thanks,
Shiqing

lijinxia:
Verified+1

PR #599, opened 2018-07-06
HV: ioapic: number of interrupt pins should be max_rte + 1

junjiemao1:
The 23:16 bits in the IOAPIC version register is the highest entry in the I/O
redirection table, which is 1 smaller than the number of interrupt input
pins. But currently we use these bits directly as the number of pins.

This patch abstracts the pin count calculation code and adds 1 to the max rte
entry number to get the right pin count.


NanlinXie:

lijinxia:
Verified+1

PR #601, opened 2018-07-06
launch_uos: remove enable_initial_modeset for UOS

feijiang1:
Some UOS need use FB device to display, while initial_modeset will disable fb,
so we need remove correlation between plane restriction and initial modeset
in kernel driver, also don't need set enable_initial_modeset in launch script.

Reviewed-by: Min He <min.he@intel.com>


NanlinXie:

lijinxia:
Verified+1

PR #602, opened 2018-07-06
HV:CPU: Add 'U/UL' for unsigned const value

wuxyintel:
According to MISRA C:2012, suffix 'U/UL' shall be for
unsigned const value, the member of enum variable should
not be used to compare with integer variable.

Add 'U/UL' for unsigned const value in the CPU module;
Use Macro insteading of enum feature_word since the member
of feature_word is used to compare with integer variable;
Use hex number insteading of Macro in the assembly code.

V1-->V2:

        Update the suffix of some constant value as 'UL'
        according to its'storage variable;
        Split MACRO updates used in the assembly code
	in other patch.


NanlinXie:

lijinxia:
Verified+1

PR #605, opened 2018-07-06
HV:CPU:Constant values replace with CPU MACRO

wuxyintel:
MISRA C requires that all unsigned constants should have
the suffix 'U/UL'(e.g. 0xffU), but the assembler may not
accept such C-style constants.

To work this around, all  unsigned constants must be
explicitly spells out in assembly with a comment tracking
the original expression from which the magic number is
calculated.


NanlinXie:

PR #608, opened 2018-07-06
virtual RPMB key handover

YadongQi:


NanlinXie:

hypervisor/arch/x86/trusty.c
```diff
@@ -461,6 +468,14 @@ bool initialize_trusty(struct vcpu *vcpu, uint64_t param)
 	/* save Normal World context */
 	save_world_ctx(&vcpu->arch_vcpu.contexts[NORMAL_WORLD]);
 
+	if (boot_param.version == TRUSTY_VERSION_2) {
+		/* copy rpmb_key from OSloader */
+		memcpy_s(&g_key_info.rpmb_key[0][0], 64,
+				&boot_param.rpmb_key[0], 64);
+		memset(&boot_param.rpmb_key[0], 0, 64);
```
dongyaozu: 2nd / 3th parameter of memset is uint8/ size_t, which is unsigned integer. Please use 0U, 64U...  This is MISRA-C requirement, and we need to enforce now.

hypervisor/arch/x86/trusty.c
```diff
@@ -7,7 +7,8 @@
 #include <hypervisor.h>
 #include <hkdf.h>
 
-#define TRUSTY_VERSION 1
+#define TRUSTY_VERSION   1
+#define TRUSTY_VERSION_2 2
```
dongyaozu: version is uint32_t, right? use 1U if it is.

hypervisor/arch/x86/trusty.c
```diff
@@ -425,14 +426,20 @@ bool initialize_trusty(struct vcpu *vcpu, uint64_t param)
...
+	 *        the interface, need to add back after the integration done.
+	 */
+	/*
 	if (sizeof(struct trusty_boot_param) !=
 			boot_param.size_of_this_struct) {
 		pr_err("%s: sizeof(struct trusty_boot_param) mismatch!\n",
 			__func__);
 		return false;
 	}
+	*/
```
dongyaozu: Why leave deadcode here?

hypervisor/arch/x86/trusty.c
YadongQi: @dongyaozu Has fixed in latest version.

hypervisor/arch/x86/trusty.c
YadongQi: Fixed in latest version.

hypervisor/arch/x86/trusty.c
YadongQi: We need to add back these code after OSloader update the version to 2. Currently, both version 1 and 2 need to be supported, otherwise, the boot will fail.

Please check the Comments before these codes:
+	/* FIXME: Temporarily comments out the size check since need to extend
+	 *        the interface, need to add back after the integration done.
+	 */
+	/*

YadongQi:
@dongyaozu Yes, 2 Versions for temporary backward compatibility.

lijinxia:
OK to verify

PR #616, opened 2018-07-09
add a log change file for v0.1 release

ailun258:
from this log change, people should see detail update from March RC release to v0.1


ailun258:
those commits ID should have a specific matched issue fix or feature, but today we are not using github issue to track issues and features, so we don't have complete info.
@dbkinder @gvancuts @hwang37 @jren1 
so i am suggesting to keep commit ID and simple description on it  in release notes only.  
let's improve step by step,

jren1:
I agreed with David, that long git history is really useful here. Intead, we need to abstract the key information from that long list.

dbkinder:
we'll include this into the 0.1 release notes (for now)

PR #623, opened 2018-07-09
HV: instr_emul: Return -EINVAL if inst_len equals 0

KaigeFu:
inst_len with 0 length should be treated as invaild value. This patch move 0 length
check to the begainning of the function and return -EINVAL if inst_len == 0.


NanlinXie:

lijinxia:
Verified+1

PR #624, opened 2018-07-09
HV: ioapic: integral type cleanup

junjiemao1:
This series cleans up the MISRA C violations related to integral types, including the conversion of ioapic pins to uint8_t in addition to a few random implicit variable narrowing and signedness conversions.


NanlinXie:

PR #625, opened 2018-07-09
lpc: resolve UOS boot-up issue caused by lpc.

liuyuan-GitHub:
The lpc dm causes that UOS can't boot if the parameters are set incorrectly,
it is not friendly to users.

This patch optimizes the lpc error handle flow. UOS always can boot successfully
whatever the lpc settings are.


NanlinXie:

lijinxia:
Verified+1

PR #626, opened 2018-07-09
Clean up integer type conversion violations in CPU module

wuxyintel:
Check CPU module by static analysis, there are some 
integer type conversion violates mainly including narrow 
integer conversions and miss 'U/UL' suffix for unsigned 
constant value.

(1) Resovle narrow integer conversions by using explicit 
    conversion or add 'U/UL';
(2) Resovle miss 'U/UL' suffix for unsigned constant value
    by add 'U/UL' as needed.


NanlinXie:

PR #633, opened 2018-07-09
HV: No assignment inside while loop condition

yuchuyang:
The assigment should be done outside while loop condition. To fix
it, one assigment initializaion and update statement of for loop
have been applied. The only while loop reminds to avoid very long
for loop expression.


lijinxia:
Verified+1

PR #634, opened 2018-07-09
HV: printf.c: Do not return expression contains subexpression

yuchuyang:
Operation should be done outside return expression. Moved the assignment out from return expression.


NanlinXie:

lijinxia:
OK to verify

lijinxia:
Verified+1

PR #636, opened 2018-07-09
HV: Assignment should not mix with operator

yuchuyang:
Removed the postfix and prefix operation in assignment expression.
Noncompliant code example:
1) *a++ = *b ++;
2) a = arr[--b];


lijinxia:
Verified+1

PR #637, opened 2018-07-10
samples: add cma reservation into sos cmdline

JasonChenCJ:
it's needed for trusty memory reservation.


lijinxia:
Verified+1

gvancuts:
I'm a little confused by this one, I thought we were moving away from `cma` to exclusively use HugeTLB instead. This seems contrary to that direction, can you explain why?

dongyaozu:
@gvancuts  Yes, we moved to hugetlb solution. But, now there is a need to allocate contiguous memory pages for trusty world.  It is 16MB in size, and we want the VHM to allocate that part, as one of the simplest solution.  Do you have any suggestion ?

gvancuts:
Hi @dongyaozu , I don't have any better suggestion :-)

I was really more curious about it and wanted to make sure I understood the rationale behind it. Thanks for the explanation!

dongyaozu:
@gvancuts in this case, I will merge this patch first, and revisit later on.

PR #638, opened 2018-07-10
Dm lanuch

yakuizhao:
Add the boot option if i915 guc for UOS. 
One is to fix the incorrect memory allocation on UOS
Another is to disable the guc_submission explicitly so that it is aligned with SOS.


NanlinXie:

lijinxia:
Verified+1

PR #641, opened 2018-07-10
HV: treewide: cleanup type-related violations in headers

junjiemao1:
This series fix violations to MISRA C rules related to integral types in the
headers. Violations that are identified in *.c files are not in the scope.

v1 -> v2:

    * Change the prototype of mmio-related functions, instead of keeping the
      explicit conversions internal.
    * Drop INT_ROUNDUPx macros since they are never used.


NanlinXie:

junjiemao1:
Tags removed.

lijinxia:
Verified+1

PR #648, opened 2018-07-10
HV:MM:add 'U/UL' suffix for unsigned contant value

wuxyintel:
In the current memory module, there are many constatn value
without U/UL suffix, it is reported as MISRA C violations by
static analysis tool.

Add 'U/UL' suffix for unsigned contant value in memory module
as needed.

Note:In the most case, CPU_PAGE_SIZE(0x1000) is used as
unsigned integer contant value, so CPU_PAGE_SIZE is defined
as unsigned integer contant value, and it is safety converted
into unsigned long type according to MISRA C standard.


lijinxia:
Verified+1

PR #651, opened 2018-07-10
hv: fix 'Pointer arithmetic is not on array'

shiqingg:
Violation 'Pointer arithmetic is not on array' occurs in following
statement:
        *flags |= fl[pos - flagchars];

char flagchars[] is a well defined array.
It could be used directly to avoid the pointer arithmetic.

v1 -> v2:
 * use uint32_t rather than uint8_t for the index in order to let the
    type aligned with sizeof(flagchars)
 * add brackets to the then-block


NanlinXie:

PR #656, opened 2018-07-10
Dockerfile: remove security by-pass in Clear Linux Dockerfile

### LINKED ISSUE #654: The Dockerfile turns off security
RUN git config --global http.sslVerify false

that's obviously never an acceptable solution for anything

gvancuts:
"RUN git config --global http.sslVerify false" was needed with
a cloud-based CI system else it failed to clone the repository.

The project does not use that cloud-based CI system and hence
this workaround (security by-pass) is not needed.

Fix #654 


NanlinXie:

PR #660, opened 2018-07-11
HV: vmx: Change variable field to uint32_t

KaigeFu:
With field defined as uint64_t, there are some implicit type conversions when
calling exec_vmwrite/read functions which takes a uint32_t for the first parameter.

This patch fix it by change field to uint32_t.

v1 - > v2:
  - Keep the field as it is and change the defination of field to uint32_t.


NanlinXie:

TianhuaSun:
OK to verify

lijinxia:
OK to verify

PR #662, opened 2018-07-11
HV:check secure/normal world for EPTP when gpa2hpa

mcao6:
for secure and normal world has different EPTP, in secure world,
it could trap to hypervisor and call gpa2hpa function. So it need
check if it is in normal or secure world for EPTP selection.

Detailed explanation:
Currently, trusty OS does not handle interrupt, so when an interrupt
is delivering in trusty, trusty will set it to LAPIC IRR register by
inject self_ipi() and then call world_switch to switch to Android,
So android will receive/handle the interrupt properly.Since the vLAPIC
is enabled in ACRN, so when trusty try to inject self_ipi(), APIC-ACCESS
vmexit will happen. Then ACRN will do instruction decode/emulation, so
the GPA2HPA will fail since page walk is using nworld_eptp. It is probability
an issue.

This patch will handle this issue.


NanlinXie:

TianhuaSun:
OK to verify

lijinxia:
OK to verify

PR #664, opened 2018-07-11
HV: build: make relocation-related code configurable

junjiemao1:
The relocation feature relies on the ld option "-z noreloc-overflow" which is
only available for binutils >= 2.27, while on Ubuntu 16.04 or older the default
version of binutils is 2.26.

This patch wraps the relocation code with a configurable macro and make it
undefined by default to avoid default build failures.

NOTE: This is just a hotfix. The code dropped with undefined CONFIG_RELOC needs
to be reviewed by the original author of this feature. Checks to the binutils
version will also follow up.

### DIFF
--- a/hypervisor/Makefile
+++ b/hypervisor/Makefile
@@ -63,12 +63,14 @@ LDFLAGS += -Wl,--gc-sections -nostartfiles -nostdlib
 LDFLAGS += -Wl,-n,-z,max-page-size=0x1000
 LDFLAGS += -Wl,-z,noexecstack
 
+ifeq (y, $(CONFIG_RELOC))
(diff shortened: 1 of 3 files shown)


NanlinXie:

ZideChen0:
Hi Junjie, thank you very much for the fix. I will I done more research on the binutils. The way to wrap the CONFIG_RELOC  looks good to me. Or in this way. 

void reloc()
{
#ifdef CONFIG_RELOC
....
#endif
} 

Thanks,
Zide

jren1:
@NanlinXie  That patch must be merged for v0.1

TianhuaSun:
OK to verify

TianhuaSun:
OK to verify

PR #665, opened 2018-07-11
Dockerfiles: don't enforce http.sslVerify to be false

jren1:


NanlinXie:

jren1:
Since @gvancuts has another patch to fix it, close this one

PR #667, opened 2018-07-11
HV:treewide:Add 16-bit atomic operations and update vpid type

wuxyintel:
There are some integer type conversions reported by static
analysis tool for vcpu id, number of created vcpus, and
vpid, to reduce these type conversions, redesign vcpu id,
number of created vcpus, and vpid type as uint16_t as per
their usage, related 16-bit atomic operations shall be
added in HV.
MISRA C requires that all unsigned constants should have the suffix 'U'
(e.g. 0xffU), but the assembler may not accept such C-style constants.

Add 16-bit atomic add/dec/store operations;
Update temporary variables type and parameters type of
related caller;
Update vpid type as uint16_t;
Replace Macro with constant value for CPU_PAGE_SIZE.

Note: According to SDM A.10, there are some bits defined
in the IA32_VMX_EPT_VPID_CAP MSR to support the INVVPID
instruction, these bits don't mean actual VPID, so
the vpid field in the data struct vmx_capability doesn't
be updated.

V1--V2:
	update comments for assembly code as per coding style;


NanlinXie:

TianhuaSun:
OK to verify

PR #669, opened 2018-07-11
 tools:acrn-crashlog: Detect and classify the crash in ACRN and kernel

liuxinw:
Since ACRN has the capability to reboot and reboot reason is available
in SOS, acrnprobe could detect the crash of acrn and SOS kernel.

List of added crash types:

1. ACRNCRASH		- crashed in hypervisor, this detection depends on
 			  files in /tmp/acrnlog_last(provided by acrnlog).
2. IPANIC		- crashed in SOS kernel, this detection depends on
			  pstore.
3. SWWDT_IPANIC		- crashed in SOS kernel and reboot reason is wdt.
4. HWWDT_UNHANDLE	- only recognize reboot reason is global, there is no
 			  further clues that it's a SOS kernel crash or a
			  hypervisor crash.
5. SWWDT_UNHANDLE	- only recognize reboot reason is wdt, there is no
			  further clues that it's a SOS kernel crash or a
			  hypervisor crash.
 6. UNKNOWN		- only recognize reboot reason is warm, there is no
			  further clues that it's a SOS kernel crash or a
			  hypervisor crash.

### DIFF
--- a/tools/acrn-crashlog/acrnprobe/android_events.c
+++ b/tools/acrn-crashlog/acrnprobe/android_events.c
@@ -400,6 +400,7 @@ static void get_last_line_synced(const struct sender_t *sender)
 	int sid;
 	int ret;
 	struct vm_t *vm;
+	char *p;
 	char vmkey[ANDROID_WORD_LEN];
 	char vm_name[32];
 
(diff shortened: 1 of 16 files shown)


NanlinXie:

TianhuaSun:
OK to verify

lijinxia:
OK to verify

PR #670, opened 2018-07-11
HV: trusty: revise trusty_boot_param structure

YadongQi:
Per new design of trusty memory allocation: VHM will reserve contiguous
memory for trusty when DM launch guest with trusty enabled. And OSloader
will relocate trusty to 511G directly and pass trusty's base/entry to HV
by trusty_boot_param when call HC_INIITIALIZE_TRUSTY.

So in this patch:
  1. Extend trusty_boot_param to support addr/entry above 4G.
  2. Remove size check for old version compatibility.

Acked-by: Zhu Bing <bing.zhu@intel.com>

### DIFF
--- a/hypervisor/arch/x86/trusty.c
+++ b/hypervisor/arch/x86/trusty.c
@@ -417,6 +417,7 @@ static bool init_secure_world_env(struct vcpu *vcpu,
 bool initialize_trusty(struct vcpu *vcpu, uint64_t param)
 {
 	uint64_t trusty_entry_gpa, trusty_base_gpa, trusty_base_hpa;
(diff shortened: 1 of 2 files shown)


NanlinXie:

TianhuaSun:
OK to verify

YadongQi:
@anthonyzxu the param is already copied to HV before use it. Please check line#425 as below.
```
 	if (copy_from_gpa(vcpu->vm, &boot_param, param, sizeof(boot_param))) {
 		pr_err("%s: Unable to copy trusty_boot_param\n", __func__);
 		return false;
 	}
```

PR #672, opened 2018-07-11
hv: irq: fix 'Pointer arithmetic is not on array'

shiqingg:
- Define the array for 'struct irq_desc' with its size
  The array size for 'struct irq_desc' is constant and will not be
   changed.

- Use array rather than a pointer to avoid the pointer arithmetic.


NanlinXie:

TianhuaSun:
OK to verify

PR #680, opened 2018-07-12
hv: fix bug in some embedded assembly code in vmx

binbinwu1:
The patch fixes the issue when SOS can't boot using HV release version.

In current code, the assembly code for "sgdt" & "sidt" is not right.
The operand is output, not input.

Also, current code use "rdmsr" instruction to read MSR_IA32_SYSENTER_CS,
which doesn't sepcify the clobbered registers it uses.
This patch uses API msr_read to read MSR_IA32_SYSENTER_CS.

### DIFF
--- a/hypervisor/arch/x86/vmx.c
+++ b/hypervisor/arch/x86/vmx.c
@@ -702,7 +702,7 @@ static void init_guest_state(struct vcpu *vcpu)
 
 		/* Base *//* TODO: Should guest GDTB point to host GDTB ? */
 		/* Obtain the current global descriptor table base */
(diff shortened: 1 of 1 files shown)


NanlinXie:

binbinwu1:
This patch is to fix the issue that sos can't boot when using HV release version.
So please help to check the patch fixed the issue using HV release version.

liuwei987:
latest source code with the patch, can fix the "sos not boot with release version". 
So v+1.

lijinxia:
Verified+1

PR #687, opened 2018-07-12
hv: further fix to configurable relocatoin

ZideChen0:
commit ia23549aa915e7dc2c ("build: make relocation-related code
configurable") adds CONFIG_RELOC to make relocation configurable

This patch corrects the behavior when CONFIG_RELOC is disabled
- Don't use "CFLAGS += -fpie" and put back "LDFLAGS += -static"
- __emalloc(): forced to allocate exactly the asked address, which is
  CONFIG_RAM_START

### DIFF
--- a/hypervisor/Makefile
+++ b/hypervisor/Makefile
@@ -39,10 +39,15 @@ CFLAGS += -ffunction-sections -fdata-sections
 CFLAGS += -fshort-wchar -ffreestanding
 CFLAGS += -m64 -mno-mmx -mno-sse -mno-sse2 -mno-80387 -mno-fp-ret-in-387
 CFLAGS += -mno-red-zone
-CFLAGS += -static -nostdinc -nostdlib -fno-common
+CFLAGS += -nostdinc -nostdlib -fno-common
 CFLAGS += -O2 -D_FORTIFY_SOURCE=2
 CFLAGS += -Wformat -Wformat-security
+ifeq (y, $(CONFIG_RELOC))
 CFLAGS += -fpie
+else
+CFLAGS += -static
+endif
+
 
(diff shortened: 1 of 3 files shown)


NanlinXie:

dongyaozu:
@binbinwu1 @junjiemao1  can you review?

hypervisor/Makefile
```diff
@@ -42,7 +42,10 @@ CFLAGS += -mno-red-zone
 CFLAGS += -static -nostdinc -nostdlib -fno-common
```
junjiemao1: Shall we add '-static' only when CONFIG_RELOC is not defined, just like what you've done for the linking options?

ZideChen0: The GCC manual states that "-static":  On systems that support dynamic linking, this prevents linking with the shared libraries. On other systems, this option has no eﬀect.

So it seems to me that this doesn't matter to the hypervisor code relocation.

junjiemao1: Per LD manual, "-static" means exactly the same thing for the linker (i.e. "Do not link against shared libraries"), and the description of this option does not mention anything about relocation, either. Would you please describe the impact of "-static" on relocation a bit further so that everyone can understand why it is critical for the linker but not the compiler?

ZideChen0: Hi Junjie,

You are right. It's safer to wrap "CFLAGS += -static" with CONFIG_RELOC. Thanks for pointing it out.

Thanks,
Zide

junjiemao1:
Looks good to me. And @rarindam has verified that this patch fixes his linking issue.

lijinxia:
Verified+1

PR #693, opened 2018-07-13
HV: vlapic: integral type cleanup

junjiemao1:
A summary of changes made in this patch:

* Use uint32_t for register offsets.
* Use uint32_t for isrvec_stk_top which equals to the number of vectors in the
  stack.
* Use local variables to hold struct members before using them, to avoid
  confusions from static checkers.
* Use uint32_t for tpr, though the higher 24 bits are reserved.
* For-loops with loop variables counting down to 0 are converted to while-loops
  to make the loop variable unsigned when necessary.
* 'U' suffix for constants in unsigned contexts.
* Local variables are typed so that explicit casts are minimized.

Issues yet to be closed include

* Conversions due to atomic APIs taking/returning signed values.
* Conversions due to exec_[vmread|vmwrite] always work on 64-bit integers.
* IOAPIC RTE constants are 64-bit instead of 32-bit.

### DIFF
--- a/hypervisor/arch/x86/guest/vlapic.c
+++ b/hypervisor/arch/x86/guest/vlapic.c
@@ -51,22 +51,23 @@
 #if VLAPIC_VERBOS
 static inline void vlapic_dump_irr(struct vlapic *vlapic, char *msg)
 {
-	int i;
+	uint32_t i;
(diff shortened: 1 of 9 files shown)


NanlinXie:

lijinxia:
Verified+1

PR #698, opened 2018-07-13
hv: cpu: fix 'Pointer arithmetic is not on array'

shiqingg:
Use the array for lapic_id directly to avoid the unnecessary pointer
arithmetic.

With current implementation,
  lapic_id_base is always a byte array with CPU_PAGE_SIZE elements

What this patch does:
 - replace 'uint8_t *lapic_id_base' with 'uint8_t
   lapic_id_array[CPU_PAGE_SIZE]' to make the boundary explicit
 - add a range check to ensure that there is no overflow

 v2 -> v3:
 * update the array size of lapic_id_array per discussion with Fengwei

 v1 -> v2:
 * remove the unnecessary range check in parse_madt in cpu.c

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -28,7 +28,7 @@ volatile uint16_t up_count = 0U;
 uint64_t pcpu_active_bitmap = 0UL;
 
 /* TODO: add more capability per requirement */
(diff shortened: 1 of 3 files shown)


NanlinXie:

lijinxia:
Verified+1

PR #700, opened 2018-07-13
dm: virtio-net: add variable name in function declaration

ssqre:
We should keep variable name in function declaration. It makes
things clearer and easier to be understood.


lijinxia:
Verified+1

PR #704, opened 2018-07-13
hv: fix the potential dead loop in _parse_madt

shiqingg:
With current implementation:
If the case 'entry->length < sizeof(struct acpi_subtable_header)' is
triggered, it will lead to a dead loop.

What this patch does:
Gracefully return when this case is triggered.

Why:
If 'entry->length < sizeof(struct acpi_subtable_header)', it means that
there is no valid 'struct acpi_subtable_header' starting from the entry.
There must be something wrong. It should not happen in normal case.

v1 -> v2:
 * Remove the unacceptable ASSERT, just gracefully return


NanlinXie:

PR #708, opened 2018-07-13
hv: set guest segment base to zero if VCPU does not start in real mode

ZideChen0:
In non real mode, for segment registers other than CS, the guest segment
base should be zero, otherwise the guest's segmentation results in wrong
effective addresses.

Linux boots with the wrong segment registers (base address in hidden part),
because it happens that it assigns the segment registers before using any
of them, which effectively reloads the segment base addresses from GDT.


NanlinXie:

lijinxia:
Verified+1

PR #712, opened 2018-07-16
HV: consider header depenencies when rebuilding

junjiemao1:
With Kconfig, it would be common to rebuild the hypervisor with a few
configuration symbols changed. But for a proper rebuild, users are required to
execute ''make clean'' first, which deletes the configuration file at all.

This patch leverages the compiler to generate a target that add involved headers
as dependencies to a specific file. The inclusion of version.h is dropped in
per_cpu.h to avoid forcefully rebuild of all files since version.h is created on
build.


NanlinXie:

dongyaozu:
@jren1 @binbinwu1 please comment on the patches

PR #713, opened 2018-07-16
hv: cpu state update should be moved just before halt.

fyin1:
The cpu offline requester monitor the target cpu state
to detect whether target cpu is put to offline already.

So we should only update the cpu state to offline after
all other operations are finished.

Reviewed-by: Eddie Dong <Eddie.dong@intel.com>


NanlinXie:

lijinxia:
Verified+1

PR #715, opened 2018-07-16
Acrnd

taoyuhong:
Acrn-manager: add Acrnd daemon
 
acrnd: a service daemon which coordinates VM lifecycle with SOS lifecycle,
handles IOC timed wakeup/poweron

Within this PR we updated comments and debug strings


NanlinXie:

PR #720, opened 2018-07-16
HV: Fixes index out of bounds for addressing irq.

javedm:
NR_MAX_IRQS is defined as 272 and IRQ_INVALID as 273 which implies
that 272 is a valid irq number. Code in hypervisor/arch/x86/irq.c
also checks that if irq > NR_MAX_IRQS then return error. But the
irq arrays that are allocated address the range 0-271. In this
case, an illegal access can occur at run time when irq_desc_array[]
or irq_count[] is accessed with index 272. This fix updates the size
of these arrays to allow indexing the range 0-272.

Signed-off-by: Madeeha Javed <madeeha_javed@mentor.com>


NanlinXie:

dongyaozu:
Good catch up, thanks!
The intention is not to view NR_MAX_IRQS as a valid IRQ.   @junjiemao1 can you comment?

javedm:
Thanks @dongyaozu. I can update the patch if NR_MAX_IRQS is intended to be invalid.

junjiemao1:
Thanks for pointing this out, @javedm! As @dongyaozu says, NR_MAX_IRQS is the number of valid IRQs from 0. Thus valid IRQs range from 0 to (NR_MAX_IRQS-1).

Actually two adjacent names are quite misleading: NR_MAX_VECTOR is the largest valid vector, while NR_MAX_IRQS is the number of valid IRQs and equals to the first invalid IRQ. Following the naming convention of Linux, we tend to use NR_IRQS for the total number of IRQs, and MAX_IRQ for the largest valid IRQ number. Thus I would suggest the following changes as a thorough fix to this issue:

1. Rename NR_MAX_IRQS to NR_IRQS,
2. For range checks, use 'x >= NR_IRQS' to check if x is invalid, 'x < NR_IRQS' to check if x is valid.
3. Optionally define IRQ_INVALID as 0xffffffffU (instead of NR_IRQS+1 or NR_IRQS), following our convention that special values for data of unsigned type uses the largest numbers of that type. (INVALID_CPU_ID, VPIC_INVALID_PIN, IOAPIC_INVALID_PIN as examples.)

javedm:
@dongyaozu yes, I havent updated the patch yet. Working on it.

lijinxia:
Verified+1

PR #725, opened 2018-07-17
Uos system reset

fyin1:
This change needs HV and DM both should be updated together on your device. Otherwise, the UOS can't boot correctly.


NanlinXie:

lijinxia:
OK to verify

PR #726, opened 2018-07-17
HV: vioapic: fixes to integral-type-related violations

junjiemao1:
This patch series fix MISRA C violations related to integral types, with:

    * the first unifying access patterns to ioapic/vioapic RTEs to avoid
      implicit narrowing, and

    * the second fixing rest of the violations that can be fixed locally.

v1 -> v2:

    * Instead of two 32-bit ''low'' and ''high'', define a union that allows
      either 32-bit or 64-bit accesses to RTEs.
    * Drop duplicated definitions to IOAPIC register offsets.
    * Drop the ''size'' parameter of vioapic_mmio_[read|write] and
      vioapic_mmio_rw since vioapic_mmio_access_handler() ensures that size is
      always 4.


lijinxia:
Verified+1

PR #727, opened 2018-07-17
 HV:vtd change the macro to the inline function

HuiHuangShi:
Function like macro changed to be inline function to limit
the return type and parameter type.


NanlinXie:

lijinxia:
Verified+1

PR #731, opened 2018-07-18
HV: instr_emul: Replace ASSERT/panic with pr_err

KaigeFu:
ASSERT/panic are called when we get invalid values. It is a little bit strict.
This patch replaces ASSERT/panic with pr_err and return -EINVAL.

v1 -> v2:
  - v1 patch name (HV: instr_emul: Remove unnecessary check in decode_xxx)
  - keep the check and replace ASSERT/panic with pr_err.


NanlinXie:

lijinxia:
Verified+1

PR #732, opened 2018-07-18
tools: replace payload[0] of struct mngr_msg with an union

taoyuhong:
We replace payload[0] of struct mngr_msg with an union, which contains all payload data. So that type cast for mngr_send_msg() is no longer needed. And we can avoid potential out-of-boundary memory accessing and using of uninitialized variable.


NanlinXie:

lijinxia:
Verified+1

PR #741, opened 2018-07-19
HV: treewide: terminate 'if .. else if' constructs with 'else'

junjiemao1:
MISRA C requires that a 'if' statement followed by one or more 'else if'
statement shall be terminated by an 'else' statement which contains either
side-effect or a comment, to ensure that conditions are considered
exhaustively.

Note that a simple 'if' statement is not required to be terminated by 'else'.

This patch fixes such violations by either refactoring the code or add the
'else' statement with either a comment (describing why this case can be skipped)
or logging the event. It may not be satisfactory for the release version where
logging is no-op, but properly handling these non-trivial cases is out of the
scope of this patch.

v1 -> v2:

    * Fix unintended semantic changes in add_(msix|intx)_remapping and
      io_instr_vmexit_handler.
    * Simplify boolean checks in vpic_ocw2.
    * Rephrase the comment in strtol_deci.


NanlinXie:

lijinxia:
Verified+1

PR #742, opened 2018-07-19
hv: check eptp value before calling free_ept_mem()

lyan3:
There is chance to destroy vm whose ept table was not created, for
example, dm fails to setup mem for vm. In this case, it's better
to check the value of eptp before calling free_ept_mem() rather than
relying on assertion in the function.


NanlinXie:

lijinxia:
Verified+1

PR #748, opened 2018-07-19
HV: cpuid: Disable Intel RDT for guest OS

shuox:
Now the Intel RDT emulation is working in progress. So disable it
for temporary solution to avoid guest OS running with incorrect RDT
configuration.


NanlinXie:

PR #749, opened 2018-07-20
hv: correct the way to check if a MSR is a fixed MTRR register

ZideChen0:
The fixed MTRR MSR numbers are not contiguous, so it's not correct
to justify it by checking if it falls in certain range.

This patch fixes this issue by removing is_fixed_range_mtrr()
and use get_index_of_fixed_mtrr() to loop fixed_mtrr_map[]
and compare individual MSR values.


NanlinXie:

lijinxia:
@ZideChen0 Could you please rebase and fix the conflicts?

PR #751, opened 2018-07-20
HV: final cleanup of arch/x86/guest/ and common/

junjiemao1:
This series unifies the type of bit-field members and hypercall return values as
well as suppress most integral-related violation reports by adding explicit
conversions, abstracting sub-expressions, adding parenthesis, etc.

There are still a small number of reports that cannot be suppressed in a trival
way, mostly involving pre-conditions and assumed implementation-defined
behavior. They'll fall in the scope of a separate document for explanation.


NanlinXie:

PR #753, opened 2018-07-20
hv: correct the way to check if a MSR is a fixed MTRR register

ZideChen0:
The fixed MTRR MSR numbers are not contiguous, so it's not correct
to justify it by checking if it falls in certain range.

This patch fixes this issue by removing is_fixed_range_mtrr()
and use get_index_of_fixed_mtrr() to loop fixed_mtrr_map[]
and compare individual MSR values.

Also removed the unused function get_subrange_end_of_fixed_mtrr()

### DIFF
--- a/hypervisor/arch/x86/mtrr.c
+++ b/hypervisor/arch/x86/mtrr.c
@@ -23,6 +23,7 @@ struct fixed_range_mtrr_maps {
 };
 
 #define MAX_FIXED_RANGE_ADDR	0x100000UL
+#define FIXED_MTRR_INVALID_INDEX	-1U
 static struct fixed_range_mtrr_maps fixed_mtrr_map[FIXED_RANGE_MTRR_NUM] = {
 	{ MSR_IA32_MTRR_FIX64K_00000, 0x0U, 0x10000U },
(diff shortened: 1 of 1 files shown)


NanlinXie:

hypervisor/arch/x86/mtrr.c
```diff
@@ -37,21 +37,17 @@ static struct fixed_range_mtrr_maps fixed_mtrr_map[FIXED_RANGE_MTRR_NUM] = {
...
 
-static bool is_fixed_range_mtrr(uint32_t msr)
-{
-	return (msr >= fixed_mtrr_map[0].msr)
-		&& (msr <= fixed_mtrr_map[FIXED_RANGE_MTRR_NUM - 1U].msr);
-}
-
-static uint32_t get_index_of_fixed_mtrr(uint32_t msr)
+static bool get_index_of_fixed_mtrr(uint32_t msr, int *index)
 {
```
dongyaozu: I prefer u return index here, but define an XXX_INVALID_INDEX...  The caller need to check if it is INVALID_INDEX

ZideChen0: Sure, will do

lijinxia:
Verified+1

PR #757, opened 2018-07-21
HV: fix bug of restore rsp context

jsun26intel:
We should use movq to restore rsp instead of mov.


NanlinXie:

lijinxia:
Verified+1

PR #759, opened 2018-07-23
DOC:GSG: Fix few mistakes about updating acrn.conf and efibootmgr opt…

wuxyintel:
…ions

During setting up ACRN hypervisor, SOS and UOC on the
intel® NUC (NUC6CAYH) board, there are few mistakes
about updating acrn.conf and efibootmgr options, this
will lead to fail to set up ACRN hypervisor, SOS and UOS
on the NUC.

About updating acrn.conf, add "hugepagesz=1G hugepages=2"
About efibootmgr options, string parameter of option "-L"
needs to add double quotation marks; "uart=disabled" in
the option "-u" since there is no serial port on NUC.

### DIFF
--- a/doc/getting-started/apl-nuc.rst
+++ b/doc/getting-started/apl-nuc.rst
@@ -150,7 +150,7 @@ partition. Follow these steps:
 
    .. code-block:: none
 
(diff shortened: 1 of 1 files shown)


ailun258:
looks good to me

PR #760, opened 2018-07-23
HV: make: check CONFIG_RELEASE=y for release build

junjiemao1:
Currently we reply on CONFIG_RELEASE=n to determine if a debug version should be
built. Though this holds due to the configuration overriding in kconfig.mk, this
only applies to make >= 3.82. For make 3.81, $(eval $(call xxx)) does not work
as expected, and thus CONFIG_RELEASE is undefined for debug builds.

Instead of checking CONFIG_RELEASE=n, this patch checks CONFIG_RELEASE=y
instead, which is guaranteed without relying on the behavior of $(eval $(call
xxx)).

### DIFF
--- a/hypervisor/Makefile
+++ b/hypervisor/Makefile
@@ -199,7 +199,7 @@ endif
 endif
 
 C_OBJS := $(patsubst %.c,$(HV_OBJDIR)/%.o,$(C_SRCS))
-ifeq ($(CONFIG_RELEASE),n)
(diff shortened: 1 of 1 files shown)


NanlinXie:

dongyaozu:
@jren1 @binbinwu1 @gvancuts

binbinwu1:
Reviewed +1

lijinxia:
OK to verify

PR #761, opened 2018-07-23
dm: vsbl: update vsbl to support watchdog reset

binbinwu1:
Update vsbl image to get watchdog reset status according to the
emulated watchdog device (i6300esb)


NanlinXie:

lijinxia:
OK to verify

dongyaozu:
binary check in need to be very conservative... Can we postpone this to some time later so that we may merge more change request. @binbinwu1

PR #762, opened 2018-07-23
HV:treewide:Fixing pointer castings

wuxyintel:
In the hypervisor, there are many casts from
an void pointer to integer pointer, then from
integer pointer to structure pointer.
These pointer castings are detected by static analysis
tool. All pointer casts are violations, There are
some duplicated pointer cast. This will make deviation
analysis complex.
BTW, there are one useless pointer casting and one
wrong pointer casting in the hypervisor.

Remvoe duplicated pointer casts to make deviation analysis
simple;
Remove one useless pointer casting;
Update one wrong pointer casting.

Note: There are many void type pointer casts, non-void type
pointer is casted to void type pointer, char type pointer casts,
non-char type pointer is casted to char type pointer. These pointer
casting is need by the memory management module, IO moudle etc.
Deviation analysis will be made and recoded in the analysis report.

V1-->V2:
	Fix mixing pointer and array voilation.
V2-->V3:
	Remvoe pointer casting from integer pointer into
	non-void/non-char pointer directly;
	Remove redundant type conversion.

### DIFF
--- a/hypervisor/arch/x86/guest/guest.c
+++ b/hypervisor/arch/x86/guest/guest.c
@@ -140,7 +140,7 @@ static int _gva2gpa_common(struct vcpu *vcpu, struct page_walk_info *pw_info,
 	uint32_t i;
 	uint64_t index;
 	uint32_t shift;
-	uint8_t *base;
+	void *base;
 	uint64_t entry;
 	uint64_t addr, page_size;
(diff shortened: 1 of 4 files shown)


lijinxia:
OK to verify

lijinxia:
Verified+1

PR #771, opened 2018-07-25
tools: acrntrace: save trace data file under current dir by default

lyan3:
Current implementation save trace data files on tmpfs by default, acrntrace
would use up all the physical mem, which can affect system functioning.
This commit changes default location for trace data file to current dir, and
removes the codes for space reservation, which is not necessary.


lijinxia:
Verified+1

PR #773, opened 2018-07-25
HV: align general-purpose reg layout with vmx

binbinwu1:
In current code, general-purpose register layout in run_context is not align
 with the  general-purpose register index when vmexit. So hv needs to map the
 index used during  vmexit to the index of the general-purpose register in
 run_context.
 This patch adjusts general-purpose register layout in run_context. So no need to
 map general-purpose regs during vmexit.
This patch also unifies the layout of general-purpose regs in interrupt/excpetion context.


NanlinXie:

lijinxia:
Verified+1

PR #778, opened 2018-07-25
HV:treewide:fix "Attempt to change parameter passed by value"

HuiHuangShi:
In the function scope,the parameter should not be
changed as Misra required.
V1->V2 recover some violations because of ldra's false positive.
V2->V3 sync local variable' type to parameter's type with the prefix of const.


NanlinXie:

lijinxia:
Verified+1

PR #780, opened 2018-07-26
HV:refine 'create_vm()' to avoid potential crash and memory leak

yonghuah:
hypervisor will crash if user try to call hypercall
  HC_CREATE_VM in infinite style.

  actually, the number of VMs hypervisor can support
  depends on the bit width of 'vmid_bitmap'.should
  return error in case of overflow.

  other cleanup for this function to avoid memory leak
  in case of failure.


NanlinXie:

lijinxia:
OK to verify

lijinxia:
OK to verify

PR #781, opened 2018-07-26
DM: add param: -V 5 to auto check/boot UOS image

mcao6:
it can auto check android image, if exists, then launch it;
if not, it will try to luanch clearlinux. it can be used by
acrn_guest.service to auto boot UOS.


NanlinXie:

mcao6:
there is another commit could be in another PR, will push later

lijinxia:
OK to verify

PR #787, opened 2018-07-26
HV: Rename functions beginning with "_"

rarindam:
V1:
In order to remove MISRA C violations for rule
219S, rename functions, macros whose name
starts with "_".
Also removed some __mmio functions because they are
duplicates and are not used anywhere.
Renamed functions like __assert, to asm_assert,
because they were only wrappers around asm calls.

V2:
Over and above the changes on V1, modified bitmap
functions names to lock (corresponding to unlock)
introduced in V1


lijinxia:
@rarindam Could you please rebase and fix the conflicts?

rarindam:
Submitted a new PR

PR #788, opened 2018-07-27
Rename struct vlapic and struct vpic

wuxyintel:
There are many naming violations detected by static analysis
tool. Data structure type name is the same as variable name.
According to naming convention rules, If the data structure
type is used by multi modules and its name meaning is
simplistic (such as vcpu, vm), its name needs prefix "acrn_".
Variable name should be different from data structure type name
and other member variable of its data structure.


NanlinXie:

lijinxia:
@wuxyintel Could you please rebase and fix the conflicts?

PR #793, opened 2018-07-27
HV: clear memory region used by UOS before it exit

yonghuah:
this patch fixes such an information leaking issue:
   in case that after a UOS is destroyed, its memroy
   will be reclaimed and maybe re-allocated for a new UOS,
   then the previous UOS sensitive data in memory may
   be leaked to the new UOS.


NanlinXie:

lijinxia:
Verified+1

PR #794, opened 2018-07-27
Rename struct timer, vie, vie_op, emul_ctxt

wuxyintel:
For data structure types "struct timer, struct vie", "struct emul_ctxt",
its name is identical with variable name in the same scope. This MISRA C 
violation is detected by static analysis tool.

Naming convention rule:If the data structure type is used by multi modules, 
its corresponding logic resource is only used by hypervisor/host and isn't 
exposed to external components (such as SOS, UOS), its name meaning is  
simplistic (such as timer), its name needs prefix "hv_". 

Naming convention rule:If the data structure type is used by only one 
module and its name meaning is simplistic, its name needs prefix 
shorten module name.

Follow naming convention rule to udpate data structure type "struct vie_op".

The following udpates are made:
struct timer-->struct hv_timer
struct vie-->struct instr_emul_vie
struct vie_op-->struct instr_emul_vie_op
struct emul_ctxt-->struct instr_emul_ctxt

### DIFF
--- a/hypervisor/arch/x86/guest/instr_emul.c
+++ b/hypervisor/arch/x86/guest/instr_emul.c
@@ -58,7 +58,7 @@
 #define	VIE_OP_F_NO_MODRM	(1U << 3)
 #define	VIE_OP_F_NO_GLA_VERIFICATION (1U << 4)
 
-static const struct vie_op two_byte_opcodes[256] = {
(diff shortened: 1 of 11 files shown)


NanlinXie:

lijinxia:
Verified+1

PR #802, opened 2018-07-30
DM: fix make install issue in auto boot UOS service

NanlinXie:
We need create system directory before we copy into $(DESTDIR)/usr/lib/systemd/system/, otherwise it will fail in clearlinux make build.

Signed-off-by: Nanlin Xie <nanlin.xie@intel.com>


mcao6:
@NanlinXie  thanks for your fixing. It is fine to me.

dongyaozu:
@jren1 please review

PR #804, opened 2018-07-30
IOC mediator: fix IOC mediator blocks acrn-dm shutdown flow

liuyuan-GitHub:
This patch resolves IOC mediator deinit function is blocked due to IOC mediator
core thread enters into sleep by epoll_wait, then pthread_join cannot return.

Trigger an event to wakeup core thread when IOC mediator deinit is invoked.


lijinxia:
Verified+1

PR #808, opened 2018-07-30
hv: use macro instead of specify number

liuwei987:
use macro instead of specify number


NanlinXie:

lijinxia:
OK to verify

PR #811, opened 2018-07-30
dm: bios: update vSBL binary to v0.8

binbinwu1:
Update vSBL with the following new features:
	Support watchdog status report when boot.
	Update vFastboot version to 0.8
Add a description file about the vSBL images.


devicemodel/bios/VSBL.txt
```diff
@@ -0,0 +1,7 @@
+Version 0.8
```
gvancuts: How about we split the content of this file into two separate files; one called `changelog` and another one called `MD5SUM`? That would be a little more self-explanatory than this file today and the `MD5SUM` could also be used directly (`md5sum -c MD5SUM`)

lijinxia:
Verified+1

PR #817, opened 2018-07-31
Revisit sostirq

lifeix:
v3-v4:
bug fix: remove undefined SOFTIRQ_ATOMIC in init_softirq

v2-v3:
1. remove SOFTIRQ_ATOMIC
2. enable irq when vm exit

v1-v2:
1. rename open_softirq to register_softirq; raise_softirq to fire_softirq
2. add pre-assumption for register_softirq and fire_softirq
3. some minor modify from comments

v1:
1. move softirq to common directory since it's not architecture related.
2. add open_softirq to register softirq handler function
3. rename exec_softirq to do_softirq. and do_softirq would be called twice:
the first would be called when irq is enabled; the second would be called
when irq is disabled to handle unhandled risk softirq.


lijinxia:
Verified+1

PR #818, opened 2018-07-31
HV: rename resume_vm to start_vm in hypercall api

jsun26intel:
Currently we don't support resume VM in HC API, the real meaning
of the code is to start VM.


lijinxia:
OK to verify

PR #819, opened 2018-07-31
IOC mediator: Add parking brake and Hvac signals

liuyuan-GitHub:
This patch adds new signals about parking brake and Hvac in the signal
definition and signal whitelist.


lijinxia:
Verified+1

PR #824, opened 2018-08-01
acrn-manager: create acrn-hypervisor-dev package

alekdu:
As first step to separate misc folder into github/intel/ioc-cbc-tools we
need let acrn create devel package. This package will let ioc-cbc-tools
use acrn manager interface API to control lifecycle.

The Clear Linux autospec script will *detect* /usr/include/acrn/*.h and
thus help us create the devel package. This devel pacakge will be in the
build_req for ioc-cbc-tools autospec.

Reviewed-by: Like Yan <like.yan@intel.com>


lijinxia:
Verified+1

PR #828, opened 2018-08-01
Enable basic UOS S3 support

jsun26intel:
This patch series enable the UOS S3 support. And with patch [2/6],
the system reset also could work.

For most virtio and emulated virtual devices, we don't need to do
anything in BE side for S3. There is only 1 exception: watchdog.
We have to move watchdog reset out of guest to make sure watchdog
always takes effect for guest to handle guest hang.

Now, there are some issues related with S3 we are still working
on:
1. PT dev issue for S3. Some PT dev can't work after UOS and UOS
   enter/exit S3.
2. Trusy S3 handling. We are waiting for dependency patch in. Then
   we will submit the trusty S3 handling patches.
3. Some mediator may have issue with UOS S3. We need to enable
   the basic UOS S3 first. So mediator owner could test/debug
   the mediator for UOS S3.

ChangeLog:
v1 -> v2:
  - Remove the suspend/resume callback added in v1 because most
    virtual devices don't need it.
  - Add global function to handle watchdog for S3.
    NOTE:
    We now only have one watchdog virtual device now. If we will
    have multi watchdog virtual devices in the future, we should
    move the global watchdog function to specific files.

### DIFF
--- a/devicemodel/Makefile
+++ b/devicemodel/Makefile
@@ -104,6 +104,7 @@ SRCS += core/sw_load_vsbl.c
 SRCS += core/smbiostbl.c
 SRCS += core/mevent.c
 SRCS += core/gc.c
+SRCS += core/pm.c
 SRCS += core/console.c
 SRCS += core/inout.c
 SRCS += core/mem.c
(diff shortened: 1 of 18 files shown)


lijinxia:
OK to verify

PR #835, opened 2018-08-01
HV: Add Partitioning mode option for ACRN

sainath14:
Adding Kconfig option to choose and compile partitioning mode for ACRN.
Current implementation does not allow ACRN to support sharing mode
and partitioning mode out of a single binary.

### DIFF
--- a/hypervisor/arch/x86/Kconfig
+++ b/hypervisor/arch/x86/Kconfig
@@ -11,6 +11,18 @@ config PLATFORM_SBL
 
 endchoice
 
+choice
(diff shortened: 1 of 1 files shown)


gvancuts:
Do we have any documentation on this new "Partition Mode"? If not, we really ought to add some (at least by the time this new mode is functional).

@dbkinder , FYI

dongyaozu:
Yes. We need the document by the time...

lijinxia:
OK to verify

PR #836, opened 2018-08-01
Ensure we have MMC device as in GP2.0

miguelinux:
Ensure we have a /dev/mmcblk1p3 device before we can launch
the guest OS service.

This will be help to not launch the acrn guest service on devices
which don't have MMC.

Exmaples are NUC with sata devices or installation on non APL MRB.

### DIFF
--- a/devicemodel/samples/apl-mrb/acrn_guest.service
+++ b/devicemodel/samples/apl-mrb/acrn_guest.service
@@ -3,6 +3,7 @@ Description=Start ACRN UOS
 After=systemd-networkd.service
 
 ConditionPathExists=/sys/kernel/gvt
+ConditionPathExists=/dev/mmcblk1p3
 
 [Service]
 Type=simple


dongyaozu:
@jren1 can u review?

gvancuts:
Shouldn't this file (`acrn_guest.service`) be only installed on the APL MRB system anyway (it's located inside `samples/apl-mrb/` after all)? I'm thinking we may have a packaging issue here is we install this on different systems.

This patch does not do any harm however and if anything at all is a good sanity check before starting the service. What it does not check however is what is on that partition (and if it's useable at all).

dongyaozu:
@binbinwu1 can u review?

miguelinux:
The reason is because this commit d40a6b9e9329 ("DM: add service to support auot boot UOS")

miguelinux:
And somebody ask to add it at bootime

binbinwu1:
Maybe we need to fix devicemodel Makefile instead?

dongyaozu:
@miguelinux can you response to @binbinwu1 comments?

miguelinux:
@mcao6 Can you help to explain the needs of this change.

mcao6:
This service is just for MRB, not for other platform, so it is under "apl-mrb", add a check is OK, if not add also no big issue, it will failed when boot, not affect system.

mcao6:
for the block check will also be done in launch_uos.sh script.

mcao6:
after think twice, the check condition add into this service, it is not general, for that check will be done in launch_uos.sh, so better not add it here, even the function is OK.

miguelinux:
Ok, then lets fix the [apl-mrb/launch_uos.sh](../blob/master/devicemodel/samples/apl-mrb/launch_uos.sh)  file.

Just recall that if [apl-mrb/acrn_guest.service](../blob/master/devicemodel/samples/apl-mrb/acrn_guest.service) is configured to launch at boot time, it will be lauched on every platform where Clear Linux is installed with service-os bundle. 

Please ensure [apl-mrb/launch_uos.sh](../blob/master/devicemodel/samples/apl-mrb/launch_uos.sh)  do not touch nothing in others systems, because  line 3333, [apl-mrb/launch_uos.sh#L333](../blob/master/devicemodel/samples/apl-mrb/launch_uos.sh#L333)  is not validated and always executed.

gvancuts:
Since we have different scripts for different platforms, how can we package these so only the relevant one are installed and enabled? One way would to create two different bundles but is there a cleverer way that would leave us with a single service-os bundle?

miguelinux:
Yes,

We could have as many services enabled but we need to ensure they are loaded in the specific platform and we can use the conditional constraint from the systemd services.

```
ConditionArchitecture=, ConditionVirtualization=, ConditionHost=, ConditionKernelCommandLine=,
 ConditionKernelVersion=, ConditionSecurity=, ConditionCapability=, ConditionACPower=, 
ConditionNeedsUpdate=, ConditionFirstBoot=, ConditionPathExists=, ConditionPathExistsGlob=, 
ConditionPathIsDirectory=, ConditionPathIsSymbolicLink=, ConditionPathIsMountPoint=, 
ConditionPathIsReadWrite=, ConditionDirectoryNotEmpty=, ConditionFileNotEmpty=, 
ConditionFileIsExecutable=, ConditionUser=, ConditionGroup=, ConditionControlGroupController=
```
currently ACRN lack of `ConditionVirtualization`

mcao6:
It is OK. I'll add more dependency into it for general usage.

PR #837, opened 2018-08-02
HV:treewide:avoid using multiple # or ## in a macro

wuxyintel:
In the C99 standard, the order of evaluation associated with
multiple #, multiple ## or a mix of # and ## preprocessor
operator is unspecified. For this case, gcc 7.3.0 manual
does not specify related implementation. So it is unsafe
to use multiple # or ## in a macro.
BTW, there are some macros with one or more "##" which are
not used by hypervisor.

Update relate codes to avoid using multiple # or ## in a macro;
Remove unused macros with one or more "##";
Remove "struct __hack;" at the end of GETCC since it is useless.

Note:
     '##' operator usage constraints: A ## preprocessing token shall
     not occur at the beginning or at the end of a replacement list
     for either form of macro definition.
V1--V2:
	Update relate codes to avoid using multiple # or ## in a macro.
V2-->V3:
	Remove unused macros with one or more "##";
	Remove "struct __hack;" at the end of GETCC since it is useless.

Signed-off-by: Xiangyang Wu <xiangyang.wu@linux.intel.com>

### DIFF
--- a/hypervisor/arch/x86/guest/instr_emul.c
+++ b/hypervisor/arch/x86/guest/instr_emul.c
@@ -366,21 +366,19 @@ static int vie_update_rflags(struct vcpu *vcpu, uint64_t rflags2, uint64_t psl)
 /*
  * Return the status flags that would result from doing (x - y).
  */
-#define	GETCC(sz)							\
-static uint64_t								\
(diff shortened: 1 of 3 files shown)


lijinxia:
OK to verify

PR #839, opened 2018-08-02
HV: Refine handling invalid parameter handling in hypervisor shell

yonghuah:
- print error message for command parameter error
   when callback/handler is called.


lijinxia:
Verified+1

PR #842, opened 2018-08-02
hv: treewide: fix 'Array has no bounds specified'

shiqingg:
MISRAC requires that the array size should be declared explicitly.

This patch fixes the issues caused by the arrays that are defined in
link_ram.ld.in or assembly file.

v1 -> v2:
 * Update the solution based on the info from the following link.
   https://sourceware.org/binutils/docs/ld/Source-Code-Reference.html

   Fix pattern is like below:
   extern char start_of_ROM, end_of_ROM, start_of_FLASH;
   memcpy (& start_of_FLASH, & start_of_ROM, & end_of_ROM - &
start_of_ROM);

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -369,7 +369,8 @@ void bsp_boot_init(void)
 	start_tsc = rdtsc();
 
 	/* Clear BSS */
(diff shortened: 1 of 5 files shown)


lijinxia:
Verified+1

PR #844, opened 2018-08-02
dm: uart: fix acrn-dm crash issue when invoke uart_closetty function

liuyuan-GitHub:
mevent is added only if uart backend fd refers TTY in uart_opentty.
So we should only delete mevent if uart backend fd refers TTY in uart_closetty.

This issue can be reproduced by below steps
1) acrnd starts UOS
2) run poweroff command in UOS
3) crash happens


dongyaozu:
@fyin1 can you ack ?

fyin1:
Eddie, I ack the patch on the mail list already. :)

lijinxia:
Verified+1

PR #845, opened 2018-08-02
HV: Rename functions, variables starting with "_"

rarindam:
In order to comply with MISRA C rules, renamed vairables
and function names starting with "_".
The major changes invloves mostly static function
names, as they are being called inside the same file
by a wrapper function.


lijinxia:
Verified+1

PR #846, opened 2018-08-02
HV: Adding mptable support for partition mode ACRN

sainath14:
Partitioning mode of ACRN needs to build mptable for UOS.
UOS uses mptable instead of ACPI tables.


lijinxia:
Verified+1

PR #856, opened 2018-08-06
DM: add tag info while no repo in release

liuwei987:
there is no git repo, which download from github release,
however, acrn-dm -v will get tag info from this repo, in
such case, the tag info was null.
this patch will fix nul tag, which get it from CL mock build.


dongyaozu:
@jren1 @binbinwu1 can u review?

miguelinux:
+1

This will help when building using the tarball from GitHub (as is in autospec)

PR #857, opened 2018-08-06
hv: timer: request timer irq once only

lyan3:
Since global vector table is being used on all pcpus, it's not necessary to request timer irq
at each cpu init. With this change, per_cpu timer nodes are removed, and only BSP registers
and unregisters timer irq.


lijinxia:
Verified+1

PR #858, opened 2018-08-06
HV:treewide:fix value outside range of underlying type

wuxyintel:
There are potential value outside range of underlying type
in some assignment expressions. This violates Rule 10.3 or
Rule 10.4 of MISRA C:2012.
BTW, all operations shall be conducted in exactly the same
arithmetic (underlying) type, otherwise, there is a
value outside range violation.

Update related assignment expressions.

V1-->V2:
	* Fix potential overflow in "pit_calibrate_tsc";
	* Move PTDEV_INVALID_PIN definition before
	  get_entry_info since this MACRO is only used by
	  debug function.

Signed-off-by: Xiangyang Wu <xiangyang.wu@linux.intel.com>

### DIFF
--- a/hypervisor/arch/x86/assign.c
+++ b/hypervisor/arch/x86/assign.c
@@ -951,9 +951,10 @@ void ptdev_remove_msix_remapping(struct vm *vm, uint16_t virt_bdf,
 }
 
 #ifdef HV_DEBUG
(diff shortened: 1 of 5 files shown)


lijinxia:
Verified+1

PR #862, opened 2018-08-06
dm: uart: add state check of backend tty before uart_closetty

liuyuan-GitHub:
This patch resolves one arcn-dm crash issue when rebooting UOS.

The rootcause is that uart releases unopened backend tty.

One reproduced case is that the board does not support IOC but IOC
feature is enabled in the acrn-dm. After rebooting UOS, crash will
happen.

NOTE: This issue is not related to IOC, it also can be reproduced
with NON-IOC scenario. Just set one invalid PTY to the lpc, then
this issue should be reproduced.
We need re-visit the whole policy for such scenario in future.


lijinxia:
Verified+1

PR #863, opened 2018-08-06
hv:removed assert in free_ept_mem

mingqiangchi:
Removed assert since the caller has checked the parameter.


lijinxia:
Verified+1

PR #870, opened 2018-08-07
hv: treewide: fix 'Potential side effect problem in expression'

shiqingg:
MISRA-C requirement:
The value of an expression shall be the same under any order of
evaluation.
The order in which side effects take place is unspecified and may lead
to unexpected results.

This patch add a temporary variable for temporary storage to avoid the
side effects of evaluation order.


shiqingg:
@dongyaozu could you please approve this patch?

lijinxia:
Verified+1

PR #873, opened 2018-08-08
add cpu_do_idle to handle idle

JasonChenCJ:
add wrap function cpu_do_idle in default_idle to handle arch cpu specific
idle operation.


lijinxia:
Verified+1

Issue #875, opened 2018-08-08
IO+MMIO emulation code reshuffle


mwang106:
The current implementation of I/O emulation is over-complicated as there are duplicated code scattering in different files and VHM request states are represented using two integers instead of one. A reshuffle is needed to make the underlying logic clearer.

mwang106:
[External_System_ID] 247219

Mingyuan18:
Can't reproduce this issue now.

PR #893, opened 2018-08-08
DM: adapt to the new VHM request state transitions

junjiemao1:
This is the counterpart in DM to the VHM request state update in the
hypervisor. Wrappers to gcc built-in atomic operations are introduced at the
same time to hide the difference between __sync and __atomic series.

v1 -> v2:

    * Use macro-defined constants for the default values for invalid PIO/MMIO
      reads.
    * Change the return type of vmexit_handler_t in DM to void as the return
      values are no longer necessary.
    * Remove VM_EXITCODE that are no longer used.

Tracked-On: #875


lijinxia:
Verified+1

lijinxia:
Verified+1

PR #902, opened 2018-08-08
hv: Adding a wrapper on top of prepare_vm0

sainath14:
Added prepare_vm function as a wrapper function on top of prepare_vm0.
This makes adding support for partition mode ACRN to boot multiple
VMs from HV cleaner.


lijinxia:
Verified+1

PR #908, opened 2018-08-08
DM: check more in guest service & launch script

mcao6:
some exceptions can happen if westnon or network name service
not started before launch uos guest, like gvt and network exceptions,
which can cause SOS reboot. so add the dependency.

add data partiton check in launch script, for it can be called in every
clearlinux platform.


lijinxia:
Verified+1

Issue #895, opened 2018-08-08
Power Management: S3


hwang37:
The hypervisor and VM's shall support the ability for the entire platform to enter/exit from S3 suspend/resume, when requested by the Service OS.

hwang37:
[External_System_ID]212379

mwang106:
This feature verified on MRB

PR #973, opened 2018-08-09
DM: notify VHM request complete after pausing the VM

junjiemao1:
It is necessary to notify the VHM and hypervisor on the completion of a VHM
request even when the UOS is in suspend or system reset mode because the VHM and
hypervisor rely on the notification to reset their own states on the
request.

Currently the VHM request state is checked against REQ_STATE_PROCESSING instead
of REQ_STATE_COMPLETE when handling system reset or suspend/resume, leading to a
completed request unnotified, and causing the HV to complain on an occupied VHM
request when it raises a new one.

This patch fixes this issue by properly notifying completed requests to the VHM
& hypervisor. Some concerns are raised during a discussion on the potential
races which does not hurt for now but may in the future. These considerations
and potential solutions are documented as comments for future reference.

Tracked-On: #895


lijinxia:
Verified+1

Issue #971, opened 2018-08-09
acrncrashlog funcitons need to be enhance


mwang106:
1. should only modify the outparam in successful cases.
2. should return -1 if error happens
3. should check the return value of strrchr

mwang106:
[External_System_ID] 251096

mwang106:
patch merged and verified

PR #974, opened 2018-08-09
tools:acrn-crashlog: Enhance some functions

liuxinw:
The changes include:

1. modify the outparam only in successful cases.
2. return -1 instead of  a errno-style value if error happens.
3. check return value of strrchr.

Tracked-On: #971
Signed-off-by: Liu, Xinwu <xinwu.liu@intel.com>
Acked-by: Chen Gang <gang.c.chen@intel.com>


lijinxia:
Verified+1

PR #976, opened 2018-08-09
Provide virtual S3 support by Trusty.

YadongQi:
This patch series provide virtual S3 support by Trusty.

Main parts of this patch:
1. secure_world_control structure refine
   Previous definition of secure_world_control is not very accurate, break
   sworld_enable into 2 bit flags: supported and active. Check active bit
   instead of sworld_eptp. Record base_gpa_in_sos and base_gpa_in_uos since
   they will be used in virtual S3/warm reset path. This assumes the trusty's
   memory address is contiguous in both SOS and physical side.
2. Remove unused hypercall id: HC_GET_SEC_INFO
3. Log printing cleanup
4. New hypercall: HC_SAVE_RESTORE_SWORLD_CTX
   In UOS S3 suspend path: trusty kernel driver will call this hypercall
   to require Hypervisor save context of secure world.
   In UOS S3 resume path: virtual firmware will call this hypercall to
   require Hypervisor restore context of secure world.

### DIFF
--- a/hypervisor/arch/x86/ept.c
+++ b/hypervisor/arch/x86/ept.c
@@ -107,9 +107,8 @@ void destroy_ept(struct vm *vm)
 	 *  - trusty is enabled. But not initialized yet.
 	 *    Check vm->arch_vm.sworld_eptp.
 	 */
-	if (vm->sworld_control.sworld_enabled &&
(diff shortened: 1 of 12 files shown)


lijinxia:
Verified+1

PR #977, opened 2018-08-09
HV: reshuffle guest init code

binbinwu1:
This patch series adds code to save boot context, which is prepare by the platfrom
bootloader or BIOS, the context can be used to init sos vmcs, such as GDT, IDT,
segment selectors, control registers, ia32_efer. In this way, HV can
leverage the data structures built by bootloader or BIOS, without
creating them in HV.
Basing on the saved context, VM0 BSP start mode can be detected runtime
instead of hardcoded.
Also restructure the code of guest init code.
In this patch, a vcpu run_context is initilaized first according to vcpu mode.
Then write the value to vmcs according to run_context value.
Remove the code of creating init page table for VM0, since it can use the page table
created by bootloader or BIOS. HV doesn't need to create page table for it.


lijinxia:
Verified+1

PR #980, opened 2018-08-09
hv: pirq: do not indicate priority when allocate vector

lyan3:
It's not necessary to specify priority for dynamic allocated vector, because
nested irq is not supported on HV, and irq of any priority would cause VM exit.

This commit makes the following changes:
 - remove the argument indicating priority in struct irq_request_info and
   following functions:
    ptdev_activate_entry()
    find_available_vector()
    irq_desc_alloc_vector()
    normal_register_handler()
 - change the macro of vector ranges:
    VECTOR_DYNAMIC_START/END for dynamically allocable vectors;
    VECTOR_FIXED_START/END for fixed allocated vectors, such as vector for
    timer etc.


lijinxia:
Verified+1

PR #982, opened 2018-08-09
hv: treewide: fix 'inline function should be declared static'

shiqingg:
MISRAC does not allow the use of an inline function with external
linkage.

What this patch does:
- Add the static keyword for the function that is only used in the
  definition file.
- Remove the inline keyword for the function that is used in multiple
  files.

v1 -> v2:
 * Move some functions to headers as static inline function if it is
    possible


lijinxia:
Verified+1

PR #983, opened 2018-08-09
hv:Refine destroy_secure_world API

mingqiangchi:
-- add clear trusty memory flag
  In some cases such as UOS power off or UOS full reset,
  need to clear trusty memory,no need to clear memory such as
  UOS S3 or UOS system reset,then add a flag to distinguish it
  when destroy secure world.
-- Restore trusty memory to guest normal world.
-- Moved free trusty EPT inside destroy_secure_world
  In some cases such as UOS S3 or UOS system reset,
  only need to free trusty EPT, this patch move free
  trusty EPT inside destroy_secure_world.
  Because PD/PT are shared in both secure world's EPT
  and normal world's EPT,before freeing trusty EPT,
  it will memset all PDPTEs except trusty memory,
  then call 'free_ept_mem', it can only free trusty EPT,
  and does't affect shared normal world EPT.

v2-->v3:
    -- Used new mmu api ept_mr_add when restore trusty memory
       to SOS and normal world
    -- Dropped this patch "Removed reverted page tables for trusty memory"
       because map_mem will be removed in future
       It will have a patch, need to update this api(ept_mr_add),
       it will not create inverted page tables for trusty memory.

v1-->v2:
   -- free trusty ept
       still use free_ept_mem, not add a new api,but need to
       memset pdptes except trusty memory
   -- Removed reverted page tables for trusty memory.

### DIFF
--- a/hypervisor/arch/x86/ept.c
+++ b/hypervisor/arch/x86/ept.c
@@ -45,7 +45,7 @@ static uint64_t find_next_table(uint32_t table_offset, void *table_base)
 /**
  * @pre pml4_addr != NULL
  */
-static void free_ept_mem(void *pml4_addr)
+void free_ept_mem(void *pml4_addr)
 {
 	void *pdpt_addr;
 	void *pde_addr;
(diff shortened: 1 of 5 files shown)


lijinxia:
Verified+1

PR #988, opened 2018-08-09
hv: Interrupt handling in ACRN partition mode

sainath14:
ACRN in partition mode does not have vector and APIC ID remapping for
device interrupts. Only MSIs are supported. No IOAPIC and legacy interrupts
for the VMs in ACRN partition mode.


lijinxia:
Verified+1

PR #989, opened 2018-08-09
HV: Add vm_id entry to VM description in partitioning mode

sainath14:
ACRN boots multiple OS in partitioning mode. This patch adds code to assign
vm_id in the vm data structure to be same as the one assigned at compile time.
This makes the vm id deterministic for each VM booted from HV directly.


lijinxia:
Verified+1

PR #990, opened 2018-08-09
hv: Handling IO exits in ACRN for partition mode

sainath14:
There is no SOS and device model in strict partition mode. ACRN emulates IO for
virtual devices. Any access to IO not backed by HV should return all FFs on read
and writes should be discarded.


lijinxia:
Verified+1

PR #992, opened 2018-08-10
dm: virtio_rnd: use delayed blocking IO to make virtio_rnd works on L…

ssqre:
…inux based SOS

Randomness sourced from /dev/random which does not block
once it has been seeded at bootup and you will always get
something when you read from that file. This is true on
Freebsd but unfortunately things are not the same on Linux.
Most cases, you can't read anything from /dev/random especially
on current acrn platform which lacking random events.
virtio_rnd inherted from freebsd doesn't work anymore.

This patch makes virtio_rnd working on Linux based SOS. It uses
blocking IO to sevice the front-end random driver and delays the
read operation into a new thread to avoid blocking the main
notify thread.

Reviewed-by: Shuo Liu <shuo.a.liu@intel.com>

### DIFF
--- a/devicemodel/hw/pci/virtio/virtio_rnd.c
+++ b/devicemodel/hw/pci/virtio/virtio_rnd.c
@@ -27,8 +27,6 @@
 
 /*
  * virtio entropy device emulation.
- * Randomness is sourced from /dev/random which does not block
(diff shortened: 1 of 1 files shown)


lijinxia:
Verified+1

PR #996, opened 2018-08-10
hv: treewide: fix 'Empty parameter list to procedure/function'

shiqingg:
Use func(void) rather than func() for the function declaration and
definition based on MISRAC requirement.


lijinxia:
Verified+1

PR #998, opened 2018-08-10
hv: Partition mode ACRN -kernel load and bootargs load address

sainath14:
For Partition mode ACRN, kernel load address and Boot args load address
are hardcoded. Boot args are currently passed from vm description for
each VM. Renamed init_vm0_boot_info to init_vm_boot_info.


lijinxia:
Verified+1

PR #999, opened 2018-08-10
hv: Add vuart flag to VM descriptions in partition mode

sainath14:
ACRN in partition mode emulates UART for the VMs directly launced
from HV. This patch adds an option to enable/disable vUART for a
VM.


lijinxia:
Verified+1

PR #1000, opened 2018-08-10
hv: Fixing build issue with PARTITION_MODE

sainath14:
Modified the vector MACRO that is failing build with PARTITION_MODE selected


lijinxia:
Verified+1

PR #1001, opened 2018-08-12
remove 2M hugepages

ailun258:
remove 2M hugepages setting here because it will not use any more. 
# make sure there is enough 2M hugepages in the pool
echo 1024 > /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages

Signed-off-by: Ailun258 <ailin.yang@intel.com>


dongyaozu:
@JasonChenCJ can u review?

PR #1002, opened 2018-08-12
fixed cpu info incorrect and remove 2M hugepages

ailun258:
with original script, user can not get 3 CPU cores info with cat/proc/cpuinfo in UOS.
it just show 1 only.  with this patch,  default,  user can see 1 cpu info in SOS and 3 cores info in UOS.
and remove 2M hugepages settings because it is not using any more.
# make sure there is enough 2M hugepages in the pool
echo 1024 > /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages
also, this patch is setting default UOS memory to 2G

Signed-off-by: Ailun258 <ailin.yang@intel.com>

### DIFF
--- a/devicemodel/samples/nuc/launch_uos.sh
+++ b/devicemodel/samples/nuc/launch_uos.sh
@@ -14,10 +14,7 @@ if [[ "$result" != "" ]]; then
 fi
 
 #for memsize setting
-mem_size=1000M
-
-# make sure there is enough 2M hugepages in the pool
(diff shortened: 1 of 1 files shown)


ailun258:
fixing this issue:
https://github.com/projectacrn/acrn-hypervisor/issues/1003

dongyaozu:
@JasonChenCJ can u review?

lijinxia:
Verified+1

PR #1007, opened 2018-08-13
Add the emulation of Cpuid 0x16 for guest to obtain cpu hz by calling cpuid

yakuizhao:
This is the patch set that adds the emulation of cupid 0x16 so that the guest can obtain the cpu hz by calling cupid.
At the same time HV also obtains the cpu hz by using cupid 0x16 if zero tsc is returned by cupid 0x15.

### DIFF
--- a/hypervisor/arch/x86/cpuid.c
+++ b/hypervisor/arch/x86/cpuid.c
@@ -119,6 +119,21 @@ static void init_vcpuid_entry(__unused struct vm *vm,
 		entry->ecx = 0U;
 		entry->edx = 0U;
 		break;
+	case 0x16U:
+		if (boot_cpu_data.cpuid_level >= 0x16U) {
+			/* call the cpuid when 0x16 is supported */
+			cpuid_subleaf(leaf, subleaf,
+				&entry->eax, &entry->ebx,
+				&entry->ecx, &entry->edx);
+		} else {
+			/* Use the tsc to derive the emulated 0x16U cpuid. */
+			entry->eax = (uint32_t) (tsc_khz / 1000U);
+			entry->ebx = entry->eax;
+			/* Bus frequency: hard coded to 100M */
+			entry->ecx = 100U;
+			entry->edx = 0U;
+		}
+		break;
 
 	/*
 	 * Leaf 0x40000000
@@ -173,6 +188,11 @@ int set_vcpuid_entries(struct vm *vm)
 	uint32_t i, j;
 
 	init_vcpuid_entry(vm, 0U, 0U, 0U, &entry);
+	if (boot_cpu_data.cpuid_level < 0x16U) {
+		/* The cpuid with zero leaf returns the max level.
+		 * Emulate that the 0x16U is supported */
+		entry.eax = 0x16U;
+	}
 	result = set_vcpuid_entry(vm, &entry);
 	if (result != 0) {
 		return result;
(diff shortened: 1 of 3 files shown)


hypervisor/arch/x86/cpuid.c
```diff
@@ -158,9 +173,36 @@ static void init_vcpuid_entry(__unused struct vm *vm,
 		break;
 
 	default:
-		cpuid_subleaf(leaf, subleaf,
-				&entry->eax, &entry->ebx,
-				&entry->ecx, &entry->edx);
+		if (leaf >= 0x80000000U) {
+			/* unsupported extended cpuid level */
```
dongyaozu: As Anthony commented in mailinglist, do we need to change leaf > 0x8000000000 case?

hypervisor/arch/x86/cpuid.c
```diff
@@ -158,9 +173,36 @@ static void init_vcpuid_entry(__unused struct vm *vm,
...
+				entry->ebx = 0U;
+				entry->ecx = 0U;
+				entry->edx = 0U;
+			} else {
+				cpuid_subleaf(leaf, subleaf,
+					&entry->eax, &entry->ebx,
+					&entry->ecx, &entry->edx);
+			}
+		} else if (leaf >= 0x40000000U) {
+			entry->eax = 0U;
```
dongyaozu: In this case, we don't present the leaf 0x400000000 ? or we only support 0x40000000U?

hypervisor/arch/x86/cpuid.c
```diff
@@ -158,9 +173,36 @@ static void init_vcpuid_entry(__unused struct vm *vm,
 		break;
 
 	default:
-		cpuid_subleaf(leaf, subleaf,
-				&entry->eax, &entry->ebx,
-				&entry->ecx, &entry->edx);
+		if (leaf >= 0x80000000U) {
+			/* unsupported extended cpuid level */
+			if (leaf > boot_cpu_data.extended_cpuid_level) {
+				entry->eax = 0U;
```
dongyaozu: The hardware must support leaf 0x15. The only one presented to guest (but not in hardware) is leaf 0x16 but it is already covered before this code. Do we need this code?

hypervisor/arch/x86/timer.c
```diff
@@ -278,23 +278,30 @@ static uint64_t pit_calibrate_tsc(uint32_t cal_ms_arg)
...
 		cpuid(0x15U, &eax_denominator, &ebx_numerator,
 			&ecx_hz, &reserved);
 
 		if (eax_denominator != 0U && ebx_numerator != 0U) {
-			return ((uint64_t) ecx_hz *
+			tsc_hz = ((uint64_t) ecx_hz *
 				ebx_numerator) / eax_denominator;
 		}
 	}
+	if ((tsc_hz == 0) && (boot_cpu_data.cpuid_level >= 0x16U)) {
```
dongyaozu: should we move this earlier than leaf 0x15? That will be slightly simple

yakuizhao: TSC based on 0x15 leaf is more precise than that based on 0x16 leaf.
So it will be better that it checks the 0x16 leaf after zero tsc is returned by 0x15 leaf.

dongyaozu: OK, then we can go with leaf 0x15 first.

hypervisor/arch/x86/cpuid.c
dongyaozu: remove dead code to keep small

hypervisor/arch/x86/cpuid.c
dongyaozu: remove it.

lijinxia:
Verified+1

PR #1012, opened 2018-08-13
Vcpu dumpreg

JasonChenCJ:


hypervisor/arch/x86/lapic.c
```diff
@@ -413,18 +413,16 @@ send_startup_ipi(enum intr_cpu_startup_shorthand cpu_startup_shorthand,
...
-	dest_lapic_id = per_cpu(lapic_id, pcpu_id);
+	uint32_t hi_32, lo_32;
 
 	/* Set the target processor. */
-	hi_32 = dest_lapic_id << 24;
+	hi_32 = dest_mask << 24;
 
 	/* Set the vector ID. */
-	lo_32 = vector;
+	lo_32 = vector | dest << 11;
```
dongyaozu: 11U.  And add () for the dest<<11U.

hypervisor/arch/x86/lapic.c
```diff
@@ -413,18 +413,16 @@ send_startup_ipi(enum intr_cpu_startup_shorthand cpu_startup_shorthand,
...
 {
-	uint32_t dest_lapic_id, hi_32, lo_32;
-
-	/* Get the lapic ID of the destination processor. */
-	dest_lapic_id = per_cpu(lapic_id, pcpu_id);
+	uint32_t hi_32, lo_32;
 
 	/* Set the target processor. */
-	hi_32 = dest_lapic_id << 24;
+	hi_32 = dest_mask << 24;
```
dongyaozu: 24U

hypervisor/arch/x86/lapic.c
```diff
@@ -413,18 +413,16 @@ send_startup_ipi(enum intr_cpu_startup_shorthand cpu_startup_shorthand,
 	return status;
 }
 
-void send_single_ipi(uint16_t pcpu_id, uint32_t vector)
+/* dest must be INTR_LAPIC_ICR_PHYSICAL(0x0U) or INTR_LAPIC_ICR_LOGICAL(0x1U) */
+void send_dest_ipi(uint32_t dest_mask, uint32_t vector, uint32_t dest)
```
dongyaozu: The 1st parameter is actually dest, not mask.
The last parameter is unknown to me.

hypervisor/arch/x86/cpu.c
```diff
@@ -641,7 +641,7 @@ static void cpu_secondary_post(void)
 	timer_init();
 
 	/* Wait for boot processor to signal all secondary cores to continue */
-	pcpu_sync_sleep(&pcpu_sync, 0UL);
+	wait_sync_change(&pcpu_sync, PCPU_SYNC_RUN);
```
dongyaozu: I prefer to use 0UL here. 0 is something special, it becomes unclear if we use a MACRO

hypervisor/arch/x86/cpu.c
```diff
@@ -558,9 +561,6 @@ static void bsp_boot_post(void)
 	/* Start all secondary cores */
 	start_cpus();
 
-	/* Trigger event to allow secondary CPUs to continue */
-	bitmap_set_nolock(0U, &pcpu_sync);
-
```
dongyaozu: why we clear the reset part?

hypervisor/arch/x86/cpu.c
```diff
@@ -714,6 +717,9 @@ void start_cpus(void)
 		do {
 		} while (1);
 	}
+
+	/* Trigger event to allow secondary CPUs to continue */
+	atomic_store64(&pcpu_sync, PCPU_SYNC_RUN);
```
dongyaozu: prefer 0UL here.

hypervisor/arch/x86/notify.c
```diff
@@ -8,15 +8,58 @@
...
 	return 0;
 }
 
+void smp_call_function(uint64_t mask, smp_call_func_t func, void *data)
+{
+	uint16_t pcpu_id;
+	struct smp_call_info_data *smp_call;
+
+	/* wait for previous smp call complete, which may run on other cpus */
+	wait_sync_change(&smp_call_mask, 0UL);
```
dongyaozu: This is redundant?

hypervisor/arch/x86/notify.c
```diff
@@ -8,15 +8,58 @@
...
+		} else if (bitmap_test(pcpu_id, &pcpu_active_bitmap)) {
+			smp_call = &per_cpu(smp_call_info, pcpu_id);
+			smp_call->func = func;
+			smp_call->data = data;
+		} else {
+			/* pcpu is not in active, print error */
+			pr_err("pcpu_id %d not in active!", pcpu_id);
+			bitmap_clear_nolock(pcpu_id, &smp_call_mask);
+		}
+	}
```
dongyaozu: from this code, it seems each PCPU executes the func() with same parameter... If this is the case, why we  need above code?

hypervisor/arch/x86/lapic.c
JasonChenCJ: OK

hypervisor/arch/x86/lapic.c
JasonChenCJ: OK

hypervisor/arch/x86/lapic.c
JasonChenCJ: 1st is dest_mask which is destination filed, last parameter is destination mode. I will update the name.

hypervisor/arch/x86/cpu.c
JasonChenCJ: Ok.

hypervisor/arch/x86/cpu.c
JasonChenCJ: It's done by atomic_store64(&pcpu_sync, PCPU_SYNC_RUN); in start_cpus.

hypervisor/arch/x86/cpu.c
JasonChenCJ: ok

hypervisor/arch/x86/notify.c
JasonChenCJ: no, it's used to sync smp_call which may come from other cpus when one cpu is running on it.

hypervisor/arch/x86/notify.c
JasonChenCJ: we would like to call func directly on current cpu (no need IPI), and we would like to check dest mask with active cpu bitmap.

hypervisor/arch/x86/lapic.c
dongyaozu: then use dest and del_mode?

hypervisor/arch/x86/notify.c
dongyaozu: if  you use smp_call_mask==0 to indicate the caller can start a new SMP call, that is dangerous

hypervisor/arch/x86/notify.c
dongyaozu: I am not saying this...  The caller usually won't smp itself, and even if it smp itself, it is OK.

hypervisor/arch/x86/notify.c
JasonChenCJ: Ok, so you mean multi cpus could do it at exact same time, right? Yes, as wait_sync_change & atomic_store64 is not an atomic operation, I am doing it just want to avoid to use a lock, if so, I will add the lock back.

hypervisor/arch/x86/lapic.c
JasonChenCJ: Will use dest and dest_mode.

hypervisor/arch/x86/notify.c
dongyaozu: Not necessary to use lock, but make sure you use test_and_exchange atmoic operation.

hypervisor/arch/x86/notify.c
JasonChenCJ: ok, let me do it like this:

if (bitmap_test(pcpu_id, &pcpu_active_bitmap)) {
    smp_call = &per_cpu(smp_call_info, pcpu_id);
    smp_call->func = func;
    smp_call->data = data;
} else {
    pr_err("pcpu_id %d not in active!", pcpu_id);
    bitmap_clear_nolock(pcpu_id, &smp_call_mask);
}

hypervisor/arch/x86/notify.c
JasonChenCJ: Sure.

dongyaozu: atomic_cmpxchg64

hypervisor/arch/x86/cpu.c
```diff
@@ -641,7 +638,7 @@ static void cpu_secondary_post(void)
 	timer_init();
 
 	/* Wait for boot processor to signal all secondary cores to continue */
-	pcpu_sync_sleep(&pcpu_sync, 0UL);
+	wait_sync_change(&pcpu_sync, 0U);
```
dongyaozu: should be 0UL since it is uint64_t

lijinxia:
Verified+1

PR #1013, opened 2018-08-13
Irqv5

lyan3:
This two commits remove the irq sharing support,  main changes including:
 - removed the dev_list field in irq_desc, and clean up codes for the list operation; 
 - replace IRQ_ASSIGNED_SHARED and IRQ_ASSIGNED_NOSHARE with IRQ_ASSIGNED;
 - remove "shared" argument in irq APIs;
 - remove dev_handler_node struct and change the return type of pri_/normal_register_handler() from dev_handler_node* to int32_t, which is irq num (>= 0) on success, and errno (> 0) on failure;
 - changes unregister_irq_handler() to take argument unint32_t instead of dev_handler_node*;

In addition,  irq request flow for pt devices is revised, in order to remove dependency on irq sharing:  register irq on adding remapping entery and unregister irq on removal an entry, and do not register/unregister at remapping an entry.

### DIFF
--- a/hypervisor/arch/x86/irq.c
+++ b/hypervisor/arch/x86/irq.c
@@ -63,7 +63,7 @@ uint32_t irq_mark_used(uint32_t irq)
 	desc = &irq_desc_array[irq];
 	spinlock_irqsave_obtain(&desc->irq_lock);
 	if (desc->used == IRQ_NOT_ASSIGNED) {
-		desc->used = IRQ_ASSIGNED_NOSHARE;
+		desc->used = IRQ_ASSIGNED;
 	}
 	spinlock_irqrestore_release(&desc->irq_lock);
 	return irq;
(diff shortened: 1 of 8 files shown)


hypervisor/arch/x86/irq.c
```diff
@@ -151,43 +151,18 @@ static void disable_pic_irq(void)
 	pio_write8(0xffU, 0x21U);
 }
 
-static bool
-irq_desc_append_dev(struct irq_desc *desc, void *node, bool share)
+static void 
+irq_desc_append_dev(struct irq_desc *desc, void *node)
```
dongyaozu: append irq or dev? We don't have device concept

dongyaozu: it is more like to register a handler?

dongyaozu: replace node with action? and should not be void *

lyan3: This function would be removed completely in the second patch, so I hadn't paid much attention to the name. 
If you insist, I will change the name to set_irq_action(struct irq_desc *desc, void *node)

lyan3:
Hi @dongyaozu , 
Jinxia needs your approval to merge this PR since it has been updated.
Thanks,
Like

lijinxia:
Verified+1

PR #1014, opened 2018-08-13
hv: add compile time assert

binbinwu1:
The patch series moves the definitions of offset of fields in boot_ctx to the same header file that the struct boot_ctx is in, to reduce the possibility that some modification would make the offset inconsistant with the fields within the structure.

Also, the patch series adds two files to do compile time assert.
One is arch specific, and put in hypervisor/arch/x86/.
The other one is common, and put in hypervisor/common/.
If the statement is not true, there will be error during compile time.
The source files will not increase the size of HV binary.


dongyaozu:
@binbinwu1  can u check the failure?

binbinwu1:
@dongyaozu Seems that the pre-merge integration system has some issue.

lijinxia:
Verified+1

PR #1017, opened 2018-08-14
Logicalconjunctionbracket

yuchuyang:
Added brackets to logical conjunctions which have precedence level is less than 13.


wenlingz:
Verified+1

PR #1020, opened 2018-08-14
hv: treewide: fix 'Shifting value too far'

shiqingg:
MISRA-C requires that shift operation cannot exceed the word length.

What this patch does:
- Add the pre condition for 'init_lapic' regarding to 'pcpu_id'
  Currently, max 8 physical cpus are supported.
  Re-design will be required if we would like to support more physical
   cpus.
  So, add the pre condition here to avoid the unintentional shift
   operation mistakes.

- Replace the id type with uint8_t in 'vlapic_build_id'
  - For VM0, it uses 'lapic_id' as its id, which is uint8_t.
  - For non VM0, it uses 'vcpu_id' as its id, which is uint16_t.
    Cast this id to uint8_t to make sure there is no loss of data after
     left shifting 24U.


lijinxia:
Verified+1

PR #1021, opened 2018-08-14
HV: io: drop REQ_STATE_FAILED

junjiemao1:
Now the DM has adopted the new VHM request state transitions and
REQ_STATE_FAILED is obsolete since neither VHM nor kernel mediators will set the
state to FAILED.

This patch drops the definition to REQ_STATE_FAILED in the hypervisor, makes
''processed'' unsigned to make the compiler happy about typing and simplifies
error handling in the following ways.

* (dm_)emulate_(pio|mmio)_post no longer returns an error code, by introducing a
  constraint that these functions must be called after an I/O request
  completes (which is the case in the current design) and assuming
  handlers/VHM/DM will always give a value for reads (typically all 1's if the
  requested address is invalid).

* emulate_io() now returns a positive value IOREQ_PENDING to indicate that the
  request is sent to VHM. This mitigates a potential race between
  dm_emulate_pio() and pio_instr_vmexit_handler() which can cause
  emulate_pio_post() being called twice for the same request.

* Remove the ''processed'' member in io_request. Previously this mirrors the
  state of the VHM request which terminates at either COMPLETE or FAILED. After
  the FAILED state is removed, the terminal state will always be constantly
  COMPLETE. Thus the mirrored ''processed'' member is no longer useful.

Note that emulate_instruction() will always succeed after a reshuffle, and this
patch takes that assumption in advance. This does not hurt as that returned
value is not currently handled.

This patch makes it explicit that I/O emulation is not expected to fail. One
issue remains, though, which occurs when a non-aligned cross-boundary access
happens. Currently the hypervisor, VHM and DM adopts different policy:

* Hypervisor: inject #GP if it detects that the access crossed boundary

* VHM: deliver to DM if the access does not complete falls in the range of a
  client

* DM: a handler covering part of the to-be-accessed region is picked and
  assertion failure can be triggered.

A high-level design covering all these components (in addition to instruction
emulation) is needed for this. Thus this patch does not yet cover the issue.

Tracked-On: #875

### DIFF
--- a/hypervisor/arch/x86/ept.c
+++ b/hypervisor/arch/x86/ept.c
@@ -156,7 +156,6 @@ int ept_violation_vmexit_handler(struct vcpu *vcpu)
 	exit_qual = vcpu->arch_vcpu.exit_qualification;
 
 	io_req->type = REQ_MMIO;
-	io_req->processed = REQ_STATE_PENDING;
 
 	/* Specify if read or write operation */
 	if ((exit_qual & 0x2UL) != 0UL) {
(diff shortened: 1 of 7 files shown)


lijinxia:
OK to verify

PR #1026, opened 2018-08-14
HV: pirq: renamed irq APIs

lyan3:
Two commits included:
hv: pirq: add static irq:vector mappings
 - to remove "vector" in irq API for special irq with vector determined statically;

hv: pirq: rename common irq APIs
 - renamed the APIs to request_irq()/free_irq()


lijinxia:
Verified+1

PR #1029, opened 2018-08-15
hv: vioapic: bug fix update PTDEV RTE

lifeix:
Now the guest may change "Destination Field", "Trigger Mode",
"Interrupt Input Pin Polarity" even "Interrupt Vector" when
"Interrupt Mask" not masked. So we should update the pass through
device interrupt pin rte in this situation. The old logic would
update it only when "Interrupt Mask" or "Trigger Mode" or
"Interrupt Input Pin Polarity" was changed.

update ptdev native ioapic rte when (a) something changed and
(b) the old rte interrupt mask is not masked or the new rte interrupt
mask is not masked.

### DIFF
--- a/hypervisor/dm/vioapic.c
+++ b/hypervisor/dm/vioapic.c
@@ -34,7 +34,7 @@
 
 #define REDIR_ENTRIES_HW	120U /* SOS align with native ioapic */
 #define	RTBL_RO_BITS	(uint32_t)(IOAPIC_RTE_REM_IRR | IOAPIC_RTE_DELIVS)
-#define NEED_TMR_UPDATE (~(IOAPIC_RTE_INTMASK | IOAPIC_RTE_INTPOL))
+#define NEED_TMR_UPDATE (IOAPIC_RTE_TRGRMOD | IOAPIC_RTE_DELMOD | IOAPIC_RTE_INTVEC)
 
 #define ACRN_DBG_IOAPIC	6U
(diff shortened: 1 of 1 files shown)


hypervisor/dm/vioapic.c
```diff
@@ -405,15 +401,9 @@ vioapic_indirect_write(struct vioapic *vioapic, uint32_t addr, uint32_t data)
...
-		 * remap for deactive: interrupt mask & vector set to 0
-		 * remap for trigger mode change
-		 * remap for polarity change
-		 */
-		if ( ((changed & IOAPIC_RTE_INTMASK) != 0UL) ||
-		     ((changed & IOAPIC_RTE_TRGRMOD) != 0UL) ||
-		     ((changed & IOAPIC_RTE_INTPOL ) != 0UL) ) {
-
+		/* remap for ptdev */
+		if (((new.full & IOAPIC_RTE_INTMASK) == 0U) ||
```
dongyaozu: 0UL

hypervisor/dm/vioapic.c
```diff
@@ -405,15 +401,9 @@ vioapic_indirect_write(struct vioapic *vioapic, uint32_t addr, uint32_t data)
...
+				((last.full & IOAPIC_RTE_INTMASK) == 0U)) {
```
dongyaozu: 0UL

lijinxia:
Verified+1

Issue #861, opened 2018-08-06
MISRA C compliance


hwang37:
[External_System_ID] 247031

ACRN code should support MISRA C compliance coding style.
Note: MISRA-C is a standard guideline for Safety Critical Software.

hwang37:
[External_System_ID] 247031

PR #1031, opened 2018-08-15
hv: vuart: fix 'Shifting value too far'

shiqingg:
MISRA-C requires that shift operation cannot exceed the word length.

What this patch does:
- Fix the bug in 'vuart_init'
  The register 'dll' and 'dlh' should be uint8_t rather than char.
  'dll' is the lower 8-bit of divisor.
  'dlh' is the higher 8-bit of divisor.
  So, the shift value should be 8U rather than 16U.
- Fix other data type issues regarding to the registers in 'struct
  vuart'
  The registers should be unsigned variables.

v1 -> v2:
 * Use a local variable 'uint8_t value_u8 = (uint8_t)value' to avoid
   mutiple times type conversion
 * Use '(uint8_t)divisor' rather than '(uint8_t)(divisor & 0xFFU)' to
   get the lower 8 bit of 'divisor'
   Direct type conversion is safe and simple per Xiangyang's suggestion.

Tracked-On: #861

### DIFF
--- a/hypervisor/debug/vuart.c
+++ b/hypervisor/debug/vuart.c
@@ -133,12 +133,13 @@ static void vuart_toggle_intr(struct vuart *vu)
 	}
 }
 
-static void vuart_write(__unused struct vm_io_handler *hdlr,
-		struct vm *vm, uint16_t offset_arg,
-		__unused size_t width, uint32_t value)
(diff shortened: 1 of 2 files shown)


lijinxia:
Verified+1

PR #1038, opened 2018-08-15
hv: Build fix for Partition mode

sainath14:
struct io_request gets rid of "processed" element. Modified io exit handler
for partition mode to accomodate the change.


lijinxia:
Verified+1

PR #1040, opened 2018-08-16
DM: adding boot option of "i915.enable_guc=0"

ailun258:
Adding boot option of "i915.enable_guc=0" to disable Guc on UOS for latest kernel

Signed-off-by: ailun258 <ailin.yang@intel.com>


ailun258:
@feijiang1 @yakuizhao  please help to review, this is to align with your patches for NUC

PR #1045, opened 2018-08-16
hv: sw_loader for VMs in ACRN partition mode

sainath14:
ACRN in partition mode needs to boot VMs from HV. This patch make ACRN
treat all VMs in the same way (as vm0)  w.r.t. setting up the guest memory
for kernel, bootargs and zeropage


wenlingz:
Verified +1

PR #1047, opened 2018-08-16
Add the boot option to disable guc on SOS/UOS

yakuizhao:
Add the boot option to disable guc on SOS/UOS kernel


lijinxia:
Verified+1

PR #1048, opened 2018-08-16
hv: fix size issue in mptable guest copy - ACRN partition mode

sainath14:
1)Fix to take into account the size of proc entries in MPtable.
2)Change APIC ids to virtual APIC IDs

### DIFF
--- a/hypervisor/arch/x86/guest/mptable.c
+++ b/hypervisor/arch/x86/guest/mptable.c
@@ -13,8 +13,15 @@
 
 #define MPTABLE_BASE		0xF0000U
 
-/* floating pointer length + maximum length of configuration table */
-#define	MPTABLE_MAX_LENGTH	(65536 + 16)U
+/* 
+ * floating pointer length + maximum length of configuration table
+ * ACRN uses contiguous guest memory from 0xF0000 to place floating pointer
+ * structure and config table. Maximum length of config table is 64K. So the
+ * maximum length of combined floating pointer and config table can go up to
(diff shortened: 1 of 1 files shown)


hypervisor/arch/x86/guest/mptable.c
```diff
@@ -292,12 +292,16 @@ int mptable_build(struct vm *vm)
...
 	struct mpfps            *mpfp;
+	size_t			mptable_length, table_length;
 
 	startaddr = (char *)GPA2HVA(vm, MPTABLE_BASE);
+
+	table_length = vm->vm_desc->mptable->mpch.base_table_length;
+	mptable_length = sizeof(struct mpfps) + table_length;
 	/* Copy mptable info into guest memory */
-	(void)memcpy_s((void *)startaddr, sizeof(struct mptable_info),
+	(void)memcpy_s((void *)startaddr, MPTABLE_MAX_LENGTH,
```
dongyaozu: Why MPTABLE_MAX_LENGTH is (65536 + 16)U ?
And I am not sure if (65536 + 16)U is an valid expression.

sainath14:
the length field in MP config header is 16 bit long. So that should be the reason the max length is set to 65536 bytes. And 16 bytes is the size of MP floating pointer structure. So together it is 65536+16. We can add logic to determine the length and reserve only the exact length in e820. Right now we are reserving 65536+16 in e820.

hypervisor/arch/x86/guest/mptable.c
anthonyzxu: sainath fixed it.

wenlingz:
Verified +1

PR #1049, opened 2018-08-16
hv: Support HV console for multiple VMs - ACRN partition mode

sainath14:
ACRN in partition mode provides vUART for all VMs. This patch adds
support to add console redirection for multiple VMs.


anthonyzxu:
Both linux and android boot are successful as UOS on GP,
Linux boot failed as UOS on NUC,

It is unlikely caused by this patch.

Can you do the test on NUC without this patch?

wenlingz:
Verified+1

PR #1051, opened 2018-08-16
hv: vm_description fix for partition ACRN

sainath14:
1) Changed VM memory size from 2 GB to 512 MB
	Some platforms might not have a total of 4 GB above 4 GB in host e820
	With this change, we need a total of only 1 GB above 4 GB in host e820
2) Update e820 entries for the above change
3) type fix for VM1_CPUS and VM2_CPUS
4) Add mptable pointer info to each VM vm_desc


wenlingz:
Verified +1

PR #1053, opened 2018-08-16
hv: ptdev: simplify ptdev_intx_pin_remap logic

lifeix:
Since this interrupt pin is passed through to the guest, we could set
it as we like. There's no need a flag to emulate the masked status.

### DIFF
--- a/hypervisor/arch/x86/assign.c
+++ b/hypervisor/arch/x86/assign.c
@@ -663,6 +663,7 @@ static void activate_physical_ioapic(struct vm *vm,
 {
 	union ioapic_rte rte;
 	uint32_t phys_irq = entry->allocated_pirq;
+	uint32_t intr_mask;
 	bool is_lvl_trigger = false;
 
 	/* disable interrupt */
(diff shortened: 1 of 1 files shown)


hypervisor/arch/x86/assign.c
```diff
@@ -671,18 +671,14 @@ static void activate_physical_ioapic(struct vm *vm,
...
-	ioapic_set_rte(phys_irq, rte);
-
 	/* update irq trigger mode according to info in guest */
 	if ((rte.full & IOAPIC_RTE_TRGRMOD) == IOAPIC_RTE_TRGRLVL) {
 		is_lvl_trigger = true;
 	}
 	set_irq_trigger_mode(phys_irq, is_lvl_trigger);
 
-	/* enable interrupt */
-	GSI_UNMASK_IRQ(phys_irq);
```
anthonyzxu: phys_irq is masked in this function, I think we need unmask here.

lifeix:
@anthonyzxu @JasonChenCJ 
We can just simple the logic to call activate_physical_ioapic in ptdev_intx_pin_remap without change ptdev_intx_pin_remap. But if the guest mask the interrupt and the device could raise interrupt, our hypervisor would waste time to deal with this although we drop the interrupt at last.

lifeix:
@NanlinXie @lijinxia 
Please help to have a test
Thanks.

lijinxia:
Verified+1

PR #1056, opened 2018-08-16
hv: Fix comments referring to wrong hypervisor name

sainath14:
Fix comments referring to wrong hypervisor name


wenlingz:
Verified +1

PR #1057, opened 2018-08-16
HV: Add const qualifiers where required

rarindam:
V1:
In order to better comply with MISRA C,
add const qualifiers whereever required.
In the patch, these are being added to pointers
which are normally used in "get" functions.

V2: Corrected the issues in the patch
pointed by Junjie in his review comments.
Moved the const qualifiers to the correct
places. Removed some changes which are not
needed.

V3: Updated patch comment.
This modifies a subset of all the functions
which might need constant qualifiers
for the their parameters.
This is not and exhaustive patch. This only
targets obvious places where we can use
the const qualifier. More changes will be
submitted in future patches, if required.

### DIFF
--- a/hypervisor/arch/x86/cpu_state_tbl.c
+++ b/hypervisor/arch/x86/cpu_state_tbl.c
@@ -89,7 +89,7 @@ static const struct cpu_state_table {
 	}
 };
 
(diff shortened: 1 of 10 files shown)


wenlingz:
Verified +1

PR #1058, opened 2018-08-17
hv: fixed compiling warning

mingqiangchi:
removed some unnecessary variables and functions.

v1-->v2:
   Replace div-by-zero with an inline ASM code


lijinxia:
Verified+1

PR #1059, opened 2018-08-17
WAVE-2: 2nd patch set for vioapic/vlapic/vpic reshuffle

ywan170:
The 2nd patch set is majorly for vmx vapic reshuffle. The lvt, nmi and split vmx_vapic are still WIP. 

-refine the interrupt-window related code.
-fix several bugs.
-remove x2apic, APIC ID register related code.
-cleanup the APIC-access VM exit handler.
-rename the apicv to vmx_vapic in vlapic.c.
-rename the vapic to vmx_vapic in virq.c.

v2:
-5 acked patches already submit pr, so removed them in the v2.
-change the name from vapic to apicv
-update the apicv APIs with vlapic_apicv prefix -keep irq_window_enabled and make it useful -removed several patches which are not real necessary

This is the 2nd wave for this patch set


hypervisor/arch/x86/virq.c
```diff
@@ -469,11 +448,32 @@ int acrn_handle_pending_request(struct vcpu *vcpu)
...
+						pending_req_bits) == 0U) &&
+		((is_apicv_intr_delivery_supported() == true) ||
+		 (vlapic_pending_intr(vcpu->arch_vcpu.vlapic,
+							  &vector) == 0U))) {
+		return ret;
+	}
 
 	/* Enable interrupt window exiting if pending */
-	if (intr_pending && arch_vcpu->irq_window_enabled == 0U) {
+	if (arch_vcpu->irq_window_enabled == 0U) {
```
anthonyzxu: if (arch_vcpu->irq_window_enabled == 0U) {
this can be removed, it is checked here
 if ((arch_vcpu->irq_window_enabled == 1U) ||
		(is_guest_irq_enabled(vcpu) == true)) {
		return ret;
	}

ywan170: Correct. I will remove it

ywan170:
@anthonyzxu Just upload the new version. Please help check..

ywan170:
@anthonyzxu please hold first. I found that the second patch cause the system can't boot if disabled delivery feature. Let me check it first.

ywan170:
I just did some investigation. 
The statement "When new ACRN_REQUEST_EVENT be triggered, it was already generate IPI to
the target pcpu." is correct. But these has one scenario, need call vcpu_pending_request to set ACRN_REQUEST_EVENT again.

If we disabled "virtual-interrupt delivery", then hypervisor has to evaluate&deliver interrupts by itself.
1, trying to send vector 253 to vcpu0, set the ACRN_REQUEST_EVENT with ipi. And the vector 253 priority is 0xf0.
2, clear and handle ACRN_REQUEST_EVENT. 
2.1, due to 0xf0 is higher than current PPR(0x0, due to TPR is 0x0), so update the ISR with 253.
2.2, update PPR to 0xf0.
3, trying to send vector 239 to vcpu0, set the ACRN_REQUEST_EVENT with ipi. The vector 239 priority is 0xe0.
4, clear and handle ACRN_REQUEST_EVENT. 
4.1, due to 0xe0 is lower than current PPR(0xf0), so be ignored, and ISR hasn't update.
5, vector 253 EOI is coming, apic write vmexit be triggered.
5.1, clear vector 253 of ISR.
5.2, update PPR to 0x00 due to TPR is 0x0.
6, need check if any IRR, if yes. Then need to set ACRN_REQUEST_EVENT flag and trigger ipi if needed.
7, re-evaluate the vector 239. Due to its priority 0xe0 is higher than 0x0, so set ISR, PPR and start to inject interrupt...

The step 6 is done by vcpu_pending_request called in acrn_handle_pending_request...

So we need update the check condition to cover this case... I will upload one new version ..

ywan170:
For step 6, I prepare to put it during update PPR(vlapic_update_ppr). If the new PPR lower than old PPR value, then to check IRR if need to set ACRN_REQUEST_EVENT with ipi. 

For delivery enable case, I need to do more investigation about which is the better place to sync pir to irr.

ywan170:
@anthonyzxu as we talked, let's merge last three patches first. For the first three patches, I will do more investigation which should be more changes are needed.

dongyaozu:
@ywan170   let us merge the mature part first. Then you need to submit a separate PR. The system cannot merge part of a PR.

anthonyzxu:
@ywan170 step 6 should be done in EOI VM_exit handler.

anthonyzxu:
@ywan170 https://github.com/projectacrn/acrn-hypervisor/pull/1085 is approved, seems we don't need this PR.

anthonyzxu:
@ywan170  if virtual interrupt delivery is not enabled, LAPIC EOI write causes APIC-write VM Exit. in both case, interrupt should be evaluated, step 6 should be executed.

ywan170:
@anthonyzxu Yes. That is almost same like what I mentioned. The APIC-WRITE VM Exit will call apic_write_vmexit_handler->vlapic_process_eoi->vlapic_update_ppr

The vlapic_update_ppr will update ppr according to current pending irr. If the updated ppr value lower than old ppr value then trigger ACRN_REQUEST_EVENT with IPI if any pending irr..

ywan170:
@anthonyzxu Just re-checked the code. Maybe you are right, we need do step 6 outside of vlapic_update_ppr due to it will also be called in other scenario, for example acrn_handle_pending_request.

I will continue handle these cases in next week. This week is busying on other urgent works...

PR #1063, opened 2018-08-17
samples: let nuc uos only start with 1 cpu

JasonChenCJ:
there are only 2 CPUs on UP2 platform, so we can only start 1 cpu for
uos on UP2 platform.

as this is only an example script, user can modify the uos start CPU
number according his requirement.


lijinxia:
Verified+1

PR #1065, opened 2018-08-17
hv:fixed several return value violations

mingqiangchi:
-- change two timer callbacks to void type
-- ignore the return value for add_timer
-- add (void) before several functions(memset/memcpy/
   vcpu_get_xxx)

v1-->v2:
   ignore the return value for add_timer


lijinxia:
Verified+1

PR #1067, opened 2018-08-20
HV: fix potential memory leak in 'vlapic_create()'

yonghuah:
- should free allocated memory in case of registering
   mmio handler failure.

### DIFF
--- a/hypervisor/arch/x86/guest/vlapic.c
+++ b/hypervisor/arch/x86/guest/vlapic.c
@@ -52,7 +52,7 @@
 static inline void vlapic_dump_irr(struct acrn_vlapic *vlapic, char *msg)
 {
 	uint32_t i;
-	struct lapic_reg *irrptr = &(vlapic)->apic_page->irr[0];
+	struct lapic_reg *irrptr = &(vlapic->apic_page.irr[0]);
 
 	for (i = 0U; i < 8U; i++)
 		dev_dbg(ACRN_DBG_LAPIC, "%s irr%u 0x%08x",
@@ -62,7 +62,7 @@ static inline void vlapic_dump_irr(struct acrn_vlapic *vlapic, char *msg)
 static inline void vlapic_dump_isr(struct acrn_vlapic *vlapic, char *msg)
 {
 	uint32_t i;
-	struct lapic_reg *isrptr = &(vlapic)->apic_page->isr[0];
+	struct lapic_reg *isrptr = &(vlapic->apic_page.isr[0]);
 
 	for (i = 0U; i < 8U; i++) {
 		dev_dbg(ACRN_DBG_LAPIC, "%s isr%u 0x%08x",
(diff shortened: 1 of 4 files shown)


hypervisor/arch/x86/guest/vlapic.c
```diff
@@ -2086,6 +2086,8 @@ int vlapic_create(struct vcpu *vcpu)
 				(uint64_t)DEFAULT_APIC_BASE +
```
dongyaozu: Even at very beginning, it may fail..

int vlapic_create(struct vcpu *vcpu)
{
        void *apic_page = alloc_page();
        struct acrn_vlapic *vlapic = calloc(1U, sizeof(struct acrn_vlapic));


Given that we want to use static memory allocation more, how about we do along this direction?
for now, let us move apic_page into vlapic structure...

dongyaozu:
Yes, this is good direction. We can gradually move other dynamically allocated memory to be static too. Please double check with Junjie

yonghuah:
@dongyaozu  
have updated, thanks. -Yonghua

hypervisor/arch/x86/guest/vlapic.c
```diff
@@ -447,7 +447,7 @@ vlapic_esr_write_handler(struct acrn_vlapic *vlapic)
 {
 	struct lapic_regs *lapic;
 
-	lapic = vlapic->apic_page;
+	lapic = &vlapic->apic_page;
```
dongyaozu: do we need to use   &(vlapic->apic_page)?   check with Junjie.

hypervisor/arch/x86/guest/vlapic_priv.h
```diff
@@ -115,9 +115,10 @@ struct vlapic_timer {
 };
 
 struct acrn_vlapic {
+	/* Keep 'apic_page' field be the first one */
+	struct lapic_regs	apic_page;
```
dongyaozu: apic_page must be 4KB aligned, keep 'apic_page' field be the first for this.

hypervisor/arch/x86/guest/vlapic.c
junjiemao1: The code without parentheses can seem confusing to me if someone is not clear on the precedence of & and ->. However MISRA is neutral on using the parentheses here and we don't have any static checker that report on their missing. How about we make it an extra rule in the coding guideline for ACRN and I go find some alternatives that help us looking for any missed parenthesis?

dongyaozu: Agree! Yonghua will be the 1st to start this rule. And we will enforce later on.

yonghuah: @junjiemao1 @dongyaozu 
Thanks for comments, will update following this rule.

lijinxia:
Verified+1

PR #1068, opened 2018-08-20
dm: pass vrpmb key via cmos interface

binbinwu1:
CMOS offset from 0x20 to 0x9F is used to store rpmb key information.
vsbl loader will init vrpmb key in CMOS when boot/reboot.
vsbl loader will not init vrpmb key during S3 resume.
vsbl will read vrpmb key via CMOS interface.
After reading, the key value is cleared in CMOS. So the key can only be
read once until next boot.

### DIFF
--- a/devicemodel/hw/platform/cmos_io.c
+++ b/devicemodel/hw/platform/cmos_io.c
@@ -21,20 +21,26 @@
  * writing.
  *************************************************************************/
 
-/* cmos io device is used for android device reboot to bootloader or
- * recovery or normal boot usage
+/* cmos io device
+ * - nvram 0x10 ~ 0x1F is used for android device reboot to bootloader or
+ *   recovery or normal boot usage
+ * - vrpmb 0x20 ~ 0x9F is used to store vrpmb for guest, read to clear
  */
 
 #include <stdio.h>
 #include <assert.h>
+#include <stdbool.h>
 
 #include "inout.h"
+#include "vmmapi.h"
+#include "vrpmb.h"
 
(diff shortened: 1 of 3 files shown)


lijinxia:
Verified+1

devicemodel/include/vmmapi.h
```diff
@@ -61,6 +63,10 @@ struct vmctx {
 	void *atkbdc_base;
 	void *vrtc;
 	void *ioc_dev;
+	/* cmos buffer used to store write/read contents,
+	 * and it should not be cleared when reboot
```
bzhu5: "should not" or "should"?

devicemodel/hw/platform/cmos_io.c
```diff
@@ -88,10 +89,15 @@ cmos_io_handler(struct vmctx *ctx, int vcpu, int in, int port, int bytes,
 			return -1;
 		}
 
-		if (in)
-			*eax = cmos_buffer[buf_offset];
+		if (in) {
+			*eax = ctx->cmos_buffer[buf_offset];
```
bzhu5: i'm confused here. *eax can hold 256 bit data?

by the way, can you let me know where is the code or patch in vSBL to retrieve/read vRPMB key from CMOS? pretty thanks.

devicemodel/include/vmmapi.h
binbinwu1: The data should not be cleared when reboot.
For the content for vrpmb, the data is always cleared after read.

PR #1069, opened 2018-08-20
hv: vioapic: fix 'No definition in system for prototyped procedure'

shiqingg:
Remove two APIs without definition


lijinxia:
Verified+1

PR #1070, opened 2018-08-20
fixing cpu cores allocation for APL NUC

ailun258:
by default,  in the script ``launch_uos.sh`` file,  the last line -  it is ``launch_clear 1 1 "64 448 8" 0x070F00 clear``
it means there are 2 cpu cores available, one for SOS and another for UOS.  but APL NUC has 4 cpu cores provided, 
so we need to modify this to use 4 cpu cores,  for example, 1 for SOS, other 3 for UOS. so we modify this line to 
``launch_clear 1 1 "64 448 8" 0x070F00 clear ``

Signed-off-by: ailun258 <ailin.yang@intel.com>

### DIFF
--- a/doc/getting-started/apl-nuc.rst
+++ b/doc/getting-started/apl-nuc.rst
@@ -331,7 +331,7 @@ Set up Reference UOS
       :language: bash
       :emphasize-lines: 26,28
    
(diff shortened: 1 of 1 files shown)


ailun258:
modify it to ``launch_clear 1 3 "64 448 8" 0x070F00 clear``

PR #1072, opened 2018-08-20
hv:fixed MISRA-C return value violations

mingqiangchi:
-- change send_start_ipi/do_copy_earlylog to void type
-- drop the return value for vcpu_queue_execption
   when inject GP/PF/UD/AC/SS


lijinxia:
Verified+1

PR #1075, opened 2018-08-21
hv: vmx: fix 'Array has no bounds specified'

shiqingg:
MISRAC requires that the array size should be declared explicitly.

This patch fixes the issues caused by vm0_boot_context.

Fix pattern is like below:
   extern char start_of_ROM, end_of_ROM, start_of_FLASH;
   memcpy (& start_of_FLASH, & start_of_ROM, & end_of_ROM - &
start_of_ROM);

Tracked-On: #861


lijinxia:
Verified+1

PR #1076, opened 2018-08-21
DM USB: xHCI: Change the default USB xHCI support to pass through.

YangLiang3:
So far, the xHCI mediator does not support USB HUB over usb pmapper. But
there are some USB touch devices integrated internal USB HUB, then
cause such touch models can't be supported. Will consider to re-enable
xHCI mediator by default once HUB support related patches get merged in
the near future.

Signed-off-by: Liang Yang <liang3.yang@intel.com>
Reviewed-by: Binbin Wu<binbin.wu@intel.com>


wenlingz:
Verified+1

PR #1078, opened 2018-08-21
hv: vtd: use EPT as translation table for PTDev in SOS

binbinwu1:
In current code, the address translation type for passthrough devices
in SOS is passthrough type, which means the device in SOS can access
the whole physcial memory.
This patch uses SOS EPT to translate address for passthrough
devices. So that the memory of VMM or trusty can be protected from
SOS PTDev.

### DIFF
--- a/hypervisor/arch/x86/vtd.c
+++ b/hypervisor/arch/x86/vtd.c
@@ -168,7 +168,8 @@ static uint32_t dmar_hdrh_unit_count;
 static uint32_t max_domain_id = 63U;
 static uint64_t domain_bitmap;
 static spinlock_t domain_lock;
-static struct iommu_domain *host_domain;
+
+static struct iommu_domain *vm0_domain;
 static struct list_head iommu_domains;
 
 static void dmar_register_hrhd(struct dmar_drhd_rt *dmar_uint);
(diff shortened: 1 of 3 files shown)


binbinwu1:
I am checking the failure on NUC test.

hypervisor/arch/x86/vtd.c
```diff
@@ -1284,22 +1285,34 @@ void resume_iommu(void)
...
 
-	host_domain = create_host_domain();
+	vm0_domain = (struct iommu_domain *) vm0->iommu;
 
 	for (bus = 0U; bus <= IOMMU_INIT_BUS_LIMIT; bus++) {
 		for (devfun = 0U; devfun <= 255U; devfun++) {
-			add_iommu_device(host_domain, 0U,
-					 (uint8_t)bus, (uint8_t)devfun);
+#ifdef HV_DEBUG
+			if ((bus == platform_uart_bus) &&
```
dongyaozu: Having a 2nd thinking, we are setting up IOMMU tables for all possible BDFs...  This is not wrong, but redundant already. Why we care about adding one more entry for UART BDF?

I thought if we skip this, we can simplify the code.

hypervisor/arch/x86/Kconfig
```diff
@@ -139,3 +139,13 @@ config MTRR_ENABLED
...
+
+config UEFI_UART_BDF
+	hex
+	depends on PLATFORM_UEFI
+	default 0x000000C1
+
+config SBL_UART_BDF
+	hex
+	depends on PLATFORM_SBL
+	default 0x000000C2
```
dongyaozu: If we add this, should we add the CONFIG_SERIAL_MMIO_BASE definition here too?

hypervisor/arch/x86/vtd.c
anthonyzxu: Since Hypervisor doesn't do UART DMA, it is safe to skip it.

binbinwu1:
@dongyaozu Need Eddie's decision to decide to merge it or not. Because the fix for memory conherecy of EPT table  issue is a temp solution.

dongyaozu:
@binbinwu1  can you check how we modify EPT? Anyway, this is a general issue that we must solve. I remember only write-protection of guest page table need to do so, in this case, we don't need to flush IOMMU.

dongyaozu:
There are three options for this:
Op1: remove the PCIe device from SOS VTD table, we need to know UART BDF and configure per platform.
Op2: If we leave as it is (SOS VTD table covers the PCIe UART), it is safe since UART doesn't use DMA.
Opt3: Not present IOMMU translation table for BDF of UART PCIe – this still require us to have the code to let the HV know the BDF of UART.

Let us go with opt2 for now. We can revisit later on.

wenlingz:
Verified +1

PR #1079, opened 2018-08-22
hv: debug: add the hypervisor NPK log

zjin7:
This series of patches implement a log destination for the hypervisor,
npk_log, which is similar to the console_log and the mem_log.
The Intel Trace Hub (aka. North Peak, NPK) is a trace aggregator for
Software, Firmware, and Hardware. The npk_log utilizes the NPK by
writing the hypervisor logs directly to the MMIO address of the
reserved NPK master, so that the hypervisor logs can be consolidated
together with any other logs sent to the NPK from other parts of the
virtualization platform with an unified timestamp.
In order to enable/disable/configure the npk_log from the SOS kernel,
a hypercall HC_SETUP_HV_NPK_LOG (the first patch of the series) and a
kernel driver (covered by another patch series) are also added.

### DIFF
--- a/hypervisor/arch/x86/Kconfig
+++ b/hypervisor/arch/x86/Kconfig
@@ -50,7 +50,7 @@ config LOG_BUF_SIZE
 
 config LOG_DESTINATION
 	int "Bitmap of consoles where logs are printed"
-	default 3
+	default 7
 
 config CPU_UP_TIMEOUT
 	int "Timeout in ms when bringing up secondary CPUs"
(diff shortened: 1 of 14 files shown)


hypervisor/arch/x86/Kconfig
```diff
@@ -50,7 +50,7 @@ config LOG_BUF_SIZE
 
 config LOG_DESTINATION
 	int "Bitmap of consoles where logs are printed"
-	default 3
+	default 7
```
dongyaozu: Why we change the default log level ?  Turn on NKP should be some time later..

zjin7: LOG_DESTINATION is not the default log level.
The npk_log is disabled by default, and can be enabled by hypercall in runtime.

wenlingz:
Verified+1

PR #1081, opened 2018-08-22
dm: cmos: fix a logic error for read to clear range

binbinwu1:


lijinxia:
Verified+1

PR #1083, opened 2018-08-22
hv: vcpuid: disable some features in cpuid

lifeix:
We didn't emualte them well, so just disable them.
Here are PMU CPUID.0A, Intel RDT CPUID.0F and CPUID.10, Intel SGX CPUID.12,
Intel Processor Trace CPUID.14


lijinxia:
Verified+1

PR #1085, opened 2018-08-22
Apicv 2nd wave new

ywan170:
These three patches from “2nd patch set for vioapic/vlapic/vpic reshuffle”. The remain 3 patches in that set need to continue investigation, but these 3 patches are mature to be merged.


ywan170:
I will do the rebase..

lijinxia:
Verified+1

PR #1086, opened 2018-08-22
HV: Removed unused parameters and unions from gdt

yuchuyang:
In the gdt.h, it defines unions and parameters which is not used
in hypervisor. Removed them to keep the gdt definition simple.

### DIFF
--- a/hypervisor/include/arch/x86/gdt.h
+++ b/hypervisor/include/arch/x86/gdt.h
@@ -79,146 +79,13 @@
 #define TSS_AVAIL  (9U)
 
 /*
- * Definition of an 8 byte code segment descriptor.
- */
-union code_segment_descriptor {
-	uint64_t value;
-	struct {
-		union {
-			uint32_t value;
-			struct {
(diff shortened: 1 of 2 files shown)


hypervisor/include/arch/x86/gdt.h
```diff
@@ -78,146 +78,15 @@
...
 		uint32_t base_addr_63_32;
-		union {
-			uint32_t value;
-			struct {
-				uint32_t rsvd_7_0:8;
-				uint32_t bits_12_8_clr:4;
-				uint32_t rsvd_31_13:20;
-			} bits;
-		} offset_12;
 	} fields;
```
anthonyzxu: offset_12 is removed, then the size of tss_64_descriptor is 12 byte, which is supposed to be 16 byte.
you can use below definition.
struct tss_64_descriptor {
		uint32_t low32_value;
		uint32_t high32_value;
		uint32_t base_addr_63_32;
                uint32_t offset_12;
}

lijinxia:
Verified+1

PR #1087, opened 2018-08-23
HV: remove callbacks registration for APICv functions

yonghuah:
- call these functions directly, no need to register
     callbacks.


wenlingz:
Verified+1

PR #1088, opened 2018-08-23
ptdev cleanup

lifeix:
Try to simplify the ptdev logic:
1. simplify the lookup method: use source id to identify the ptdev;
2. only check whether phys_pin is valid in add_intx_remapping
3. remove unnecessary variable from structure ptdev_msi_info


wenlingz:
Verified+1

PR #1096, opened 2018-08-24
Mmu: cleanup mmu.h

lifeix:
Since we reshuffle mmu code, there're many MACRO or structure we don't use 
any more in mmu.h. This serial try to remove them and replace the outdated
with the new one.

### DIFF
--- a/hypervisor/arch/x86/cpu_primary.S
+++ b/hypervisor/arch/x86/cpu_primary.S
@@ -225,14 +225,14 @@ cpu_primary32_gdt_ptr:
     .align  0x1000
     .global cpu_boot32_page_tables_start
 cpu_boot32_page_tables_start:
-    /* 0x3 = (IA32E_COMM_P_BIT | IA32E_COMM_RW_BIT) */
+    /* 0x3 = (PAGE_PRESENT | PAGE_RW) */
     .quad   cpu_primary32_pdpt_addr + 0x3
     /*0x1000 = CPU_PAGE_SIZE*/
     .align  0x1000
 cpu_primary32_pdpt_addr:
     address = 0
     .rept   4
-    /* 0x3 = (IA32E_COMM_P_BIT | IA32E_COMM_RW_BIT) */
+    /* 0x3 = (PAGE_PRESENT | PAGE_RW) */
     .quad   cpu_primary32_pdt_addr + address + 0x3
     /*0x1000 = CPU_PAGE_SIZE*/
     address = address + 0x1000
@@ -242,7 +242,7 @@ cpu_primary32_pdpt_addr:
 cpu_primary32_pdt_addr:
     address = 0
     .rept  2048
-    /* 0x83 = (IA32E_PDPTE_PS_BIT | IA32E_COMM_P_BIT | IA32E_COMM_RW_BIT) */
+    /* 0x83 = (PAGE_PSE | PAGE_PRESENT | PAGE_RW) */
     .quad  address + 0x83
     address = address + 0x200000
     .endr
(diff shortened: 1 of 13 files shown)


hypervisor/arch/x86/cpu_primary.S
```diff
@@ -225,14 +225,14 @@ cpu_primary32_gdt_ptr:
     .align  0x1000
     .global cpu_boot32_page_tables_start
 cpu_boot32_page_tables_start:
-    /* 0x3 = (IA32E_COMM_P_BIT | IA32E_COMM_RW_BIT) */
+    /* 0x3 = (PAGE_PRESENT | PAGE_RW) */
```
dongyaozu: This is not to remove the unused MACRO definition...  It mismatch with the commit message.
BTW, why we need to change the name here? Can you explain a bit?

hypervisor/include/arch/x86/mmu.h
```diff
@@ -10,214 +10,68 @@
...
-#define MMU_32BIT_PDE_D       0x00000040U
-/** The 'Page Size' bit in a 32 bit paging page directory entry */
-#define MMU_32BIT_PDE_PS      0x00000080U
-/** The 'Global' bit in a 32 bit paging page directory entry */
-#define MMU_32BIT_PDE_G       0x00000100U
-/** The 'PAT' bit in a page 32 bit paging directory entry */
-#define MMU_32BIT_PDE_PAT     0x00001000U
+
 /** The flag that indicates that the page fault was caused by a non present
  * page.
```
dongyaozu: This comments seems to be confusing

hypervisor/include/arch/x86/mmu.h
```diff
@@ -10,214 +10,68 @@
...
  */
-#define PAGE_FAULT_P_FLAG    0x00000001U
+#define PAGE_FAULT_P_FLAG	0x00000001U
```
dongyaozu: Is this TAB or space?

hypervisor/arch/x86/trusty.c
```diff
@@ -106,9 +106,9 @@ static void create_secure_world_ept(struct vm *vm, uint64_t gpa_orig,
 	 */
 	sub_table_addr = alloc_paging_struct();
 	sworld_pml4e = HVA2HPA(sub_table_addr) | table_present;
-	mem_write64(pml4_base, sworld_pml4e);
+	*(uint64_t *)pml4_base = sworld_pml4e;
```
dongyaozu: why not use set_pgentry too?

hypervisor/arch/x86/trusty.c
```diff
@@ -106,9 +106,9 @@ static void create_secure_world_ept(struct vm *vm, uint64_t gpa_orig,
...
 
-	nworld_pml4e = mem_read64(vm->arch_vm.nworld_eptp);
+	nworld_pml4e = *(uint64_t *)vm->arch_vm.nworld_eptp;
```
dongyaozu: ditto

hypervisor/arch/x86/cpu_primary.S
lifeix: This is wants to set the hypervisor MMU page table attributes not guest EPT page table attributes.
It should use the Macro definition for hypervisor MMU page table attributes.

hypervisor/include/arch/x86/mmu.h
lifeix: Replace space with TAB

hypervisor/arch/x86/trusty.c
lifeix: You're right. I'm not aware of this.
Thanks.

hypervisor/include/arch/x86/mmu.h
lifeix: I think it wants to say "was caused by the present bit not set". 
This patch doesn't plan to resolve all the issues in mmu.h. Let's by pass this now.

PR #1097, opened 2018-08-24
hv: fix MISRA-C issues related to for loop

shiqingg:
This patch fixes the following issues:
- Assignment operation in expression.
- For loop incrementation is not simple.
- No brackets to loop body.
- Use of comma operator.

Tracked-On: #861


hypervisor/arch/x86/notify.c
```diff
@@ -35,7 +35,8 @@ void smp_call_function(uint64_t mask, smp_call_func_t func, void *data)
 
 	/* wait for previous smp call complete, which may run on other cpus */
 	while (atomic_cmpxchg64(&smp_call_mask, 0UL, mask & INVALID_BIT_INDEX));
-	while ((pcpu_id = ffs64(mask)) != INVALID_BIT_INDEX) {
+	pcpu_id = ffs64(mask);
+	while (pcpu_id != INVALID_BIT_INDEX) {
```
dongyaozu: for this kind of issues, previously we solve it using:
for (pcpu_id = ffs64(mask); pcpu_id != INVALID_BIT_INDEX; pcpu_id = ffs64(mask) {
}

can this solve the problem? it is just one LOC.

hypervisor/arch/x86/pagetable.c
```diff
@@ -264,7 +264,7 @@ int mmu_modify_or_del(uint64_t *pml4_page,
 	dev_dbg(ACRN_DBG_MMU, "%s, vaddr: 0x%llx, size: 0x%llx\n",
 		__func__, vaddr, size);
 	vaddr_end = vaddr + size;
-	for (; vaddr < vaddr_end; vaddr = vaddr_next) {
+	while (vaddr < vaddr_end) {
```
dongyaozu: so you mean we cann't use empty operations in “for". Is this the case?
If yes, how about we do:
for (vaddr_end = vaddr + size; vaddr < vaddr_end; vaddr = vaddr_next) {
}
??

hypervisor/arch/x86/vmx.c
```diff
@@ -606,15 +607,15 @@ static void init_guest_context_vm0_bsp(struct vcpu *vcpu)
 	struct ext_context *ectx =
 		&vcpu->arch_vcpu.contexts[vcpu->arch_vcpu.cur_context].ext_ctx;
 	struct boot_ctx * init_ctx = (struct boot_ctx *)(&vm0_boot_context);
-	uint16_t *sel;
+	uint16_t *sel = &init_ctx->cs_sel;
```
dongyaozu: we added another rule to use () for  & x->y.   please use &(x->y).

hypervisor/arch/x86/notify.c
shiqingg: The static checker requires that the loop incrementation should be simple, like ++ or -- operators.
If we use 'pcpu_id = ffs64(mask)' as the loop incrementation. The static checker will report the violation.

hypervisor/arch/x86/pagetable.c
shiqingg: The static checker requires that the loop incrementation should be simple, like ++ or -- operators.
So, 'vaddr = vaddr_next' is not a good loop incrementation cause.
I discussed this kind of issue with Junjie. We think that changing the for loop to a while loop does not affect the code readability and could help eliminate this kind of violation.

hypervisor/arch/x86/vmx.c
shiqingg: Got it. Will fix it in next version.

shiqingg:
New version is pushed to the same PR.

v1 -> v2:
 * Replace &x->y with &(x->y) based on our new coding rule

PR #1098, opened 2018-08-24
tools: acrn-manager: fix several warnings

lyan3:
This commit eliminates several compiler warnings via:
 - added 'const' quilifier to the 1st argument of send_msg();
 - added 'const' quilifier to vm ops functions.


lijinxia:
Verified+1

PR #1100, opened 2018-08-24
hv: Device MSIs in partition mode ACRN

sainath14:
VMs in partition mode uses a kernel patch to switch to physical mode xAPIC.
The kernel patch together with this patch makes sure device MSIs use
physical mode as DM and physical APIC IDs in Destination field. bootargs in
partition/vm_description.c uses a new parameter xapic_phys for the kernel patch
to take effect.


wenlingz:
Ready to merge

wenlingz:
Ready to merge

PR #1102, opened 2018-08-24
hv: init_iommu rework to enable IOMMU for partition mode ACRN

sainath14:
Current code enables IOMMU from init_iommu_vm0_domain which is called
from vm0 boot sequence. For partition mode VMs, this is not called as
VMs are numbered from 1.

This patch adds support to initialize root table pointer for each IOMMU
and enable all IOMMUs from init_iommu. Hence IOMMUs are enabled even
though ACRN does not boot vm0.


lijinxia:
SOS boot failed on NUC

sainath14:
is there a way to see boot log of SOS?

anthonyzxu:
There is a new PR for this. so close it

PR #1110, opened 2018-08-28
hv:No need to create inverted page tables for trusty memory

mingqiangchi:
There is no requirement to  hpa2gpa for trusty memory,
now we use the same inverted eptp for normal world and secure world,
that is it can get some trusty memory information by the inverted
page tables in normal world,then this patch removed the inverted
page tables for trusty memory.


wenlingz:
Ready to merge

Issue #1118, opened 2018-08-29
[Hypervisor]UOS boot in incorrect mode after reboot.


mwang106:
Version Tag: 2018w34.4.140000p
Reproduce steps:
1. UOS is Android OS.
2. Execute 'adb reboot' related commands:
adb reboot bootloader
adb reboot recovery

[ACTUAL OUTCOME]
Device boot in Android OS instead of fastboot or recovery mode.

mwang106:
Cannot be reproduced with ww36.3 aaag image.
Image info:
SOS version: 20180901-24780
acrn-hypervisor 2018w35.5.140000p
linux-pk414 4.14.66 90
SBL IFWI Version： SBL_APL_ACRN_1_0_2_127
IOC Version ：4. 0. 7

PR #1119, opened 2018-08-29
dm: cmos: move cmos storage out of vmctx

binbinwu1:
The whole vmctx will be cleared during cold reset.
cmos data should not be cleared during cold reset.
Move cmos data out of vmctx.

Tracked-On: #1118


wenlingz:
Verified+1

PR #1144, opened 2018-09-03
tools: acrn-crashlog: add usercrash_c in the pipe of core_pattern

chengangc:
This patch adds the usercrash client in the pipe of core_pattern
without affecting default core_pattern program.

In acrnprobe_prepare.sh, core_pattern will be set as usercrash-wrapper
with all of the arguments, which parses the parameters of the default
core_pattern program and the usercrash client, and then invokes them
separately.


chengangc:
Patch has been updated with Tracked-On info.

Issue #1146, opened 2018-09-03
UOS warm reset failure


gzhai7:
UOS warm reset fails as new EPT API refuse to remap same mem range.

mwang106:
AaaG build
Image info:
SOS version:   20180908-24920
acrn-hypervisor    2018w36.5.140000p
linux-pk414            4.14.67           94

PR #1153, opened 2018-09-05
DM: unmap ptdev BAR when deinit

gzhai7:
New EPT api treat remapping the same memseg as error, which causes UOS
warm restart failure with ptdev. This patch unmap ptdev BAR when
deinit to fix it.

Tracked-on: #1146


gzhai7:
Eddie,
Changed commits according to your comments.

jianlaipinan:
Ready to merge

PR #1167, opened 2018-09-06
hv: treewide: fix 'Macro parameter not in brackets'

shiqingg:
Add the brackets for Macro parameter to avoid the unintentional
mistakes.

A simple example that may cause mistakes:
        #define minus(x) -x
When the following call is made,
        z = minus(a-b)
it becomes:
        z = -a-b;
where "-a - b" is equivalent to "(-a) - b" rather than "- (a - b)", as
expected.

v1 -> v2:
 * Remove some changes to function like macro since MISRA-C requires to
   use inline functions if it is possible.
   These MACRO brackets violations will be fixed together when fixing
   other issues related to function like macro.

Tracked-On: #861


hypervisor/arch/x86/vtd.c
```diff
@@ -85,18 +85,18 @@ dmar_set_bitslice(uint64_t var, uint64_t mask,
 
 #define DMAR_OP_TIMEOUT CYCLES_PER_MS
 
-#define DMAR_WAIT_COMPLETION(offset, condition, status) \
-	do {                                                \
+#define DMAR_WAIT_COMPLETION(offset, condition, status)			\
```
dongyaozu: This is also API like macro, can we change it to inline API?

hypervisor/dm/vpic.c
```diff
@@ -51,10 +51,10 @@ enum irqstate {
 /*
  * Loop over all the pins in priority order from highest to lowest.
  */
-#define	PIC_PIN_FOREACH(pinvar, i8259, tmpvar)			\
-	for (tmpvar = 0U, pinvar = (i8259->lowprio + 1U) & 0x7U;	\
-	    tmpvar < NR_VPIC_PINS_PER_CHIP;				\
-	    tmpvar++, pinvar = (pinvar + 1U) & 0x7U)
+#define	PIC_PIN_FOREACH(pinvar, i8259, tmpvar)				\
```
dongyaozu: ditto

hypervisor/include/arch/x86/cpu.h
```diff
@@ -334,10 +334,10 @@ void stop_cpus(void);
 void wait_sync_change(uint64_t *sync, uint64_t wake_sync);
 
 /* Read control register */
-#define CPU_CR_READ(cr, result_ptr)                         \
-{                                                           \
-	asm volatile ("mov %%" __CPP_STRING(cr) ", %0"      \
-			: "=r"(*result_ptr));               \
+#define CPU_CR_READ(cr, result_ptr)				\
```
dongyaozu: ditto

hypervisor/arch/x86/vtd.c
shiqingg: 'status' is used to calculate the value of 'condition' when we use DMAR_WAIT_COMPLETION. And how to calculate 'condition' is case by case. We could not convert it to a API.

hypervisor/dm/vpic.c
```diff
@@ -51,10 +51,10 @@ enum irqstate {
...
+	for ((tmpvar) = 0U, (pinvar) = ((i8259)->lowprio + 1U) & 0x7U;	\
+	    (tmpvar) < NR_VPIC_PINS_PER_CHIP;				\
+	    (tmpvar)++, (pinvar) = ((pinvar) + 1U) & 0x7U)
```
dongyaozu: This MACRO has other issues:   1) 2 operations in  for ( X, Y; ..; M, N).  I think we also need to solve this issue

hypervisor/include/arch/x86/guest/guest.h
```diff
@@ -18,10 +18,10 @@
 
 #include <mmu.h>
 
-#define foreach_vcpu(idx, vm, vcpu)				\
-	for (idx = 0U, vcpu = vm->hw.vcpu_array[idx];		\
-		idx < vm->hw.num_vcpus;				\
-		idx++, vcpu = vm->hw.vcpu_array[idx])		\
+#define foreach_vcpu(idx, vm, vcpu)					\
+	for ((idx) = 0U, vcpu = vm->hw.vcpu_array[(idx)];		\
+		(idx) < vm->hw.num_vcpus;				\
```
dongyaozu: ditto

hypervisor/include/arch/x86/per_cpu.h
```diff
@@ -57,7 +57,7 @@ extern uint64_t pcpu_active_bitmap;
  * get percpu data for pcpu_id.
  */
 #define per_cpu(name, pcpu_id)	\
-	(per_cpu_data_base_ptr[pcpu_id].name)
+	(per_cpu_data_base_ptr[(pcpu_id)].name)
```
dongyaozu: This is not necessary?  [] has similar behavior

hypervisor/include/debug/logmsg.h
```diff
@@ -134,7 +134,7 @@ static inline int vprintf(__unused const char *fmt, __unused va_list args)
 
 #define dev_dbg(lvl, ...)					\
 	do {							\
-		do_logmsg(lvl, pr_prefix __VA_ARGS__);	\
+		do_logmsg((lvl), pr_prefix __VA_ARGS__);	\
```
dongyaozu: This is not necessary too? "," has same result.

hypervisor/include/lib/list.h
```diff
@@ -112,11 +112,11 @@ static inline void list_splice_init(struct list_head *list,
...
 
 #define list_for_each(pos, head) \
-	for (pos = (head)->next; pos != (head); pos = pos->next)
+	for ((pos) = (head)->next; (pos) != (head); (pos) = (pos)->next)
 
 #define list_for_each_safe(pos, n, head) \
-	for (pos = (head)->next, n = pos->next; pos != (head); \
-		pos = n, n = pos->next)
+	for ((pos) = (head)->next, (n) = (pos)->next; (pos) != (head); \
+		(pos) = (n), (n) = (pos)->next)
```
dongyaozu: multiple operations in “for"

hypervisor/arch/x86/vtd.c
dongyaozu: we can change the semantics of MACO and/or code logic for this.

hypervisor/dm/vpic.c
dongyaozu: ok

hypervisor/dm/vpic.c
junjiemao1: A 'foreach' macro for array/list enumeration compacts multiple expressions needed into one for-loop header, inevitably introducing violations with regard to the for-loop well-formedness. 

However, enforcing well-formed for loop in such case means the introduction of duplicated code to every place using these 'foreach' helpers, which can hurt readability. As we have discussed earlier, we are considering deviating the well-formed rule, but only for these 'foreach' cases.

shiqingg:
New version has been updated to the same PR.

v2 -> v3:
 * convert DMAR_WAIT_COMPLETION to inline function
 * remove the macro PIC_PIN_FOREACH and implement the well-formed
   for loop in each case
 * replace __CPP_STRING with STRINGIFY and remove the unused CPP_STRING

jianlaipinan:
Ready to merge

PR #1169, opened 2018-09-06
hv: replace 'return' with 'panic' in bsp_boot_post

mingqiangchi:
If hardware_detect_support() failed,as it return to
bsp_boot_init with a different stack and thus returning
from bsp_boot_init becomes an undefined behavior.

Tracked-On: #861


wenlingz:
Verify+1

PR #1177, opened 2018-09-07
Doc: add interrupt storm mitigation explanation.

mcao6:
Tracked-On: #866

### DIFF
--- a/doc/developer-guides/interrupt-hld.rst
+++ b/doc/developer-guides/interrupt-hld.rst
@@ -451,6 +451,16 @@ a guest's vMSI. The handler then injects the corresponding virtual
 interrupt into the guest, based on vPIC/vIOAPIC/vLAPIC APIs described
 earlier.
 
+Interrupt Storm Mitigation
+==========================
+
+When the Device Model (DM) launches a User OS (UOS), the ACRN hypervisor
+will remap the interrupt for this user OS's pass-through devices. When
+an interrupt occurs for a pass-through device, the CPU core is assigned
+to that User OS gets trapped into the hypervisor. The benefit of such a
+mechanism is that, should an interrupt storm happen in a particular UOS,
+it will have only a minimal effect on the performance of the Service OS.
+
 Interrupt/Exception Injection Process
 =====================================
 


doc/developer-guides/interrupt-hld.rst
```diff
@@ -451,6 +451,13 @@ a guest's vMSI. The handler then injects the corresponding virtual
 interrupt into the guest, based on vPIC/vIOAPIC/vLAPIC APIs described
 earlier.
 
+Interrupt Storm Mitigation: When device model launch a User OS, ACRN will
+remap interrupt for this user OS's pass-through devices. So when an interrupt
+occurs for such device, the CPU core that is assigned to that User OS gets
+trapped into hypervisor. The benefit of such mechanism is that should an
+interrupt storm happen in a particular UOS, it will have little effect on
+the performance of the Service OS.
+
```
dongyaozu: I think the rational is that the processing of software_IRQ  for the UOS will take global lock, which will block the processing of software_IRQ in SOS side.  Because previously we use global lock...  Is it the case?

mcao6: yes, correct. my another patch is to remove the global lock, replaced by a per VM lock. So the explanation for our design is matched.

doc/developer-guides/interrupt-hld.rst
```diff
@@ -451,6 +451,13 @@ a guest's vMSI. The handler then injects the corresponding virtual
 interrupt into the guest, based on vPIC/vIOAPIC/vLAPIC APIs described
 earlier.
 
+Interrupt Storm Mitigation: When device model launch a User OS, ACRN will
```
dbkinder: This seems to be a new section, so how about we change this to be:

    Interrupt Storm Mitigation
    ==========================

    When the Device Model (DM) launches a User OS (UOS), the ACRN hypervisor will
    remap the interrupt for this user OS's pass-through devices. When an interrupt
    occurs for a pass-through device, the CPU core is assigned to that User OS gets
    trapped into the hypervisor. The benefit of such a mechanism is that, should an
    interrupt storm happen in a particular UOS, it will have only a minimal effect on
    the performance of the Service OS.

doc/developer-guides/interrupt-hld.rst
dbkinder: Are you saying we should hold merging this documentation PR until some code change happens?

dongyaozu: The patch will be merged soon. So, no worry, we can keep revising this patch.

dongyaozu:
@dbkinder In this case, how about to name it as "Free of Interferon" ?

mcao6:
@dbkinder OK, will do as your suggestion.

mcao6:
just updated this PR.

dbkinder:
@mcao6 @dongyaozu Is it OK to merge and publish this doc change now?

dongyaozu:
@wenlingz  please merge this doc change. Let us stay with this description for now,

wenlingz:
Verify+1

PR #1184, opened 2018-09-07
HV: improve pass-thru device interrupt process

mcao6:
this patch is used to avoid pass-thru devices interrupt 
interferences between different VMs.

for each pass-thru device and its entry owned by one VM, 
so change the pass-thru device's softirq lock & entry list 
into per VM, so one VM's PT device interrupt process will 
not interfere with other VMs; especially in case one UOS
"interrupt storm" happens, it will have little effect on SOS.

Tracked-On: #866


jianlaipinan:
Ready to merge

jianlaipinan:
Ready to merge

PR #1198, opened 2018-09-08
Tiny fix

JasonChenCJ:


hypervisor/arch/x86/io.c
```diff
@@ -312,9 +312,15 @@ emulate_io(struct vcpu *vcpu, struct io_request *io_req)
...
+			 * size definition is same
+			 */
 			struct pio_request *pio_req = &io_req->reqs.pio;
-			pr_fatal("Err:IO %s access to port 0x%04lx, size=%lu",
-				(pio_req->direction != REQUEST_READ) ? "read" : "write",
+			pr_fatal("Err: %s access to %s 0x%llx, size=%lu",
+				(pio_req->direction == REQUEST_READ) ?
+				"read" : "write",
+				(io_req->type == REQ_PORTIO) ?
+				"io port" : "mmio",
```
dongyaozu: Since this is for developers, I intend to use binary more than "string". Can we simply print  pio_req->direction & type?  We can have a dedicated pattern string at beginning to replace "Err:"

JasonChenCJ: Ok, let me update

PR #1203, opened 2018-09-10
doc: Add fixed issues in v0.2 release note

NanlinXie:
Filter v0.2 fixed issues from github closed issue list.

Signed-off-by: Nanlin Xie <nanlin.xie@intel.com>

### DIFF
--- a/doc/release_notes_0.2.rst
+++ b/doc/release_notes_0.2.rst
@@ -17,6 +17,34 @@ ACRN device model, and documentation.
 Version 0.2 new features
 ************************
 
+Fixed Issues
+============
+:acrn-issue:`663` - Black screen displayed after booting SOS/UOS
+
+:acrn-issue:`676` - Hypervisor and DM version numbers incorrect
+
+:acrn-issue:`1126` - VPCI coding style and bugs fixes for partition mode
+
+:acrn-issue:`1125` - VPCI coding style and bugs fixes found in integration testing for partition mode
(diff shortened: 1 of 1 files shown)


doc/release_notes_0.2.rst
```diff
@@ -17,6 +17,21 @@ ACRN device model, and documentation.
 Version 0.2 new features
 ************************
 
+Fixed Issues
+============
+:acrn-issue:`663` - Black screen displayed after booting SOS/UOS
```
dbkinder: The intention was to also add a description of the issue and if there is a workaround (see the 0.1 release notes for an example).

Each of these issues needs to have a blank line between them (or they'll run together as one paragraph).  (Click on the VIEW button here in GitHub to see what I mean.)

dbkinder: The VIEW button when you're looking at the file changed:  https://github.com/projectacrn/acrn-hypervisor/blob/7fc0fd8bd1f12328bb492ff0f3e891656462f21a/doc/release_notes_0.2.rst

NanlinXie:
@dbkinder, These are for fixed issues which need not workaround, should we still need short description for them?

doc/release_notes_0.2.rst
dbkinder: doh.  You're right, I was thinking these are Known issues not fixed ones.

Issue #1209, opened 2018-09-11
specific PCI device failed to passthrough to UOS


mwang106:
Aaeon’s AI Core, which integrates the Movidius Myriad2 chip and a USB3 controller on a PCIe card failed to passthrough to UOS

mwang106:
[External_System_ID] ACRN-1441

Mingyuan18:
Verified by Community User

PR #1210, opened 2018-09-11
dm: passthru: fix a bug in msix read/write

binbinwu1:
In current code, when read/write msix table, it first handle the case
the offset is in pba range when msix talbe and pba share the same bar.
But the code didn't add the condition whether pba bar equals msix table
bar. It will cause problems for the passthrugh devices,whose pba and msix
table don't share the same bar.

Tracked-On: #1209


jianlaipinan:
Ready to merge

jianlaipinan:
Ready to merge

Issue #1124, opened 2018-08-30
MMU code reshuffle


mwang106:
Although the MMU feature is done, but it is worthy to revisit the code to check whether our implementation is good enough by referring to x86 SDM and other hypervisor implemention.

mwang106:
[External_System_ID] 247220

PR #1227, opened 2018-09-13
hypercall: no need check HV memory overlap for MR_DEL

JasonChenCJ:
in local_set_vm_memory_region, hpa is not used for MR_DEL, and it does
not need check HV memory overlap for a MR_DEL operation.

this is a bug exposed by remapping SOS first page, as the remap page whose
gpa = 0 actually fall into HV memory area; while during MR_DEL, the input
param vm0_gpa is set to 0 by default.

Tracked-On: #1124


lifeix:
LGTM

Issue #1235, opened 2018-09-13
explicitly remove the possibility of out-of-range access to irq_alloc_bitmap[]


lyan3:

mwang106:
[External_System_ID] 254613

Mingyuan18:
Index to the bitmap might be out-of-range.
commit ccf562402ef3ddcfe5e58858a72e8d7f51f20df7

PR #1236, opened 2018-09-13
hv:irq: avoid out-of-range access to irq_alloc_bitmap[]

lyan3:
Logically, out-of-range access won't happen at these places. However, it
depends on the behaviour of other codes.
This commit makes changes to explicitly eliminate the possibility in these
functions.

Tracked-On: #1235

### DIFF
--- a/hypervisor/arch/x86/irq.c
+++ b/hypervisor/arch/x86/irq.c
@@ -41,12 +41,13 @@ uint32_t alloc_irq_num(uint32_t req_irq)
 	if (irq == IRQ_INVALID) {
 		/* if no valid irq num given, find a free one */
 		irq = ffz64_ex(irq_alloc_bitmap, NR_IRQS);
-		irq = (irq == NR_IRQS) ? IRQ_INVALID : irq;
 	}
 
-	if (irq != IRQ_INVALID) {
+	if (irq >= NR_IRQS) {
+		irq = IRQ_INVALID;
+	} else {
 		bitmap_set_nolock((uint16_t)(irq & 0x3FU),
-			      irq_alloc_bitmap + (irq >> 6U));
+				irq_alloc_bitmap + (irq >> 6U));
 	}
 	spinlock_irqrestore_release(&irq_alloc_spinlock, rflags);
 	return irq;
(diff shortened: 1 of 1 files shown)


hypervisor/arch/x86/irq.c
```diff
@@ -322,7 +322,7 @@ void dispatch_interrupt(struct intr_excp_ctx *ctx)
 	uint32_t irq = vector_to_irq[vr];
 	struct irq_desc *desc;
 
-	if (irq == IRQ_INVALID) {
+	if (irq >= NR_IRQS) {
 		goto ERR;
```
JasonChenCJ: do we really need such change? As vector_to_irq[irq] will set to IRQ_INVALID for unused one.

hypervisor/arch/x86/irq.c
```diff
@@ -41,7 +41,7 @@ uint32_t alloc_irq_num(uint32_t req_irq)
 	if (irq == IRQ_INVALID) {
 		/* if no valid irq num given, find a free one */
 		irq = ffz64_ex(irq_alloc_bitmap, NR_IRQS);
-		irq = (irq == NR_IRQS) ? IRQ_INVALID : irq;
+		irq = (irq >= NR_IRQS) ? IRQ_INVALID : irq;
```
dongyaozu: Should we change the code to:
if (...)
    irq = IRQ_INVALID.

This way, in normal case, we avoid a redundant operation of irq = irq.

dongyaozu: can you check the entire HV to see if this kind of issues exist?

hypervisor/arch/x86/irq.c
dongyaozu: Here, we assume IRQ_INVALID is >= NR_IRQS. This may be not the case.

Here the irq comes from vector_to_irq tables, it should contain 0--NR_IRQS-1 or INVALID.  Can you add a pre-assumption here?

dongyaozu: Agree with Jason, this is not necessary.

haoweiah:
Ready to merge

PR #1244, opened 2018-09-14
Enable support for USB hub and 'cold plug' mode

xiaoguangwu:
This patch series is used to support 'flat mode' hub emulation and 'cold
plug' feature.

1. Flat mode hub emulation means DM only emulates USB devices under hub,
   the Guest OS cannot see emulated hub. In the perspective of Guest OS,
   all the emulated devices are under root hub.

2. Cold plug mode refers to USB devices are plugged in before Guest OS
   is booted. In this case, the libusb will not report 'connect' event
   to DM, and hence the Guest OS will have no opportunity to do enumeration
   for those devices. The second patch is used to fix this issue.


haoweiah:
Ready to merge

Issue #1252, opened 2018-09-17
Buffer Overflow Vulnerabilities in ACRN hypervisor & Device Model


yonghuah:
During code analysis, a buffer overflow condition was detected that could reach out of bounds for the index of the array object in
the function. Many such instances were found to be present in the source code for the ACRN kernel and the hypervisor.

mwang106:
[External_System_ID] 255251

mwang106:
No regression issue

PR #1253, opened 2018-09-17
hv: vtd: fix potential buffer overflow in suspend/resume

binbinwu1:
In current code of suspend_iommu/resume_iommu, there is potential buffer overflow
according to the code.
This patch put the buffer to struct dmar_drhd_rt, so that no need to access the buffer
via index.

Tracked-On: #1252

### DIFF
--- a/hypervisor/arch/x86/vtd.c
+++ b/hypervisor/arch/x86/vtd.c
@@ -30,6 +30,10 @@
 #define ROOT_ENTRY_LOWER_CTP_POS            (12U)
 #define ROOT_ENTRY_LOWER_CTP_MASK           (0xFFFFFFFFFFFFFUL)
 
+/* 4 iommu fault register state */
+#define	IOMMU_FAULT_REGISTER_STATE_NUM	4U
+#define	IOMMU_FAULT_REGISTER_SIZE	4U
+
 #define CTX_ENTRY_UPPER_AW_POS          (0U)
 #define CTX_ENTRY_UPPER_AW_MASK         \
 		(0x7UL << CTX_ENTRY_UPPER_AW_POS)
@@ -118,6 +122,7 @@ struct dmar_drhd_rt {
 	uint16_t cap_num_fault_regs;
 	uint16_t cap_fault_reg_offset;
 	uint16_t ecap_iotlb_offset;
+	uint32_t fault_state[IOMMU_FAULT_REGISTER_STATE_NUM]; /* 32bit registers */
 };
 
 struct dmar_root_entry {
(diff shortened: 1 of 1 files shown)


hypervisor/arch/x86/vtd.c
```diff
@@ -1230,21 +1230,12 @@ void suspend_iommu(void)
...
 		dmar_disable_translation(dmar_unit);
-
-		/* If the number of real iommu devices is larger than we
-		 * defined in kconfig.
-		 */
-		if (iommu_idx > CONFIG_MAX_IOMMU_NUM) {
-			pr_err("iommu dev number is larger than pre-defined");
-			break;
-		}
-		iommu_idx++;
```
dongyaozu: should we remove definition of iommu_idx?

hypervisor/arch/x86/vtd.c
```diff
@@ -30,6 +30,10 @@
 #define ROOT_ENTRY_LOWER_CTP_POS            (12U)
 #define ROOT_ENTRY_LOWER_CTP_MASK           (0xFFFFFFFFFFFFFUL)
 
+/* 4 iommu fault register state */
+#define	IOMMU_FAULT_REGISTER_STATE_NUM	4U
+#define	IOMMU_FAULT_REGISTER_STATE_SIZE	4U
```
dongyaozu: This MACRO is actually  sizeof(failt_state), can we do in this way?
It helps to couple them together.

hypervisor/arch/x86/vtd.c
binbinwu1: will fix it.

hypervisor/arch/x86/vtd.c
binbinwu1: It is the register width/register address offset of these fault registers. And it is fixed according to vtd spec. sizeof(fault_state) kind of related the hardware register offset to the data structure we defined, which I not prefer to. If we want to get rid of a new macro, I would prefer to use sizeof (uint32_t).

hypervisor/arch/x86/vtd.c
```diff
@@ -1230,29 +1229,20 @@ void suspend_iommu(void)
 
 		/* save IOMMU fault register state */
 		for (i = 0U; i < IOMMU_FAULT_REGISTER_STATE_NUM; i++) {
-			iommu_fault_state[iommu_idx][i] =
-				iommu_read32(dmar_unit, DMAR_FECTL_REG +
-					(i * IOMMU_FAULT_REGISTER_STATE_NUM));
+			dmar_unit->fault_state[i] =  iommu_read32(dmar_unit,
+				DMAR_FECTL_REG + (i * sizeof(uint32_t)));
```
dongyaozu: sizeof(uint32_t) is not good --- if the definition side changes, it still lead to bug. should be sizeof(fault_state).
for example  sizeof(  ((struct dmar_drhd_rt*) NULL) ->fault_state[0]).

binbinwu1: This value is not related to definition side of fault_state, it should be aligned with hardware spec as below, so it should be a constant value:
#define DMAR_FSTS_REG   0x34U    /* Fault Status register */
#define DMAR_FECTL_REG  0x38U    /* Fault control register */
#define DMAR_FEDATA_REG 0x3cU    /* Fault event interrupt data register */
#define DMAR_FEADDR_REG 0x40U    /* Fault event interrupt addr register */

dongyaozu: Spec is the source of the requirement. data structure definition is following the spec.
The MACRO could derive from the data structure or from the original spec. Either way is better than sizeof(uint32_t) here, which follows neither of the source.

binbinwu1: fixed

haoweiah:
Ready to merge

PR #1255, opened 2018-09-17
vioapic: add irr to indicate assert/deassert

JasonChenCJ:
a guest may use per-cpu vector allocation mechanism which may lead to
same vector on different vcpu for different vioapic pins.
when we broadcast EOI from vlapic to vioapic, it could trigger
ptdev_intx_ack for incorrect vioapic pin which will lead to incorrect
vioapic_irq_deassert.

old implementation is recording assert & deassert by acnt, like:
- assert -> acnt++
- deassert -> acnt--
which means an incorrect deassert may descrease acnt from 0 to -1, it leads
to error.

this patch change the recording method for a pin assert/deassert, it
just use irr bit set/clear to indicate assert/deassert:
- assert -> irr set
- deassert -> irr clear

it can avoid the error of incorrect deassert, and for ptdev sharing irq,
the dev hardware could re-trigger interrupt after deassert.
but for emulated device, there is no mechanism to re-trigger interrupt if
there is irq sharing - for DM, we already has constraint to restrict vGSI
sharing, so there is no problem of it.


hypervisor/dm/vioapic.c
```diff
@@ -89,87 +89,60 @@ vioapic_send_intr(struct acrn_vioapic *vioapic, uint32_t pin)
...
+				vioapic->irr + (pin >> 6U));
+		return;
 	}
-	newcnt = vioapic->acnt[pin];
 
-	if (newcnt < 0) {
-		pr_err("ioapic pin%hhu: bad acnt %d", pin, newcnt);
-	}
+	/* assert and deliver intrrupt */
+	oldstate = bitmap_test(pin & 0x3FU, vioapic->irr + (pin >> 6U));
```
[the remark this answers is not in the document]
JasonChenCJ: ok

hypervisor/dm/vioapic.c
```diff
@@ -89,87 +89,60 @@ vioapic_send_intr(struct acrn_vioapic *vioapic, uint32_t pin)
...
-	bool needintr;
+	bool oldstate;
 
-	oldcnt = vioapic->acnt[pin];
-	if (newstate) {
-		vioapic->acnt[pin]++;
-	} else {
-		vioapic->acnt[pin]--;
+	/* deassert and return */
+	if (newstate == false) {
```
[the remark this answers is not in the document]
JasonChenCJ: IOAPIC support polarity, but my question is that do we real need support this for virtual IOAPIC? We just need make sure when assert, interrupt trigger to vlapic.

hypervisor/dm/vioapic.c
```diff
@@ -89,87 +89,60 @@ vioapic_send_intr(struct acrn_vioapic *vioapic, uint32_t pin)
...
+		return;
 	}
-	newcnt = vioapic->acnt[pin];
 
-	if (newcnt < 0) {
-		pr_err("ioapic pin%hhu: bad acnt %d", pin, newcnt);
-	}
+	/* assert and deliver intrrupt */
+	oldstate = bitmap_test(pin & 0x3FU, vioapic->irr + (pin >> 6U));
+	bitmap_set_nolock(pin & 0x3FU, vioapic->irr + (pin >> 6U));
```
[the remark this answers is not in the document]
JasonChenCJ: no.
newstate = 0 means deassert -> bitmap_clr
newstate = 1 means assert -> bitmap_set

hypervisor/dm/vioapic.c
```diff
@@ -85,91 +85,69 @@ vioapic_send_intr(struct acrn_vioapic *vioapic, uint32_t pin)
 
 /**
  * @pre pin < vioapic_pincount(vm)
+ *
+ * here the assumption is that IRQ only fires at raising edge for edge
+ * triggered interrupt, and high level for level triggered interrupt.
+ * actually, the virtual interrupt asserting for vioapic does not consider
+ * the polarity of the RTE setting.
  */
```
dongyaozu: Then please check the guest write of RTE to assert or fail on guest change of polarity.

hypervisor/include/arch/x86/guest/vioapic.h
```diff
@@ -37,15 +37,16 @@
...
 struct acrn_vioapic {
 	struct vm	*vm;
 	spinlock_t	mtx;
 	uint32_t	id;
 	uint32_t	ioregsel;
 	union ioapic_rte rtbl[REDIR_ENTRIES_HW];
-	/* sum of pin asserts (+1) and deasserts (-1) */
-	int32_t acnt[REDIR_ENTRIES_HW];
+	/* assert status bitmap: 1 - assert, 0 - deassert */
+	uint64_t assert[ASSERT_BITMAP_SIZE];
```
dongyaozu: level is better than assert.  pin_level/pin_state is fine too.

hypervisor/dm/vioapic.c
JasonChenCJ: will update

hypervisor/include/arch/x86/guest/vioapic.h
JasonChenCJ: will update

hypervisor/dm/vioapic.c
JasonChenCJ: @dongyaozu , there is one problem here, for ptdev, we need update polarity configuration from vioapic to native ioapic, so we cannot assert guest writing RTE for polarity setting.  
Per my view, we can emulate virtual pin state in vioapic fully, but it may bring other complex, for example:
if the assumption is like
- vioapic_assert_irq -> pin_state to high level
- vioapic_deassert_irq -> pin_state to low level
then how should we trigger a low level active interrupt? We should keep the pin_state as high at the beginning, and the emulated device should call vioapic_deassert_irq to trigger interrupt, and then call vioapic_assert_irq to stop triggering.

dongyaozu: I am suggesting to support one polarity only -- this is what your patch assumes. In this case, if the guest change polarity, we should assert. I don't see any issue.

If we fully emulate the polarity, and propagate the vioapic polarity to physical IOAPIC. At the time of PIRQ, we propagate the physical IOAPIC level to vioapic level.  That works fine too.

JasonChenCJ: For PTDEV INTx interrupt, we must keep polarity configuration from vioapic as it must be set to native ioapic accordingly, otherwise, this device may fail to work. So we cannot support one polarity only.

And yes, we can fully emulate the polarity, but it bring complicated change for both ptdev and emulated device irq assert/deassert. My question is that, why do we must take care polarity, when there is interrupt come, then we just assert it and trigger corresponding virtual interrupt, it should be enough.

hypervisor/dm/vioapic.c
```diff
@@ -87,78 +82,89 @@ vioapic_send_intr(struct acrn_vioapic *vioapic, uint32_t pin)
...
-	} else if ((oldcnt == 1) && (newcnt == 0)) {
-		dev_dbg(ACRN_DBG_IOAPIC, "ioapic pin%hhu: deasserted", pin);
+	uint32_t old_lvl;
+	union ioapic_rte rte = vioapic->rtbl[pin];
+
+	old_lvl = bitmap_test(pin & 0x3FU, &vioapic->pin_state[pin >> 6U]);
+	/* clear pin_state and deliver interrupt according to polarity */
+	if (level == 0U) {
+		bitmap_test_and_clear_nolock(pin & 0x3FU,
+				&vioapic->pin_state[pin >> 6U]);
```
dongyaozu: bitmap_clr_nolock

hypervisor/dm/vioapic.c
```diff
@@ -87,78 +82,89 @@ vioapic_send_intr(struct acrn_vioapic *vioapic, uint32_t pin)
...
+		if (((rte.full & IOAPIC_RTE_INTPOL) != 0UL) && old_lvl == 1U)
+			vioapic_send_intr(vioapic, pin);
```
dongyaozu: old_lvl != level

hypervisor/dm/vioapic.c
```diff
@@ -87,78 +82,89 @@ vioapic_send_intr(struct acrn_vioapic *vioapic, uint32_t pin)
...
 	} else {
-		dev_dbg(ACRN_DBG_IOAPIC, "ioapic pin%hhu: %s, ignored, acnt %d",
-		    pin, pinstate_str(newstate), newcnt);
-	}
-
-	if (needintr) {
-		vioapic_send_intr(vioapic, pin);
+	/* set pin_state and deliver intrrupt according to polarity */
+		bitmap_set_nolock(pin & 0x3FU, &vioapic->pin_state[pin >> 6U]);
+		if (((rte.full & IOAPIC_RTE_INTPOL) == 0UL) && old_lvl == 0U)
```
dongyaozu: ditto

hypervisor/debug/vuart.c
```diff
@@ -133,13 +133,9 @@ static void vuart_toggle_intr(struct acrn_vuart *vu)
...
-		vpic_set_irq(vu->vm, COM1_IRQ, GSI_SET_HIGH);
-
-		vioapic_set_irq(vu->vm, COM1_IRQ, GSI_SET_HIGH);
-
-		vpic_set_irq(vu->vm, COM1_IRQ, GSI_SET_LOW);
-
-		vioapic_set_irq(vu->vm, COM1_IRQ, GSI_SET_LOW);
+		/* active low for COM1_IRQ polarity */
+		vpic_set_irq(vu->vm, COM1_IRQ, GSI_FALLING_PULSE);
+		vioapic_set_irq(vu->vm, COM1_IRQ, GSI_FALLING_PULSE);
```
dongyaozu: http://www.ti.com/lit/ds/symlink/pc16550d.pdf   for the specification.
The INTR is active high.  

If we still have issue in this patch, please remove this one for now.  We can revisit later on.

dongyaozu: static void vuart_toggle_intr(struct acrn_vuart *vu)
{
        uint8_t intr_reason;

        intr_reason = vuart_intr_reason(vu);

        if (intr_reason != IIR_NOPEND) {
                vpic_set_irq(vu->vm, COM1_IRQ, GSI_SET_HIGH);

                vioapic_set_irq(vu->vm, COM1_IRQ, GSI_SET_HIGH);

                vpic_set_irq(vu->vm, COM1_IRQ, GSI_SET_LOW);

                vioapic_set_irq(vu->vm, COM1_IRQ, GSI_SET_LOW);
        }
	=
		If (  //has IRQ ) {
			SET_HIGH
		}
		Else
			SET_LOW.
}

hypervisor/dm/vioapic.c
JasonChenCJ: sure

hypervisor/dm/vioapic.c
JasonChenCJ: sure

hypervisor/dm/vioapic.c
JasonChenCJ: sure

hypervisor/arch/x86/assign.c
```diff
@@ -359,11 +372,19 @@ static void ptdev_intr_handle_irq(struct vm *vm,
 		}
 
 		if (trigger_lvl) {
-			vioapic_set_irq(vm, virt_sid->intx_id.pin,
-					GSI_SET_HIGH);
+			if (entry->active_polarity)
```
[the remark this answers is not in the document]
JasonChenCJ: this already be done based on polarity

hypervisor/include/common/ptdev.h
```diff
@@ -58,6 +58,7 @@ struct ptdev_remapping_info {
 	struct vm *vm;
 	uint32_t active;	/* 1=active, 0=inactive and to free*/
 	uint32_t allocated_pirq;
+	uint32_t active_polarity; /* 0=active high, 1=active low*/
```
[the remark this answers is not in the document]
JasonChenCJ: sure

hypervisor/debug/vuart.c
JasonChenCJ: SOS guest actually set vIOAPIC to low active, if there is no such change, the series console can not work..

PR #1271, opened 2018-09-18
tools: acrnd: bugfix: service lack of prerequisition

taoyuhong:
As a system service, acrnd will launch UOS, that must be done
after all required services and conditions are ready, such as
acrnprobe, weston, etc.

Tracked-On: #1278 

### DIFF
--- a/tools/acrn-manager/acrnd.service
+++ b/tools/acrn-manager/acrnd.service
@@ -1,9 +1,14 @@
 [Unit]
 Description=ACRN manager deamon
+After=weston.service
+After=systemd-resolved.service
+ConditionPathExists=/sys/kernel/gvt
+ConditionPathExists=/dev/acrn_vhm
 
 [Service]
 Type=simple
 ExecStart=/usr/bin/acrnd
+ExecStop=/usr/bin/killall -s TERM acrnd
 
 [Install]
 WantedBy=multi-user.target


dongyaozu:
@lyan3 can you review?

tools/acrn-manager/acrnd.service
```diff
@@ -1,9 +1,15 @@
 [Unit]
 Description=ACRN manager deamon
+After=acrnprobe.service
```
lyan3: Why this service? It looks it is not available.

tools/acrn-manager/acrnd.service
```diff
@@ -1,9 +1,15 @@
...
+After=weston.service
+After=systemd-resolved.service
```
lyan3: why is there dependency on "Network Name Resolution" service?

tools/acrn-manager/acrnd.service
taoyuhong: OK， remove this line

tools/acrn-manager/acrnd.service
taoyuhong: I think it's better to keep this line, because UOS launch script will set up network on SOS...

taoyuhong:
removed acrnprobe
UOS launch script will set up network, which may relays on systemd-resolved

haoweiah:
Ready to merge

PR #1273, opened 2018-09-18
Malloc

guoch2:
The patch series try to remove dynamic memory allocateion in EFI boot code of EFI stub module. Compared to V1, the changes are:

  a)  Allocate boot related struct right after hypervisor memory, merge then into one call to UEFI memory allocation; After this, the EFI stub code (which boot hypervisor) will only have 3 calls to dynamic memory:
      1. the call for hv binary and the 3 struct;
      2. the call to CPU boot trampoline code;
      3. the call to alloc mmap buf when inquery memory layout from UEFI FW

  b) add a print/warning in case that UEFI fw allocate zero-page to us;

> Q from Jason: 
>>To fix L1TF issue, we must make sure trampoline buffer reserved by
>>  UEFI stub do not use first page (0~4K), could you help confirm if
>>  allocate_pages could do such constraint?
memory allocation is under the control of UEFI firmware, we assume that UEFI FW knows aobut L1TF and won't alloc zero-page for boot loader/code.
Just add a warning when zero-page is allocated to CPU trampoline code,and warn users of this. If the case does occur in the future, we can add some workaround to solve this issue, like: first request zero page, hold it and then do other memory allocation, at last, right before jump into hypervisor, release the zero-page. 

> Q from Anthony:
>>  How do you know allocate_pool allocate memory in range 0f 0x80000000~0xffffffff?
    Oops.  misunderstood the argment. Now, in V2/V3 patches, we don't use allocation_pool() for EFI boot related struct, instead, only one call to the func, but the memory is temporary, and will b released in EFI stub code before jumping into hypervisor.

>> RELOC is enabled by default, can we assume it? then we don't need 
>> emalloc_fixed_addr

A:  RELOC config is still in git repo. users can disable it. So, we leave it then, and remove it later if CONFIG_RELOC is removed from config file in the future

> Q from Eddie:
>> For FuSa purpose, we won't remove the dynamic memory allocation... 
>> So, UEFI bootloader part needs to do something similar...
>> We know this may be difficult.. But want to evolve toward this direction.
Merged the 3 calls to dynamic memory for boot related structs into one and do it together with HV binary memory when EFI stub is trying to relocate HV binary. Since dynamic memory allocation is needed for HV binary relocation on UEFI platform. 

Chaohong guo (6):
  Remove emalloc_for_low_mem() routine in EFI boot code of HV
  Replace __emalloc() with a call to uefi allocate_page()
  allocate boot related struct right after hypervisor memory
  Replace the call to emalloc() to uefi pool allocation
  Removed dead funcs in EFI stub module
  Add a comment for switch fallthrough to quiet lint warning

 hypervisor/bsp/uefi/efi/boot.c      |  71 ++++----
 hypervisor/bsp/uefi/efi/boot.h      |  18 ++
 hypervisor/bsp/uefi/efi/efilinux.h  |  35 ++++
 hypervisor/bsp/uefi/efi/malloc.c    | 262 ----------------------------
 hypervisor/bsp/uefi/efi/multiboot.h |   2 +-
 hypervisor/bsp/uefi/efi/stdlib.h    |  41 -----
 hypervisor/debug/npk_log.c          |   1 +
 hypervisor/debug/shell.c            |   2 +
 8 files changed, 91 insertions(+), 341 deletions(-)

--
2.17.1


guoch2:
@lijinxia any error ?  description about the failure ?

lijinxia:
manually check PASS

PR #1294, opened 2018-09-20
Makefile: fix cross-compiling issues

liuming50:
To be able to support cross compilation, some fixes are needed as
follows:
- Replace all the host paths with a prefix sysroot, it defaults to be
  empty.
- Use weak assignment to compiler variables, like CC, this allows them
  to be overridden by environment variables passed to make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ TOOLS_OUT := $(ROOT_OUT)/tools
 DOC_OUT := $(ROOT_OUT)/doc
 BUILD_VERSION ?=
 BUILD_TAG ?=
+DESTDIR ?=
+SYSROOTDIR ?=
 export TOOLS_OUT
 
 .PHONY: all hypervisor devicemodel tools doc
--- a/hypervisor/bsp/uefi/efi/Makefile
+++ b/hypervisor/bsp/uefi/efi/Makefile
@@ -53,12 +53,12 @@ endif
 # its tools and libraries in different folders. The next couple of
 # variables will determine and set the right path for both the
 # tools $(GNUEFI_DIR) and libraries $(LIBDIR)
-GNUEFI_DIR := $(shell find /usr/lib* -name elf_$(ARCH)_efi.lds -type f | xargs dirname)
+GNUEFI_DIR := $(shell find $(SYSROOTDIR)/usr/lib* -name elf_$(ARCH)_efi.lds -type f | xargs dirname)
 LIBDIR := $(subst gnuefi,,$(GNUEFI_DIR))
 CRT0 := $(GNUEFI_DIR)/crt0-efi-$(ARCH).o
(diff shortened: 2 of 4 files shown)


NanlinXie:

Makefile
```diff
@@ -12,6 +12,8 @@ TOOLS_OUT := $(ROOT_OUT)/tools
 DOC_OUT := $(ROOT_OUT)/doc
 BUILD_VERSION ?=
 BUILD_TAG ?=
+DESTDIR ?=
```
binbinwu1: This var not used?

liuming50: It's being referred by install targets in Makefiles, such as:
...
./devicemodel/Makefile:	install -D --mode=0755 $(DM_OBJDIR)/$(PROGRAM) $(DESTDIR)/usr/bin/$(PROGRAM)
./devicemodel/Makefile:	install -D -t $(DESTDIR)/usr/share/acrn/samples/nuc $^
./devicemodel/Makefile:	install -D -t $(DESTDIR)/usr/share/acrn/samples/apl-mrb $^
./devicemodel/Makefile:	install -d $(DESTDIR)/usr/lib/systemd/system/
./devicemodel/Makefile:	install -p -D -m 0644 ./samples/apl-mrb/acrn_guest.service $(DESTDIR)/usr/lib/systemd/system
./devicemodel/Makefile:	install -d $(DESTDIR)/usr/share/acrn/bios
./devicemodel/Makefile:	install -D --mode=0664 -t $(DESTDIR)/usr/share/acrn/bios $^
...

here I just want give it a default empty value in the top Makefile, it's not a bug fix, but I think it would be easier to be read.

binbinwu1:

liuming50:
Will send a V2 with "Tracked-On" tag.

Makefile
gvancuts: `DESTDIR` should not be directly set in the `Makefile`, it is a built-in variable that is only used (and only useful) to the `(un)install` targets of a `Makefile` - see https://www.gnu.org/prep/standards/html_node/DESTDIR.html#DESTDIR

You're not setting it to any value but it's awkward to have it in the Makefile, it is expected that readers of Makefiles know this is a built-in variable. Can you remove it in your v2?

gvancuts:
> Will send a V2 with "Tracked-On" tag.

You should force-push an updated version of your patch, else it becomes really hard to follow the history of what happened to a particular conversation. I don't know for example if you have already sent a v2 patch. Can you make sure that you update (and force-push) your patches to as to re-use the same PR until the code gets merged? Thanks!

(if you need help/instructions on how to do that, we have some here: https://projectacrn.github.io/latest/developer-guides/contribute_guidelines.html#contribution-workflow)

Makefile
liuming50: You are absolutely right, will remove it.

liuming50:
OK, will try to do that. thanks for the explanation. But I had sent out a V2, will send a V3 on top of that one.

Issue #1295, opened 2018-09-20
ACRN ramdump have regression after switched to ABL


mwang106:
After switched to ABL, we are using SPI crashmode.
crashmode can work successfully, and dump is also ok

SOS also can be decoded successfully.
However, ACRN have issues, below is the error:

file '/home/btian1/work/data/ramdump/hv/ram.sparse.bin' (SPARSE) loaded.
loading ELF/DWARF
file '/home/btian1/work/data/ramdump/hv/acrn.out' (ELF/DWARF2) loaded.
cpu_regs variable not found..

Attachments

Mingyuan18:
commit id: 2908f09f88e47f297fa4bbfc6e9cce946671c2e5
Can't reproduce this issue

PR #1296, opened 2018-09-20
hv: fix ramdump regression

zhouji3x:
this patch change type of global variable for saving registers on execption.

global variable 'crash_ctx' should not be set to static.
crash_ctx is for offline analysis when system crashed, not for runtime usage.
as crash_ctx is only be set without being read, compiler will regard
crash_ctx as an useless variable if it is set to static, and will not
generate code for it.

Tracked-On: #1295
Signed-off-by: zhouji3x <jianfengx.zhou@intel.com>


haoweiah:
Ready to merge

PR #1298, opened 2018-09-20
hv:Fix unused var value on all paths

junjunshan1:
MISRAC checks whether a variable is assigned a value not used in
all branches of a program. Var value which is unused on all paths
can be removed with a consequent improvement in the readability
and efficiency of the code. This patch is used to fix these
violations.

Tracked-On: #861
Signed-off-by: Junjun Shan <junjun.shan@intel.com>


haoweiah:
Ready to merge

PR #1302, opened 2018-09-20
hv: fix 'Static procedure is not explicitly called in code analysed'

shiqingg:
MISRA-C states that redundant code reduce the maintainability of code.

In some cases, we would like to keep the current unused static functions
for code completeness, such as checking register info. These functions
might be used later.

This patch removes the unused static function 'mmu_pt_for_pde'.
Looks like it is legacy code and not being used in our project.

Tracked-On: #861


haoweiah:
Ready to merge

PR #1311, opened 2018-09-21
doc: Add change log in v0.2 release note.

NanlinXie:
Add change log in v0.2 release note.

Signed-off-by: Xie, nanlin <nanlin.xie@intel.com>


ailun258:
Looks good to me,  just one comments- 
can we add specific v0.2 release tag here? so the people can clone code with exact URL directly,

 for example:
 ACRN reference code of v0.2 is here:
 https://github.com/projectacrn/acrn-hypervisor/tree/v0.2  It includes the ACRN hypervisor, the
 ACRN device model, and documentation.

Issue #1307, opened 2018-09-20
cross-compiling issues 


mwang106:
To be able to support cross compilation, some fixes are needed as
follows:

- Replace all the host paths with a prefix sysroot, it defaults to be empty.
- Use weak assignment to compiler variables, like CC, this allows them to be overridden by environment variables passed to make.

mwang106:
[External_System_ID] ACRN-1472

Issue #1122, opened 2018-08-30
Security: Enable compiler and linker setting-flags to harden software


wws666:
Below settings/flags should be enabled for ACRN: 
  1) Stack execution protection;
  2) Data relocation and protection (RELRO);        
  3) Stack-based Buffer Overrun Detection;          
  4) Position Independent Execution (PIE);           
  5) Fortify source;                                
  6) Format string vulnerabilities;

hwang37:
[External_System_ID] 253110

PR #1312, opened 2018-09-21
Makefile: fix cross-compiling issues

liuming50:
To be able to support cross compilation, some fixes are needed as
follows:
- Replace all the host paths with a prefix sysroot, it defaults to be
  empty.
- Use weak assignment to compiler variables, like CC, this allows them
  to be overridden by environment variables passed to make.

Tracked-On: #1307

### DIFF
--- a/hypervisor/bsp/uefi/efi/Makefile
+++ b/hypervisor/bsp/uefi/efi/Makefile
@@ -53,12 +53,12 @@ endif
 # its tools and libraries in different folders. The next couple of
 # variables will determine and set the right path for both the
 # tools $(GNUEFI_DIR) and libraries $(LIBDIR)
-GNUEFI_DIR := $(shell find /usr/lib* -name elf_$(ARCH)_efi.lds -type f | xargs dirname)
+GNUEFI_DIR := $(shell find $(SYSROOT)/usr/lib* -name elf_$(ARCH)_efi.lds -type f | xargs dirname)
 LIBDIR := $(subst gnuefi,,$(GNUEFI_DIR))
 CRT0 := $(GNUEFI_DIR)/crt0-efi-$(ARCH).o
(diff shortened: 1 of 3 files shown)


NanlinXie:

Makefile
```diff
@@ -12,6 +12,8 @@ TOOLS_OUT := $(ROOT_OUT)/tools
 DOC_OUT := $(ROOT_OUT)/doc
 BUILD_VERSION ?=
 BUILD_TAG ?=
+DESTDIR ?=
+SYSROOTDIR ?=
```
gvancuts: How do you use this `SYSROOTDIR` variable? Is this something that you expect people to set via the command-line?

I'm asking but I find it odd to have a weak assignment in this Makefile but the variable is not being used at all (neither passed on when calling other Makefiles, nor exported as a variable) so this seems useless to have it here.

liuming50: I supposed to use it by passing it to make command or exporting it in shell in which make is being executed.

I think you are right, it does not have to be in the top Makefile, it's redundant and useless, will drop it in V3, and I am also going to rename it to SYSROOT, will send a V3 soon.

liuming50:
I have updated this V3 according to gvancuts's comments, please help me to review it.

lijinxia:
OK to verify

liuming50:
Hi,

I have to update the patch again, since I noticed it's conflicting with a recent merged commit: 
...
commit 52ee615428963c4bdc20dbac7926e43f0fcdded4
Author: Yan, Like <like.yan@intel.com>
Date:   Thu Sep 20 09:37:11 2018 +0800

    tools: acrnlog: update Makefile
    
    In order to sync the compiler options with devicemode and enable options
    to harden software.
    
    Tracked-On: #1122
...

haoweiah:
Ready to merge

Issue #1313, opened 2018-09-21
[APL][IO Mediator] Remove unused netmap/vale in virtio-net


mwang106:
Remove unused netmap/vale

PR #1314, opened 2018-09-21
dm: virtio_net: remove netmap/vale backend support

ssqre:
Netmap/vale attempts to speed up network communication by bypassing the
TCP/IP network stack, which requires patching the physical NIC driver so
that applications developed based on netmap can interact directly with
the physical NIC driver. It may make sense for some specific scenarios
which requres very high bandwith (10Gb/s or 100Gb/s), we can even put up
with the complexity and compatibility introduced by this techology.
However for ACRN, a virtualization solution for IoT, there is no need to
support this backend. For 1Gb NICs or below, the VBS-U/tap solution
can already achieve near-native bandwidth. To keep simplicity and
improve compatibility, remove the netmap/vale support in dm.

Tracked-On: #1313

### DIFF
--- a/devicemodel/hw/pci/virtio/virtio_net.c
+++ b/devicemodel/hw/pci/virtio/virtio_net.c
@@ -28,9 +28,6 @@
 
 #include <sys/uio.h>
 #include <net/ethernet.h>
-#ifndef NETMAP_WITH_LIBS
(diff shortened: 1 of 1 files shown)


haoweiah:
Ready to merge

PR #1317, opened 2018-09-21
hv: fix 'Switch case not terminated with break'

shiqingg:
MISRA-C requires that every switch case shall be terminated with break
to avoid the unintentional fall through.

The code will become redundant if we enforce this rule.
So, we will keep the current implementation for the following two cases.
1. The fall through is intentional.
2. The function is returned in the switch case.
    If we decide to eliminate the mutiple returns in one function later,
    this case would be handled properly at that time.

What this patch does:
- add the mssing break for the default case
- add the pre condition for some functions and remove the corresponding
  panic which will never happen since the function caller could guarantee
  the pre condition based on the code implementation

v1 -> v2:
 * remove the redundant cases above default in 'vlapic_get_lvtptr'
 * add the similar pre condition for 'lvt_off_to_idx' as
   'vlapic_get_lvtptr' since all the function callers could guarantee it
 * remove the assertion in 'lvt_off_to_idx' since the pre condition
   could guarantee that the assertion will never happen
 * add the similar pre condition for 'vpic_set_irqstate' as
   'vioapic_set_irqstate' since all the function callers could guarantee it
 * remove the assertion in 'vpic_set_irqstate' since the pre condition
   could guarantee that the assertion will never happen

Tracked-On: #861

### DIFF
--- a/hypervisor/arch/x86/guest/vlapic.c
+++ b/hypervisor/arch/x86/guest/vlapic.c
@@ -518,10 +518,20 @@ vlapic_set_intr_ready(struct acrn_vlapic *vlapic, uint32_t vector, bool level)
 	return 1;
 }
 
+/**
+ * @pre offset value shall be one of the folllowing values:
+ *	APIC_OFFSET_CMCI_LVT
+ *	APIC_OFFSET_TIMER_LVT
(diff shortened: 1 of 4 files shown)


haoweiah:
Ready to merge

PR #1327, opened 2018-09-25
hv: arch: fix 'Unused procedure parameter'

shiqingg:
MISRA-C requires that there should be no unused parameters in
functions.

In some cases, we will keep the unused parameters.
vmexit handler is one example. It is used as function pointer.
Some of the vmexit handlers use the input parameter 'vcpu', some of
them don't. We still need to keep the unused parameters 'vcpu' for
those handlers don't use 'vcpu'.

This patch removes the unused parameters that is not being used
unconditionally.

v1 -> v2:
 * remove the non-implemented API 'vlapic_id_write_handler'

Tracked-On: #861


haoweiah:
Ready to merge

PR #1334, opened 2018-09-26
Documentation: updates to Getting Started Guide

gvancuts:


dongyaozu:
@ywan170  can you have a look?

doc/getting-started/apl-nuc.rst
```diff
@@ -281,6 +281,18 @@ partition. Follow these steps:
 #. From here you can login as root using the password you set previously when
    you installed Clear Linux.
 
+#. (**Optional**) The ``software-defined-cockpit`` bundle installs the
+   ``ioc-cbc-tools`` on the system and activates three ``systemd`` services.
+   Those services do not work on off-the-shelf platforms such as NUCs, UP2.
```
dbkinder: Seems wrong to have software installed and services run that aren't supported on the open source defined hardware platforms.

dbkinder:
Are we waiting for approval by @alekdu to merge this change?

doc/getting-started/apl-nuc.rst
gvancuts: It is wrong but until we get these service files fixed, this is my best recommendation.

alekdu:
> Looks good to me.
> @alekdu Can you please help double confirm?

Looks good to me too.

doc/getting-started/apl-nuc.rst
alekdu: Agree, blindly running cbc service and waiting for service to die (30 seconds later) is not good solution. But there isn't really good way to detect IOC chip, unless ACPI table could give us some hints.

gvancuts: Is there any entry from the ACPI table we could use for that, something like a PlatformID or anything else? (disclaimer: I really don't know much about ACPI tables)

PR #1341, opened 2018-09-26
Documentation: add instruction to use a specific version of Clear

gvancuts:
Add a note (and instruction) on how to use a specific version of
Clear Linux on a system. Picking up the image of a specific release
is not sufficient as Clear Linux always grabs the very latest image
when installing.

### DIFF
--- a/doc/getting-started/apl-nuc.rst
+++ b/doc/getting-started/apl-nuc.rst
@@ -62,6 +62,12 @@ complete this setup.
 
       # swupd autoupdate --disable
 
+   .. note::
+      The Clear Linux installer will automatically check for updates and install the
+      latest version available on your system. If you wish to use a specific version
+      (such as 25130), you can achieve that after the installation has completed using
+      ``swupd verify --fix --picky -m 25130``
+
 #. If you have an older version of Clear Linux already installed
    on your hardware, use this command to upgrade Clear Linux
    to version 25130 (or newer):


doc/getting-started/apl-nuc.rst
```diff
@@ -62,6 +62,12 @@ complete this setup.
 
       # swupd autoupdate --disable
 
+   .. note::
+
+      Clear Linux will automatically grab the latest version available during the
+      installation process and put it on your system. If you wish to use a specific
+      version (such as 25130), you can enforce that using ``swupd verify --fix --picky -m 25130``
```
dbkinder: Your choice of saying "enforce" makes me wonder, will this command **prevent** an update to a newer version, or just **restore** the system to the specified version if it (accidentally) gets changed to a newer one?

gvancuts: It will restore the system to that specific version. The Clear Linux installer **will** pull down the latest version of Clear and install that one (it's not accidental), this is how you can force your system to go back to the version we specify.

gvancuts: I have changed the phrasing, can you let me know if you like it better?

Adding new bundles to your system does not update it, it installs the bundle for the current version.

Restoring Clear Linux to an earlier version will preserver all bundles that are installed. What I don't know is what happens if a bundle did not exist in the version you're restoring it to.

PR #1346, opened 2018-09-27
hv: drop the macro arguments acting as formal parameter names

shiqingg:
This patch fixes the following issue pointed by Xiangyang and Junjie.
There are some macro arguments acting as formal parameter names.
Drop such arguments since they make no difference to the expanded
implementation and they might confuse some developers.

Here is an example.
'ptr' is dropped in this patch, which is acting as a formal parameter
name and make no difference to the expanded implementation.

-#define build_atomic_load(name, size, type, ptr)       \
+#define build_atomic_load(name, size, type)            \
 static inline type name(const volatile type *ptr)      \
{                                                       \
        type ret;                                       \
        asm volatile("mov" size " %1,%0"                \
                        : "=r" (ret)                    \
                        : "m" (*ptr)                    \
                        : "cc", "memory");              \
        return ret;                                     \
}

Some minor coding style fixes are also included in this patch.
- use TAB for the alignment rather than mixing TAB with space
- fix some typo in the comments

Tracked-On: #861

### DIFF
--- a/hypervisor/arch/x86/guest/instr_emul.c
+++ b/hypervisor/arch/x86/guest/instr_emul.c
@@ -633,19 +633,19 @@ static void vie_update_rflags(struct vcpu *vcpu, uint64_t rflags2, uint64_t psl)
 /*
  * Return the status flags that would result from doing (x - y).
  */
-#define build_getcc(name, type, x, y)	\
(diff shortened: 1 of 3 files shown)


haoweiah:
Ready to merge

Issue #1387, opened 2018-10-01
Newer versions of `kconfiglib` break the documentation generation process


gvancuts:
If you install `kconfiglib` today (10/01/2018) using `pip3 install --user kconfiglib`, you will get version 10.9.1. Using that version to generate the documentation pages will lead to the following error:
```
gvancuts@gvancuts-MOBL3:~/acrn-hypervisor$ make doc
make -C /home/gvancuts/acrn-hypervisor/doc html BUILDDIR=/home/gvancuts/acrn-hypervisor/build/doc
make[1]: Entering directory '/home/gvancuts/acrn-hypervisor/doc'
Traceback (most recent call last):
  File "scripts/genrest.py", line 377, in <module>
    write_kconfig_rst()
  File "scripts/genrest.py", line 97, in write_kconfig_rst
    write_sym_rst(sym, out_dir)
  File "scripts/genrest.py", line 123, in write_sym_rst
    kconfig_definition_rst(sym))
  File "scripts/genrest.py", line 318, in kconfig_definition_rst
    for node in sc.nodes)
  File "scripts/genrest.py", line 318, in <genexpr>
    for node in sc.nodes)
  File "/home/gvancuts/.local/lib/python3.5/site-packages/kconfiglib.py", line 4989, in __str__
    return self.custom_str(standard_sc_expr_str)
  File "/home/gvancuts/.local/lib/python3.5/site-packages/kconfiglib.py", line 4998, in custom_str
    self._sym_choice_node_str(sc_expr_str_fn)
  File "/home/gvancuts/.local/lib/python3.5/site-packages/kconfiglib.py", line 5061, in _sym_choice_node_str
    indent_add_cond("default " + expr_str(default, sc_expr_str_fn),
TypeError: expr_str_rst() takes 1 positional argument but 2 were given
Makefile:48: recipe for target 'kconfig' failed
make[1]: *** [kconfig] Error 1
make[1]: Leaving directory '/home/gvancuts/acrn-hypervisor/doc'
Makefile:38: recipe for target 'doc' failed
make: *** [doc] Error 2
```

Installing an older version (such as 7.0.0: `pip3 install --user kconfiglib==7.0.0`) solves this issue.

I do not know what version broke this, I am only referencing 7.0.0 because that's on one of my other dev system where I can generate the documentation successfully.

I do not know either whether we can correctly configure the ACRN hypervisor using this latest version, this is something we should check too.

gvancuts:
[...]

PR #1388, opened 2018-10-01
doc: update genrest script for latest kconfiglib

dbkinder:
The genrest.py script (used to generate config documentation from the
Kconfig files) broke when upgrading to the latest kconfiglib (10.9.1).
Copied the latest genrest.py script from the Zephyr project (where this
script and processing was developed) and things work again.

Also update requirements.txt to match the kconfiglib version verified to
work.

fixes: #1387


gvancuts:
Thanks @dbkinder , I'll test this now.

You;ll need to change the `fixes:` to `Tracked-On` for our CI system to be happy though

dbkinder:
DO NOT MERGE until #1389 is addressed or all CI builds that do a `make docs` will fail.

gvancuts:
@dbkinder , I have thinking and thinkering with this a bit this morning. We should augment this PR to also add a check for the `kconfiglib` in your `doc/scripts/show-versions.py` script.

Thinking about it some more, it would actually be better for the `show-versions.py` script to gather all requirements captured in the file and print out the versions that are actually installed... so I came up with this alternative script:
```
#!/usr/bin/env python3
#
# Copyright (c) 2018, Intel Corporation
#
#
# Show installed versions of doc building tools

import os.path
import sys
import pkg_resources
import subprocess

class color:
   PURPLE = '\033[95m'
   CYAN = '\033[96m'
   DARKCYAN = '\033[36m'
   BLUE = '\033[94m'
   GREEN = '\033[92m'
   YELLOW = '\033[93m'
   RED = '\033[91m'
   BOLD = '\033[1m'
   UNDERLINE = '\033[4m'
   END = '\033[0m'

# Check all requirements listed in requirements.txt and print out version installed (if any)
print (color.BOLD + color.BLUE + "Listing out dependencies found on your sytem..." + color.END)

rf = open(os.path.join(sys.path[0], "requirements.txt"),"r")

for reqs in pkg_resources.parse_requirements(rf):
    try:
        ver = pkg_resources.get_distribution(reqs.project_name).version
        print (color.DARKCYAN + reqs.project_name + " version: " + ver + color.END)
    except:
        print (color.RED + color.BOLD + reqs.project_name + " is missing." + color.END +
                " (Hint: install all dependencies with " + color.YELLOW +
                "\"pip3 install --user -r requirements.txt\"" + color.END + ")")

rf.close()

# Print out the version of Doxygen (not installed via pip3)
print (color.DARKCYAN + "doxygen version: " + subprocess.check_output(["doxygen", "-v"]).decode("utf-8") + color.END)
```

This is an example of the output (minus the colouring and formatting):
```
Listing out dependencies found on your sytem...
breathe version: 4.9.1
sphinx version: 1.7.7
docutils is missing. (Hint: install all dependencies with "pip3 install --user -r requirements.txt")
sphinx-rtd-theme version: 0.4.0
kconfiglib version: 10.9.2
doxygen version: 1.8.13

```

What do you think?

dbkinder:
Thanks. Fixed a misspelling (sytem), removed colors (other than the error message), and added some indenting and padding to make it a bit easier to read.  (The blue and dark cyan were hard to read on a black background terminal.)

gvancuts:
Is there a real risk that `pkg_resources` returns a different version than the one being used? We can keep your version if you think the risk is real. My starting point when going down that path was really that I wanted to be able to extract from the `requirements.txt` file what the deps were and print out the versions for each (as opposed to maintaining two lists in different files). It seems cleaner but keeping these two lists in sync isn't a whole lot of work either so I'm fine either way.

dbkinder:
I'm OK with using pkg_resources, just was an observation.  And I like getting the list of dependencies from requirements.txt, so it works for me.

PR #1429, opened 2018-10-11
V3: HV: parse seed from ABL

YadongQi:
ABL pass seed_lists to HV through different interface/structures.
So, in this patch, add interface to retrieve seed from ABL and refactor
the seed parsing logic.


YadongQi:
misoperation, please ignore.

PR #1439, opened 2018-10-11
hv: fix 'No brackets to then/else'

shiqingg:
- add missing brackets for 'if/else' statements based on MISRA-C
  requirements

v1 -> v2:
 * add brackets for each conditions in 'if' statements to improve
   the readability
 * modify 'ptdev_init' to make the logic clearer

Tracked-On: #861

### DIFF
--- a/hypervisor/common/ptdev.c
+++ b/hypervisor/common/ptdev.c
@@ -127,8 +127,9 @@ release_all_entries(struct vm *vm)
 	list_for_each_safe(pos, tmp, &ptdev_list) {
 		entry = list_entry(pos, struct ptdev_remapping_info,
 				entry_node);
-		if (entry->vm == vm)
+		if (entry->vm == vm) {
 			release_entry(entry);
+		}
 	}
 }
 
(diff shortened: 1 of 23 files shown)


hypervisor/common/ptdev.c
```diff
@@ -191,8 +192,9 @@ ptdev_deactivate_entry(struct ptdev_remapping_info *entry)
 
 void ptdev_init(void)
 {
-	if (get_cpu_id() > 0)
+	if (get_cpu_id() != BOOT_CPU_ID) {
```
wuxyintel: Hi Shiqing, Is the ptdev_init only called by the BSP? Can we use precondition instead of this checking? Please check it, thanks!

shiqingg: @wuxyintel Yes, ptdev_init is only called by BSP.
@dongyaozu Eddie, do you agree that we should use precondition to replace this check?

haoweiah:
Ready to merge

dongyaozu:
I am fine with either choice.

dongyaozu:
I am fine with either choice.

shiqingg:
Since this patch has been merged, will keep it as it is.

PR #1457, opened 2018-10-14
Fixed Acrnd crash caused by fd leaking

lyan3:
Fixed Acrnd crash caused by fd leaking and removed an unnecessary pre-assumption that fd num less than 1024.
Tracked-On: #1477
Signed-off-by: Tao, Yuhong <yuhong.tao@intel.com>


wenlingz:
@lyan3 please add Tracked-On info, thanks.

Issue #1363, opened 2018-09-28
SOS+UOS glmark2-es-wayland causes hypervisor crash


mattcurfman:
NUC7i7DNHE (Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz, Mesa DRI Intel(R) UHD Graphics 620)
Clear Linux 25180
HV version 0.3-unstable-2018-09-25 10:26:50- DBG 

SOS: 1 CPU
UOS: 1 CPU

Run 'weston-es2-wayland' simultaneously in both SOS and UOS.  After 5-20 seconds, hypervisor reports 'vcpu handling pending request fail' in console and entire system hard-hangs (console frozen, SOS and UOS frozen)  100% reproducible

@hwang37

mattcurfman:
I am not able to reproduce this crash on an UP2 board

hwang37:
We'll take a look and assign to our Zhipeng.

dandanyx:
Cannot reproduce this issue on SOS #25860, please help to close it.

mattcurfman:
@dandanyx, I will re-try SOS #25860 and close if no longer reproducible. Thx!

mattcurfman:
@dandanyx, #25860 is much worse: it crashes hypervisor when booting SOS.  Since you closed this defect without waiting for my feedback, I will open a new issue.

```
ACRN Hypervisor
calibrate_tsc, tsc_khz=2700000
[7631962us][cpu=0][sev=2][seq=3]:HV version 0.3-unstable-2018-10-26 12:55:07- DBG (daily tag:) build by mockbuild, start time 7627046us
[7645087us][cpu=0][sev=2][seq=4]:API version 1.0
[7649969us][cpu=0][sev=2][seq=5]:Detect processor: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
[7658982us][cpu=0][sev=2][seq=6]:hardware support HV
[7674659us][cpu=0][sev=3][seq=12]:System S3/S5 is NOT supported.
[7681034us][cpu=0][sev=2][seq=18]:Start VM0
[7685402us][cpu=0][sev=3][seq=19]:VMX ctrl 0x48b not fully enabled: request 0x18b but get 0x8b

ACRN:\>[17851978us][cpu=0][sev=3][seq=9600]:vlapic: Start Secondary VCPU1 for VM[0]...
[...]

PR #1462, opened 2018-10-15
HV: allow no IRR when pending bit set if no APIC-V

mcao6:
An issue reported on KBL NUC (No APIC-V), when runing GVT test
case, it is easily to cause SOS/UOS hung. This patch is to root
cause it and how to avoid it.

On some platforms with no APIC-V support, this modidication can
avoid SOS hung when no IRR but event pending bit set.

If no APIC-V, interrupt injection will use IRR in vLAPIC and
event pending bitmap; work logic as following (set ABC for notes):
1. in ISR or when UOS send an interrupt to SOS/VHM (like IO request),
    step A: set IRR --> step B: set event pending bit;
2. in SOS event handing,  step C: check/clear event pending bit -->
  step D: probe/get IRR --> step E: clear/handle IRR.
3. after that, it will probe IRR again to check if other IRR left:
  step F: probe IRR --> step G: set event pending bit --> step H:
  enable IRQ window in VMX.

Before, from step C to D, if pending bit checked, but no IRR, it will
return -1, then the CPU will goto ZOMBIE status. It can cause
SOS hung. It can happen occasionally under following case :
4. between UOS step A and step B, SOS can goto step F, so it
gets IRR, and step G --> step H, then SOS enter non-root mode, for
IRQ window enabled, it will cause vmexit to continue step C/D/E.
5. then UOS does step B, to set pending bit, but no IRR now. When next
vmexit, SOS does step C and D, it can't get IRR, failure happens.

In summary, a failed case steps: A-->F-->G-->H-->C-->D-->E-->B-->C-->D

So we allow that when event pending bit checked, IRR could be cleared
already.It just wastes one probe time occasionally.

Tracked-On: #1363

### DIFF
--- a/hypervisor/arch/x86/virq.c
+++ b/hypervisor/arch/x86/virq.c
@@ -132,7 +132,7 @@ static int vcpu_inject_vlapic_int(struct vcpu *vcpu)
 	 *   through the local APIC.
 	 */
 	if (ret == 0) {
-		return -1;
+		return 0;
 	}
 
 	if (!(vector >= 16U && vector <= 255U)) {


HuiHuangShi:
1462<br>No new violations to the coding guideline detected.

PR #1491, opened 2018-10-17
DM: add acrn_timer api for timer emulation

jsun26intel:
We will use timerfd and epoll mechanism to emulate kinds of timers like
PIT/RTC/WDT/PMTIMER/... in device model under Linux. The api is unified
in this patch.

Compare with sigevent mechanism, timerfd has a advantage that it could
avoid race condition on resource accessing in the async sigev thread.

This patch set replace timer api for RTC and WDT.

change log:
	v1 -> v2: add NULL pointer check for function parameter;
	v2 -> v3: rename file name of vtimer.* to timer.*;
		  rename structure name of vtimer to acrn_timer;
		  add read() in timer handler to consume I/O event;
	v3 -> v4: replace bool clock_realtime with int clockid;
		  close acrn_timer->fd properly;

### DIFF
--- a/devicemodel/Makefile
+++ b/devicemodel/Makefile
@@ -122,6 +122,7 @@ SRCS += core/mptbl.c
 SRCS += core/main.c
 SRCS += core/hugetlb.c
 SRCS += core/vrpmb.c
+SRCS += core/timer.c
 
 # arch
 SRCS += arch/x86/pm.c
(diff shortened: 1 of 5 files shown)


HuiHuangShi:
1491 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

ClaudZhang1995:
Ready to merge

PR #1497, opened 2018-10-17
Add string api, replace banned functions for virtio-blk

conghuic23:
1.  As function sscanf is banned, to get value from parameter buffer,strto*
    is recommended. To reduce the inspection code when using strto*, it's
    better to use a string convert API.
2.  replace banned functions for virtio-blk.


HuiHuangShi:
1497 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

Issue #1498, opened 2018-10-17
add watchdog MSI and INTR support in DM


jsun26intel:
The i6300esb watchdog has a feature that it could generate a interrupt on stage1 timeout, the reset occurs on stage2 timeout. This feature could be used for debug that user could dump system info in the interrupt.
We need to enable the feature in DM.

jsun26intel:
[External_System_ID] 260218

Mingyuan18:
No regression issue found in test

PR #1499, opened 2018-10-17
DM: add MSI and INTR support for i6300esb watchdog

jsun26intel:
Per i6300esb spec, when WDT_INT_TYPE(bit 0 and 1 of WDT config register)
are set to 00, IRQ feature should be supported.

The WDT_INT_ACTIVE bit is set when the first stage of the 35-bit
down-counter reaches zero. An interrupt will be generated if WDT_INT_TYPE
is configured to do so (See WDT Configuration Register). This is a sticky
bit and is only cleared by writing a 1.

SMI feature(WDT_INT_TYPE are set to 0x10) is not supported.

Tracked-On: #1498


HuiHuangShi:
1499 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

ClaudZhang1995:
Ready to merge

Issue #1536, opened 2018-10-20
dm: add virtio_mei mediator


tomasbw:
The MEI device mediator works over virtualization enabled mei device and mei driver.
Currently it only supports MEI on APL-I with HBM version 2.2
And as well it requires an SOS kernel with mei virtualization support.
For other platforms only passthru  mode is enabled.

mwang106:
[External_System_ID] 261352

PR #1537, opened 2018-10-20
dm: add virtio_mei mediator

tomasbw:
The MEI device mediator works over virtualization enabled mei device and mei driver.
Currently it only supports MEI on APL-I with HBM version 2.2
And as well it requires an SOS kernel with mei virtualization support.
For other platforms only passthru  mode is enabled.
Tracked-On: #1536


ClaudZhang1995:
Ready to merge

PR #1560, opened 2018-10-23
hv: fix problemtic logic in create_vcpu()

yonghuah:
"hw.create_vcpus++" should be after checking that resource is
 enough to create one more vcpu.

Tracked-On: #1252

### DIFF
--- a/hypervisor/arch/x86/guest/vcpu.c
+++ b/hypervisor/arch/x86/guest/vcpu.c
@@ -315,15 +315,15 @@ int create_vcpu(uint16_t pcpu_id, struct vm *vm, struct vcpu **rtn_vcpu_handle)
 
 	pr_info("Creating VCPU working on PCPU%hu", pcpu_id);
 
+	if (vm->hw.created_vcpus >= CONFIG_MAX_VCPUS_PER_VM) {
+		pr_err("%s, no more vcpus can be supported!\n", __func__);
+		return -EINVAL;
+	}
 	/*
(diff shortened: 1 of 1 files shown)


hypervisor/arch/x86/guest/vcpu.c
```diff
@@ -315,15 +315,15 @@ int create_vcpu(uint16_t pcpu_id, struct vm *vm, struct vcpu **rtn_vcpu_handle)
...
+	}
 	/*
 	 * vcpu->vcpu_id = vm->hw.created_vcpus;
 	 * vm->hw.created_vcpus++;
 	 */
 	vcpu_id = atomic_xadd16(&vm->hw.created_vcpus, 1U);
-	if (vcpu_id >= CONFIG_MAX_VCPUS_PER_VM) {
-		pr_err("%s, vcpu id is invalid!\n", __func__);
-		return -EINVAL;
-	}
```
dongyaozu: Had a second thinking. If there are 2 threads executing  this code, and they both passed the above check of created_vcpus (created_vcpus = CONFIG_MAX_VCPUS_PER_VM-1) .  And both comes here.  It is still possible vcpu_id may be CONFIG_MAX_VCPUS_PER_VM + 1.

We seems need a lock somewhere to protect hypercalls... Need more thinking.,.,

Will pend this patch

yonghuah: @dongyaozu  OK,  Got it, will revise, thanks. thanks.

Issue #1563, opened 2018-10-23
tools: acrnd: Stop all vms when SOS shutdown/reboot


KaigeFu:
When SOS shutdown/reboot, systemd will send SIGTERM to acrnd. We should catch up this signal and stop all vms gracefully.

mwang106:
[External_System_ID] 261412

PR #1564, opened 2018-10-23
tools: acrnd: Stop all vms when SOS shutdown/reboot

KaigeFu:
When SOS shutdown/reboot, systemd will send SIGTERM to acrnd. We should catch up
this signal and stop all vms gracefully.

BTW, this path also fix the following error when stop acrnd service by removing
ExecStop config. Systemd will send SIGTERM signal to process by default.

systemd[9378]: acrnd.service: Failed to execute command: No such file or directory
systemd[9378]: acrnd.service: Failed at step EXEC spawning /usr/bin/killall: No such file or directory
-- Subject: Process /usr/bin/killall could not be executed
-- Defined-By: systemd
-- Support: https://lists.freedesktop.org/mailman/listinfo/systemd-devel
--
-- The process /usr/bin/killall could not be executed and failed.
--
-- The error number returned by this process is 2.

Tracked-On: #1563

### DIFF
--- a/tools/acrn-manager/acrnd.c
+++ b/tools/acrn-manager/acrnd.c
@@ -5,6 +5,7 @@
 
 #include <time.h>
 #include <pthread.h>
+#include <signal.h>
 #include <sys/queue.h>
 #include <sys/stat.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


ClaudZhang1995:
Ready to merge

ClaudZhang1995:
Ready to merge

Issue #1570, opened 2018-10-23
dm:mei: HBM disconnect doesn't have address set.


tomasbw:
MEI HBM the disconnect replay doesn't have the address set, breaking the protocol.

mwang106:
[External_System_ID] ACRN-1621

tomasbw:
Yes, sorry I'm sorry I'm not clear about the process, I supposed that the validation is closing the tickets.

PR #1583, opened 2018-10-24
dm: mei: set addresses in the hbm disconnect reply

tomasbw:
MEI HBM the disconnect replay doesn't have the address set,
breaking the protocol, fix the issue.

Tracked-On: #1570
Signed-off-by: Alexander Usyskin <alexander.usyskin@intel.com>


tomasbw:
What is ext-id?

Issue #1254, opened 2018-09-17
 Replace unbounded functions with safe string and buffer handling  functions in ACRN DM & Tools


yonghuah:
During static code analysis of the ACRN hypervisor & kernel source code, excessive counts of unsafe functions were found in use. Use of deprecated and dangerous functions is not recommended as they often have known security weaknesses, leading to
resultant weaknesses such as buffer overflows and other unpredictable behavior. Unsafe functions also allows the attackers to exploit software through various binary-level attacks

mwang106:
[External_System_ID] 254443

mwang106:
all banned functions  have been removed for device model and tools modules. and also we refined the usage for functions that can be used conditionally.

PR #1586, opened 2018-10-24
tools: acrn-crashlog: remove unsafe apis in usercrash

xiaojingrich:
Since strlen/vsnprintf/ato* api are not safe, so use strnlen instead of
strlen, use vasprintf instead of vsnprintf and use strtol instead of
atoi.

Tracked-On: #1254
Signed-off-by: xiaojin2 <xiaojing.liu@intel.com>
Reviewed-by: Huang Yonghua <yonghua.huang@intel.com>
Reviewed-by: Liu, Xinwu <xinwu.liu@intel.com>
Acked-by: Chen Gang <gang.c.chen@intel.com>


ClaudZhang1995:
Ready to merge

Issue #1160, opened 2018-09-05
Add support for partition mode


zds100:
Add support for partition mode on ACRN hypervisor.  With the partition mode enabled, Hypevisor assigns the PCI passthru devices to UOS directly, achieving a near-native performance.  To support partition mode,  we need to add/port the following to Hypervisor:
1.  VPCI: needs to capture the PCI config space access and emulate a virtual PCI bus to UOS.
The per-vm PCI passhtru device list needs to be defined to specify a list of the PCI passthru devices assigned to UOS. 
2. VRTC: needs to emulate a virtual RTC to UOS
3. MP table:  pass MP table to UOS, ACPI will not be used
4. LAPIC passthru/interrupt remapping:  pass LAPIC directly to UOS to achieve exit-less interrupt handling, no interrupt remapping on Hypervisor side
5. Eliminate DM: do not use DM to launch UOS, launch UOS directly in hypervisor instead.

KaigeFu:
[External_System_ID] 255255

Mingyuan18:
The patch was merged.

PR #1601, opened 2018-10-25
HV: Fix boot failure of partition mode

KaigeFu:
When drdh_info_array[0].ignore = false with partition mode enabled,the following
assert error will be triggered when trying to boot vm:

Assertion failed in file arch/x86/vtd.c: fatal error

The root cause is that root_table_addr reading from CPU3(vm2's bsp) is NULL.

This patch move the init_iommu which initialize root_table_addr before starting
secondary cores and flush the cache at the end if the function to avoid getting
NULL pointer from secondary cores.
Tracked-On: #1160
Reviewed-by: Binin Wu <binbin.wu@intel.com>

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -482,6 +482,8 @@ static void bsp_boot_post(void)
 	/* Initialize interrupts */
 	interrupt_init(BOOT_CPU_ID);
 
+	init_iommu();
+
 	timer_init();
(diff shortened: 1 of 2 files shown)


KaigeFu:
Tracked-On: #1160

HuiHuangShi:
1601<br>No new violations to the coding guideline detected.

ClaudZhang1995:
Ready to merge

Issue #1606, opened 2018-10-25
kernel hang on smp_call_function   


mchinth:
race condition created as smp_call_mask  can be modified by more than one process which can cause kernel hang.

mwang106:
[External_System_ID] ACRN-1607

Mingyuan18:
This is an old issue and has been fixed for a long time.

PR #1608, opened 2018-10-25
Fix to kernel hang in smp_call_function

mchinth:
smp_call_function:
 smp_call_mask can be modified by more than one process which can cause kernel hang and timeout

Adding a lock to smp_call_mask to prevent race condition

Tracked-On: #1606
Acked-by:   Xu, Anthony <anthony.xu.intel.com>
Signed-off-by: Manisha Chinthapally <manisha.chinthapally@intel.com>

### DIFF
--- a/hypervisor/arch/x86/notify.c
+++ b/hypervisor/arch/x86/notify.c
@@ -25,7 +25,7 @@ static void kick_notification(__unused uint32_t irq, __unused void *data)
 		if (smp_call->func != NULL) {
 			smp_call->func(smp_call->data);
 		}
-		bitmap_clear_nolock(pcpu_id, &smp_call_mask);
+		bitmap_clear_lock(pcpu_id, &smp_call_mask);
 	}
 }
 


hypervisor/arch/x86/notify.c
```diff
@@ -25,7 +25,7 @@ static void kick_notification(__unused uint32_t irq, __unused void *data)
 		if (smp_call->func != NULL) {
 			smp_call->func(smp_call->data);
 		}
-		bitmap_clear_nolock(pcpu_id, &smp_call_mask);
+		bitmap_clear_lock(pcpu_id, &smp_call_mask);
```
dongyaozu: There are 2 places which do bitmap_clear, should we change both of them to use lock API?

anthonyzxu: the other is called inside smp_call_function, we assume smp_call_function won't be called in multiple processors. otherwise we need to redesign smp_call.

anthonyzxu:
This patch should cause below failure, please check.

"make clean && make sos_kernel hypervisor PLATFORM=uefi
Makefile:255: *** missing separator (did you mean TAB instead of 8 spaces?).  Stop.
Build step 'Execute shell' marked build as failure"

HuiHuangShi:
1608<br>No new violations to the coding guideline detected.

PR #1622, opened 2018-10-26
Documentation: tweak 'partition mode' tutorial

gvancuts:
Tweak the 'Using partition mode on UP2' tutorial by removing
a couple of Grub menu entries that are not present by default
(nor introduced as part of the tutorial).

Also make it more obvious that there is still a default entry
called 'Ubuntu' which is the one by default.

### DIFF
--- a/doc/tutorials/using_partition_mode_on_up2.rst
+++ b/doc/tutorials/using_partition_mode_on_up2.rst
@@ -187,9 +187,8 @@ Enabling partition mode
   .. code-block:: console
      :emphasize-lines: 2
 
-      ACRN Partition Mode with Zephyr
-     *ACRN Partition Mode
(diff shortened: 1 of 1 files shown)


doc/tutorials/using_partition_mode_on_up2.rst
```diff
@@ -187,9 +187,8 @@ Enabling partition mode
   .. code-block:: console
      :emphasize-lines: 2
 
-      ACRN Partition Mode with Zephyr
-     *ACRN Partition Mode
-      ADG Partition Mode
+     *Ubuntu
```
dbkinder: OK, I just made a text version of the askew screen photo that was there.  You need to change it to ``:emphasis-lines: 1`` above since it's now the first line of the block.

Issue #1632, opened 2018-10-28
dm: mei: fix firmware reset race.


tomasbw:
The FW reset is currently detected from two points upon
    a read failure from native read and from the reset
    handler.
    The fix removes the detection from the mevent rx callback,
    leaving a single detection point.
    To prevent reset hiccup, hw_ready is not set if a full rescan is
    performed, it will be set only when virtio FW will request the FW reset.

mwang106:
[External_System_ID] ACRN-1623

Mingyuan18:
This is an old issue and has been fixed for a long time.

PR #1633, opened 2018-10-28
Virtio mei reset Tracked-On: #1632

tomasbw:
The FW reset is currently detected from two points upon
    a read failure from native read and from the reset
    handler.
    The fix removes the detection from the mevent rx callback,
    leaving a single detection point.
    To prevent reset hiccup, hw_ready is not set if a full rescan is
    performed, it will be set only when virtio FW will request the FW reset.

Tracked-On: #1570
Signed-off-by: Alexander Usyskin <alexander.usyskin@intel.com>


HuiHuangShi:
1633 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

Issue #1659, opened 2018-10-30
AaaG user build boot failure in some case


binbinwu1:
When more than 2GB memory allocated to AaaG, and AaaG is flash unlocked, AaaG user build will fail to boot.

binbinwu1:
[External_System_ID] 261644

PR #1660, opened 2018-10-30
dm: bios: update to version 1.0.1

binbinwu1:
Bug fix:
- Fix AaaG user build boot failure with  memory size > 2GB and flash unlocked.

Tracked-On: #1659


binbinwu1:
[External_System_ID] 261644

Issue #1691, opened 2018-11-01
System hang after resume from S3


fyin1:
There is possible the system hang after resume from S3 with following log:
[393948us][cpu=0][sev=1][seq=119]:Secondary CPUs failed to come up

fyin1:
[External_System_ID] ACRN-1637

fyin1:
The patch was merged and we could close this issue.

PR #1700, opened 2018-11-02
hv: Remove the up_count_spinlock and use atomic for up_count

fyin1:
It's possible that the up_count_spinlock is not release during
system enter S3. The case is like following:

   BSP                                  AP
 stop_cpus                            cpu_dead
                                         cpu_set_current_state
                                            spinlock_abtain
                                                up_count--

   wait_for up_count == 1
 enter S3

                                            spinlock_release

Especially, considering the real spinlock release could be delayed
by cache.

Actually, the most content protected by up_count_spinlock is per
cpu data and could be changed without lock. Only left is up_count.
This patchset remove the up_count_spinlock and use atomic API
for up_count changing.

Tracked-On: #1691

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -14,15 +14,10 @@ spinlock_t trampoline_spinlock = {
 	.tail = 0U
 };
 
-static spinlock_t up_count_spinlock = {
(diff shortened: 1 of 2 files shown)


fyin1:
Checked with Huihuang. This violation could be skipped.

Issue #1670, opened 2018-10-31
[IO Mediator][SIT] touch with USB3.0 port in uos, when touch the screen , it always print Error Infor: not enough avail descs	


lidongwangx:
Environment Image info:
acrn-hypervisor: acrn-2018w44.1-140000p
linux-pk414: Linux sos-dom0 4.19.0-rc8-quilt-2e5dc0ac-00010-g689bffb94893 #1 SMP PREEMPT Tue Oct 30 01:30:38 UTC 2018 x86_64 GNU/Linux
SOS version: 25920
linux-pk version: 4.19.0-rc8-quilt-2e5dc0ac-g284805005401
UOS OS Type: AaaG
Linux localhost 4.19.0-rc8-quilt-2e5dc0ac-g284805005401 #1 SMP PREEMPT Wed Oct 24 19:29:04 UTC 2018 x86_64
Reproduce rate: 100%

Reproduce steps:
-------------
1: insert touch label into usb3.0 port or usb2.0 port.

2. check input device info. And find the input event in SOS:
root@sos-dom0 /home # cat /proc/bus/input/devices
I: Bus=0003 Vendor=04f3 Product=24e1 Version=0110
N: Name="ELAN Touchscreen"
P: Phys=usb-0000:00:15.0-2/input0
S: Sysfs=/devices/pci0000:00/0000:00:15.0/usb1/1-2/1-2:1.0/0003:04F3:24E1.0004/input/input11
U: Uniq=
H: Handlers=mouse0 event1
B: PROP=2
B: EV=1b
B: KEY=400 0 0 0 0 0
B: ABS=3273800000000003
B: MSC=20

4. add virtio-input parameter in launch.sh:
-s 7,virtio-input,/dev/input/event1
5. boot up uos and check touch take effect on uos
when touch the screen, it always print Error Infor:console:/ $virtio_input_send_event: not enough avail descs, dropped:69

Expected result:
Step 5,when touch the screen, it should not print related error information.

Actual result:
Step 5,when touch the screen, it always print Error Infor: console:/ $virtio_input_send_event: not enough avail descs, dropped:69

Debugging info:
console:/ $virtio_input_send_event: not enough avail descs, dropped:69

Attachement:
no need the attachment. console:/ $virtio_input_send_event: not enough avail descs, dropped:69

lidongwangx:
[External_System_ID] ACRN-1619

Mingyuan18:
[...]

PR #1705, opened 2018-11-02
dm: virtio-input: ignore MSC_TIMESTAMP from guest

chejianj:
(EV_MSC, MSC_TIMESTAMP) is added to each frame just before the
SYN event since kernel 4.15. EV_MSC is configured as
INPUT_PASS_TO_ALL. In the use case of virtio-input, there is
a loop as follows:
- A mt frame with (EV_MSC, MSC_TIMESTAMP) is passed to FE.
- FE will call virtinput_status to pass (EV_MSC, MSC_TIMESTAMP)
  back to BE.
- BE writes this event to evdev. Because (EV_MSC, MSC_TIMESTAMP)
  is configured as INPUT_PASS_TO_ALL, it will be written into
  the event buffer of evdev then be read out by BE without
  SYN followed.
- Each mt frame will introduce one (EV_MSC, MSC_TIMESTAMP).
  Later the frame becomes larger and larger...

This patch fixed above issue by ignoring MSC_TIMESTAMP from guest.
Besides that timestamp is added for every status event from guest
before writing to evdev.

Tracked-On: #1670

### DIFF
--- a/devicemodel/hw/pci/virtio/virtio_input.c
+++ b/devicemodel/hw/pci/virtio/virtio_input.c
@@ -208,6 +208,33 @@ virtio_input_cfgwrite(void *vdev, int offset, int size, uint32_t val)
 	return 0;
 }
 
+static bool
+virtio_input_ignore_event(struct virtio_input_event *event)
+{
+	if (!event)
+		return true;
+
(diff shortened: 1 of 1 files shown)


HuiHuangShi:
1705 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

PR #1709, opened 2018-11-02
HV:treewide:rename data structure

wuxyintel:
In the hypervisor, some names of data structure type is
identical with variable name in the same scope. This is
a MISRA C  violation.

This patch is to fix these naming MISRA C violation.

Tracked-On: #861


wenlingz:
OK to verify

lijinxia:
OK to verify

wenlingz:
OK to verify

PR #1742, opened 2018-11-07
hv: fix a minor bug of static checks

lifeix:
When a macro is expanded, the two tokens on either side of each ‘##’ operator
are combined into a single token, which then replaces the ‘##’ and the two
original tokens in the macro expansion. So we need use CAT__ to expand the
__LINE__ MACRO and use CAT_ to combine the expaneded MACRO.


HuiHuangShi:
1742<br>No new violations to the coding guideline detected.

PR #1754, opened 2018-11-08
hv: fix 'Expression is not Boolean'

shiqingg:
MISRA-C requires that the controlling expression of an if statement or
an iteration-statement shall be Boolean type.

v1 -> v2:
* update the U suffix of constant to let the type of operands
  to bit operations (&) is the same.
* update the type of 'i' from uint64_t to uint16_t in
  'profiling_set_control' since it is limited by phys_cpu_num,
  which is uint16_t.

Tracked-On: #861


ClaudZhang1995:
Ready to merge

PR #1800, opened 2018-11-12
Documentation: adjust "Tracked-On" capitalization in documentation

gvancuts:
Update the capitalization of "Tracked-On" in the documentation.
"Tracked-on" should work but it throws a warning so it's better
to make the documentation focus on using "Tracked-On" only.

### DIFF
--- a/doc/developer-guides/contribute_guidelines.rst
+++ b/doc/developer-guides/contribute_guidelines.rst
@@ -176,7 +176,7 @@ is set up correctly by using:
    git config --global user.name "David Developer"
    git config --global user.email "david.developer@company.com"
 
-Tracked-on
+Tracked-On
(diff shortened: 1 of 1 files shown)


gvancuts:
Here is where it was mentioned the CI system will throw a warning: https://github.com/projectacrn/acrn-kernel/pull/137#issuecomment-437283669

dbkinder:
OK, but I don't recall getting a warning when using "Tracked-on" vs. "Tracked-On"...

gvancuts:
> OK, but I don't recall getting a warning when using "Tracked-on" vs. "Tracked-On"...

That surprised me too. @NanlinXie was there any changes to the CI system made recently that would explain the warning now?

PR #1822, opened 2018-11-15
hv: partition mode also needs free vm id when shutdown vm

shuox:
Even in partition mode, we also need to clear vmid in vmid_bitmap to
indicate the VM is not present.

Tracked-On: projectacrn#1821


HuiHuangShi:
1822<br>No new violations to the coding guideline detected.

Issue #1837, opened 2018-11-16
'acrnctl list' shows incomplete VM names


liuming50:
I am currently using Yocto build and deploy UOS images, it would be much more convenient to let the VMs to have the same names with when they are being built. For instance, a typical image name in Yocto is like: uos-image-weston, but it can not show the name completely with 'acrnctl list', since the length limit is 16.

Increasing the limit to 32 would fix this problem.

mwang106:
[External_System_ID] ACRN-1744

lijinxia:
[External_System_ID] ACRN-1746

ftangx:
The bug cannot be reproduced in NUC and UP2.
Installed version: 26410
Version URL: http://linux-ftp.sh.intel.com/pub/mirrors/clearlinux/update
Content URL: http://linux-ftp.sh.intel.com/pub/mirrors/clearlinux/update

liuming50:
It's reproducible with the build from https://github.com/liuming50/meta-acrn-manifest, the patch had been merged so close it.

PR #1838, opened 2018-11-16
tools: acrn-manager: set MAX_NAME_LEN to 32

liuming50:
I am currently using Yocto build and deploy UOS images, it would be
much more convenient to let the VMs to have the same names with when
they are being built. For instance, a typical image name in Yocto is
like: uos-image-weston, but it can not show the name completely with
'acrnctl list', since the length limit is 16.

Increasing the limit to 32 would fix this problem.

Tracked-On: #1837


HuiHuangShi:
1838 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

HuiHuangShi:
1838 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

Issue #1853, opened 2018-11-20
Put partition mode board specific files under partition/$(CONFIG_BOARD)/ and added partition mode support for cb2_dnv


donsheng:
Put partition mode board specific files under partition/$(CONFIG_BOARD)/ and added partition mode support for cb2_dnv.

donsheng:
[External_System_ID] ACRN-1899

Mingyuan18:
No regression found, closed this issue.

Issue #1852, opened 2018-11-20
Partition mode source code file layout reorg


donsheng:
Partition mode source code file layout reorg

lijinxia:
[External_System_ID] ACRN-1906

NanlinXie:
No regression found, close this issue.

PR #1861, opened 2018-11-21
Put partition mode board specific files under partition/$(CONFIG_BOARD)/ and added partition mode support for cb2_dnv

donsheng:
Put partition mode board specific files under partition/$(CONFIG_BOARD)/ and added partition mode support for cb2_dnv

Tracked-On: #1852 #1853


hypervisor/arch/x86/configs/cb2_dnv.config
```diff
@@ -0,0 +1,7 @@
+CONFIG_BOARD="cb2_dnv"
```
gvancuts: It would have been better to follow the same (albeit implicit) convention here and use `dnv-cb2`.

donsheng: OK, we will discuss, and it this makes sense, we will change it

Issue #1595, opened 2018-10-25
doc: add comprehensive API documentation in code


eternalNight:
API documentation shall be maintained in the code and get updated whenever the documented code changes.

For the hypervisor, device model and tools we use doxygen-style comments. Here are templates for functions.

```
/**
 * @brief Initialize environment for Trusty-OS on a VCPU.
 *
 * More info here.
 *
 * @param[in]    vcpu Pointer to VCPU data structure
 * @param[inout] param guest physical address. This gpa points to
 *                     struct trusty_boot_param
 *
 * @retval 0 on success.
 * @retval -EIO description when this error can happen
 * @retval -EINVAL description …
 *
 * @pre vcpu != NULL
 * @pre param != 0UL
 *
 * @post the return value is non-zero if param is ....
 * @post …
 *
 * @remark The api must be invoked with interrupt disabled.
 * @remark (Other usage constraints here)
 */
```

Please note that there are two ways to describe the return values, either briefly describe the return value with a single `@return`, or use one or multiple adjacent `@retval` to list and describe the possible return values. Multiple `@retval` is preferred when a function returns an integer that indicate the success/failure of some operations. Here are the examples.

```
 * @retval 0 on success.
 * @retval -EIO description when this error can happen
 * @retval -EINVAL description …
```

or 

```
* @return the ID of the given \p vm
```

When a function returns nothing (i.e. the return type is void), use the following comment to explicitly state this.

```
* @return None
```

The following shows an example of doxygen-style document for a data structure.

```
/**
 * @brief An mmio request.
 *
 * More info here.
 */
struct mmio_request {
        uint32_t direction;  /**< Direction of this request. */
        uint32_t reserved; /**< Reserved. */
        int64_t address; /**< gpa of the register to be accessed. */
 
        /**
         * @brief Width of the register to be accessed. 
         * 
         * More multiple-paragraph descriptions.
         */
        int64_t size; 

        int64_t value; /**< Single-line comments can be appended after the field. */
} __aligned(8);
```

For kernel modules such as VHM, the [kernel-doc](https://www.kernel.org/doc/html/latest/doc-guide/kernel-doc.html) format shall be followed.

junjiemao1:
[External_System_ID] 262833

gvancuts:
@eternalNight could we update the template to have a single invokation of `@return`, even for multiple values? Otherwise we get as many "return" sections in the HTML output which isn't very readable.

So something like:
```
 * @return
[...]

PR #1872, opened 2018-11-22
doc: fix inconsistency and format in comments on return values

junjiemao1:
This series fixes the following issues reported by @gvancuts.

1. Absence of return values are stated in inconsistent ways in the doxygen-style comments.
1. Adjacent `@return`s are not joined by breathe, causing multiple `Return` sections in the generated doc.

Tracked-On: #1595

### DIFF
--- a/hypervisor/arch/x86/guest/vlapic.c
+++ b/hypervisor/arch/x86/guest/vlapic.c
@@ -1314,8 +1314,8 @@ vlapic_icrlo_write_handler(struct acrn_vlapic *vlapic)
  * @param[inout] vecptr Pointer to vector buffer and will be filled
  *               with eligible vector if any.
  *
- * @return 0 - There is no eligible pending vector.
- * @return 1 - There is pending vector.
+ * @retval 0 There is no eligible pending vector.
+ * @retval 1 There is pending vector.
  *
  * @remark The vector does not automatically transition to the ISR as a
  *	   result of calling this function.
@@ -1954,8 +1954,8 @@ vlapic_set_intr(struct acrn_vcpu *vcpu, uint32_t vector, bool level)
  *			   interrupt to all vCPUs.
  * @param[in] vector       Vector to be fired.
  *
- * @return 0 on success.
- * @return -EINVAL on error that vcpu_id_arg or vector is invalid.
+ * @retval 0 on success.
+ * @retval -EINVAL on error that vcpu_id_arg or vector is invalid.
  *
  * @pre vm != NULL
  */
@@ -1997,8 +1997,8 @@ vlapic_set_local_intr(struct acrn_vm *vm, uint16_t vcpu_id_arg, uint32_t vector)
  * @param[in] addr MSI address.
  * @param[in] msg  MSI data.
  *
- * @return 0 on success.
- * @return non-zero on error that addr is invalid.
+ * @retval 0 on success.
+ * @retval -1 on error that addr is invalid.
  *
  * @pre vm != NULL
  */
--- a/hypervisor/arch/x86/guest/vlapic.c
+++ b/hypervisor/arch/x86/guest/vlapic.c
@@ -540,7 +540,7 @@ vlapic_set_intr_ready(struct acrn_vlapic *vlapic, uint32_t vector, bool level)
  *
  * @param[in] dest_pcpu_id Target CPU ID.
  *
- * @return void
+ * @return None
  */
 void vlapic_post_intr(uint16_t dest_pcpu_id)
 {
@@ -1363,7 +1363,7 @@ vlapic_pending_intr(const struct acrn_vlapic *vlapic, uint32_t *vecptr)
  * @param[in] vlapic Pointer to target vLAPIC data structure
  * @param[in] vector Target virtual interrupt vector
  *
- * @return void
+ * @return None
  *
  * @pre vlapic != NULL
  */
--- a/hypervisor/arch/x86/io.c
+++ b/hypervisor/arch/x86/io.c
@@ -186,9 +186,9 @@ void emulate_io_post(struct acrn_vcpu *vcpu)
  *
  * @pre io_req->type == REQ_PORTIO
  *
- * @return 0       - Successfully emulated by registered handlers.
- * @return -ENODEV - No proper handler found.
- * @return -EIO    - The request spans multiple devices and cannot be emulated.
+ * @retval 0 Successfully emulated by registered handlers.
+ * @retval -ENODEV No proper handler found.
+ * @retval -EIO The request spans multiple devices and cannot be emulated.
  */
 int32_t
 hv_emulate_pio(const struct acrn_vcpu *vcpu, struct io_request *io_req)
@@ -236,9 +236,9 @@ hv_emulate_pio(const struct acrn_vcpu *vcpu, struct io_request *io_req)
  *
  * @pre io_req->type == REQ_MMIO
  *
- * @return 0       - Successfully emulated by registered handlers.
- * @return -ENODEV - No proper handler found.
- * @return -EIO    - The request spans multiple devices and cannot be emulated.
+ * @retval 0 Successfully emulated by registered handlers.
+ * @retval -ENODEV No proper handler found.
+ * @retval -EIO The request spans multiple devices and cannot be emulated.
  */
 static int32_t
 hv_emulate_mmio(struct acrn_vcpu *vcpu, struct io_request *io_req)
@@ -285,11 +285,11 @@ hv_emulate_mmio(struct acrn_vcpu *vcpu, struct io_request *io_req)
  * @param vcpu The virtual CPU that triggers the MMIO access
  * @param io_req The I/O request holding the details of the MMIO access
  *
- * @return 0       - Successfully emulated by registered handlers.
- * @return IOREQ_PENDING - The I/O request is delivered to VHM.
- * @return -EIO    - The request spans multiple devices and cannot be emulated.
- * @return -EINVAL - \p io_req has an invalid type.
- * @return Negative on other errors during emulation.
+ * @retval 0 Successfully emulated by registered handlers.
+ * @retval IOREQ_PENDING The I/O request is delivered to VHM.
+ * @retval -EIO The request spans multiple devices and cannot be emulated.
+ * @retval -EINVAL \p io_req has an invalid type.
+ * @retval <0 on other errors during emulation.
  */
 int32_t
 emulate_io(struct acrn_vcpu *vcpu, struct io_request *io_req)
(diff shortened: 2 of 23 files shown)


hypervisor/include/arch/x86/mmu.h
```diff
@@ -221,8 +221,8 @@ void destroy_ept(struct acrn_vm *vm);
  * @param[in] vm the pointer that points to VM data structure
  * @param[in] gpa the specified guest-physical address
  *
- * @return INVALID_HPA - the HPA of parameter gpa is unmapping
- * @return hpa - the HPA of parameter gpa is hpa
+ * @retval INVALID_HPA - the HPA of parameter gpa is unmapping
+ * @return the HPA of parameter gpa is hpa
```
gvancuts: Combining `@return` and `@retval` creates two sections in the documentation. It may be OK to have both but I think we should still have one additional `@retval` line that reads something like:
`@retval hpa the host physical address mapping to the \p gpa`

hypervisor/include/arch/x86/mmu.h
```diff
@@ -233,8 +233,8 @@ uint64_t gpa2hpa(struct acrn_vm *vm, uint64_t gpa);
  * @param[out] size the pointer that returns the page size of
  *                  the page in which the gpa is
  *
- * @return INVALID_HPA - the HPA of parameter gpa is unmapping
- * @return hpa - the HPA of parameter gpa is hpa
+ * @retval INVALID_HPA - the HPA of parameter gpa is unmapping
+ * @return the HPA of parameter gpa is hpa
```
gvancuts: Same here (although that function does not appear in the online doc)

hypervisor/include/arch/x86/mmu.h
```diff
@@ -117,15 +117,15 @@ void mmu_modify_or_del(uint64_t *pml4_page, uint64_t vaddr_base, uint64_t size,
...
+ * @retval -ENODEV Don't support EPT or VPID capability
  */
 int check_vmx_mmu_cap(void);
 /**
  * @brief VPID allocation
  *
- * @return 0 - VPID overflow
- * @return more than 0 - the valid VPID
+ * @retval 0 VPID overflow
+ * @retval positive the valid VPID
```
gvancuts: How about changing positive to int? (that keyword is formatted as if it is a parameter)
```suggestion
 * @retval int the valid VPID
```

hypervisor/include/arch/x86/ioreq.h
```diff
@@ -262,11 +262,11 @@ void dm_emulate_mmio_post(struct acrn_vcpu *vcpu);
...
- * @return 0       - Successfully emulated by registered handlers.
- * @return IOREQ_PENDING - The I/O request is delivered to VHM.
- * @return -EIO    - The request spans multiple devices and cannot be emulated.
- * @return -EINVAL - \p io_req has an invalid type.
- * @return Negative on other errors during emulation.
+ * @retval 0       Successfully emulated by registered handlers.
+ * @retval IOREQ_PENDING The I/O request is delivered to VHM.
+ * @retval -EIO    The request spans multiple devices and cannot be emulated.
+ * @retval -EINVAL \p io_req has an invalid type.
+ * @retval Negative on other errors during emulation.
```
gvancuts: How about this?
```suggestion
 * @retval <0 on other errors during emulation.
```
I find this more explicit since that word (`Negative` or `<0` as proposed) is formatted as a parameter

hypervisor/include/arch/x86/guest/vlapic.h
```diff
@@ -240,8 +240,8 @@ int vlapic_set_local_intr(struct acrn_vm *vm, uint16_t vcpu_id_arg, uint32_t vec
  * @param[in] addr MSI address.
  * @param[in] msg  MSI data.
  *
- * @return 0 on success.
- * @return non-zero on error that addr is invalid.
+ * @retval 0 on success.
```
gvancuts:
```suggestion
 * @retval 0 on success, non-zero otherwise (\p addr is invalid)
```

hypervisor/include/arch/x86/guest/vlapic.h
```diff
@@ -240,8 +240,8 @@ int vlapic_set_local_intr(struct acrn_vm *vm, uint16_t vcpu_id_arg, uint32_t vec
...
+ * @retval non-zero on error that addr is invalid.
```
gvancuts:
```suggestion
```

hypervisor/arch/x86/io.c
```diff
@@ -285,11 +285,11 @@ hv_emulate_mmio(struct acrn_vcpu *vcpu, struct io_request *io_req)
...
- * @return 0       - Successfully emulated by registered handlers.
- * @return IOREQ_PENDING - The I/O request is delivered to VHM.
- * @return -EIO    - The request spans multiple devices and cannot be emulated.
- * @return -EINVAL - \p io_req has an invalid type.
- * @return Negative on other errors during emulation.
+ * @retval 0 Successfully emulated by registered handlers.
+ * @retval IOREQ_PENDING The I/O request is delivered to VHM.
+ * @retval -EIO The request spans multiple devices and cannot be emulated.
+ * @retval -EINVAL \p io_req has an invalid type.
+ * @retval negative on other errors during emulation.
```
gvancuts:
```suggestion
 * @retval <0 on other errors during emulation.
```

hypervisor/arch/x86/guest/vlapic.c
```diff
@@ -1997,8 +1997,8 @@ vlapic_set_local_intr(struct acrn_vm *vm, uint16_t vcpu_id_arg, uint32_t vector)
  * @param[in] addr MSI address.
  * @param[in] msg  MSI data.
  *
- * @return 0 on success.
- * @return non-zero on error that addr is invalid.
+ * @retval 0 on success.
```
gvancuts:
```suggestion
 * @retval 0 on success, non-zero otherwise (\p addr is invalid)
```

hypervisor/arch/x86/guest/vlapic.c
```diff
@@ -1997,8 +1997,8 @@ vlapic_set_local_intr(struct acrn_vm *vm, uint16_t vcpu_id_arg, uint32_t vector)
...
+ * @retval non-zero on error that addr is invalid.
```
gvancuts:
```suggestion
```

hypervisor/include/arch/x86/vtd.h
```diff
@@ -531,7 +531,7 @@ int unassign_iommu_device(const struct iommu_domain *domain, uint8_t bus, uint8_
  * @param[in] addr_width address width of the VM
  *
  * @return Pointer pointer to iommu_domain
- * @return NULL if translation_table is 0
+ * @retval NULL if translation_table is 0
```
gvancuts: This creates two distinct sections (called "return" and "return value") in the HTML doc, but not all possible return values are listed in `retval` which is a bit confusing.

hypervisor/include/arch/x86/vtd.h
```diff
@@ -588,8 +588,8 @@ void resume_iommu(void);
  *
  * Register DMAR units on the platform according to the pre-parsed information or DMAR table.
  *
- * @return 0 - on success
- * @return non-zero - iommu is a must have feature, if init_iommu failed, the system should not continue booting
+ * @retval 0 on success
```
gvancuts:
```suggestion
 * @retval 0 on success, non-zero otherwise. The IOMMU is a must-have feature, if init_iommu failed, the system should not continue booting
```

hypervisor/include/arch/x86/vtd.h
```diff
@@ -588,8 +588,8 @@ void resume_iommu(void);
...
+ * @retval non-zero iommu is a must have feature, if init_iommu failed, the system should not continue booting
```
gvancuts:
```suggestion
```

hypervisor/include/arch/x86/mmu.h
junjiemao1: Thanks for pointing this out! The `@return` seems redundant after adding the additional `@retval`. I added the `@retval` and dropped the `@return`. Does that look good to you?

hypervisor/include/arch/x86/mmu.h
junjiemao1: Updated similarly.

hypervisor/include/arch/x86/mmu.h
junjiemao1: `int` is broader than `positive` and may be confusing because the return value has type uint16_t while `int` can be negative. How about using `>0` here, like replacing `negative` with `<0` below?

hypervisor/include/arch/x86/ioreq.h
junjiemao1: Updated. Thanks.

hypervisor/include/arch/x86/guest/vlapic.h
junjiemao1: Combining different return values in a single `@retval` directive makes the return value list incomplete at first glance. How about replacing `non-zero` with `<0` (the return value is always negative on error actually) to make to more explicit?

hypervisor/arch/x86/io.c
junjiemao1: Updated. Thanks.

hypervisor/include/arch/x86/vtd.h
junjiemao1: I added another `@retval` to cover the non-NULL case. Does that look good to you?

hypervisor/include/arch/x86/vtd.h
junjiemao1: Having the success and failure cases separated as two `@retval` seems more explicit to me. How about updating the `non-zero` with `<0`, like we are doing above?

hypervisor/include/arch/x86/vtd.h
```diff
@@ -530,8 +530,10 @@ int unassign_iommu_device(const struct iommu_domain *domain, uint8_t bus, uint8_
  * @param[in] translation_table the physcial address of EPT table of the VM specified by the vm_id
  * @param[in] addr_width address width of the VM
  *
- * @return Pointer pointer to iommu_domain
- * @return NULL if translation_table is 0
+ * @return Pointer to the created iommu_domain
+ *
+ * @retval NULL when \p translation_table is 0
+ * @retval Non-NULL when \p translation_table is not 0
```
gvancuts: Could we use `<0` here instead of `Non-NULL` (assuming the return value is always negative)?

(the automatic formatting will make whatever follows `retval` formatted like a parameter, or code so this is why I prefer to avoid plain English here)

gvancuts:
> @gvancuts Thanks for the comments! I've updated the missed `@return N/A`.

Thanks!

> As for the duplicated comments, they are preserved intentionally. We would like to use the comments in source files to generate the interface specs in module design in the future, which is required to achieve functional safety. Without these comments, it would be very tricky to generate such doc.

Fine by me!

hypervisor/include/arch/x86/vtd.h
junjiemao1: `<0` may be confusing here because the return  value is a pointer, not an integer.

hypervisor/include/common/irq.h
```diff
@@ -50,8 +50,8 @@ struct irq_desc {
  *			IRQF_LEVEL - 1: level trigger; 0: edge trigger;
  *    			IRQF_PT    - 1: for passthrough dev
  *
- * @return valid irq num - on success
- * @return IRQ_INVALID - on failure
+ * @retval non-negative on success
```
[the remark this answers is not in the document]
junjiemao1: Updated. Thanks.

junjiemao1:
@gvancuts Would you please reapprove this PR as some minor tweaks are made since your last approval? Thanks!

Issue #1882, opened 2018-11-23
Extend the SOS CMA range from 64M to 128M


fyin1:
In case too many DMA allocation from SOS.

fyin1:
[External_System_ID] ACRN-1907

mwang106:
We choose to move the trusty memory allocation from SOS kernel to hypervisor. So we don't need to this work any more. Won't merge to master branch

PR #1883, opened 2018-11-23
SOS: extend the size of SOS CMA from 64M to 128M

fyin1:
We observed some dma allocation in SOS could failed because no
enough memory available. Let's extend the SOS CMA size a little bit

Tracked-On: #1882


HuiHuangShi:
1883 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

ClaudZhang1995:
Ready to merge

gvancuts:
We don't have such `cma` option by default when using an UEFI platform, should we add one?

fyin1:
I don't need it on UEFI platform for now. Because it's used for trusty which is not enabled on UEFI platform yet.

gvancuts:
> I don't need it on UEFI platform for now. Because it's used for trusty which is not enabled on UEFI platform yet.

Thanks for the confirmation!

PR #1884, opened 2018-11-23
Fix some USB virtualization issues

xiaoguangwu:
1. Losing USB device(s) after resuming from S3;
2. Fail to enumerate USB device after many times of suspending and resuming;
3. Long dalay (~20 seconds) for USB headset enumeration.


HuiHuangShi:
1884 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

Issue #1868, opened 2018-11-21
acrn-dm: monitor: should wait for thread cancel finish.


fyin1:
We need to wait for monitor thread cancel finish. And then continue other cleanup work.

fyin1:
[External_System_ID] ACRN-1907

PR #1889, opened 2018-11-23
acrn-dm: wait for monitor thread canceling finish

fyin1:
When thread_cancel() is called to exit pthread, we can't know
when the thread exit done unless pthread_join() is used to
wait for thread exit.

Tracked-On: #1868


ClaudZhang1995:
Ready to merge

PR #1949, opened 2018-11-30
hv: fix type conversion violations

shiqingg:
This patch fixes the following violations:
 1. Implicit conversion: actual to formal param
 2. Value is not of appropriate type
 3. No cast for widening complex int expression
 4. Widening cast on complex integer expression
 5. Narrower int conversion without cast.

Tracked-On: #861


ClaudZhang1995:
Ready to merge

Issue #1975, opened 2018-12-05
Minor refinement and enhancement about acrntrace/log


KaigeFu:
1. In current implementation, those tools will run without an error thrown out when no devices found. It is a little bit confusing.
2. We use get_cpu_num to get devices count. It is more suitable to rename it to get_dev_cnt.

Mingyuan18:
[External_System_ID] ACRN-2018

Mingyuan18:
It's fixed by using below build: 
ACRN-Hypervisor :  2018w51.2.140000p_134
SOS : 26830

PR #1976, opened 2018-12-05
Refine acrntrace acrnlog

KaigeFu:
If module acrn_trace is configured as "m" and we don't insert it before
launching acrntrace tool, acrntrace will run normally instead of throw
out an error in current implementation. And no trace files are generated.
It will cause confusing here.

So, this patch throw out an error massage to user about the absence of acrn
trace devices.

BTW, this patch replaces cpuid/num related concept to devid/cnt which is more
suitable for /dev/acrn_trace_xxx.

Tracked-On: #1975

### DIFF
--- a/tools/acrnlog/acrnlog.c
+++ b/tools/acrnlog/acrnlog.c
@@ -19,13 +19,12 @@
 
 #define LOG_ELEMENT_SIZE        80
 #define LOG_MSG_SIZE		480
-#define PCPU_NUM		4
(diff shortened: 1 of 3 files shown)


HuiHuangShi:
1976 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

HuiHuangShi:
1976 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

Issue #1979, opened 2018-12-05
add 'no-omit-frame-pointer' in debug version


mingqiangchi:
Hypervisor uses '-O2' compiler option,
it will omit frame pointer by default for '-O2',
need to add 'no-omit-frame-pointer' in debug version.

Mingyuan18:
[External_System_ID] ACRN-2026

Mingyuan18:
No regression issue found in test. So close it.

PR #1980, opened 2018-12-05
hv: add 'no-omit-frame-pointer' in debug version

mingqiangchi:
Hypervisor uses '-O2' compiler option,
it will omit frame pointer by default for '-O2',
This patch add 'no-omit-frame-pointer' in debug version.

Tracked-On: #1979


HuiHuangShi:
1980 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

Issue #2013, opened 2018-12-09
[MRB][HV][SOS] Acrnctl add vm failed with absolute path


taoyuhong:
There is a bug to run 'acrnctl add /[path]/script.sh', when the launch script has an absolute path. Acrnctl will generate wrong path for temp files and fail to add VM.
[External_System_ID] ACRN-1998

PR #2014, opened 2018-12-09
tools: acrnctl: Fix path error when run "acrnctl add" cmd

taoyuhong:
There is a bug to run 'acrnctl add /[path]/script.sh', when the
launch script has an absolute path. Acrnctl will generate wrong path
for temp files and fail to add VM.

And message '/opt/acrn/conf: No such file or directory' always comes
out, until user once successfully run 'acrnctl add' cmd. That is reported
by _scan_added_vm(), because 'opt/acrn' is missing, only 'acrnctl add'
can create it, we should also check it in _scan_added_vm().

Tracked-On: #2013

### DIFF
--- a/tools/acrn-manager/acrn_vm_ops.c
+++ b/tools/acrn-manager/acrn_vm_ops.c
@@ -223,6 +223,9 @@ static void _scan_added_vm(void)
 	char suffix[128];
 	int ret;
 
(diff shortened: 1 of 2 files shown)


HuiHuangShi:
2014 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

Issue #1935, opened 2018-11-29
VirtIO devices can get junk bits on BAR writes


peterfang:
OVMF's virtio drivers can write junk bits to a device during I/O instruction emulation because the BAR write function does not truncate RAX data.

junjunshan1:
[External System ID] 259022

Mingyuan18:
Patch has been merged for a long time, so close it.

PR #2051, opened 2018-12-13
dm: enforce data size when accessing PCI BARs

peterfang:
Always enforce data size when using BAR access functions.

Currently, the size serves as a hint to the BAR access functions and
these functions are expected to behave accordingly. Some of the access
functions, e.g. virtio ones, don't always truncate the data but expect
the caller to take care of the data size. This causes problems with
OVMF's virtio drivers during I/O instruction emulation because RAX can
contain junk bits that shouldn't be written to the device.

v1 -> v2:
- improve readability

Tracked-On: #1935

### DIFF
--- a/devicemodel/hw/pci/core.c
+++ b/devicemodel/hw/pci/core.c
@@ -355,6 +355,17 @@ pci_msix_pba_bar(struct pci_vdev *dev)
 		return -1;
 }
 
+static inline uint64_t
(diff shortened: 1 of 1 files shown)


HuiHuangShi:
2051 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

Issue #1893, opened 2018-11-23
USB virtualization issues


mwang106:
Losing USB device(s) after resuming from S3;
Fail to enumerate USB device after many times of suspending and resuming;
Long dalay (~20 seconds) for USB headset enumeration.

Mingyuan18:
[External_System_ID] ACRN-2014

Mingyuan18:
[External_System_ID] ACRN-2284

PR #2069, opened 2018-12-14
refine USB mediator S3 process

xiaoguangwu:
The USB VBus will be cut off when SOS enters S3 state, hence the USB drivers
will need several seconds to be ready during the SOS resuming. Before SOS USB
drivers are ready, USB mediator delays UOS resuming for 5 seconds to wait. This
will delay the whole S3 resuming time.

This patch set is used to fix this issue. The thought is let UOS resuming as
quickly as possible and delay the USB emulation. So the UOS will turn on the
screen early and it could not USB devices for several seconds. The test shows,
~2 seconds delay for only one USB device attached, ~5 seconds delay for five
USB devices attached.

Tracked-On: #1893
Signed-off-by: Xiaoguang Wu <xiaoguang.wu@intel.com>

### DIFF
--- a/devicemodel/core/main.c
+++ b/devicemodel/core/main.c
@@ -66,8 +66,6 @@
 #include "vmcfg.h"
 #include "tpm.h"
 #include "virtio.h"
-#include "usb.h"
-#include "xhci.h"
 
 #define GUEST_NIO_PORT		0x488	/* guest upcalls via i/o port */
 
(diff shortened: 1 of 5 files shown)


HuiHuangShi:
2069 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

PR #2102, opened 2018-12-18
hv: vpci: fix MISRA-C violations related to variable declarations

ZideChen0:
78D: Global variable should be declared const.

Global variables should be declared constant wherever possible to
avoid unintentional modification.

27D: Variable should be declared static
  pci_ops_vdev_msi is not accessed by other files. Remove the declaration
  from the header and define it with the static qualifier; Because it's
  referenced by populate_msi_struct(), so move the define statements forward.

33D: No real declaration for external variable
  certain variables are available in sharing mode or partition mode only,
  so that the declarations in header files must be enclosed with
  CONFIG_PARTITION_MODE

Tracked-On: #861


HuiHuangShi:
2102<br>No new violations to the coding guideline detected.

Issue #1995, opened 2018-12-06
Support SBL firmware as bootloader on Apollo Lake UP2


tw4452852:
Add UP2 with SBL support including relative configs.

totoroow:
[External_System_ID] ACRN-2191

mwang106:
One binary for up2 SBL tested ok with 27050 SOS version

PR #2228, opened 2019-01-03
config: fix no serial output with SBL on UP2

totoroow:
SBL doesn't support 40-pin header for now, so the serial PCI
device(0:18:1) on it doesn't work. Let's just workaround this by
changing hv's default serial output device to 0:18:0. In the future,
we'd rather change it back if SBL supports it.

This change is only for SBL, for UEFI, now we offer serial configuration
through command line parameter(vuart=...).

Tracked-On: #1995


wenlingz:
@dongyaozu please help approve it,thanks

Issue #2157, opened 2018-12-21
Profiling: fix the profiling tool crash by page faults


mlim19:
Profiling tools are broken, which cause page faults during collection.
The issue happens by enabling SMAP recently. Therefore, 
we use stac() and clac() to allow access to buffers allocated by guest vm.

Mingyuan18:
[External_System_ID] ACRN-2344

Mingyuan18:
@mlim19  please verify this issue, thanks!

mlim19:
Successfully verified.

PR #2264, opened 2019-01-09
Fix crash by SMAP and align data types

mlim19:
This PR includes two patches:
1. Fix crash by SMAP when performing profilings.
2. Align data types between hypervisor core and profiling to avoid unnecessary type conversions. 

Tracked-On: #2157


acrnsi:
Ok to verify

Issue #2282, opened 2019-01-10
[IO Mediator][AaaG]IOC mediator sometimes loses packets sent to UOS, including wakeup reason and signal data.[FR: 5%](apl_sdc_stable)


lidongwangx:
IOC mediator sometimes loses packets sent to UOS, including wakeup reason and signal data.[FR: 5%]

Mingyuan18:
[External_System_ID] ACRN-2427

Mingyuan18:
No regression issue found in test.

PR #2286, opened 2019-01-10
IOC mediator: fix the issue of abnormally missing tx package

liuyuan-GitHub:
The issue will result in the loss of wakeup reason and signal
data, because the CBC header was not added when generating the
tx package.

Tracked-On: #2282


Mingyuan18:
@liuyuan-GitHub , please add tracked on link in conversation and commits.

Issue #2301, opened 2019-01-11
S3 resume failed due to ioc not sending wakeup reason to UOS


Mingyuan18:
Environment:
AaaG build: 83590
HV patch:
0001-dm-refine-the-uart_core.patch
0001-sos-kernel-cmdline-add-option-to-enable-panic-dump.patch
0002-dm-Add-teardown-callback-for-mevent-in-uart_core.patch
Kernel patch:
0001-vhm-remove-flooding-log-from-vhm.patch
0002-acrn-kernel-config-Enable-xhci-and-igb-driver-as-bui.patch
0003-sdhci-fix-the-fake-timeout-bug.patch
0008-panic-add-options-to-print-system-info-when-panic-ha.patch
Bug detailed description:
In the stability test "S3 suspend/resume",failed to find "Set ram refresh action to 1" in tty USB2 console after the 24th suspend.
Reproduce steps:

boot up system
Removed IPU option in launch_uos.sh script, and then open another ssh section to SOS, input: “acrnctl add ./launch_uos.sh –V 2” (note: -V 1 for LaaG, 2 for AaaG), you will get output as below:
“vm1-xx:yy:zz added”
Note: this only need run once, next time you might get “vm1 … already exist …”,
Input "acrnctl list” to confirm VM name(i.e. vm1-xx:yy:zz);
Run “systemctl enable acrnd” command in the SOS to boot UOS.
Open another ssh section to SOS
Press the ignition button for 2s, then release button and wait for 20s in order to suspend SOS and UOS.
Connect SOS IP by ssh, expected result: failed to connect SOS IP
Detect Android guest by adb devices,expected result: failed to detect devices
Press the ignition button for 2s, then release button and wait for 20s in order to resume SOS and UOS.
Connect SOS IP by ssh, expected result: succeed to connect SOS IP
11.Detect Android guest ,expected result: succeed to detect devices
Return to step 6
Current result:
In the stability test "S3 suspend/resume",failed to find "Set ram refresh action to 1" in ttyUSB2 console after the 24th suspend.
After one night past, 10 hours past, still cannot find "Set ram refresh action to 1" in ttyUSB2 console.
Failed rate: 1/24
Expected result:
In the stability test "S3 suspend/resume",can find "Set ram refresh action to 1" in ttyUSB2 console after every time suspend.

Mingyuan18:
[External_System_ID] ACRN-2442

Mingyuan18:
[...]

PR #2302, opened 2019-01-11
DM: mei: solving the issue of closing the fd

liuyuan-GitHub-zz:
The issue leads that the IOC mediator's fd is turned off
incorrectly, causing the IOC mediator to fail.

The root cause is because after closing fd is not set
to an invalid fd.

The issue can be reproduced by the S3 stress test.

Tracked-On: #2301


HuiHuangShi:
2302 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

Issue #2328, opened 2019-01-14
Wifi Firmware initialization failed happened on Reboot test.(apl_sdc_stable)


Mingyuan18:
Wifi Firmware initialization failed happened on Reboot test

Mingyuan18:
[External_Sytem_ID] ACRN-2461

Mingyuan18:
Fix this issue in below build:
ACRN-Hypervisor: 2019w03.5.150000p_148
SOS : 27215

PR #2329, opened 2019-01-14
dm: passthru: reset wifi chip via gpio

binbinwu1:
During UOS reboot, sometime the wifi firmware status is wrong and can't
handled by UOS driver, this patch add gpio reset for the wifi chip in dm to avoid
such error.

Tracked-On: #2328 

### DIFF
--- a/devicemodel/hw/pci/passthrough.c
+++ b/devicemodel/hw/pci/passthrough.c
@@ -39,6 +39,7 @@
 #include <pciaccess.h>
 #include <fcntl.h>
 #include <unistd.h>
+#include <dm_string.h>
 
 #include "iodev.h"
 #include "vmmapi.h"
@@ -113,6 +114,7 @@ struct passthru_dev {
 	 *   need_reset - reset dev before passthrough
 	 */
 	bool need_reset;
+	int gpio_reset_pin;
 	/* The memory pages, which not overlap with MSI-X table will be
 	 * passed-through to guest, two potential ranges, before and after MSI-X
 	 * Table if any.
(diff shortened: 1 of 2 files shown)


binbinwu1:
> how can we passthru GPIO pin? Do you mean map a physical GPIO to a guest GPIO?
> 
> If this is the case, let us use a well-defined name,

Here passthru means the patch is in passthru driver in DM, not mean to passthrough gpio pin to UOS.
GPIO pin is accessed by device model after it exported to usersapce.

fyin1:
> 
> 
> how can we passthru GPIO pin? Do you mean map a physical GPIO to a guest GPIO?
> 
> If this is the case, let us use a well-defined name,
Eddie, This is not GPIO passthru. This is kind of workaround. We do GPIO reset operation in SOS side.  But this GPIO is for passthru device reset.

The correct fixing should passthru the GPIO to guest and let guest to handle it.

dongyaozu:
Let us use different name of "verb". Let "passthru" to be for PCIe passthru only. Then, USB port assignment/GPIO assignment etc.... should use a new name. How about to use present or route?  @binbinwu1 @fyin1 @hwang37

binbinwu1:
> Let us use different name of "verb". Let "passthru" to be for PCIe passthru only. Then, USB port assignment/GPIO assignment etc.... should use a new name. How about to use present or route? @binbinwu1 @fyin1 @hwang37

@dongyaozu this patch doesn't present or route gpio pin to UOS, just toggle the gpio pin  in device model to reset wifi device. Here "passthru" indicates the modification is in pci passthru driver.

fyin1:
> 
> 
> > Let us use different name of "verb". Let "passthru" to be for PCIe passthru only. Then, USB port assignment/GPIO assignment etc.... should use a new name. How about to use present or route? @binbinwu1 @fyin1 @hwang37
> 
> @dongyaozu this patch doesn't present or route gpio pin to UOS, just toggle the gpio pin in device model to reset wifi device. Here "passthru" indicates the modification is in pci passthru driver.

BTW, Binbin, my understanding is that this is a workaround instead of formal fixing. Right?

Issue #2316, opened 2019-01-11
Tools don't respect CFLAGS/LDFLAGS from environment


rossburton:
The build environment might want to pass extra CFLAGS or LDFLAGS to the build of the acrn tools.  With conventional build systems like automake, Meson, etc this is possible by just setting CFLAGS and LDFLAGS in the environment.

However as the tools are built using bare Makefiles, these environment variables are overwritten.

Mingyuan18:
[External_System_ID] ACRN-2474

Mingyuan18:
@rossburton Please verify this issues, close it if the result is pass. Thanks!

PR #2357, opened 2019-01-16
tools: respect CFLAGS and LDFLAGS from environment

rossburton:
The build environment might want to pass extra CFLAGS or LDFLAGS to the build of
the acrn tools. With conventional build systems like automake, Meson, etc this
is possible by just setting CFLAGS and LDFLAGS in the environment.

However as the tools are built using bare Makefiles, these environment variables
are overwritten.  Respect them by renaming the variables in the Makefiles to
e.g. LOG_CFLAGS and adding CFLAGS to that.

Tracked-On: #2316
Signed-off-by: Ross Burton <ross.burton@intel.com>

### DIFF
--- a/tools/acrn-manager/Makefile
+++ b/tools/acrn-manager/Makefile
@@ -2,21 +2,22 @@ T := $(CURDIR)
 OUT_DIR ?= $(shell mkdir -p $(T)/build;cd $(T)/build;pwd)
 CC ?= gcc
 
-CFLAGS := -g -O0 -std=gnu11
-CFLAGS += -D_GNU_SOURCE
-CFLAGS += -DNO_OPENSSL
-CFLAGS += -m64
-CFLAGS += -Wall -ffunction-sections
-CFLAGS += -Werror
-CFLAGS += -O2 -D_FORTIFY_SOURCE=2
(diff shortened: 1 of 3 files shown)


NanlinXie:

ClaudZhang1995:
Ok to verify

dongyaozu:
@binbinwu1

binbinwu1:
I am not sure the change is what we want.
If we want the tools compiled in a specific way, I think it is fine to override the default CFLAGS/LDFLAGS
@gvancuts  @lyan3 What do you think?

rossburton:
FWIW this is precisely what automake for example does.  The makefiles use AM_CFLAGS, leaving CFLAGS empty for the builder to override.

gvancuts:
> FWIW this is precisely what automake for example does. The makefiles use AM_CFLAGS, leaving CFLAGS empty for the builder to override.

binbinwu1:
Reviewed-by: Binbinwu <binbin.wu@intel.com>

wenlingz:
@rossburton please modify the "Tracked-on" to "Tracked-On" on  your commit, the conversation we have modified.

rossburton:
Rebased to master and fixed capitalisation.

Issue #2365, opened 2019-01-17
Aaag crash during run interruption storm test


mwang106:
1.Environment
--------------------------------------------------------------
[Board]: D0 MRB
HW: MRB D0 Premium
IOC: 4. 0. 12
ABL: 1846_GP20
Debug Image: build image with EBtools(tag: acrn-2018w50.1-140000p)
git clone https://github.intel.com/projectacrn/acrn-ebtool
git clone https://github.com/projectacrn/acrn-hypervisor.git
git clone https://github.com/projectacrn/acrn-kernel.git
SOS kernel: 4.19.0-quilt-2e5dc0ac-00001-gcc2304c338d1

2. Reproduce Steps
--------------------------------------------------------------
2.1 reboot into SOS
2.2 modify threshold/s from 10000 to 1000 in launch_uos.sh
intr_storm_monitor="--intr_monitor 1000,10,1,100"
2.3 and passthru eth to aaag with modification in launch_uos.sh
echo "8086 1533" > /sys/bus/pci/drivers/pci-stub/new_id
echo "0000:02:00.0" > /sys/bus/pci/devices/0000:02:00.0/driver/unbind
echo "0000:02:00.0" > /sys/bus/pci/drivers/pci-stub/bind
-s 9,passthru,02/00/0 \
2.4 then "acrnd &" to boot AAAG, run netserver 
./netserver
2.5 on host to check the interrupts per 1s
watch -n 1 adb shell "cat /proc/interrupts | grep eth0"

3. Expected result:
--------------------------------------------------------------
interruption storm is triggered, the interruption speed is slowed down
And system should work normally.

4. Current result:
--------------------------------------------------------------
The interruption can reach 20000+ /s
The aaag crash durring the test.

5. Fail rate: 
---------------------------------------------------------------
3/5

mwang106:
[External_System_ID] ACRN-2380

PR #2375, opened 2019-01-18
HV: check to avoid interrupt delay timer add twice

mcao6:
to edge interrupt, like eth device, it can triger the interrupt again
when its IRQ in softirq entry queue or in timer list.

in current design, for sofrirq entry, it calls "list_del" before
"list_add_tail", to avoid the entry added twice.

so for interrupt delay timer, add to check if it is started
then just drop the next one; to avoid it enqueue twice.

Tracked-On: #2365


HuiHuangShi:
2375<br>No new violations to the coding guideline detected.

Issue #2410, opened 2019-01-24
Launch UOS will occur page fault error when use the hypervisor build on Ubuntu


mwang106:
Steps:
1. Build Acrn Hypervisor on Ubuntu OS
2. Flash images
3. Launch SOS & UOS

Actual result:
Launch UOS will occur page fault error

mwang106:
[External_System_ID] ACRN-2573

PR #2413, opened 2019-01-24
fix schedule issue when hypervisor built with ubuntu

JasonChenCJ:
after compile, the compiled code could change rsp, so use pure asm code
 to avoid such problem which will cause schedule switch failure.

Tracked-On: #2410


hypervisor/arch/x86/sched.S
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+    .text
+
+   .code64
+   .align       8
+   .global      arch_switch_to
+arch_switch_to:
```
dongyaozu: Add an description for the API.
BTW, how does KVM do? I remember it uses inline ASM too.

HuiHuangShi:
2413<br>No new violations to the coding guideline detected.

hypervisor/arch/x86/sched.S
JasonChenCJ: Will add the description. For KVM, it actually is using Linux kernel process scheduler which using task structure memory to save/restore context instead of stack.

Issue #2422, opened 2019-01-25
[PATCH] profiling: fix the system freeze issue when running profiling  tool


mlim19:
The msr load/restore during vmexit/vmenty in HV by default. The profiling has assumption that it's only user for this feature, which could overwrite of HV default setting.

Mingyuan18:
[External_System_ID] ACRN-2605

Mingyuan18:
@mlim19 , Please help to check this issue is resolved or not.  If verify pass, please close it. Thanks!

PR #2423, opened 2019-01-25
profiling: fix the system freeze issue when running profiling

mlim19:
tool

The msr load/restore during vmexit/vmenty is enabled in HV by
default. The profiling has assumption that it's only user for this
feature, which could overwrite of HV default setting.

This fix combines the msr load list for vmexit when profiling.

Tracked-On: #2422


wenlingz:
OK to verify

fyin1:
Please hold on this patch merging. We could merge this one after Beta release. Thanks.

wenlingz:
@fyin1 OK, got it

fyin1:
We could merge this patch now. Suppose the Beta was cut.

PR #2428, opened 2019-01-25
doc: update coding guidelines

shiqingg:
- update and add some rules in coding guidelines

Tracked-On: #861

### DIFF
--- a/doc/developer-guides/coding_guidelines.rst
+++ b/doc/developer-guides/coding_guidelines.rst
@@ -660,6 +660,41 @@ Compliant example::
        }
 
 
+FN-10: The formal parameter type of a function shall be consistent
+==================================================================
+
+The formal parameter type of a function shall be the same between its
+declaration and definition.
+
+Compliant example::
+
+    /* In `showcase.h`: */
+    uint32_t func_showcase(uint32_t param);
+    
+    /* In `showcase.c`: */
+    #include "showcase.h"
+    
+    uint32_t func_showcase(uint32_t param)
+    {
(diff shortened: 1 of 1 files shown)


doc/developer-guides/coding_guidelines.rst
```diff
@@ -1099,6 +1134,27 @@ Compliant example::
        showcase_test = showcase++;
 
 
+ST-19: Array indexing shall be valid
```
dbkinder: Better might be:  ``Array indexing shall be in-bounds``

doc/developer-guides/coding_guidelines.rst
```diff
@@ -1099,6 +1134,27 @@ Compliant example::
...
+====================================
+
+When indexing is performed on an array, the index value shall be smaller than
```
dbkinder: A bit clearer would be:
```
An array index value shall be between zero (for the first element)
and the array size minus one (for the last element).  Out-of-bound 
array references are an undefined behavior and must be avoided.
```

doc/developer-guides/coding_guidelines.rst
```diff
@@ -1696,6 +1752,40 @@ Compliant example::
        char chr = showcase[1];
 
 
+TY-20: The actual parameter type shall be the same with the formal parameter type
```
dbkinder: change ``same with`` to ``same as``

shiqingg:
@dbkinder Thanks for your suggestion David! New version has been pushed.

Issue #2435, opened 2019-01-28
meta issue to track enabling ACRN on a new platform


totoroow:
[External_System_ID] ACRN-1977

Mingyuan18:
This is not a issue . So close it.

PR #2436, opened 2019-01-28
Fix two issues found on a new platform

totoroow:
EFI: fix potential memory overwrite due to mmap table
Some bios may have more mmap table entry than our current limitation
which is 128. This will lead to a memory overwrite, so add a check to
prevent this and enlarge the limitation to 256. This should fix most
bioses.

Kconfig: enlarge range of maximum number of IOMMU  …
Some arches have more than two IOMMUs, so change this limitation.
Tracked-On: #2435


HuiHuangShi:
2436 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

PR #2438, opened 2019-01-28
doc: change term of vm0 to sos_vm

jsun26intel:
Using term of VM0 would mislead to a VM with VM id 0 easily, whereas
VM id 0 could be used for any PRE_LAUNCHED_VM. So replace VM0 with
SOS_VM.

### DIFF
--- a/doc/developer-guides/hld/hv-memmgt.rst
+++ b/doc/developer-guides/hld/hv-memmgt.rst
@@ -516,7 +516,7 @@ Host to Guest Mapping
 =====================
 
 ACRN hypervisor creates Service OS's host (HPA) to guest (GPA) mapping
-(EPT mapping) through the function ``prepare_vm0_memmap_and_e820()``
+(EPT mapping) through the function ``prepare_sos_vm_memmap()``
 when it creates the SOS VM. It follows these rules:
 
 -  Identical mapping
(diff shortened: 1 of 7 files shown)


doc/developer-guides/hld/hv-memmgt.rst
```diff
@@ -516,7 +516,7 @@ Host to Guest Mapping
 =====================
 
 ACRN hypervisor creates Service OS's host (HPA) to guest (GPA) mapping
-(EPT mapping) through the function ``prepare_vm0_memmap_and_e820()``
+(EPT mapping) through the function ``prepare_sos_vm_memmap_and_e820()``
```
dbkinder: I can't find a function with this name, but there in one called ``prepare_sos_vm_memmap``

jsun26intel: OK, I will update.

jsun26intel:
Hi David, do you know who has the backup of these images? For me, it is too tough to modify them.

dbkinder:
@jsun26intel Alas, no.  All these images came from word documents that are stored on the google drive.  If you're lucky, the image is embedded and can be extracted to edit.

jsun26intel:
David,
Very lucky, I double checked every images that there is no SOS VM was referred to as VM0 only, some exception is safe because it marked as "Service OS" at the same time. So the patch should be safe.
Please approve it, thanks a lot!

BR,
Victor

PR #2464, opened 2019-01-31
hv: fix host call stack dump issue

totoroow:
As scheduler use its own stack now and let's take this into account when unwinding host call stack.


HuiHuangShi:
2464 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

PR #2465, opened 2019-01-31
efi-stub: reset all APs after entering guest mode

totoroow:
In current hv implementation, we assume all AP have no context before
jumping into guest mode. But this is not true in all UEFI bios. BIOS
could have enabled all (or some of) APs at first. These APs could stay
in a run loop or wait for a semaphore. But after hv takes over control from
efi-stub, all of these AP environments will be simply dropped because
we don't support AP context save/restore for now. As a result,
BSP's ExitBootService will hang forver because it's waiting for AP in its
way (by waiting for a semaphore for example), unfortunately APs are now
in the context that hv provides in which they usually stay in idle loop.

To fix the issue above, we could have two solutions:
1. Save AP's runtime context before entering hv and restore context
after hv jumps back.
2. After hv jumps back, reset all the APs in the UEFI way, so the
previous context will be thrown away and a fresh new starts. Moreover
this new one is under virtualization.

Currently, we adopt the second one by disabling all the APs before
virtualization and then enabling them after hv jumps back. A reset
will be triggered. And this is guaranteed by UEFI MP Service protocol.

Tracked-On: #2435

### DIFF
--- a/efi-stub/MpService.h
+++ b/efi-stub/MpService.h
@@ -1,0 +1,633 @@
+/** @file
+  When installed, the MP Services Protocol produces a collection of services
+  that are needed for MP management.
+
+  The MP Services Protocol provides a generalized way of performing following tasks:
+    - Retrieving information of multi-processor environment and MP-related status of
+      specific processors.
+    - Dispatching user-provided function to APs.
+    - Maintain MP-related processor status.
+
+  The MP Services Protocol must be produced on any system with more than one logical
+  processor.
+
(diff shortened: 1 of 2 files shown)


HuiHuangShi:
2465 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

JasonChenCJ:
@totoroow , seemed the patch name is not very good, a better name?

PR #2467, opened 2019-01-31
Need to delete '# CONFIG_BLK_DEV_NVME is not set' to enable NVME driver

lijianpi:
Need to delete '# CONFIG_BLK_DEV_NVME is not set' to enable NVME driver


NanlinXie:

Issue #2474, opened 2019-01-31
Need to capture dropped sample info while profiling


mlim19:
Since the profiling may experience dropped samples due to the limited size of buffer, capturing the information is required to understand the reliability of the data collected by the profiling tool.

Mingyuan18:
[External_System_ID] ACRN-2636

Mingyuan18:
@mlim19 , Could you please verify this issue? If it cannot be reproduced, please close this issue, thank you!

mlim19:
Verified.

PR #2475, opened 2019-01-31
profiling: enable to capture dropped samples while buffering

mlim19:
Since the profiling utilizes the limited size of buffer to capture
sample data, dropping samples could happen while collecting data
if data is generated faster than flushing by consumer. Capturing
the dropped sample info is critical to understand how much the data
is reliable to use.

To capture the information, the new hypercall "PROFILING_GET_STATUS"
is introduced.

Tracked-On: #2474
Signed-off-by: Manisha Chinthapally <manisha.chinthapally@intel.com>


HuiHuangShi:
2475 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

PR #2492, opened 2019-02-01
Porting the patch from master branch

mlim19:
The previous patch was revised base on Eddie Dong's recommendation. The diff needs to ported to the stable branch to make both branches have the same change.
Tracked-On: #2422


wenlingz:
@fyin1 , please help approve it, thanks.

PR #2505, opened 2019-02-03
doc: add more rules in coding guidelines

shiqingg:
This patch adds more rules in coding guidelines.

Tracked-On: #861

### DIFF
--- a/doc/developer-guides/coding_guidelines.rst
+++ b/doc/developer-guides/coding_guidelines.rst
@@ -24,7 +24,7 @@ Compliant example::
 
     #define CONCAT(x, y) x ## y
     
-    uint32_t ab = 32;
+    uint32_t ab = 32U;
     printf("%d \n", CONCAT(a, b));
 
 .. rst-class:: non-compliant-code
@@ -33,7 +33,7 @@ Compliant example::
 
        #define CONCAT(x, y, z) x ## y ## z
        
-       uint32_t abc = 32;
+       uint32_t abc = 32U;
        printf("%d \n", CONCAT(a, b, c));
 
 
@@ -80,6 +80,20 @@ Compliant example::
        
        /* header contents without any protection */
        uint32_t func_showcase(uint32_t param);
+
+
+PP-04: Parentheses shall be used when referencing a MACRO parameter
+===================================================================
+
+Compliant example::
+
+    #define NEGATING(x) -(x)
+
+.. rst-class:: non-compliant-code
+
+   Non-compliant example::
+
+       #define NEGATING(x) -x
 
 
 
@@ -407,6 +421,21 @@ Compliant example::
        uint32_t showcase_array[5] = {0, 1};
 
 
+DI-08: An array declaration shall use a constant for the size
+=============================================================
+
+Compliant example::
+
+    uint32_t array_showcase[10];
+
+.. rst-class:: non-compliant-code
+
+   Non-compliant example::
+
+       uint32_t array_size = 10U;
+       uint32_t array_showcase[array_size];
+
+
 
 Functions
 *********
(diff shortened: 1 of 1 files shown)


doc/developer-guides/coding_guidelines.rst
```diff
@@ -695,6 +724,294 @@ Compliant example::
...
+       uint32_t func_showcase(uint32_t param)
+       {
+               return param;
+       }
+
+
+FN-12: Banned functions shall not be used
+=========================================
+
+The following cases shall be covered:
```
dbkinder: This would read better as:
```
a) These dynamic memory allocation functions shall not be used:
    ``calloc``, ``malloc``, ``realloc``, and ``free``.  Dynamic memory
   allocation shall be replaced with static memory allocation.
b) The functions ``va_arg``, ``va_start``, and ``va_end`` shall
   only be used within variadic functions (functions taking a
   variable number of parameters) such as ``printf``.
```

doc/developer-guides/coding_guidelines.rst
```diff
@@ -695,6 +724,294 @@ Compliant example::
...
+    uint32_t showcase_array[32];
+
+.. rst-class:: non-compliant-code
+
+   Non-compliant example::
+
+       uint32_t *showcase_ptr = (uint32_t *)malloc(32U * sizeof(uint32_t));
+
+
+FN-13: A function with declaration shall have the corresponding definition
```
dbkinder: Maybe instead: ``All declared functions shall have a corresponding definition``

doc/developer-guides/coding_guidelines.rst
```diff
@@ -695,6 +724,294 @@ Compliant example::
...
+   Non-compliant example::
+
+       /* In `showcase.h`: */
+       /* declaration */
+       uint32_t func_showcase(uint32_t param);
+       
+       /* There is no definition of `func_showcase` anywhere in all source files */
+
+
+FN-14: All defined functions shall be used
```
dbkinder: This requirement confuses me a bit.  This would prohibit writing a library of useful functions (such as a math library) and then being required to remove functions that weren't called.  What's the intent of this requirement?  Must all code in the application have a chance to be executed (100% coverage)?

doc/developer-guides/coding_guidelines.rst
```diff
@@ -695,6 +724,294 @@ Compliant example::
...
+       #include "showcase.h"
+       
+       /* There is no usage of `func_showcase` anywhere in all source files */
+       uint32_t func_showcase(uint32_t param)
+       {
+           return param;
+       }
+
+
+FN-15: Returning a structure from a function shall be used with restrictions
```
dbkinder: This requirement is really about pointers and not structures.  A function shall not return a pointer to a local object, either directly or within a returned structure or array.

doc/developer-guides/coding_guidelines.rst
```diff
@@ -1156,54 +1473,205 @@ Compliant example::
        char showcase = showcase_array[10];
 
 
+ST-20: The else statement shall not be empty if it is following a else if
```
dbkinder:
```suggestion
ST-20: The else statement shall not be empty if it is following an else if
```

doc/developer-guides/coding_guidelines.rst
```diff
@@ -1156,54 +1473,205 @@ Compliant example::
...
+           break;
+       case 'A':
+           /* do something */
+           break;
+       }
+
+
+ST-22: Every switch clause shall be terminated with a break statement
+=====================================================================
+
```
dbkinder: Maybe add an explanation: ``Falling through a case is not allowed.``

This would prohibit writing something like this then:
```
switch (someoption) {
  case 'a':
  case 'b':
  case 'c':
    // do something
    break;
  case 'd':
  case 'e':
    // do something else
    break;
}
```

doc/developer-guides/coding_guidelines.rst
```diff
@@ -1281,6 +1749,68 @@ Compliant example::
...
+    int32_t showcase = -10;
+
+.. rst-class:: non-compliant-code
+
+   Non-compliant example::
+
+       int32_t showcase = -10U;
+
+
+EP-08: A pointer shall not be assigned to another pointer whose lifetime is longer
```
dbkinder: Seems like the requirement is on the value of the pointer variable rather than the lifetime of the pointer variable. Your non-compliant example below is storing the address of a local variable into a more global pointer variable.  It's not assigning one pointer to another as your stated requirement says.

doc/developer-guides/coding_guidelines.rst
```diff
@@ -1695,8 +2225,14 @@ Compliant example::
...
 
-TY-17: An integer shall not be cast to a pointer
-================================================
+TY-17: A pointer shall not be cast from any other types
+=======================================================
+
+Only the following pointer assignment shall be allowed:
+
+a) Assignment shall be allowed via the address operator &;
+b) Assignment shall be allowed if the object pointed by the two pointers are
```
dbkinder: This should read:
```
Assignment shall be allowed if the objects pointed to by the two
pointers are of the same type.
```

doc/developer-guides/coding_guidelines.rst
```diff
@@ -1787,6 +2323,197 @@ Compliant example::
        }
 
 
+TY-21: The bit-filed shall either be signed integer or unsigned integer or bool
```
dbkinder: I suspect this should be ``bit-field`` and should be changed to:
``A bit-field shall be a signed integer, unsigned integer, or bool``

doc/developer-guides/coding_guidelines.rst
```diff
@@ -2023,3 +2750,31 @@ Compliant example::
...
+    typedef unsigned short uint16_t;
+
+.. rst-class:: non-compliant-code
+
+   Non-compliant example::
+
+       typedef unsigned short ushort_t;
+
+
+ID-10: The C keyword shall not be re-defined by a MACRO
```
dbkinder:
```suggestion
ID-10: A C keyword shall not be re-defined by a MACRO
```

doc/developer-guides/coding_guidelines.rst
```diff
@@ -1787,6 +2323,197 @@ Compliant example::
...
+    uint32_t showcase = 32U;
+
+.. rst-class:: non-compliant-code
+
+   Non-compliant example::
+
+       uint32_t showcase = 32UL;
+
+
+TY-28: The operands type of arithmetic operations shall be the same
```
dbkinder: This would read better as ``The operands of arithmetic operations shall be the same type``

doc/developer-guides/coding_guidelines.rst
```diff
@@ -1787,6 +2323,197 @@ Compliant example::
...
+    uint32_t showcase = 32U;
+
+.. rst-class:: non-compliant-code
+
+   Non-compliant example::
+
+       unsigned int showcase = 32U;
+
+
+TY-27: The operands type of the assignment operator shall be the same
```
dbkinder: This would read better as ``The operands of an assignment operator shall be the same type``

dbkinder:
@shiqiangyu @hwang37 Please resolve comments if you want this update included in the 0.6 release docs, otherwise it can be merged and published later.

hwang37:
@shiqingg +correct owner: Gao, Shiqing to address the comments

doc/developer-guides/coding_guidelines.rst
shiqingg: This rule is to eliminate the unreachable code and improve the code maintainability.

An exception shall be added here. I will update this rule in v2.

doc/developer-guides/coding_guidelines.rst
shiqingg: @dbkinder 
A pointer is a variable whose value is the address of another variable.
In the non-compliant example, the address of a local variable is a local pointer and is assigned to a global variable.
I will add a variable `uint32_t *showcase_ptr_local` to make the example more straightforward.

doc/developer-guides/coding_guidelines.rst
shiqingg: @dbkinder 
This rule is to eliminate the unreachable code and improve the code maintainability.
An exception shall be added here. I will update this rule in v2.

shiqingg:
@dbkinder Sorry for the late response, just come back from Chinese New Year vacation.
New version has been updated. Could you please help review again?

doc/developer-guides/coding_guidelines.rst
```diff
@@ -695,6 +724,299 @@ Compliant example::
...
+
+FN-14: All defined functions shall be used
+==========================================
+
+All defined functions shall be used, either called explicitly or indirectly via
+its address. Otherwise, the function shall be removed. The following case is an
+exception. Some extra functions may be kept in order to provide a more complete
+library of APIs. These functions may have been implemented but not used
+currently. These functions will come in handy in the future. In this case, these
+functions may not be removed.
```
dbkinder: This wording prohibits removing these "library" functions when your intention is to say they may remain:
```suggestion
functions may remain.
```

doc/developer-guides/coding_guidelines.rst
dbkinder: I still think the requirement needs to talk about the value of the pointer, not the lifetime of the pointer variable.  As written, I can't assign the address of a global variable to a local pointer and then assign this pointer value to a global pointer (a perfectly valid thing to do).

Is this a compliant example:
```
      uint32_t *showcase_ptr_global;
      uint32_t showcase_global = 100U;
       
       void func_showcase(void)
       {
           uint32_t *showcase_ptr_local;
 
           showcase_ptr_local = &showcase_global;
           showcase_ptr_global = showcase_ptr_local;   
       } 
       void main(void)
       {
           func_showcase();
           printf("*showcase_ptr_global %d \n", *showcase_ptr_global);
       }  
```

shiqingg:
> One last typo to fix. EP-08 still bothers me.

@dbkinder 
Got your point on EP-08.
Just pushed a new version. Could you please review it again to see if the statement is proper?

PR #2531, opened 2019-02-14
doc: use weston instead of ias by bundle add x11-server

wenlingz:
use weston instead of ias by bundle add x11-server


wenlingz:
@dbkinder please help review and approve, merge this PR, which will release on v0.6(plan ww07.5), thanks

Issue #2405, opened 2019-01-24
 exception injection 


mwang106:
Issue: low priority exception is not handled properly

mwang106:
[External_System_ID] ACRN-2031

Mingyuan18:
fixed bug by changed code

PR #2544, opened 2019-02-18
vmcall handling fix

binbinwu1:
ACRN HV hide VMX capability from guest. Only vmcall from SOS or some
specific vmcall from UOS are allowed.
Unsupported vmcall from UOS should be considered a "not in VMX
operation" case, and should be handled first according to SDM Vol. 3C 30-9.

HV passes the return value of vmcall by register RAX unconditionally.
However, if the vmcall is undefined for a guest, RAX value of guest vcpu
should not be changed.

According to SDM Vol. 3C 30-9, VMCALL is allowed from any CPL in guest.
VMCALL is NOT allowed from  CPL > 0 in vmx root mode.
ACRN hypervisor doesn't call VMCALL in vmx root mode, though.
In current code, ACRN also deny VMCALL from CPL > 0 in guest.
So for this case, #GP will not be injected, instead, modify the RAX to notify
the return value.

Tracked-On: #2405

### DIFF
--- a/hypervisor/arch/x86/guest/vmcall.c
+++ b/hypervisor/arch/x86/guest/vmcall.c
@@ -192,14 +192,14 @@ int32_t vmcall_vmexit_handler(struct acrn_vcpu *vcpu)
 	/* hypercall ID from guest*/
 	uint64_t hypcall_id = vcpu_get_gpreg(vcpu, CPU_REG_R8);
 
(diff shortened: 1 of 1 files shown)


HuiHuangShi:
2544<br>No new violations to the coding guideline detected.

HuiHuangShi:
2544<br>No new violations to the coding guideline detected.

Issue #2543, opened 2019-02-15
vLAPIC: DCR not properly initialized


peterfang:
DCR is initialized to 0 which means divisor shift is 1. Currently, both are initialized to 0 which result in incorrect APIC timer counts if the vLAPIC's DCR is never programmed.

peterfang:
[External_System_ID] ACRN-2688

Mingyuan18:
No regression with 
swupd info
Installed version: 28010

PR #2545, opened 2019-02-18
hv: vlapic: properly initialize DCR

peterfang:
DCR is initialized to 0 which means divisor shift is 1. Currently, both are initialized to 0 which result in incorrect APIC timer counts if the vLAPIC's DCR is never programmed.

This bug was exposed by OVMF because OVMF does not program DCR during LAPIC initialization.

Tracked-On: #2543


HuiHuangShi:
2545<br>No new violations to the coding guideline detected.

PR #2546, opened 2019-02-18
add features list on release_note_0.6

wenlingz:
add features list on release_note_0.6


NanlinXie:
Merged due to Dave is on vacation.

PR #2551, opened 2019-02-19
Documentation: tweak GSG

wenlingz:
* Remove the older kernel version(pk414)
* Add a one-line command to insert the PARTUUID into the acrn.conf
  file (without copying it all manually)

### DIFF
--- a/doc/getting-started/apl-nuc.rst
+++ b/doc/getting-started/apl-nuc.rst
@@ -42,10 +42,10 @@ complete this setup.
    Please refer to the ACRN :ref:`release_notes` for the Clear Linux OS
    version number tested with a specific ACRN release.  Adjust the
    instruction below to reference the appropriate version number of Clear
-   Linux OS (we use version 27230 as an example).
+   Linux OS (we use version 27960 as an example).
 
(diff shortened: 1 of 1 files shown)


wenlingz:
@gvancuts please also help review this PR which is the your original PR 2533, thanks.

doc/getting-started/apl-nuc.rst
```diff
@@ -124,7 +124,6 @@ partition. Follow these steps:
       kernel-org.clearlinux.native.4.20.2-683
       kernel-org.clearlinux.iot-lts2018-sos.4.19.13-1901141830
```
gvancuts: Which Clear Linux version are we using as a reference here?

I suspect the `release` field of the kernel is of the new type (much shorter) so the kernel number looks like `kernel-org.clearlinux.iot-lts2018-sos.4.19.19-8`

wenlingz: We will update one reference here, thanks

wenlingz: @gvancuts done, please help check, thanks

Issue #2558, opened 2019-02-19
isrvec_stk_top could be zero once we enable APICv(apl_sdc_stable)


Mingyuan18:
isrvec_stk_top could be zero once we enable APICv

Mingyuan18:
[External_System_ID] ACRN-2705

Mingyuan18:
Code review pass and no regression issue found.

PR #2559, opened 2019-02-19
hv: isrvec_stk_top could be zero once we enable APICv

liuwei987:
With APICv enabled, it's possible that the interrupt is not
injected through VMX_ENTRY_INT_INFO_FIELD. In that case, the
isrvec_stk_top will not be increased.

Tracked-On: #2558


wenlingz:
OK to verify

Issue #2595, opened 2019-02-22
Update vsbl to v1.2(apl_sdc_stable)


binbinwu1:
Update vFastboot to v1.2
Newly changed:
        Optimize fastboot erase for virtual IO block medium.
        Update openssl to version 1.0.2q.

Mingyuan18:
[External_System_ID] ACRN-2745

Mingyuan18:
Cannot reproduced this issue in below build:
SOS version: 28100

PR #2596, opened 2019-02-22
Update vsbl to version 1.2

binbinwu1:
Update vFastboot to v1.2
Newly changed:
        Optimize fastboot erase for virtual IO block medium.
        Update openssl to version 1.0.2q.

Tracked-On: #2595


gvancuts:
Both `*.bin` files are 0 byte in size?

binbinwu1:
> Both `*.bin` files are 0 byte in size?

I think its +0 byte. The size is not changed.

gvancuts:
> I think its +0 byte. The size is not changed.

Yes, good catch :)

Issue #2020, opened 2018-12-10
DM: Enable QoS in ACRN, based on runC container


Mingyuan18:
DM: Enable QoS in ACRN, based on runC container

Mingyuan18:
[System_External_ID] ACRN-2192

Mingyuan18:
Based on runC container, verify some I/O P1 cases on each platform devices, and the function works well.
SOS: 28600
ACRN tag: acrn-2019w14.1-140000p
Kernel: 2018 4.19.31-39

PR #2603, opened 2019-02-25
ACRN Device-Model QoS patch

Ragnarliu:
The patch implements acrn-dm QoS. We use runC container to implement
acrn-dm QoS, the patch adds the way to start a VM in runC container.
So we can isolate the resouce of service OS.

Tracked-On: #2020
Signed-off-by: Liu, Long  long.liu@intel.com


HuiHuangShi:
2603 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

HuiHuangShi:
2603 <br>The PR does not touch any hypervisor source file. Skip coding guideline checks.<br>

wenlingz:
@Ragnarliu please add Track-On info,thanks

wenlingz:
@dongyaozu please help approve it, thanks

tools/acrn-manager/acrnctl.c
```diff
@@ -193,6 +193,21 @@ static int acrnctl_do_add(int argc, char *argv[])
...
 			return -1;
 		}
+
+		/*
+		 * If there's "-C" parameter in acrnctl add command
+		 * check if the SoS support runC container at first, then
+		 * strip "-C" and set the flag.
+		*/
+		if (strncmp(argv[i], "-C", 2) == 0) {
+			if (access("/sbin/runc", F_OK) != 0) {
```
gvancuts: I'm concerned that this would easily break on systems that do have `runc` installed but possibly on a different path (e.g., I believe Ubuntu ships `runc` in `/usr/sbin`).

tools/acrn-manager/acrnctl.c
```diff
@@ -193,6 +193,21 @@ static int acrnctl_do_add(int argc, char *argv[])
...
+				printf("runC command not supproted\n");
```
gvancuts: How about: `Your system does not appear to have runc installed, please install it first. Exiting.`

tools/acrn-manager/acrnctl.c
Ragnarliu: This message only appears when the user use acrnctl add command with "-C" parameter.

tools/acrn-manager/acrnctl.c
Ragnarliu: The function only works on SOS which is above specific version,  so even users install the runc, it will not work, so if there's  no runc binary in /sbin, we suggest user don't use "-C" parameter.

gvancuts: We need to be more agnostic to the SOS version. Other OSs (Linux distros) could be used instead of Clear Linux which we use as the reference today. Hard-coding things the way it's done is not good practice. This, for example, will break things with Ubuntu for no other reason that we're checking the presence of a tool in the wrong location (Ubuntu puts it under `/usr/sbin` and not `/sbin`), but the feature would otherwise be fully functional.

Ragnarliu: OK, I will change the logical in the next patch. 
Thanks

PR #2607, opened 2019-02-25
ABL seed parse new argument

YadongQi:
Due to ABL design change, it will reword the "dev_sec_info.param_addr="
to "ABL.svnseed" in command line.

Tracked-On: projectacrn#2605
Acked-by: Zhu Bing <bing.zhu@intel.com>

### DIFF
--- a/hypervisor/boot/sbl/abl_seed_parse.c
+++ b/hypervisor/boot/sbl/abl_seed_parse.c
@@ -22,7 +22,11 @@ struct dev_sec_info {
 	struct abl_seed_info seed_list[ABL_SEED_LIST_MAX];
 };
 
-static const char *dev_sec_info_arg = "dev_sec_info.param_addr=";
+static const char *abl_seed_arg[] = {
+		"ABL.svnseed=",
+		"dev_sec_info.param_addr=",
+		NULL
+};
 
 static void parse_seed_list_abl(void *param_addr)
 {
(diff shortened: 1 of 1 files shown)


hypervisor/boot/sbl/abl_seed_parse.c
```diff
@@ -30,56 +30,49 @@ static void parse_seed_list_abl(void *param_addr)
...
-	 * seeds will be generated which will contain the legacy seed.
-	 * Here find the legacy seed index first.
-	 */
-	for (i = 1U; i < sec_info->num_seeds; i++) {
-		if (sec_info->seed_list[i].svn <
-				sec_info->seed_list[legacy_seed_index].svn) {
-			legacy_seed_index = i;
+	bool parse_success = false;
+
+	if ((sec_info != NULL) && (sec_info->num_seeds >= 2U) && (sec_info->num_seeds <= ABL_SEED_LIST_MAX)) {
```
bzhu5: is this ABL_SEED_LIST_MAX defined as 4?

I was told that this could be upgraded to 5 in future. 

what's the impact if you update this to 5? if no for now, we should change it to 5

bzhu5: anyway, it should be fine if we still use 4 now

YadongQi: I think we would do future things in future? since this commit is cherry-picked from master branch. we'd better to make an additional patch instead of modify this patch.

bzhu5: ok, i would recommend that we should do this in future on "master" branch. (in case that if cse fw upgrades to max 5, then we would fail at that time)

PR #2616, opened 2019-02-25
Add fatal warning PCI MSIX table count > conf max

DunderRoffe:
Add a fatal warning when an PCI device is added with an MSI-X table_count
larger than CONFIG_MAX_MSIX_TABLE_NUM. This is fatal since the init function
in the handler for MSI-X (vmsix_init) only looks at table_count when
populating the table. Since CONFIG_MAX_MSIX_TABLE_NUM is the max size of the
table array entry in the pci_msix struct. This will cause the msix handler
to write outside of the table array.
Tracked-On: #2624

### DIFF
--- a/hypervisor/hw/pci.c
+++ b/hypervisor/hw/pci.c
@@ -351,6 +351,8 @@ static void pci_read_cap(struct pci_pdev *pdev)
 				pdev->msix.table_offset = table_info & ~PCIM_MSIX_BIR_MASK;
 				pdev->msix.table_count = (msgctrl & PCIM_MSIXCTRL_TABLE_SIZE) + 1U;
 
+				ASSERT(pdev->msix.table_count <= CONFIG_MAX_MSIX_TABLE_NUM);
+
 				/* Copy MSIX capability struct into buffer */
 				for (idx = 0U; idx < len; idx++) {
 					pdev->msix.cap[idx] = (uint8_t)pci_pdev_read_cfg(pdev->bdf, offset + idx, 1U);


NanlinXie:

hypervisor/hw/pci.c
```diff
@@ -351,6 +351,11 @@ static void pci_read_cap(struct pci_pdev *pdev)
 				pdev->msix.table_offset = table_info & ~PCIM_MSIX_BIR_MASK;
 				pdev->msix.table_count = (msgctrl & PCIM_MSIXCTRL_TABLE_SIZE) + 1U;
 
+				if (pdev->msix.table_count > CONFIG_MAX_MSIX_TABLE_NUM) {
+					pr_fatal("Table count for PCI device with BDF 0x%04x is higher than the configured maximum (%d > %d)!",
+							pdev->bdf, pdev->msix.table_count, CONFIG_MAX_MSIX_TABLE_NUM);
+				}
+
```
dongyaozu: From FuSa p.o.v, this is deadcode and must be removed.  We should make it as pre-assumption and let FuSa test case make sure it.

We are defining standard error processing method with FuSa ..

ClaudZhang1995:
OK to verify

HuiHuangShi:

wenlingz:
@DunderRoffe please add Tracked-On info,thanks

DunderRoffe:
Added an issue and linked to it as "Tacked-On: #2624"

acrnsi:
OK to verify

wenlingz:
@DunderRoffe please add Tracked-On info, thanks

wenlingz:
@DunderRoffe please refer eddie's comment,thanks, and add review-by info next time

hypervisor/hw/pci.c
DunderRoffe: I am sorry but could you provide a link to the meaning of FuSa? I can not seem to recall the term.

I agree that this code is not contributing to any features. However, it helps identify a case where the compile-time configuration is not matching the system the hypervisor is intended to run on and will lead to pagefaults. Also this code only runs once for each physical PCI device which max 96 times with default configuration.

Is there a way this could comply to FuSa? As in having it terminate the hypervisor or something along those lines?

Sorry again for my ignorance of not knowing FuSa.

dongyaozu: @wuxyintel  can you suggest if we should use @pre here or debug only code or normal code? We may publish the error condition processing  rule to outside too.

Fusa means functional safety certification.

wuxyintel: Hi Eddie, in this case, "pdev->msix.table_count" is read from physical PCI configuration space, as we known, in the current code, physical PCI is configured by bootloader, the hypervisor read them through IO port. So physical PCI configuration information is external resource provided by bootloader. Suggest to check configuration error during platform initialization, if it detects some errors then panic, and define some constraints in the safety manual. But in this case, we maybe need to consider more other error cases and do checking, it is hard to do complete checking for physical PCI configuration information. For example, in this patch, this checking can't guarantee that "pdev->msix.table_count" is right. BTW, can we consider to statically define all physical PCI configuration of the platform and verify them manually?

pre:
> can you suggest if we should use @pre here or debug only code or normal code?

I don’t consider @pre necessary here.

hypervisor/hw/pci.c
donsheng: Currently in pci.c, during startup it will scan the physical pci bus ans store the results (bar base, bar size, capacity structure, etc) into global variables, then this info is used by other vpci modules.
In the future, we can statically defining all the pci config rather than scanning it dynamically, or  during startup, compare the statically defined pci confi with these dynamically retrieved results and throw an error if an mismatch is found, etc

wuxyintel: @DunderRoffe After reconsider more about this case from safety view, physical PCI configuration is provided by Board Support Package or firmware,  it is hard to do complete checking for physical PCI configuration information by hypervisor, so it is better to define application constraints on the hypervisor safety manual and Board Support Package or firmware should guarantee physical PCI configuration meet related application constraints, then the hypervisor can trust this configuration. 
So in this case, for debug purpose, we can use ASSERT to check whether MSI-X table_count
is larger than CONFIG_MAX_MSIX_TABLE_NUM or not.

DunderRoffe: So you mean change to
```c
assert(pdev->msix.table_count > CONFIG_MAX_MSIX_TABLE_NUM);
```
instead of the if statement?

wuxyintel: @DunderRoffe  Please use assert(pdev->msix.table_count <= CONFIG_MAX_MSIX_TABLE_NUM), thanks!

DunderRoffe: Updated per your request :)

HuiHuangShi:

Issue #2618, opened 2019-02-26
possibility of use uninitialized params in abl_seed_parse()(apl_sdc_stable).


YadongQi:
'len' and 'arg' might be used uninitialized in abl_seed_parse().

Mingyuan18:
[External_System_ID] ACRN-2814

Mingyuan18:
Verified pass, so close it.

PR #2619, opened 2019-02-26
hv: fix possibility of using uninitialized params

YadongQi:
'len' and 'arg' might be used uninitialized in abl_seed_parse().
Fix them by setting initial value.

 Tracked-On: #2618


ClaudZhang1995:
OK to verify

Issue #2598, opened 2019-02-22
38 sub test cases fail in CtsSimpleperfTestCases9(apl_sdc_stable).


minhe1:
Summery:
[ACRN][Gordon_Peak][CTS] 38 sub test cases fail in CtsSimpleperfTestCases
Get DUT prepared with CTS execution environment followed with: https://source.android.com/compatibility/cts/setup.html

Steps to Reproduce:
1. Start the cts-tradefed with: ./android-cts/tools/cts-tradefed
2. Run the CTS with cts-tf > run cts --module/-m <module> --test/-t <test_name> -s <YOUR_DEVICE_ID> --skip-preconditions 
3. Check test result

Expected test result:
pass
Actual test result:
fail
fail list
ReportCommandTest.dwarf_callgraph
ReportCommandTest.exclude_kernel_callchain
record_cmd.callchain_joiner_options
record_cmd.dashdash
record_cmd.donot_stop_when_having_targets
record_cmd.dump_build_id_feature
record_cmd.dump_kernel_mmap
record_cmd.duration_option
record_cmd.dwarf_callchain_sampling
record_cmd.event_modifier
record_cmd.existing_processes
record_cmd.existing_threads
record_cmd.exit_with_parent_option
record_cmd.fp_callchain_sampling
record_cmd.freq_option
record_cmd.generate_samples_by_hw_counters
record_cmd.group_option
record_cmd.handle_SIGHUP
record_cmd.kernel_symbol
record_cmd.mmap_page_option
record_cmd.more_than_one_event_types
record_cmd.multiple_freq_or_sample_period_option
record_cmd.no_dump_symbols
record_cmd.no_options
record_cmd.no_post_unwind_option
record_cmd.no_unwind_option
record_cmd.output_file_option
record_cmd.record_meta_info_feature
record_cmd.sample_period_option
record_cmd.start_profiling_fd_option
record_cmd.stop_when_no_more_targets
record_cmd.symfs_option
record_cmd.system_wide_dwarf_callchain_sampling
record_cmd.trace_offcpu_option
stat_cmd.auto_generated_summary
stat_cmd.event_modifier
stat_cmd.group_option
stat_cmd.sample_speed_should_be_zero

Additional Information:
Operator Network: Yes
Test Location: SH
CTS version : http://gas-report-00.sh.intel.com/DATA/Binary/CTS/9.0_r3/android-cts-9.0_r3-linux_x86-x86.zip
[...]

PR #2620, opened 2019-02-26
Patches to passthru PMU to UOS

minhe1:
make hypervisor PROFILING=0, it will disable profiling and pmu will be pass-thru and perf tools can be run directly inside UOS.

Tracked-On:#2598


acrnsi:
OK to verify

gvancuts:
Is there any reason we’re not making this a Kconfig parameter?

minhe1:
> Is there any reason we’re not making this a Kconfig parameter?

This patch is only for GP2.0 project and will be as a reference to the customer. In the master branch, we will make PMU-passthru a kconfig option.

PR #2636, opened 2019-02-27
sos-bootargs: change i915 initial mode setting to 0

rebeccasf:
Signed-off-by: Chang Rebecca Swee Fun <rebecca.swee.fun.chang@intel.com>


fyin1:
This is patch for stable branch.

Issue #2677, opened 2019-03-05
xDCI driver dose not work well when the 64-bit BAR is allocated above 4GB


mwang106:
Aaag faild to enter fastboot mode due to xDCI init fails

mwang106:
[External_System_ID] ACRN-2864

Mingyuan18:
Can't reproduce this issue in below build:
acrn-2019w10.2-140000p

PR #2678, opened 2019-03-05
[RevertMe] dm: pci: restore workaround when alloc pci mem64 bar

binbinwu1:
There was a workaround in DM that allocates PCI 64bit mem bar in 32bit mem space
if the bar size is within 32MB.
After the workaround being removed, there is an issue to enter fastboot
mode for inappropriate handling of 64bit mem bar in guest driver.
The patch bring the workaround back, and skip the workaround when the guest
is booted by OVMF.

Revert the patch after the guest fixs the issue of handling 64bit mem bar in
fastboot mode.

Tracked-On: #2677

### DIFF
--- a/devicemodel/core/main.c
+++ b/devicemodel/core/main.c
@@ -89,6 +89,7 @@ uint8_t trusty_enabled;
 char *mac_seed;
 bool stdio_in_use;
 bool lapic_pt;
+bool skip_pci_mem64bar_workaround = false;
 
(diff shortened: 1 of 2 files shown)


binbinwu1:
> "After the workaround being removed," can you say which walkaround?
> This is for commit message.

The

binbinwu1:
> "After the workaround being removed," can you say which walkaround?
> This is for commit message.

added

PR #2686, opened 2019-03-06
hv: refined the pmu passthru irq handling

minhe1:
This patch refined the pmu passthru irq handling, which will re-enable
the PMC interrupt in the vlapic handler. So that can avoid some
interrupt storm situation happens.

This patch can fix the remaining Android perf CTS failure.

Tracked-On: #2598

### DIFF
--- a/hypervisor/arch/x86/guest/vlapic.c
+++ b/hypervisor/arch/x86/guest/vlapic.c
@@ -714,6 +714,13 @@ vlapic_lvt_write_handler(struct acrn_vlapic *vlapic, uint32_t offset)
 	*lvtptr = val;
 	idx = lvt_off_to_idx(offset);
 	atomic_store32(&vlapic->lvt_last[idx], val);
+
+#if defined(HV_DEBUG) && !defined(PROFILING_ON)
+	if ((offset == APIC_OFFSET_PERF_LVT) &&
+			(val == APIC_LVT_DM_NMI)) {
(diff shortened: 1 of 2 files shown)


HuiHuangShi:

minhe1:
> if profiling is disabled, would this patch cause problem?

Only when PMU registers is configured, the PM interrupt will be triggered. So there should be no side effect. But to be safe, I'll add a ifdef to wrap this piece of code. 
Thanks for the suggestion.

HuiHuangShi:

anthonyzxu:
> > if profiling is disabled, would this patch cause problem?
> 
> Only when PMU registers is configured, the PM interrupt will be triggered. So there should be no side effect. But to be safe, I'll add a ifdef to wrap this piece of code.
> Thanks for the suggestion.

Seems Profiling and PMU pass-thr can't coexist, am I right?
Do we need to have a config to enable one of them?

minhe1:
> if profiling is disabled, would this patch cause problem?
> 
> Only when PMU registers is configured, the PM interrupt will be triggered. So there should be no side effect. But to be safe, I'll add a ifdef to wrap this piece of code.
> Thanks for the suggestion.
> 
> Seems Profiling and PMU pass-thr can't coexist, am I right?
> Do we need to have a config to enable one of them?

Correct, they cannot co-exist.
In this stable branch, it's controlled by the PROFILING_ON flag.
And in the master branch, I'll use a config to control the PMU mode.

Issue #2713, opened 2019-03-11
Enable ACRN to boot Zephyr


KaigeFu:
Enable ACRN to boot Zephyr as a normal VM OS.

Mingyuan18:
[External_System_ID] ACRN-2870

Mingyuan18:
Verify pass on below build:
SOS version: 28600
acrn tag: acrn-2019w14.1-140000p
kernel: 2018 4.19.31-39

PR #2745, opened 2019-03-13
DM: Set thre_int_pending as true when write IER_ETXRDY

KaigeFu:
There are uart drivers which send data to hardware in ISR, like zephyr.
Zephyr uart driver stores data in tx_ringbuf and then write the IER of
uart to trigger tx interrupt. Finally, it write the data stored in tx_ringbuf
to uart hardware in ISR.

In current design of vUART in acrn-dm, we cann't handle above situation as we
only mark thre_int_pending as true when writing to REG_DATA. But it will no
real data write to uart hardware before firing interrupt in implementation of
zephyr's uart driver. Consequently, zephyr uart driver will wait for tx
interrupt to be triggered and then wirte data to uart hardware. But acrn-dm
will wait for data arriving to trigger interrupt. Here, dead loop occurs.

This patch set uart->thre_int_pending to true if (uart->ier & IER_ETXRDY)
changes from 0 to 1 in order to tigger one tx interrupt as expect to make
zephyr uart driver happy.

Tracked-On: #2713
Signed-off-by: Anthony Xu <anthony.xu@intel.com>

### DIFF
--- a/devicemodel/hw/uart_core.c
+++ b/devicemodel/hw/uart_core.c
@@ -466,7 +466,13 @@ uart_write(struct uart_vdev *uart, int offset, uint8_t value)
 		/*
 		 * Apply mask so that bits 4-7 are 0
 		 * Also enables bits 0-3 only if they're 1
+		 *
+		 * Set uart->thre_int_pending to true if (uart->ier & IER_ETXRDY) changes
+		 * from 0 to 1 in order trigger one tx interrupt as expect.
 		 */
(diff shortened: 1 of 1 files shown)


KaigeFu:
Hi Eddie. I get lost here. What do you mean by saying "This is completely wrong"?
commit, code or both of them? Could you help me out of here?

HuiHuangShi:
out of time

Issue #2743, opened 2019-03-13
hypervisor hang with reset_ptdev_irq_info fuzzing test、


Mingyuan18:
1.Environment
 [Board]: D0 MRB
 HW: MRB D0 Premium
 IOC: 4. 0. 12
 ABL: master branch
 Debug Image: build image with EBtools(tag: acrn-2018w47.5-173000p including patch e350abe40d623a283ba4bc59511fa526d3714bcb) or latest code
 git clone [https://github.intel.com/projectacrn/acrn-ebtool]
 git clone [https://github.com/projectacrn/acrn-hypervisor.git] 
 git clone [https://github.com/projectacrn/acrn-kernel.git]
 SOS kernel: 4.19.0-quilt-2e5dc0ac

Tools setup:
 We used Syzkaller ran with hypercall unit tests to do Fuzzing test for ACRN, which ran on SOS and communicate with DM process by socket.
 enable_syscalls: ["syz_ic_reset_ptdev_irq_info"]

2. Reproduce Steps
  setup env with wiki: And sync code to local
 apply patch_for_fuzzing_on_dm.txt to devicemodel, and build acrn images
 modify acrn_build.sh based your own environment, and run it to rebuild syzkaller
 flash images, and then make uos autoboot, remove sos password
 use acrn.cfg (modify the ip to your own ip) to run syzkaller cases: ./bin/syz-manager -config=acrn.cfg debug

3. Expected result:
 SOS/UOS not crashed not hang, and SUT works well

4. Current result:
 After run: ./bin/syz-manager -config=acrn.cfg --debug
 DUT hang at HV

Hypervisor hang logs as below：
ACRN:\>[4145500297us][cpu=0][sev=3][seq=305]:hcall_reset_ptdev_intr_info: Invalid irq type: 3
 [4145537351us][cpu=0][sev=3][seq=306]:hcall_reset_ptdev_intr_info: Invalid irq type: 3
 [4145573283us][cpu=0][sev=3][seq=307]:hcall_reset_ptdev_intr_info: Invalid irq type: 3
 [4145618278us][cpu=0][sev=3][seq=308]:hcall_reset_ptdev_intr_info: Invalid irq type: 3
 [4145662286us][cpu=0][sev=3][seq=309]:hcall_reset_ptdev_intr_info: Invalid irq type: 3
 [4146451281us][cpu=0][sev=3][seq=310]:hcall_reset_ptdev_intr_info: Invalid irq type: 3
[...]

PR #2746, opened 2019-03-13
hv: vtd: check bus limit when add/remove device

binbinwu1:
Check bus limit when add/remove device in iommu.

Tracked-On: #2743

### DIFF
--- a/hypervisor/arch/x86/vtd.c
+++ b/hypervisor/arch/x86/vtd.c
@@ -959,6 +959,11 @@ static int32_t remove_iommu_device(const struct iommu_domain *domain, uint16_t s
 		return -EINVAL;
 	}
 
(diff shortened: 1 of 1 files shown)


yonghuah:
@dongyaozu  @binbinwu1 , given @pre is used,  no need this check in local function,  then, need similar check in hypercalls: hcall_assign_ptdev() &  hcall_deassign_ptdev().

binbinwu1:
> This runtime check will become deadcode in FuSa, This is an problem. Can you use @pre and debug-only ASSERT?

The parameters are passed from hypercall, which is not trust worthy. I am not sure if we can add the @pre for the case. But the check can be move to public function instead of the internal function, what do you think? @dongyaozu

HuiHuangShi:
out of time

Issue #2742, opened 2019-03-13
doc: fix wrong desciption of trusty's memory mapping


YadongQi:

Mingyuan18:
[External_System_ID] ACRN-2975

Mingyuan18:
No regression found in test:
SOS version : 28340 
ACRN TAG: 2019w11.5.140000p |

PR #2749, opened 2019-03-13
doc: fix wrong desciption of trusty's memory mapping

YadongQi:
Original design is to map trusty's memory to 512G-513G, but GR-MRB platform
does not support memory mapping bigger than 512G. So final implementation
is mapping trusty's memory to 511G-512G.

This patch update the documentation with latest implementation.

Tracked-On: #2742


hwang37:
@bzhu5 @JasonChenCJ to review the content.  @dbkinder can merge if agreed by them.

PR #2775, opened 2019-03-14
doc: add some rules related to naming convention

shiqingg:
This patch adds some rules related to naming convention.

Tracked-On: #861

### DIFF
--- a/doc/developer-guides/coding_guidelines.rst
+++ b/doc/developer-guides/coding_guidelines.rst
@@ -2776,8 +2776,6 @@ Compliant example::
        if (condition) { do_a();} else {do_b();}
 
 
-
-
 CS-03: Tabs shall be used for code indentation
 ==============================================
 
@@ -2808,8 +2806,6 @@ CS-04: Tabs shall be 8 characters wide
 
 A tab character shall be considered 8-character wide when limiting the line
 width.
-
-
 
 
 CS-05: Trailing whitespace shall not be allowed at the end of lines
@@ -2984,8 +2980,6 @@ Compliant example::
        }
 
 
-
-
 CS-13: Braces after if/switch/for/do/while shall be in the same line
 ====================================================================
 
@@ -3033,8 +3027,6 @@ Compliant example::
        uint32_t func_showcase(uint32_t param) {
                return param;
        }
-
-
 
 
 CS-15: A 'switch' statement and its subordinate 'case' shall be aligned
@@ -3092,8 +3084,6 @@ Compliant example::
                      param_3);
 
 
-
-
 CS-17:  '//' shall not be used for single-line comments
 =======================================================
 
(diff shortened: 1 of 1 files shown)


doc/developer-guides/coding_guidelines.rst
```diff
@@ -3110,3 +3100,176 @@ Compliant example::
...
+.. rst-class:: non-compliant-code
+
+   Non-compliant example::
+
+       struct timer {
+               ...
+       };
+
+
+NC-05: Data structures only used by one module shall be named with the module name as prefix
```
dbkinder: Is there a precedence hierarchy for these naming conventions?  What if I have a data structure used in only one module in the hypervisor?  Which rule prevails: using ``hv_`` prefix or the module name prefix?

doc/developer-guides/coding_guidelines.rst
```diff
@@ -3110,3 +3100,176 @@ Compliant example::
...
+.. rst-class:: non-compliant-code
+
+   Non-compliant example::
+
+       struct ctxt {
+               ...
+       };
+
+
+NC-06: Data structures related to hardware resource shall be named with the resource name as suffix
```
dbkinder: Similar question here.  If the data structure is a hardware resource that is exposed to the public, what prefix is expected: ``acrn_`` (because it's public) or the resource name (because it's a hardware resource), or both?

doc/developer-guides/coding_guidelines.rst
```diff
@@ -3110,3 +3100,176 @@ Compliant example::
...
+       struct firmware_operations firmware_sbl_ops = {
+               .init = sbl_init,
+               .get_rsdp = sbl_get_rsdp,
+       };
+
+
+NC-08: A variable name can be shortened from its data structure type name
+=========================================================================
+
+A variable name can be shortened from its data structure type name to avoid the
```
dbkinder: I'm not sure if this rule is needed because you have rule ID-08 that already forbids a variable name the same as a type name.  Are there any rules about acceptable shortened names?

doc/developer-guides/coding_guidelines.rst
shiqingg: In rule NC-03, it mentions that `The data structure is used by multiple modules;`. 
So, in this case, we should use the module name prefix.

doc/developer-guides/coding_guidelines.rst
shiqingg: This rule is more like a suggestion. We don't have the rules about acceptable shortened names.
Maybe I should remove this rule? Since its purpose is to make sure ID-08 is satisfied.

doc/developer-guides/coding_guidelines.rst
shiqingg: We should apply both rules here if it meets both requirements.

shiqingg:
@dbkinder Thanks for your review David! 
I removed NC-08 in new version and some comments are inline.

PR #2791, opened 2019-03-15
Add gsg quick setup script.

lirui34:
Working out a script to quick setup SOS or UOS.

Co-Authored-By: lirui34 <48583653+lirui34@users.noreply.github.com>


NanlinXie:

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+# This script is aim to setup or upgrade sos & uos automatically.
+
+# switch for download uos function
+skip_download_uos=0
+# turn on if you need to setup uos
+upgrade_uos=0
+
+function print_help()
+{
+echo "Usage:"
```
gvancuts: We should add some more text that explains what this script does. Also making sure that people understand the risks of running this since it will modify various system parameters with full root access.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+echo "Launch this script as: ./flash.sh -t 28100"
+echo "-t to specify clear linux version for upgrading"
```
gvancuts: It would improve readability to add a `tab` at the beginning of each parameter line, e.g. 
```suggestion
echo -e "\t-t to specify clear linux version for upgrading"
```

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+
+function upgrade_sos()
+{
+# setup mirror, proxy env if specified parameters to p & m opt.
+swupd mirror -u
+if [[ -n $mirror ]]; then
+    echo "Add swupd mirror..."
+    swupd mirror -s $mirror
+fi
+unset https_proxy
```
gvancuts: We should only do this if a proxy is provided. Usually, one expects proxy to be transparently. What I mean by that is that if the system knows the proxy information (which should represent 90+% of the cases), then we shouldn't use this setting. This should only be needed where the proxy setting were not automatically determined.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+    if [[ $number -ge 1 ]]; then
+        echo "Clean all Linux bootloader event"
+        efibootmgr | grep 'Linux bootloader' | cut -d'*' -f1 | cut -d't' -f2 | head -n$number | xargs -i efibootmgr -b {} -B
+    fi
+
+    echo "Add new ACRN efi boot event"
+    efibootmgr -c -l "\EFI\acrn\acrn.efi" -d /dev/sda -p 1 -L "ACRN"
+
+    echo "Create loader.conf"
+    rm /mnt/loader/loader.conf && touch /mnt/loader/loader.conf
```
gvancuts: How about we at least keep a backup copy of this?
```suggestion
    mv /mnt/loader/loader.conf /mnt/loader/loader.conf.bck && touch /mnt/loader/loader.conf
```

doc/getting-started/gsg_quick_setup.sh
gvancuts: Also, why do you unset this twice (it's also done below in line 108)

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+    cp -r /usr/lib/modules/"`readlink /usr/lib/kernel/default-iot-lts2018 | awk -F '2018.' '{print $2}'`.iot-lts2018" /mnt/lib/modules
+    umount /mnt
+    sync
+fi
+uos_img_path=$(find ~/ -name clear-$target_version-kvm.img)
+cp -r /usr/share/acrn/samples/nuc/launch_uos.sh ~/
+sed -i "s/\(virtio-blk.*\)\/home\/clear\/uos\/uos.img/\1$(echo $uos_img_path | sed "s/\//\\\\\//g")/" ~/launch_uos.sh
+echo "Upgrade uos done..."
+echo "Now you can run below command to start UOS..."
+echo "# cd ~/ && ./launch_uos.sh -V 1"
```
gvancuts: We do not have this `-V` parameter in the launch scripts for EFI platforms

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+else
+    cd ~
+    curl https://download.clearlinux.org/releases/$target_version/clear/clear-$target_version-kvm.img.xz -o clear-$target_version-kvm.img.xz
+    if [[ $? -eq 0 ]]; then
+        unxz clear-$target_version-kvm.img.xz
+    fi
+    if [[ $? -eq 0 ]]; then
+        losetup -f -P --show clear-$target_version-kvm.img
+    fi
+    mount /dev/loop0p3 /mnt
```
gvancuts: Remove the hardcoded `loop0` here. You need to read it from the previous command, that's what the `--show` flag from `losetup` will do for you, print out the loop device that has been used.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+# Compare with current clear linux and skip upgrade sos if get the same version.
+if [[ $(echo -e `swupd info` | cut -d' ' -f3) -eq $target_version ]]; then
+    echo "Specified to the same clear linux version, continue to setup sos env."
+else
+    echo "Upgrade clear linux version to : $target_version ..."
+    swupd verify --fix --picky -m $target_version
+fi
+
+# Do the setups if previous process succeed.
+if [[ $? -eq 0 ]]; then
```
gvancuts: Do we need to reboot after the `swupd verify fix` operation or is it not required.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+        unxz $uos_img_xz_path
+        uos_img_path=$(find ~/ -name clear-$target_version-kvm.img)
+    fi
+    echo "get uos img file in: $uos_img_path"
+    if [[ $? -eq 0 ]]; then
+        losetup -f -P --show $uos_img_path
+    else
+        echo "fail to losetup uos img, please check uos img status" && exit 1
+    fi
+    mount /dev/loop0p3 /mnt
```
gvancuts: Remove the hardcoded loop0 here. You need to read it from the previous command, that's what the --show flag from losetup will do for you, print out the loop device that has been used.

gvancuts: You should also verify that nothing is already mounted under `/mnt`

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+                        ;;
+                h) print_help
+                        ;;
+                ?) print_help
+                        ;;
+        esac
+done
+
+# -t opt is must
+if [[ -z $1 ]]; then
```
gvancuts: We should add a check to ensure we have root privileges:
```suggestion
if [[ -z $1 || $EUID -ne 0 ]]; then
```

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+
+echo "Init swupd mirror & proxy"
+swupd mirror -u
+unset https_proxy
+umount /mnt
+}
+
+function upgrade_uos()
+{
+# Add proxy if you need to download uos img automatically.
```
gvancuts: You never download the UOS image automatically in the script, or did I miss it?

If true, then no need to modify any proxy setting

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+upgrade_uos=0
+
+function print_help()
+{
+echo "Usage:"
+echo "Launch this script as: ./flash.sh -t 28100"
+echo "-t to specify clear linux version for upgrading"
+echo "-p to specify a proxy server"
+echo "-m to use swupd mirror url"
+echo "-u to upgrade uos, if you just upgrade sos using this script, you'd better to manually reboot sos then continue to use this function."
```
dbkinder: I think I'll wait to do an English edit until after @gvancuts technical issues are addressed. For example this would be better as:
```
echo "-u to upgrade UOS.  After upgrading you must manually reboot the system."
```

doc/getting-started/gsg_quick_setup.sh
dbkinder: Do you have a convention of indenting the body of a function?

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+
+# You need to specify the correct clear linux ver number.
+if [[ `echo $target_version | awk '{print length($0)}'` -ne 5 ]]; then
+    echo 'Please input right clearlinux version to upgrade' && exit 1
+fi
+
+if [[ $upgrade_uos == 1 ]]; then
+    echo "Start upgrading uos to: $target_version ..."
+    upgrade_uos
+    exit
```
dbkinder: Should you write out that the user must reboot the system now, or add a reboot to the script?

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+unset https_proxy
+if [[ -n $proxy ]]; then
+    echo "Add proxy..."
+    export https_proxy=$proxy
+fi
+echo "disable auto update..."
+swupd autoupdate --disable
+
+# Compare with current clear linux and skip upgrade sos if get the same version.
+if [[ $(echo -e `swupd info` | cut -d' ' -f3) -eq $target_version ]]; then
```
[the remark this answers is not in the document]
lirui34: Thanks for your comments, I will update it in next release

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+    echo "Upgrade clear linux version to : $target_version ..."
+    swupd verify --fix --picky -m $target_version
+fi
+
+# Do the setups if previous process succeed.
+if [[ $? -eq 0 ]]; then
+    echo "Add service-os kernel-iot-lts2018 bundle"
+    swupd bundle-add service-os kernel-iot-lts2018
+
+    echo "mount /dev/sda1 to /mnt"
```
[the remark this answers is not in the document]
lirui34: Yes, I have changed code to determine the right ESP, also it will handle multiple ESPs if they are detected via 'fdisk -l' command.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+    fi
+    echo "Check linux bootloader event"
+    number=$(expr `efibootmgr | grep 'Linux bootloader' | wc -l` - 1)
+    if [[ $number -ge 1 ]]; then
+        echo "Clean all Linux bootloader event"
+        efibootmgr | grep 'Linux bootloader' | cut -d'*' -f1 | cut -d't' -f2 | head -n$number | xargs -i efibootmgr -b {} -B
+    fi
+
+    echo "Add new ACRN efi boot event"
+    efibootmgr -c -l "\EFI\acrn\acrn.efi" -d /dev/sda -p 1 -L "ACRN"
```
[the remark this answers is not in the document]
lirui34: Yes, I have used 'fdisk -l' to determine the right ESP, could you please help to review the latest code ?

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+    echo "Check ACRN efi boot event"
+    check_acrn_bootefi=`efibootmgr | grep ACRN`
+    if [[ "$check_arcn_bootefi" -ge "ACRN" ]]; then
+        echo "Clean all ACRN efi boot event"
+        efibootmgr | grep ACRN | cut -d'*' -f1 | cut -d't' -f2 | xargs -i efibootmgr -b {} -B
+    fi
+    echo "Check linux bootloader event"
+    number=$(expr `efibootmgr | grep 'Linux bootloader' | wc -l` - 1)
+    if [[ $number -ge 1 ]]; then
+        echo "Clean all Linux bootloader event"
```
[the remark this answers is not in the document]
lirui34: I was thinking that people would like to see only one "Linux bootloader" remained in boot list. :smile:

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+    swupd verify --fix --picky -m $target_version
+fi
+
+# Do the setups if previous process succeed.
+if [[ $? -eq 0 ]]; then
+    echo "Add service-os kernel-iot-lts2018 bundle"
+    swupd bundle-add service-os kernel-iot-lts2018
+
+    echo "mount /dev/sda1 to /mnt"
+    mount /dev/sda1 /mnt
```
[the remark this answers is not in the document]
lirui34: Yes, it will be checked in the latest code.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,207 @@
...
+    echo "Add default boot to ACRN"
+    echo "default acrn" >> /mnt/loader/loader.conf
+
+    echo "Get latest sos file from EFI org.clearlinux: "
+    new_kernel=`ls /mnt/EFI/org.clearlinux/*sos* -tl | grep kernel | head -n1 | awk -F'/' '{print $5}'`
+    echo "Get current sos kernel from acrn.conf file: "
+    cur_kernel=`cat /mnt/loader/entries/acrn.conf | sed -n 2p | cut -d'/' -f4`
+
+    echo "Replace root uuid from acrn.conf"
+    sed -i "s/<UUID of rootfs partition>/`blkid -s PARTUUID -o value /dev/sda3`/g" /mnt/loader/entries/acrn.conf
```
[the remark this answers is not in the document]
lirui34: Yes, I have updated in latest code, could you please help to review it ?

doc/getting-started/gsg_quick_setup.sh
lirui34: You are right, I added more text and example usages in it.

doc/getting-started/gsg_quick_setup.sh
lirui34: Thanks for the comments, it should be removed.

doc/getting-started/gsg_quick_setup.sh
lirui34: Yes, it's reword to this 'cd ~/ && ./launch_uos.sh'

doc/getting-started/gsg_quick_setup.sh
lirui34: Thanks for the comments, it's updated in latest code, could you please help to review it ?

doc/getting-started/gsg_quick_setup.sh
lirui34: Yes, it will do reboot after sos setup is complete.

doc/getting-started/gsg_quick_setup.sh
lirui34: Yes, it should be handled in latest code.

doc/getting-started/gsg_quick_setup.sh
lirui34: I have update in this part. Now it is also support to download uos image from official site with proxy settings.

doc/getting-started/gsg_quick_setup.sh
lirui34: Rewrite upgrade uos function. Now it can be upgraded directly without checking sos clear linux version.  Also users don't need to reboot device for taking effects after it.

doc/getting-started/gsg_quick_setup.sh
lirui34: I just used tab in it...may I have any suggestion for improvement :smile:

doc/getting-started/gsg_quick_setup.sh
lirui34: Yes, I added text and some of usages in it, could you please help to check in latest code ?

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+    umount $efi_mount_point
+fi
+
+echo "Disable auto update..."
+swupd autoupdate --disable
+
+# Compare with current clear linux and skip upgrade sos if get the same version.
+source /etc/os-release
+if [[ $VERSION_ID -eq $target_version ]]; then
+    echo "Specify with the same clear linux version, continue to setup sos..."
```
dbkinder:
```suggestion
    echo "Clear Linux version $target_version is already installed. Continuing to setup SOS..."
```

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+else
+    echo "Upgrade clear linux version to : $target_version ..."
```
dbkinder:
```suggestion
    echo "Upgrading Clear Linux version from $VERSION_ID to $target_version ..."
```

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+    echo "Add /mnt/EFI/acrn folder"
+    mkdir -p /mnt/EFI/acrn
+    echo "Copy $acrn_conf_path /mnt/loader/entries/"
+    if [[ ! -f $acrn_conf_path ]]; then
+        echo "Missing acrn.conf file in folder: $acrn_conf_path"
+        umount /mnt && sync
+        exit 1
+    fi
+    cp -r $acrn_conf_path /mnt/loader/entries/
+    if [[ $? -ne 0 ]]; then echo "You didn't choose an right EFI partition." && exit 1; fi
```
dbkinder: The user can't choose the EFI partition, this is figured out by the script. What's the real error here?

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+        echo "Clean all Linux bootloader event"
+        efibootmgr | grep 'Linux bootloader' | cut -d'*' -f1 | cut -d't' -f2 | head -n$number | xargs -i efibootmgr -b {} -B
+    fi
+
+    echo "Add new ACRN efi boot event"
+    efibootmgr -c -l "\EFI\acrn\acrn.efi" -d $partition -p 1 -L "ACRN"
+
+    echo "Create loader.conf"
+    mv /mnt/loader/loader.conf /mnt/loader/loader.conf.bck && touch /mnt/loader/loader.conf
+    echo "Add default boot wait time"
```
dbkinder:
```suggestion
    echo "Add default (5 seconds) boot wait time"
```

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+    echo "Add default boot to ACRN"
+    echo "default acrn" >> /mnt/loader/loader.conf
+
+    new_kernel=`ls /mnt/EFI/org.clearlinux/*sos* -tl | grep kernel | head -n1 | awk -F'/' '{print $5}'`
+    echo "Getting latest Service OS kernel version: $new_kernel"
+    cur_kernel=`cat /mnt/loader/entries/acrn.conf | sed -n 2p | cut -d'/' -f4`
+    echo "Getting current Service OS kernel version: $cur_kernel"
+
+    echo "Replacing root partition uuid in acrn.conf"
+    sed -i "s/<UUID of rootfs partition>/`blkid -s PARTUUID -o value $root_partition`/g" /mnt/loader/entries/acrn.conf
```
dbkinder: You should add a test to check this command didn't fail.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+
+    echo "Service OS setup done!"
+    echo "Rebooting Service OS to taking effect changes."
+else
+    echo -e "Fail to upgrade sos $target_clearlinux from mirror url."
+fi
+
+umount /mnt
+sync
+reboot
```
dbkinder: If the upgrade fails, you're still rebooting the system?  The user may not get a chance to see the failure error message.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+    if [[ ! -f $uos_img_xz ]] && [[ ! -f $uos_img ]]; then
+        echo "You should download uos clear-$target_version-kvm.img.xz file firstly." && exit 1
+    fi
+    if [[ -f $uos_img_xz ]]; then
+        echo "Unxz uos file: $uos_img_xz"
+        unxz $uos_img_xz
+        uos_img=`echo $uos_img_xz | sed 's/.xz$//g'`
+    fi
+    echo "get uos img file: $uos_img"
+    if [[ $? -eq 0 ]]; then
```
dbkinder: This test of ``$?`` will return the result of the ``echo`` command execution, right?  So it will always be 0.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+    fi
+    mount $uos_partition /mnt
+    cp -r /usr/lib/modules/"`readlink /usr/lib/kernel/default-iot-lts2018 | awk -F '2018.' '{print $2}'`.iot-lts2018" /mnt/lib/modules
+else
+    cd ~
+    echo "Downloading UOS image: $uos_image_link"
+    curl $uos_image_link -o clear-$target_version-kvm.img.xz
+    uos_img=clear-$target_version-kvm.img
+    if [[ -f $uos_img ]] && [[ -f $uos_img.xz ]]; then echo "Moving $uos_img to $uos_img.old."; mv $uos_img $uos_img.old; fi
+    if [[ $? -eq 0 ]]; then
```
dbkinder: Which command execution status do you think this it checking?   It's not the ``curl``

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+        echo "Unxz UOS image: clear-$target_version-kvm.img.xz"
+        unxz clear-$target_version-kvm.img.xz
+    fi
+    if [[ $? -eq 0 ]]; then
```
dbkinder: I'm not sure which command execution status this checks either...

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+    cp -r /usr/lib/modules/"`readlink /usr/lib/kernel/default-iot-lts2018 | awk -F '2018.' '{print $2}'`.iot-lts2018" /mnt/lib/modules
+fi
+umount /mnt
+losetup -D
+sync
+
+cp -r /usr/share/acrn/samples/nuc/launch_uos.sh ~/
+sed -i "s/\(virtio-blk.*\)\/home\/clear\/uos\/uos.img/\1$(echo $uos_img | sed "s/\//\\\\\//g")/" ~/launch_uos.sh
+echo "Upgrade UOS done..."
+echo "Now you can run below command to start UOS..."
```
dbkinder:
```suggestion
echo "Now you can run this command to start UOS..."
```

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+# UOS download link
+uos_image_link="https://download.clearlinux.org/releases/$target_version/clear/clear-$target_version-kvm.img.xz"
+
+# Set proxy if needed.
+if [[ -n $proxy ]]; then
+    echo "Setting proxy to: $proxy"
+    export https_proxy=$proxy
+fi
+
+if [[ ! -z `findmnt /mnt -n` ]]; then umount /mnt; fi
```
gvancuts: I think it's a bit rude of the script to silently unmount `/mnt` if it is already mounted. Perhaps the user is using this for some other purpose. It would be better to exit gracefully and ask him to free it up himself.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+    if [[ $? -eq 0 ]]; then
+        uos_partition=`losetup -f -P --show $uos_img`p3
+    else
+        echo "Missing UOS file: $uos_img, please check if UOS is exist." && exit 1
+    fi
+    mount $uos_partition /mnt
+    cp -r /usr/lib/modules/"`readlink /usr/lib/kernel/default-iot-lts2018 | awk -F '2018.' '{print $2}'`.iot-lts2018" /mnt/lib/modules
+fi
+umount /mnt
+losetup -D
```
gvancuts: The user may be using other loop devices for other things, we shouldn't detach *all* of them unconditionally.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+# Copyright (C) 2019 Intel Corporation.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# This script is aim to quick setup sos and uos automatically.
+# You can also use this to upgrade sos with a specified clear linux version.
+# Also you should know that system may have boot problems once the script is running failed
+# since it will modify various system parameters with full root access.
+#
+# Usages:
+#   Upgrade sos to 28100
```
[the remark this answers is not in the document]
lirui34: It will update in next release.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+#     sudo <script> -t 28100
+#   Upgrade sos to 28100 with specified proxy server
+#     sudo <script> -t 28100 -p <your proxy server>:<port>
+#   Upgrade uos to 28100 with specified proxy server
+#     sudo <script> -t 28100 -p <your proxy server>:<port> -u
```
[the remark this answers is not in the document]
lirui34: Agree, I'll move it to the first option

doc/getting-started/gsg_quick_setup.sh
lirui34: OK, I will add some test in next release.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+
+
+function print_help()
+{
+echo "Usage:"
+echo "Launch this script as: sudo $0 -t 28100"
+echo -e "\t-t to specify clear linux version for upgrading"
+echo -e "\t-p to specify a proxy server"
+echo -e "\t-m to use swupd mirror url"
+echo -e "\t-u to upgrade UOS."
```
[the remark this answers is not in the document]
lirui34: I changed usage to: -s to upgrade SOS and -u to upgrade UOS.  and move original -s to -k for skip download, add -d to disable auto reboot after upgrade SOS.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+                s) skip_download_uos=1
+                        ;;
+                h) print_help
+                        ;;
+                ?) print_help
+                        ;;
+        esac
+done
+
+# -t opt is must
```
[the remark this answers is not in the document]
lirui34: Changed options check methods.

doc/getting-started/gsg_quick_setup.sh
dbkinder: Should this line ``echo "Usage:"`` (and all the script lines in the body of this function) have a tab before them?  I don't see one.

doc/getting-started/gsg_quick_setup.sh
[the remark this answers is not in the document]
lirui34: I am now considering to let user determine ESP path(SATA EFI path as default) before running this script just like this:
`# You should set EFI path before upgrading SOS.`
 `efi_partition=/dev/sda1`
` # Check EFI path exists.`
` [[ ! -b /dev/sda1 ]] && echo "You didn't choose the right EFI path." && exit 1`
For the reason, it's hard to auto determine the right ESP on a multiple OS(hard disk) device and also we need this script in automation test framework for some auto flashing cases...but I did not find a good resolution except that to handle such case.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+                        ;;
+                s) skip_download_uos=1
+                        ;;
+                h) print_help
+                        ;;
+                ?) print_help
+                        ;;
+        esac
+done
+
```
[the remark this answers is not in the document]
lirui34: Yes, this line will check if script is running with `sudo`:
`[[ $EUID -ne 0 ]] && echo "You have to run script as root." && exit 1`

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+if [[ $upgrade_uos == 1 ]]; then
+    echo "Upgrading UOS to: $target_version ..."
+    upgrade_uos
+    exit
+fi
+
+# Exit script if you specified an newer clear linux version.
+if [[ $(echo -e `swupd info` | cut -d' ' -f3) -gt $target_version ]]; then
+    echo -e 'No need to upgrade Service OS.' && exit
```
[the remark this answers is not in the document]
lirui34: Yes, update message in next release.

doc/getting-started/gsg_quick_setup.sh
lirui34: Update message in next release, it's checking whether line: `cp -r $acrn_conf_path /mnt/loader/entries/` is running successfully.

doc/getting-started/gsg_quick_setup.sh
lirui34: Add several tests for the previous processing of upgrade SOS, if any one of them fail, the script will stopped so that it will not reboot device.
Also I have added disable_reboot parameter to let user skip rebooting after upgrade.

doc/getting-started/gsg_quick_setup.sh
lirui34: Yes, it's removed in next release.

doc/getting-started/gsg_quick_setup.sh
lirui34: Remove this check event in next release.

doc/getting-started/gsg_quick_setup.sh
lirui34: Also remove it in next release.

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,277 @@
...
+        done
+    done
+else
+    efi_partition=`fdisk -l | grep "$type_efi" | awk -F' ' '{print $1}'`
+fi
+
+if [[ -z $efi_partition ]]; then echo "You didn't choose a partition, please try it again."; exit 1; fi
+efi_mount_point=`findmnt $efi_partition -n | cut -d' ' -f1`
+root_partition=`echo $efi_partition | sed 's/[0-9]$/3/g'`
+partition=`echo $efi_partition | sed 's/[0-9]$//g'`
```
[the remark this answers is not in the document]
lirui34: Fixed

doc/getting-started/gsg_quick_setup.sh
lirui34: Yes, it's fixed.

doc/getting-started/gsg_quick_setup.sh
lirui34: Removed this operation

doc/getting-started/gsg_quick_setup.sh
lirui34: I got it. it will fix in next release.

dbkinder:
@NanlinXie who should do the final approval of this PR and merge?

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,246 @@
...
+    echo "Upgrading SOS..."
+
+    # setup mirror and proxy url while specified with m and p options
+    [[ -n $mirror ]] && echo "Setting swupd mirror to: $mirror" && swupd mirror -s $mirror
+    [[ -n $proxy ]] && echo "Setting proxy to: $proxy" && export https_proxy=$proxy
+
+    # You should set EFI path before upgrading SOS.
+    efi_partition=/dev/sda1
+    # Check EFI path exists.
+    [[ ! -b $efi_partition ]] && echo "Please choose the right EFI path firstly." && exit 1
```
gvancuts: How can the user set the ESP to the right value? Is editing the script the only option?

We should add a note explaining this

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,246 @@
...
+    efi_mount_point=`findmnt $efi_partition -n | cut -d' ' -f1`
+    root_partition=`echo $efi_partition | sed 's/1$/3/g'`
+    partition=`echo $efi_partition | sed 's/1$//g;s/p$//g'`
+
+    # unmount efi partition if it's already mounted
+    if [[ -n $efi_mount_point ]]; then
+        echo "EFI partition is mounted: $efi_mount_point..."
+        echo "You should unmount it before upgrading SOS."
+        exit 1
```
gvancuts: We do not need to error out, we could simply use that mount point directly, couldn't we?

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,246 @@
...
+    if [[ $VERSION_ID -eq $sos_ver ]]; then
+        echo "Clear Linux version $sos_ver is already installed. Continuing to setup SOS..."
+    else
+        echo "Upgrading Clear Linux version from $VERSION_ID to $sos_ver ..."
+        swupd verify --fix --picky -m $sos_ver 2>/dev/null
+    fi
+
+    # Do the setups if previous process succeed.
+    if [[ $? -eq 0 ]]; then
+        echo "Add service-os kernel-iot-lts2018 bundle"
```
gvancuts:
```suggestion
        echo "Adding the service-os and kernel-iot-lts2018 bundles..."
```

doc/getting-started/gsg_quick_setup.sh
lirui34: Yes, we can use it directly.

doc/getting-started/gsg_quick_setup.sh
lirui34: Yes, so far it's the only option to edit script to set ESP.
I have also added a note at the head of the script to explain user should set the right ESP before running script.

NanlinXie:
@gvancuts , Do you have more comment for the script?

doc/getting-started/gsg_quick_setup.sh
gvancuts: We should really make this a command-line parameter and the explanation should be included in the "usage" section. People do not typically read scripts before running them. It's great to have details and explanations in them but for something that really has to be provided by the user, a parameter should be used (as opposed to editing the script).

doc/getting-started/gsg_quick_setup.sh
```diff
@@ -0,0 +1,248 @@
...
+#   Upgrade UOS to 28100 without downloading UOS image, you should put UOS image in /root directory previously.
+#     sudo <script> -u 28100 -k
+
+function print_help()
+{
+    echo "Usage:"
+    echo "Launch this script as: sudo $0 -s 28100"
+    echo -e "\t-s to upgrade SOS"
+    echo -e "\t-u to upgrade UOS"
+    echo -e "\t-p to specify a proxy server"
```
gvancuts: Nitpick:
```suggestion
    echo -e "\t-p to specify a proxy server (HTTPS)"
```

doc/getting-started/gsg_quick_setup.sh
lirui34: Thank you very much for suggestion, I have added a new parameter "-e" for user to specify the right ESP and also added "note" into print_help function.

lirui34:
> Very close!
Hi @gvancuts , I have added a new parameter "-e" for user to specify their ESP and also added "note" into print_help function.
Could you please have a look at those changes ?

Issue #2684, opened 2019-03-05
The "ACRN Shell Commands" documentation is out-of-date


gvancuts:
The [ACRN Shell Commands](https://projectacrn.github.io/latest/user-guides/acrn-shell.html) document is out of date and could do with some extra details.

Examples of outdated info: `vmexit`, `logdump` are no longer available

A couple of commands are also missing:
```
rdmsr           [-p<pcpu_id>] <msr_index> - rdmsr -p<pcpu_id> <msr_index>, msr_index in hexadecimal
wrmsr           [-p<pcpu_id>] <msr_index> <value> - wrmsr -p<pcpu_id> <msr_index> <value>, msr_index and value in hexadecimal

```

dbkinder:
Resolved.

PR #2815, opened 2019-03-19
doc: add rdmsr/wrmsr to the "ACRN Shell Commands" documentation

gvancuts:
Add 'rdmsr' and 'wrmsr' to the "ACRN Shell Commands" documentation. Both are
already described via the interactive help command but not (yet) included in our
on-line documentation.

Tracked-On: #2684

### DIFF
--- a/doc/user-guides/acrn-shell.rst
+++ b/doc/user-guides/acrn-shell.rst
@@ -44,3 +44,9 @@ The ACRN hypervisor shell supports the following commands:
      - Display the CPUID leaf [subleaf], in hexadecimal
    * - reboot
      - Trigger a system reboot (immediately)
+   * - rdmsr [-p<pcpu_id>] <msr_index>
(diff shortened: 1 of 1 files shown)


doc/user-guides/acrn-shell.rst
```diff
@@ -44,3 +44,9 @@ The ACRN hypervisor shell supports the following commands:
      - Display the CPUID leaf [subleaf], in hexadecimal
    * - reboot
      - Trigger a system reboot (immediately)
+   * - rdmsr [-p<pcpu_id>] <msr_index>
+     - Read the Model-Specific Register (MSR) at index ``msr_index`` (in hex)
```
dbkinder: better would be to use "hexadecimal" as used above (fix other uses of "hex" too)

gvancuts: > better would be to use "hexadecimal" as used above (fix other uses of "hex" too)

I agree... done

PR #2818, opened 2019-03-19
hv: minor fixes to a few calls to strncpy_s()

ZideChen0:
strncpy_s(d, dmax, s, slen): the 'dmax' includes the null terminator, while
slen doesn't. Thus if (dmax == slen == strlen(s)), strncpy_s() chooses to
discard the last character from s and instead write '\0' to d[dmax - 1].

strnlen_s(s, maxsize): if there is no terminating null character in the
first maxsize characters pointed to by s, strnlen_s() returns maxsize.

So in the following example or similar cases, we need to increase the size
of d[] by 1 to accommodate the null terminator, and add '1' to the dmax
argument to strncpy_s().

uint8_t d[MAX_LEN];
size = strnlen_s(s, MAX_LEN);
strncpy_s(d, MAX_LEN, s, size);

Tracked-On: #861

### DIFF
--- a/hypervisor/boot/sbl/multiboot.c
+++ b/hypervisor/boot/sbl/multiboot.c
@@ -18,7 +18,7 @@
  * - cmdline from acrn stitching tool. mod[0].mm_string
  * We need to merge them together
  */
-static char kernel_cmdline[MEM_2K];
+static char kernel_cmdline[MEM_2K + 1U];
 
 /* now modules support: FIRMWARE & RAMDISK & SeedList */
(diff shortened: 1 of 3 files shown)


HuiHuangShi:

hypervisor/boot/sbl/multiboot.c
```diff
@@ -18,7 +18,7 @@
  * - cmdline from acrn stitching tool. mod[0].mm_string
  * We need to merge them together
  */
-static char kernel_cmdline[MEM_2K];
+static char kernel_cmdline[MEM_2K + 1U];
```
dongyaozu: We should define a new MACRO for the size of the CMDLINE, not MEM_2K

PR #2819, opened 2019-03-20
doc: add some rules related to language extensions

shiqingg:
This patch adds some rules related to language extensions.

Tracked-On: #861

### DIFF
--- a/doc/developer-guides/coding_guidelines.rst
+++ b/doc/developer-guides/coding_guidelines.rst
@@ -3310,3 +3310,47 @@ The number of bytes in an object is implementation-defined, according to J.3.13
 item 2 in C99. For ACRN hypervisor, char is 1 byte, short is 2 bytes, int is 4
 bytes, long is 8 bytes, and long long is not used.
 
+Language Extensions
+*******************
+
+
+LE-01: Use of inline Assembly language in C Code is allowed
+===========================================================
+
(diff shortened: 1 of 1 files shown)


doc/developer-guides/coding_guidelines.rst
```diff
@@ -3310,3 +3310,47 @@ The number of bytes in an object is implementation-defined, according to J.3.13
 item 2 in C99. For ACRN hypervisor, char is 1 byte, short is 2 bytes, int is 4
 bytes, long is 8 bytes, and long long is not used.
 
+Language Extensions
+*******************
+
+
+LE-01: Inline Assembly language in C Code is used
```
dbkinder: Should these all say "may be used" (are you giving permission as part of the coding standard)?
Also, can some of these be combined, for example  put all the type attributes together, and the build-in types and functions together?

shiqingg: Combination of these features is allowed.

Could you please provide more details about this question? Are you talking about the GCC manual?
`are you giving permission as part of the coding standard?`

shiqingg:
@dbkinder Thanks for your review David!
I updated the statements to be 'Use of *** is allowed' in new version. Other comments are inline.
Could you please help review again?
Thanks!

Issue #2792, opened 2019-03-15
Pass ACRN E820 map to OVMF


peterfang:
ACRN currently uses [4GB, 5GB) as its 64-bit PCI host aperture. This is inconsistent with OVMF's assumption of its platform's memory layout, because it fetches the size of high memory from RTC CMOS.

Pass ACRN's E820 map to OVMF so it can correctly pass ACRN's memory configuration to UOS.

peterfang:
[External_System_ID] ACRN-2931

Mingyuan18:
Can't reproduce this issue in below build:
Installed version: 28600
Kernel: 4.19.31-39
daily tag:”2019w14.1.140000p”

PR #2828, opened 2019-03-20
Pass ACRN E820 map to OVMF

peterfang:
OVMF requires a more descriptive mechanism than RTC CMOS to retrieve ACRN's memory layout, so we now pass the E820 map to it, starting at 0xEF000 (ROM area).

ACRN currently uses [4GB, 5GB) as its 64-bit PCI host aperture. This is inconsistent with OVMF's assumption of its platform's memory layout, because it derives the size of high memory from RTC CMOS, which is incapable of describing the 64-bit PCI hole.

By default, OVMF uses RTC CMOS 0x5b/0x5c/0x5d to determine the size of high memory. This value only tells OVMF how much memory is above 4GB, but not the platform's memory layout above 4GB.

Using RTC CMOS works for QEMU, because QEMU places its 64-bit PCI host aperture above its highmem. Therefore, OVMF can always assume highmem is located at [4GB, 4GB + highmem), which is not where ACRN's highmem is located. For example, if we have 1GB of usable memory above 4GB, ACRN will place it at [5GB, 6GB).

This patchset allows OVMF to correctly identify the guest's memory layout. It will consider any reserved region above 4GB as 64-bit PCI host aperture. It also adds the MMIO fallback handler to the 64-bit PCI host aperture to prevent the guest from inadvertently crashing acrn-dm.

v1 -> v2:
- provide more explanation to this patchset
- add signature before E820 map for OVMF backward compatibility
- add the MMIO fallback handler to the 64-bit PCI host aperture

Tracked-On: #2792

### DIFF
--- a/devicemodel/core/mem.c
+++ b/devicemodel/core/mem.c
@@ -37,11 +37,14 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <assert.h>
+#include <string.h>
 #include <pthread.h>
 
 #include "vmm.h"
 #include "mem.h"
 #include "tree.h"
+
+#define MEMNAMESZ (80)
 
 struct mmio_rb_range {
 	RB_ENTRY(mmio_rb_range)	mr_link;	/* RB tree links */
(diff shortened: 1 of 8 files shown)


HuiHuangShi:

Issue #2708, opened 2019-03-08
one binary for SBL and UEFI


Mingyuan18:
One ACRN binary should both support UEFI and SBL firmware.

Mingyuan18:
[External_System_ID] ACRN-2939

Mingyuan18:
This issue could not be reproduced on clearlinux 28340 (acrn-hypervisor: acrn-2019w11.4-140000p).

PR #2841, opened 2019-03-22
doc: minor changes according to one binary for SBL and UEFI

totoroow:
PLATFORM=xxx is no longer supported, use BOARD instead.

Tracked-On: #2708

### DIFF
--- a/doc/tutorials/using_partition_mode_on_up2.rst
+++ b/doc/tutorials/using_partition_mode_on_up2.rst
@@ -306,7 +306,7 @@ Enable partition mode in ACRN hypervisor
 
    .. code-block:: none
 
-     $ make PLATFORM=sbl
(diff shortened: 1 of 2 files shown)


gvancuts:
This LGTM, but I would also like to have @ttzeng 's opinion on this since he's more familiar with that content than I am.

As a side note, we should probably remove the `pci_devices_ignore=(0:18:1)` argument in the Ubuntu tutorial (does not have to be done in this PR).

dbkinder:
@ttzeng @NanlinXie Can this be merged now?  @gvancuts wanted approval from @ttzeng

gvancuts:
It's been sitting there for 10 days so let's merge this for now. I'm still hoping @ttzeng will test this and he can follow-up later with patches if needed.

PR #2910, opened 2019-04-09
doc: add rules in coding guidelines

shiqingg:
This patch adds rules in coding guidelines.

Tracked-On: #861


shiqingg:
> OK, but it would seem like recursion is a valuable coding practice for some capabilities.

Hi David, sometimes, the depth of the recursion is not easy to predict. This may cause some undefined behaviors. So, we would like to disable the usage of recursion.

PR #2956, opened 2019-04-15
doc: Add enable serial console info about KBL on the GSG 

lirui34:
Add enable serial console info about KBL on the GSG


NanlinXie:

lirui34:
will create new one

Issue #1842, opened 2018-11-19
ACRN modulization


JasonChenCJ:
According to ACRN HLD https://projectacrn.github.io/latest/developer-guides/hld/hld-overview.html#acrn-architecture, we organize ACRN arch as different components like lib, hwinit, hwmgmt, vcpu, passthru, dm, io, vmmgmt and debug.
This story is used to trace the modulization progress for these different components.

JasonChenCJ:
[External_System_ID] ACRN-1750

NanlinXie:
Closed with no regression issue.

PR #2981, opened 2019-04-19
Vlapic: minor fix about APICv feature setting

lifeix:
1) Shouldn't try to set APIC-register virtualization if the physical doesn't
    support APICV advanced mode.
    2) Remove all APICv features VMCS setting when LAPIC is passed through to guest.

    However, if emulated this in MRB board, the android can't boot because when switch to trusty world, 
    it will check  "Delivery Status" in ICR first. It turns out that this Bit Test instruction is not emulated  
    correctly.  This's because not every instruction supports the operand-size bit (w). This patch try to 
    correct what done in commit-id 9df8790 by setting a flag VIE_OP_F_BYTE_OP to indicate which
    instruction supports the operand-size bit (w).

    Tracked-On: #1842

### DIFF
--- a/hypervisor/arch/x86/guest/instr_emul.c
+++ b/hypervisor/arch/x86/guest/instr_emul.c
@@ -84,15 +84,23 @@
 #define VIE_OP_TYPE_TEST	15U
 
 /* struct vie_op.op_flags */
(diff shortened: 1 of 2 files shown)


HuiHuangShi:
out of time

HuiHuangShi:

Issue #2737, opened 2019-03-12
Build ACRN failed in Ubuntu 16.04


Mingyuan18:
GSG for Build hypervisor section is out of date, need update.

Mingyuan18:
[External_System_ID] ACRN-2969

gvancuts:
@Mingyuan18 , can you be more specific as to what is out of date? (I must confess I haven't tried with Ubuntu 16.04 for a while so if you can give me some pointers as to the problems you encountered it would be helpful)

wenlingz:
Please refer the fixed method:

run these command:

apt-get install openwince-include 

cp /usr/include/openwince/arm/sa11x0/gpio.h  /usr/include/linux/

fixed.

gvancuts:
There are a number of issues with the current set of instructions for Ubuntu 16.04.
* Missing instructions on how to upgrade `binutils` to >=2.26
* Missing the `pkg-config` package
* Missing header file (for GPIO)

@wenlingz , I followed your instructions above to add the missing `gpio.h` header file but I'm now hitting the following error message:
```
In file included from hw/pci/virtio/virtio_gpio.c:20:0:
/usr/include/linux/gpio.h:42:10: fatal error: common.h: No such file or directory
 #include <common.h>
          ^~~~~~~~~~
compilation terminated.
```

gvancuts:
I did a `cp /usr/include/openwince/common.h /usr/include/` to get around this new error and I am now hitting the following error:
```
cc -g -O0 -std=gnu11 -D_GNU_SOURCE -DNO_OPENSSL -m64 -Wall -ffunction-sections -Werror -O2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -fno-strict-aliasing -fpie -I/root/acrn/devicemodel/include -I/root/acrn/devicemodel/include/public -I/root/acrn/build/devicemodel/include -I/root/acrn/build/tools -fstack-protector-strong -DDM_DEBUG -c hw/pci/virtio/virtio_gpio.c -o /root/acrn/build/devicemodel/hw/pci/virtio/virtio_gpio.o
hw/pci/virtio/virtio_gpio.c: In function 'native_gpio_update_line_info':
hw/pci/virtio/virtio_gpio.c:298:23: error: storage size of 'info' isn't known
  struct gpioline_info info;
                       ^~~~
[...]

Issue #2457, opened 2019-01-30
Compilation on Fedora 29 (for UEFI platforms) is broken


gvancuts:
Attempting to compile the ACRN hypervisor for UEFI platforms on Fedora 29 will fail as follows:
```
<snip>
echo "building hypervisor as EFI executable..."
building hypervisor as EFI executable...
make -C /home/gvancuts/work/acrn/acrn-hypervisor/efi-stub HV_OBJDIR=/home/gvancuts/work/acrn/acrn-hypervisor/build/hypervisor EFI_OBJDIR=/home/gvancuts/work/acrn/acrn-hypervisor/build/efi-stub
make[1]: Entering directory '/home/gvancuts/work/acrn/acrn-hypervisor/efi-stub'
find: ‘/usr/libexec/initscripts/legacy-actions/auditd’: Permission denied
dirname: missing operand
Try 'dirname --help' for more information.
[ ! -e /home/gvancuts/work/acrn/acrn-hypervisor/build/efi-stub/boot.o ] && mkdir -p /home/gvancuts/work/acrn/acrn-hypervisor/build/efi-stub/; \
cc  -I/home/gvancuts/work/acrn/acrn-hypervisor/build/hypervisor/include -I. -c -I. -I.. -I../hypervisor/include/arch/x86/guest -I/usr/include/efi -I/usr/include/efi/x86_64 -I../hypervisor/include/public -I../hypervisor/include/lib -I../hypervisor/bsp/include/uefi -DEFI_FUNCTION_WRAPPER -fPIC -fshort-wchar -ffreestanding -Wall -I../fs/ -Dx86_64 -O2 -include config.h -mno-mmx -mno-sse -mno-sse2 -mno-80387 -mno-fp-ret-in-387 -mno-red-zone  boot.c -o /home/gvancuts/work/acrn/acrn-hypervisor/build/efi-stub/boot.o
In file included from boot.c:34:
/usr/include/efi/efi.h:41:10: fatal error: efibind.h: No such file or directory
 #include "efibind.h"
          ^~~~~~~~~~~
compilation terminated.
make[1]: *** [Makefile:114: /home/gvancuts/work/acrn/acrn-hypervisor/build/efi-stub/boot.o] Error 1
make[1]: Leaving directory '/home/gvancuts/work/acrn/acrn-hypervisor/efi-stub'
make: *** [Makefile:52: hypervisor] Error 2
```

[...]

PR #3014, opened 2019-04-24
doc: update instructions for building ACRN on Ubuntu 16.04

gvancuts:
Update the instructions on how to build ACRN on Ubuntu 16.04,
specifically:
* Add 'pkg-config' package
* Upgrade 'binutils' to 2.27

Note: there is still an issue with regards to the GPIO header
(linux/gpio.h) which needs to be resolved.

Tracked-On: #2737

### DIFF
--- a/doc/getting-started/building-from-source.rst
+++ b/doc/getting-started/building-from-source.rst
@@ -41,7 +41,8 @@ each with their own way to install development tools:
           python3 \
           python3-pip \
           libblkid-dev \
-          e2fslibs-dev
+          e2fslibs-dev \
+          pkg-config
      $ sudo pip3 install kconfiglib
 
   .. note::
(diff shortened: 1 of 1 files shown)


doc/getting-started/building-from-source.rst
```diff
@@ -57,6 +58,20 @@ each with their own way to install development tools:
         $ sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 60 \
                              --slave /usr/bin/g++ g++ /usr/bin/g++-7
 
+  .. note::
+     You need a newer version of ``binutils`` on Ubuntu 16.04. The latest
```
dbkinder: This would be better written as:
```
ACRN development requires ``binutils`` version 2.27 (or higher).
You can verify your version of ``binutils`` with the command 
``apt-show-versions binutils``. While Ubuntu 18.04 has a new version
of ``binutils`` the default version on Ubuntu 16.04 needs updating (see
issue `#1133 <https://github.com/projectacrn/acrn-hypervisor/issues/1133>`_ ).
```

Reading through this whole doc though, I don't see us mentioning anything other than Ubuntu 16.04.  Should this doc be updated to mention 18.04?

ttzeng:
@gvancuts Hi Geoffroy, instead of fixing any incompatibility on legacy Ubuntu OSes which pulled the old toolchains, do we consider to recommend building the code in containers? that may reduce the discrepancies when users build the code not using Clear Linux?

doc/getting-started/building-from-source.rst
gvancuts: Possibly, I haven't tried building on Ubuntu 18.04 for a while and was running out of time after debugging and updating the 16.04 instructions. I decided it was better to at least update this portion for now rather than wait to have a more complete update.

It's been a while since I've tried the various distros we list to be honest, it could be that various sections need updating. Fedora 29 for example needs a patch (see #2457) but that is also not documented anywhere yet.

gvancuts: > This would be better written as:

I have updated the PR with the text you suggested.

> You can verify your version of ``binutils`` with the command 
> ``apt-show-versions binutils``

Except this part as I do not have `apt-show-versions` available in my container for example, so instead of adding the tool, I used `apt show binutils` which does the same thing out of the box.

gvancuts:
> @gvancuts Hi Geoffroy, instead of fixing any incompatibility on legacy Ubuntu OSes which pulled the old toolchains, do we consider to recommend building the code in containers? that may reduce the discrepancies when users build the code not using Clear Linux?

I think we should add such instructions, this would indeed be very valuable.

There is some merit to having instructions on how to compile natively on different Linux distributions but I'm not sure we pay enough attention to these to keep them up-to-date.

PR #3040, opened 2019-04-29
doc: How to build ACRN on Fedora 29

acrnsi:
Update FAQ about compilation issue workaround on Fedora 29

Signed-off-by: zhangyun <yunxzhang@intel.com>

Tracked-On: #2457

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -145,3 +145,39 @@ static memory allocation. This is why ACRN removed all ``malloc()``-type code,
 and why it needs to pre-identify the size of all buffers and structures used in
 the Virtual Memory Manager. For this reason, knowing the available RAM size at
 compile time is necessary to statically allocate memory usage.
+
+
+How to build ACRN on Fedora 29?
+*******************************
+
+There is a known issue when attempting to build ACRN on Fedora 29
+because of how ``gnu-efi`` is packaged in this Fedora release.
+(See the `ACRN GitHub issue
+<https://github.com/projectacrn/acrn-hypervisor/issues/2457>`_ 
+for more information.)  The following patch to ``/efi-stub/Makefile``
+fixes the problem on Fedora 29 development systems (but should
+not be used on other Linux distros)::
+
+   diff --git a/efi-stub/Makefile b/efi-stub/Makefile
+   index 5b87d49b..dfc64843 100644
+   --- a/efi-stub/Makefile
+   +++ b/efi-stub/Makefile
+   @@ -52,14 +52,14 @@ endif
+    # its tools and libraries in different folders. The next couple of
+    # variables will determine and set the right path for both the
+    # tools $(GNUEFI_DIR) and libraries $(LIBDIR)
+   -GNUEFI_DIR := $(shell find $(SYSROOT)/usr/lib* -name elf_$(ARCH)_efi.lds -type f | xargs dirname)
+   +GNUEFI_DIR := $(shell find $(SYSROOT)/usr/lib* -name elf_x64_efi.lds -type f | xargs dirname)
+   LIBDIR := $(subst gnuefi,,$(GNUEFI_DIR))
+   -CRT0 := $(GNUEFI_DIR)/crt0-efi-$(ARCH).o
+   -LDSCRIPT := $(GNUEFI_DIR)/elf_$(ARCH)_efi.lds
+   +CRT0 := $(GNUEFI_DIR)/crt0-efi-x64.o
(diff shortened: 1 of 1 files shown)


doc/faq.rst
```diff
@@ -1,147 +1,177 @@
...
+dynamic allocation failures. Real-time applications similarly benefit from
+static memory allocation. This is why ACRN removed all ``malloc()``-type code,
+and why it needs to pre-identify the size of all buffers and structures used in
+the Virtual Memory Manager. For this reason, knowing the available RAM size at
+compile time is necessary to statically allocate memory usage.
+
+How to workaround compilation issue on Fedora 29?
+*************************************************
+
+Please refer to following information for workaround::
```
gvancuts: I'll let @dbkinder recommend the exact phrasing but we should:
1. Give a short summary of the issue, e.g.: "There is a known issue when attempting to build ACRN on Fedora 29. This is due to the way `gnu-efi` is packaged up in Fedora.
2. Add a reference to the issue on Github some people can see the whole thread/discussion/history

doc/faq.rst
```diff
@@ -1,147 +1,177 @@
...
+
+Important features for ACRN include Functional Safety (FuSa) and real-time
+behavior. FuSa requires a static allocation policy to avoid the potential of
+dynamic allocation failures. Real-time applications similarly benefit from
+static memory allocation. This is why ACRN removed all ``malloc()``-type code,
+and why it needs to pre-identify the size of all buffers and structures used in
+the Virtual Memory Manager. For this reason, knowing the available RAM size at
+compile time is necessary to statically allocate memory usage.
+
+How to workaround compilation issue on Fedora 29?
```
gvancuts: How about something simpler like: `How to build ACRN on Fedora 29`?

dbkinder:
@gvancuts @NanlinXie "The odd thing about this PR is that the diff shows the entire file was changed, while only a few lines were added at the very end of it... odd."  

That's because the current checked-in file uses Linux line endings (LF), while the new file uses Windows line ending (CR LF).  We should probably add a ``.gitattributes`` file to the project to keep this straight (see https://help.github.com/en/articles/dealing-with-line-endings).  We really don't want to have a mixture of Linux/Windows text file line endings in the repo.

doc/faq.rst
dbkinder: Interesting that you changed some uses of the ``ARCH`` variable to be hard coded as ``x64``, but not all of them?  In any case, yes, this needs an intro as @gvancuts commented.  Something like:

```
How to build ACRN on Fedora 29
******************************

There is a known issue when attempting to build ACRN on Fedora 29
because of how ``gnu-efi`` is packaged in this Fedora release.
(See the `ACRN GitHub issue
<https://github.com/projectacrn/acrn-hypervisor/issues/2457>`_ 
for more information.)  The following patch to ``/efi-stub/Makefile``
fixes the problem on Fedora 29 development systems (but should
not be used on other Linux distros)::
```

gvancuts: > Interesting that you changed some uses of the `ARCH` variable to be hard coded as `x64`, but not all of them?

I wanted to fix the paths and names of utilities. The only other `ARCH` that's left is part of the `CFLAGS` and used as a macro. I don't actually know what that does, if anything at all, so I did not touch it :-)

gvancuts:
> @gvancuts @NanlinXie "The odd thing about this PR is that the diff shows the entire file was changed, while only a few lines were added at the very end of it... odd."
> 
> That's because the current checked-in file uses Linux line endings (LF), while the new file uses Windows line ending (CR LF). We should probably add a `.gitattributes` file to the project to keep this straight (see https://help.github.com/en/articles/dealing-with-line-endings). We really don't want to have a mixture of Linux/Windows text file line endings in the repo. (Currently there are no files in the ACRN repo with windows line endings.)

I think we should add this `.gitattributes` file.

acrnsi:
@gvancuts @dbkinder thanks for your suggestion. I have been updated,please review.

dbkinder:
Oh, and thanks for fixing the line endings :)

gvancuts:
> I guess folks that know how to patch a source file will know what to do with this. (I was thinking you'd provide a patch file and give them the command like to apply the patch.) @gvancuts What do you think?

I think it would be better indeed. But I also just tried to copy-paste that content in a `gnu-efi-fed29.patch` and apply it using `git apply` (or `patch -p1 < gnu-efi-fed29.patch`) and it complains that line 11 is malformed. At this stage, it's probably easier to merge this as-is, it's close enough in my opinion.

PR #3043, opened 2019-04-29
 tools: acrn-crashlog: refine IO operation

lauxinw:
This patch set aims to refine the IO operation on the eMMC.
Check a folder's size need to traverse all nodes of files and subdirs,
in the case of extreme memory strain, accessing these log file nodes
frequently will highly increase the pressure of io operation.
Previously, all log files will be accessed and counted in each space
check, this patch aims to refine it by only calculating the increment of new
logs.

In order to achieve this purpose, acrnprobe must maintain one variate to
record the total size of all logs, and modify it after each time
log changed. To avoid the using of lock, the access of this variate must
be single-threaded.
So, firstly, these serial patches unify the handling process of each kind
of events to:
analyze events - request resources -  collect logs - record current log size
and then make the change of log space check.


HuiHuangShi:

PR #3048, opened 2019-04-30
dm: uart_core: fix uart dm cause interrupt storm issue

liuyuan-GitHub:
For Linux kernel, the UART driver will disable IER.THRE. But for Windows, it
will keep IER.THRE turn on then cause interrupt storm.

From pc16550d UART spec, INTR PIN description:

"Interrupt. This pin goes high whenever any one of the following interrupt types
has an active high condition and is enabled through the IER Receiver Error Flag;
Received Data Available timeout (FIFO Mode only); Transmitter Holding Register
Empty; and MODEM Status. The INTR signal is reset low upon the appropriate
interrupt service or a Master Reset operation."

And Interrupt Reset Control of Transmitter Holding Register Empty Interrupt
description:

"Reading the IIR Register (if source of interrupt) or Writing into
the Transmitter Holding Register"

The datasheet hasn't describe very clear if the THRE interrupt will be
re-generate after "Reading the IIR Register". We assume it will not do
that, so the THRE interrupt only generate when THR turn to empty.

This patch follows above assumption to resolve the interrupt storm cause WaaG
boot failed issue.

Tracked-On: #2713

### DIFF
--- a/devicemodel/hw/uart_core.c
+++ b/devicemodel/hw/uart_core.c
@@ -434,6 +434,9 @@ uart_write(struct uart_vdev *uart, int offset, uint8_t value)
 		uart->thre_int_pending = true;
 		break;
 	case REG_IER:
+		if (((uart->ier & IER_ETXRDY) == 0) &&
+				((value & IER_ETXRDY) != 0))
+			uart->thre_int_pending = true;
 		/*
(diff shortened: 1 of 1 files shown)


HuiHuangShi:

Issue #1917, opened 2018-11-27
Documentation: What does ',b' flag do with virtio-blk?


mattcurfman:
In the launch script @ https://github.com/projectacrn/acrn-hypervisor/blob/master/devicemodel/samples/apl-mrb/launch_uos.sh#L118, there is a ‘,b’ flag added to the virtio-blk parameters for acrn-dm, but this flag is missing in the NUC launch script @ https://github.com/projectacrn/acrn-hypervisor/blob/master/devicemodel/samples/nuc/launch_uos.sh

Further, there is no documentation for this flag at https://projectacrn.github.io/latest/developer-guides/hld/virtio-blk.html

What does the ‘,b’ flag do?

yifanddd:
[External_System_ID] ACRN-3378

Mingyuan18:
@mattcurfman , Could you please confirm whether this problem still exists? If it does not exist, can you close it?

gvancuts:
This issue is still valid as far as I can see, the script uses the `,b` flag as part of the `virtio-blk` options but it is not described anywhere (that I could easily find) in our documentation.

Mingyuan18:
Patch merged. So close this issue.

PR #3086, opened 2019-05-08
Doc: virtio-blk: add description of boot device option

binbinwu1:
Add the description of option 'b' to indicate the boot device including
the bootable image.

Tracked-On: #1917

### DIFF
--- a/doc/developer-guides/hld/virtio-blk.rst
+++ b/doc/developer-guides/hld/virtio-blk.rst
@@ -63,8 +63,11 @@ Usage:
 
 The device model configuration command syntax for virtio-blk is::
 
-   -s <slot>,virtio-blk,<filepath>[,options]
+   -s <slot>,virtio-blk,[,b,]<filepath>[,options]
 
+- ``b``: when using ``vsbl`` as the virtual bootloader, use this
+  immediately after ``virtio-blk`` to specify it as a bootable
+  device and the bootable image location.
 - ``filepath`` is the path of a file or disk partition
 - ``options`` include:
 


doc/developer-guides/hld/virtio-blk.rst
```diff
@@ -65,6 +65,9 @@ The device model configuration command syntax for virtio-blk is::
 
    -s <slot>,virtio-blk,<filepath>[,options]
 
+- ``b``: needed when use vsbl as the virtual bootloader, to indicate the
```
dbkinder: Can you provide an example use of this ``b`` parameter?  Where do you put the ``b`` (and its parameters) in the syntax description above?

Also, edit the wording to be something like:
```
- ``b``: when using ``vsbl`` as the virtual bootloader, this indicates the
  bootable device and the bootable image location. If used, it must be
  the first option.
```

binbinwu1: updated, please help to have a check.

doc/developer-guides/hld/virtio-blk.rst
```diff
@@ -63,8 +63,11 @@ Usage:
 
 The device model configuration command syntax for virtio-blk is::
 
-   -s <slot>,virtio-blk,<filepath>[,options]
+   -s <slot>,virtio-blk,[,b,]<filepath>[,options]
 
+- ``b``: needed when use ``vsbl`` as the virtual bootloader, to indicate the
```
dbkinder: ah OK.  This wording would be better:
```
- ``b``: when using ``vsbl`` as the virtual bootloader, use this
  immediately after ``virtio-blk`` to specify it as a bootable
  device and the bootable image location.
```

binbinwu1: thanks for advice.

gvancuts:
I thought ACRN used `vSBL` by default to boot Linux-as-a-Guest, but we do not use this option in the launch script on the NUCs. What am I missing?

binbinwu1:
> I thought ACRN used `vSBL` by default to boot Linux-as-a-Guest, but we do not use this option in the launch script on the NUCs. What am I missing?

It is not using vSBL to boot Linux guest on NUCs. vSBL is dedicated for APL-MRB.
For open-source project, OVMF will take the role as the virtual bootloader eventually.

Issue #3085, opened 2019-05-08
Can't set tsc frequency through option -f/--frequency of acrnalyze.py


KaigeFu:
It doesn't take effect when set tsc frequency through -f/--frequency of acrnalyze.py

KaigeFu:
[External_System_ID] ACRN-3651

Mingyuan18:
No regression issue fond in QA test:
SOS Installed version: 29360
ACRN tag: 2019w19.5.140000p

PR #3087, opened 2019-05-08
tools: Fix the non-effective option -f --frequency

KaigeFu:
We use global variable TSC_FREQ defined in config.py in each of the analyzers.
And the value set to TSC_FREQ through option -f/--frequency doesn't take effect.

This patch fix it by adding one parameter to each of the analyzers and pass
the value set by -f/--frequency to it instead of use the global one.

Tracked-On: #3085


HuiHuangShi:

Issue #1337, opened 2018-09-26
ACRN instruction decoding gets wrong opsize when the real opsize is 1 byte


fyin1:
instruction decoding gets wrong operand size when the real operand size is 1 byte (like movsb instruction).

Mingyuan18:
This issue was fixed in acrn : acrn-2018w39.5-140000p; sos: 25180

PR #3101, opened 2019-05-13
hv: instr_emul: fix movzx return memory opsize wrong

lifeix:
There're some instructions which not support bit 0(w bit) flag but which
memory opcode size is fixed and the memory opcode size is not equal to the
register opcode size. In our code, there is movzx (which opcode is 0F B7)
which memory opcode size is fixed to 16 bits. So add a flag VIE_OP_F_WORD_OP
to indicate a instruction which memory opcode size is fixed to 16 bits.

Tracked-On: #1337


HuiHuangShi:

PR #3124, opened 2019-05-15
doc: Update clearlinux os installation guide link.

lirui34:
bare-metal-install is replaced with bare-metal-install-server due to an recent change by clearlinux.

Signed-off-by: ruix.li <ruix.li@intel.com>


acrnsi:
Ok to verify

PR #3176, opened 2019-05-24
doc: update function naming convention

shiqingg:
This patch updates coding guidelines about naming convention.

### DIFF
--- a/doc/developer-guides/c_coding_guidelines.rst
+++ b/doc/developer-guides/c_coding_guidelines.rst
@@ -3501,7 +3501,13 @@ function. Some detailed rules are listed below:
    enough with prior rules, it shall be named with the module name as prefix, such
    as 'vie_read_mmio'.
 5) If the function is a wrapper of inline Assembly codes, it shall be named with
-   the format 'asm_<Assembly instruction mnemonic>', such as 'asm_pause'.
+   one of the following formats:
+
+   a) asm_<Assembly instruction mnemonic>, such as 'asm_pause'.
+   b) If the Assembly instruction mnemonic does not clearly indicate the
+      purpose of the function or the function includes multiple Assembly
+      instruction statements, the function shall be named with 'asm\_' as
+      prefix and apply the other non-assembly function naming rules.
 6) <nouns> mentioned in prior rules may either be one noun or multiple nouns, as
    long as it could clearly illustrate the object.
 


acrnsi:
OK to verify

doc/developer-guides/c_coding_guidelines.rst
```diff
@@ -3501,7 +3501,13 @@ function. Some detailed rules are listed below:
    enough with prior rules, it shall be named with the module name as prefix, such
    as 'vie_read_mmio'.
 5) If the function is a wrapper of inline Assembly codes, it shall be named with
-   the format 'asm_<Assembly instruction mnemonic>', such as 'asm_pause'.
+   one of the following formats:
+
+   a) asm_<Assembly instruction mnemonic>, such as 'asm_pause'.
+   b) If the Assembly instruction mnemonic could not clearly indicate the
```
dbkinder:
```suggestion
   b) If the Assembly instruction mnemonic does not clearly indicate the
```

doc/developer-guides/c_coding_guidelines.rst
```diff
@@ -3501,7 +3501,13 @@ function. Some detailed rules are listed below:
...
+      purpose of the function or the function includes multiple Assembly
+      instruction statements, the function shall be named with 'asm' as
```
dbkinder:
```suggestion
      instruction statements, the function shall be named with 'asm_' as
```

doc/developer-guides/c_coding_guidelines.rst
```diff
@@ -3501,7 +3501,13 @@ function. Some detailed rules are listed below:
...
+      prefix and apply the prior rules.
```
dbkinder:
```suggestion
      prefix and apply the other non-assembly function naming rules.
```

doc/developer-guides/c_coding_guidelines.rst
shiqingg: Hi @dbkinder,
Do you think we should update the following two rules as well?

From:
C-NC-03: Data structures exposed to external components shall be named with prefix ‘acrn’
C-NC-04: Data structures only used by hypervisor shall be named with prefix ‘hv’
To:
C-NC-03: Data structures exposed to external components shall be named with prefix ‘acrn_’
C-NC-04: Data structures only used by hypervisor shall be named with prefix ‘hv_’

shiqingg:
@dbkinder Thanks for your suggestion! New version has been pushed.
Could you please help review it again and let me know your opinion about the inline comment?
Thanks!

doc/developer-guides/c_coding_guidelines.rst
dbkinder: Yes, that would be a good change (to add the underscore), but be sure to escape the underscore so it doesn't look like a link:  'hv\_'

Issue #2962, opened 2019-04-16
Enabling WaaG on ACRN


Mingyuan18:
Enabling WaaG on ACRN

Mingyuan18:
[External_System_ID] ACRN-3403

Mingyuan18:
[External_System_ID] ACRN-3403

Mingyuan18:
The test result is pass in V1.1 release test.

PR #3208, opened 2019-05-31
dm: pci: fix the MMIO regions overlap when getting bar size

chejianj:
PCI spec says that the procedure of sizing a BAR is as follows:
1) disable the decode via command register
2) save the original value of BAR register
3) write all-1 to the BAR register
4) read BAR register back, calculate the size
5) restore the original value to BAR register
6) re-enable the decode via command register

Some driver does not disable the decode of BAR register via the
command register before sizing a BAR. This will lead to a overlay
of the BAR addresses when trying to register the intermediate
BAR address via register_bar. A stateful variable sizing is used
to keep track of such kind of BAR address changes and workaroud
this violation.

Currently this issue is only found when audio device is passed
through to Windows 10 guest. When it is fixed in the Windows
audio driver, this patch should be reverted.

v1 -> v2:
- change the commit message to add the procedure of BAR sizing from
  PCI spec

Tracked-On: #2962

### DIFF
--- a/devicemodel/hw/pci/core.c
+++ b/devicemodel/hw/pci/core.c
@@ -595,15 +595,17 @@ memen(struct pci_vdev *dev)
  */
 static void
 update_bar_address(struct vmctx *ctx, struct pci_vdev *dev, uint64_t addr,
-	int idx, int type)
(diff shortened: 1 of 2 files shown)


HuiHuangShi:

Issue #2987, opened 2019-04-19
HV: vuart refine


conghuic23:
As there are requirements for switching VM's console at hypervisor side and providing a mechanism to connect vuart between different VMs, so refine the vuart code.

In current implementation, two vuart are added to pre-launched VM's configuration, one is for console, the other is for communication. And these can only be used in debug mode. Will move the vuart code from hypervisor/debug/ to hypervisor/dm/, so that the release version can also use it.

Mingyuan18:
[External_System_ID] ACRN-3461

PR #3211, opened 2019-05-31
 HV: vuart: Bugfix for no interrupts on THRE

conghuic23:
When enable Transmitter Holding Register Empty interrupts, vuart should
trigger interrupt when the THR is empty. In the previous code, only
after the data is written to THR, the flag thre_int_pending is set to
true. The thre_int_pending should also be true after THRE is set.
Besides, add logic to make sure interrupt is in right state when vuart
is initiated.
Add vuart_deinit to vm shutdown so that the vuart resource can be
reset, and when the Guest VM restart, it could have right state.
Tracked-On: #2987


HuiHuangShi:

Issue #3217, opened 2019-06-04
Hypervisor hang when run syz_ic_reset_ptdev_irq_info.


Mingyuan18:
1.Environment
[Board]: APL UP2
Installed version: 28800
Version URL: https://cdn.download.clearlinux.org/update/
Content URL: https://cdn.download.clearlinux.org/update/
Kenrel Version: linux-iot-lts2018-4.19.34-45
acrn hypervisor tag: acrn-2019w15.4.140000p

Tools setup wiki: https://wiki.ith.intel.com/display/OTCCWPQA/syzkaller+enabling+on+ACRN 
We used Syzkaller ran with hypercall unit tests to do Fuzzing test for ACRN, which ran on SOS and communicate with DM process by socket.
enable_syscalls: ["syz_ic_assign_ptdev", "syz_ic_deassign_ptdev", "syz_ic_vm_pci_msix_remap", "syz_ic_set_ptdev_irq_info", "syz_ic_reset_ptdev_irq_info" ]

2. Reproduce Steps
setup env with wiki: And sync latest ACRN code and fuzzing tool code to your host
apply patch_for_fuzzing_on_dm.txt to devicemodel, and build images
modify acrn_build.sh based your own environment, and run it to rebuild syzkaller tool
flash images, and then make uos autoboot, remove sos password, crashlogctl enable
use acrn.cfg (modify the ip to your own ip) to run syzkaller cases: ./bin/syz-manager -config=acrn.cfg --debug

3. Expected result:
Hypervisor  not hang or reboot.

4. Current result:
After run: ./bin/syz-manager -config=acrn.cfg --debug
Hyperivisor hang.

5. repetition rate
100%

Mingyuan18:
[External_System_ID] ACRN-3487

PR #3219, opened 2019-06-04
hv: fix possible null pointer dereference

tianhuas:
This patch fix potential null pointer dereference

1, will access null pointer if 'context' is null.
2, if entry already been added to the VM when add
   intx entry for this vm, but parameter virt_pin
   is not equal to entry->virt_sid.intx_id.pin. So
   will saves this entry address to
   vpin_to_pt_entry[entry->virt_sid.intx_id.pin] and
   vpin_to_pt_entry[virt_pin]. In this case, this entry
   will be freed twice.

Tracked-On: #3217
Signed-off-by: Tianhua Sun <tianhuax.s.sun@intel.com>

### DIFF
--- a/hypervisor/arch/x86/guest/assign.c
+++ b/hypervisor/arch/x86/guest/assign.c
@@ -358,6 +358,7 @@ static struct ptirq_remapping_info *add_intx_remapping(struct acrn_vm *vm, uint3
 		uint32_t phys_pin, bool pic_pin)
 {
 	struct ptirq_remapping_info *entry = NULL;
+	bool entry_is_updated = true;
 	uint32_t vpin_src = pic_pin ? PTDEV_VPIN_PIC : PTDEV_VPIN_IOAPIC;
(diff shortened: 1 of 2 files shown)


hypervisor/arch/x86/guest/assign.c
```diff
@@ -398,9 +399,10 @@ static struct ptirq_remapping_info *add_intx_remapping(struct acrn_vm *vm, uint3
 		} else {
 			/* The mapping has already been added to the VM. No action
 			 * required. */
+			entry_is_updated = false;
 		}
 
-		if (entry != NULL) {
+		if (entry != NULL && entry_is_updated == true) {
```
dongyaozu: entry_is_updated  is already bool. No need to use "=="

HuiHuangShi:
out of time

hypervisor/arch/x86/guest/assign.c
tianhuas: fixed on pr3223

Issue #3242, opened 2019-06-11
raise the priority of ioreq thread to reduce the trap time


yakuizhao:
Now the created kernel thread uses the default thread priority. When the SOS has some workload, the thread of handling emulated ioreq will be preempted. In such case maybe it will cause that the guest vcpu is in the trapped state.
In order to complete the emulated ioreq from guest, the priority needs to be changed.

yakuizhao:
[External_System_ID] ACRN-4023

Mingyuan18:
The patch was merged.

PR #3243, opened 2019-06-11
DM: Adjust the priority related with ioreq emulation for Android Guest

yakuizhao:
When the IO/MMIO acccess on Android guest causes the trap to hypervisor,
it will be blocked until the ioreq emulation is completed.
In order to get the better performance, it will be better that the ioreq
emulation can be completed ASAP. As it is handled in SOS kernel, the
corresponding thread priority will be raised.

Tracked-On:#3242
Signed-off-by: Liu, Shuo <shuo.liu@intel.com>

### DIFF
--- a/devicemodel/samples/apl-mrb/launch_uos.sh
+++ b/devicemodel/samples/apl-mrb/launch_uos.sh
@@ -378,7 +378,16 @@ echo "dm_run: after passthru dev preparing" > /dev/kmsg
    -l com2,/run/acrn/ioc_$vm_name \
    $boot_image_option \
    --enable_trusty \
-   -B "$kernel_cmdline" $vm_name
+   -B "$kernel_cmdline" $vm_name &
+
+sleep 10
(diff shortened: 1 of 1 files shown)


acrnsi:
OK to verify

acrnsi:
OK to verify

acrnsi:
OK to verify

dongyaozu:
@binbinwu1 can u review?

binbinwu1:
@yakuizhao @shuox Any side effect after raising the priority of these thread? I remembered that there was a patch to set a DM thread to RT, if UOS keeps polling io, it can block SOS from executing other threads.

fyin1:
> 
> 
> @yakuizhao @shuox Any side effect after raising the priority of these thread? I remembered that there 
> was a patch to set a DM thread to RT, if UOS keeps polling io, it can block SOS from executing other 
> threads.

At that time, the problem is RT schedule. This change just update the schedule priority instead of schedule policy to RT. So it should be safe.

Issue #2412, opened 2019-01-24
Top-level makefile cleans before building


rossburton:
It's impossible to do this:

```
make -C hypervisor menuconfig
make hypervisor
```

Because the first thing `make hypervisor` does is a clean, which removes the config file.

Mingyuan18:
[External_System_ID] ACRN-2606

junjiemao1:
@rossburton The top-level Makefile is specifically designed to be invoked by package builders to create, e.g. RPM, packages. If you would like to create a customized build, it would be better if you can invoke the component-specific Makefiles (like hypervisor/Makefile, devicemodel/Makefile, etc.) with command-line options.

Please kindly let me know if you have any requirement on the top-level Makefile. Thanks.

rossburton:
Speaking as a package builder, it's impossible to customise the .config before calling make hypervisor.

hwang37:
@jren1 @JasonChenCJ @dongyaozu to check this issue

hwang37:
Thanks Ross for the bug submission.
If the ask is about pre-defined configuration file, there're some ACRN configuration files defined under the folder below:
./hypervisor/arch/x86/configs/*

They are very simple, and rely the build process to generate the default configuration.
Can we just put the .config as the default configuration files? for your use?

fuzhongl:
@rossburton 
Any more concern, please update the comments. If no more feedback within two weeks, this ticket will be closed by default. 
Appreciate for your understanding.

rossburton:
My assertion that a 'build' target shouldn't enforce a clean still stands.  There should be standalone 'clean' and 'build' targets.

Mingyuan18:
Close with won't fix.

gvancuts:
@Mingyuan18 , I really wish you'd pay a little more attention to these kinds of bugs before closing them. I feel that @rossburton stance is very valid and more generally speaking, properly configuring, building and packaging ACRN is harder than it should be. Just closing the issue with no explanation at all is not how we should evaluate and discuss feedback we receive IMHO.

Mingyuan18:
@fuzhongl ,Please follow this issue. Thanks!

fuzhongl:
Will track with: https://github.com/projectacrn/acrn-hypervisor/issues/3779

mgcao:
OK, we'll consider how to decouple the "clean" with "build", it need add some dependency for Object and C/H files.

fuzhongl:
[...]

Issue #2156, opened 2018-12-21
Enhance the "Building from source" documentation


gvancuts:
We should provide more details (including workflows) on how to configure and build ACRN. Specifically, we should provide additional details on how to use new `Makefile` variables such as: `BOARD`, `FIRMWARE`, and mention `PLATFORM` is now obsolete and should not be used anymore.

As part of this enhancement, we should document how to use the hypervisor `Kconfig` system when building from the top-level folder (as building directly from within subfolders is not supported).

yifanddd:
[External_System_ID] ACRN-3377

fuzhongl:
Hi Geoffroy,
Do you still have more concern for this issue? Or this issue can be closed?
Thanks!

gvancuts:
> Do you still have more concern for this issue? Or this issue can be closed?

I still have concerns about our instructions, specifically on how to properly use `Kconfig` to change the default configuration. Here is an example of why I'm concerned:
* Clone the repo and build the entire project:
```
cd acrn-hypervisor
make BOARD=nuc6cayh
```
And you will get an `acrn.efi` file under `build/hypervisor`

Now, do the following:
```
cd acrn-hypervisor/hypervisor
make defconfig BOARD=nuc6cayh
make
```
And you don't get an `acrn.efi` file as output... (under `build/`)

Another concern I have is that if you run `make BOARD=nuc7i7bnh` from the top-level folder, you don't even get an `acrn.efi` file in that case.

fuzhongl:
> Now, do the following:
> 
> ```
> cd acrn-hypervisor/hypervisor
> make defconfig BOARD=nuc6cayh
> make
> ```
> 
> And you don't get an `acrn.efi` file as output... (under `build/`)

@gvancuts To get an `acrn.efi` file; not need cd acrn-hypervisor/hypervisor 
Please do it at  the top-level folder

> Another concern I have is that if you run `make BOARD=nuc7i7bnh` from the top-level folder, you don't even get an `acrn.efi` file in that case.

@gvancuts please using following command:
`make BOARD=nuc7i7bnh FIRMWARE =uefi'

BTW: we will restructure the make file for ARM support.

fuzhongl:
> > Do you still have more concern for this issue? Or this issue can be closed?
> 
> I still have concerns about our instructions, specifically on how to properly use `Kconfig` to change the default configuration. Here is an example of why I'm concerned:
> 
> * Clone the repo and build the entire project:
> 
[...]

PR #3246, opened 2019-06-11
doc: add an introduction for building hypervisor

shiqingg:
- add an introduction part for building ACRN hypervisor
- clarify why ACRN hypervisor does not support one binary
  for all platforms

### DIFF
--- a/doc/getting-started/building-from-source.rst
+++ b/doc/getting-started/building-from-source.rst
@@ -3,6 +3,50 @@
 Build ACRN from Source
 ######################
 
+Introduction
+************
+
+Following general embedded system programming model, ACRN hypervisor
+is designed to be customized per hardware platform and per usage scenario,
+rather than one binary for all platforms/all usage scenarios.
+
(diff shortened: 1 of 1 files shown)


acrnsi:
OK to verify

shiqingg:
> This looks good from a content perspective.
> 
> It keeps bugging me quite a lot that our instructions/mechanism on how to configure the hypervisor are really not straightforward and it will take any user some amount of trial-and-error before it will do what they want to. Could we assign someone to look into this and fix it? We can track this as part of #2156 (and #2412 is also part of the problem).

@gvancuts Thanks for your review on this PR!
This PR is created to clarify why ACRN hypervisor does not support one binary for all platforms.

Those issues you pointed out is out of scope of this PR, I will let @NanlinXie and @fuzhongl to handle it.
@NanlinXie  @fuzhongl Could you please help on it?

dbkinder:
I'll merge this and come back with some clarity edits in a separate PR.

Issue #3249, opened 2019-06-12
handle the emulation PCI cfg access from unaligned reg_offset


yakuizhao:
When the SOS kernel/pre-launched OS access the 0xCF8/0xCFC, it will cause
the vm-exit and then the hypervisor tries to emulate the PCI_cfg access.

If the aligned reg is passed in 0xCF8 register, it can work well. But when
the unaligned reg is passed in 0xCF8 register, the cached_reg + offset will cause
that the incorrect pci_cfg offset is accessed. For example:
   The cached_reg = 0x02(Device_ID offset) based on the value passed from 0xCF8
   offset = 2 based on 0xCFC-0xCFF address.
   Then cached_reg + offset is used as the offset(PCI_CMD_REG)
 
In fact the unaligned reg can work well on the real HW.

yakuizhao:
[External_System_ID] ACRN-4037

Mingyuan18:
The patch was merged and no regression in this tag:
tag:acrn-2019w25.4-140000p

PR #3250, opened 2019-06-12
ACRN/HV: emulated pcicfg uses the aligned offset to fix the unaligned…

yakuizhao:
… pci_cfg access

When the SOS kernel/pre-launched OS access the 0xCF8/0xCFC, it will cause
the vm-exit and then the hypervisor tries to emulate the PCI_cfg access.
0xCF8 write:  The bdf/reg is captured. cache_reg = value & (0xFF);
0xCFC-0xCFF read/write: offset = address - 0xCFC. Then cached_reg + offset is
used as the offset to access the pci_cfg.

If the aligned reg is passed in 0xCF8 register, it can work well. But when
the unaligned reg is passed in 0xCF8 register, the cached_reg + offset will cause
that the incorrect pci_cfg offset is accessed. For example:
   The cached_reg = 0x02(Device_ID offset) based on the value passed from 0xCF8
   offset = 2 based on 0xCFC-0xCFF address.
   Then cached_reg + offset is used as the offset(PCI_CMD_REG)

In fact the unaligned reg can work well on the real HW.

So the cached_reg should be aligned to handle the unaligned reg passed in
0xCF8 reg.

Tracked-On: #3249

### DIFF
--- a/hypervisor/dm/vpci/vpci.c
+++ b/hypervisor/dm/vpci/vpci.c
@@ -85,7 +85,7 @@ static bool pci_cfgaddr_io_write(struct acrn_vm *vm, uint16_t addr, size_t bytes
 
 	if ((addr == (uint16_t)PCI_CONFIG_ADDR) && (bytes == 4U)) {
 		pi->cached_bdf.value = (uint16_t)(val >> 8U);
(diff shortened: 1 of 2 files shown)


HuiHuangShi:

PR #3264, opened 2019-06-14
doc: add systemd-networkd-autostart bundle in APL GSG

leilu:
Clear Linux changed from systemd-networkd to NetworkManager as default
"network manager", We need to install "systemd-networkd-autostart"
bundle to enable systemd-networkd to create ACRN network bridges and Tap
devices.

Signed-off-by: Lei Lu <leix.lu@intel.com>

### DIFF
--- a/doc/getting-started/acrn_quick_setup.sh
+++ b/doc/getting-started/acrn_quick_setup.sh
@@ -77,8 +77,8 @@ function upgrade_sos()
 
     # Do the setups if previous process succeed.
     if [[ $? -eq 0 ]]; then
-        echo "Adding the service-os and kernel-iot-lts2018 bundles..."
-        swupd bundle-add service-os kernel-iot-lts2018 2>/dev/null
+        echo "Adding the service-os, kernel-iot-lts2018 and systemd-networkd-autostart bundles..."
+        swupd bundle-add service-os kernel-iot-lts2018 systemd-networkd-autostart 2>/dev/null
 
         mount $efi_partition /mnt
         echo "Add /mnt/EFI/acrn folder"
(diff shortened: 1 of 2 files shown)


NanlinXie:

ClaudZhang1995:
OK to verify

ClaudZhang1995:
OK to verify

ClaudZhang1995:
OK to verify

gvancuts:
Is there any risk that NetworkManager will attempt to manage (and mess) with that bridge? Same question regarding the ethernet interface? (I'm asking because I seem to recall that as part of the previous set-up, there was a config somewhere that specifically instructed NetworkManager to **not** touch these, and I'm wondering if we need to bring these back or if it's also part of this `systemd-networkd-autostart` bundle.

ClaudZhang1995:
OK to verify

gvancuts:
@miguelinux FYI

miguelinux:
You can base in the Clear Linux service-os config file

https://cdn.download.clearlinux.org/current/config/image/service-os.yaml

it has:
```
bundles: [                                                                                                                
    openssh-server,                                                                                                       
    os-core,                                                                                                              
    os-core-update,                                                                                                       
    service-os,                                                                                                           
    software-defined-cockpit,                                                                                             
    systemd-networkd-autostart 
  ]
```
Note: this is the CL new config file to be used with the new clr-intaller.

miguelinux:
When you install CL and wants to use ACRN, **yes**,  you need to add the `systemd-networkd-autostart `

gvancuts:
@ttzeng , do you know if we need something similar for [this script](https://github.com/projectacrn/acrn-hypervisor/blob/master/doc/tutorials/create-up2-images.sh#L151)?

Issue #3268, opened 2019-06-17
dm: add virtio-rnd device to command line


gaojunhao0504:
When FE virtio devices work in polling mode, sshd.service can't start
normally and blocks at generating random bytes. When reading from
the random source, getrandom() blocks caused by no random bytes.

Mingyuan18:
[External_System_ID] ACRN-4006

Mingyuan18:
Verify this issue the result is pass

PR #3269, opened 2019-06-17
dm: add virtio-rnd device to command line

gaojunhao0504:
When FE virtio devices work in polling mode, sshd.service can't start
normally and blocks at generating random bytes. When reading from
the random source, getrandom() blocks caused by no random bytes.

Tracked-On: #3268
Signed-off-by: Gao Junhao <junhao.gao@intel.com>


HuiHuangShi:

gaojunhao0504:
Sorry for expressing unclearly. This patch is a workaround for [ACRN-4006](https://jira.devtools.intel.com/browse/ACRN-4006). The platform usually use interrupts as raw material for generating random numbers because the arrival of interrupts are random. If a user system disables all its device interrupts(virtio-device work in polling mode in RTVM) and it doesn't have other hardware random source, it may cause the random number-based application(sshd.service) running on it to wait due to lack of random source. So we try to add a virtio-rnd device to work around for it.

Issue #3252, opened 2019-06-12
Assert usages issue in device model from KW work.


Mingyuan18:
Assert usages issue in device model from KW work.

Mingyuan18:
[External_System_ID] ACRN-4042

Mingyuan18:
The patch was merged and no regression in daily test.
ACRN Tag: acrn-2019w29.1-140000p
CL: 30340

PR #3270, opened 2019-06-17
dm: passthru: remove the use of assert()

binbinwu2:
Remove the use of assert() in passthrough driver.

Tracked-On: #3252
Reviewed-by: Shuo A Liu <shuo.a.liu@intel.com>

### DIFF
--- a/devicemodel/hw/pci/passthrough.c
+++ b/devicemodel/hw/pci/passthrough.c
@@ -71,9 +71,6 @@
 
 extern uint64_t audio_nhlt_len;
 
-/* TODO: Add support for IO BAR of PTDev */
-static int iofd = -1;
-
 /* reference count for libpciaccess init/deinit */
 static int pciaccess_ref_cnt;
 static pthread_mutex_t ref_cnt_mtx = PTHREAD_MUTEX_INITIALIZER;
@@ -404,8 +401,6 @@ init_msix_table(struct vmctx *ctx, struct passthru_dev *ptdev, uint64_t base)
 	uint16_t virt_bdf = PCI_BDF(dev->bus, dev->slot, dev->func);
 	struct ic_ptdev_irq ptirq;
 
-	assert(ptdev_msix_table_bar(ptdev) >= 0 && ptdev_msix_pba_bar(ptdev) >= 0);
-
 	b = ptdev->sel.bus;
 	s = ptdev->sel.dev;
 	f = ptdev->sel.func;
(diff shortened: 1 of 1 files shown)


NanlinXie:

devicemodel/hw/pci/passthrough.c
```diff
@@ -626,7 +621,11 @@ cfginitbar(struct vmctx *ctx, struct passthru_dev *ptdev)
 		 */
 		if (bartype == PCIBAR_MEM64) {
 			i++;
-			assert(i <= PCI_BARMAX);
+			if (i > PCI_BARMAX) {
+				warnx("BAR count out of range\n");
+				return -1;
```
dongyaozu: So if this happens, we can gracefully handle it, right?

devicemodel/hw/pci/passthrough.c
```diff
@@ -1073,21 +1072,14 @@ passthru_write(struct vmctx *ctx, int vcpu, struct pci_vdev *dev, int baridx,
...
-		assert(dev->bar[baridx].type == PCIBAR_IO);
-		bzero(&pio, sizeof(struct iodev_pio_req));
-		pio.access = IODEV_PIO_WRITE;
-		pio.port = ptdev->bar[baridx].addr + offset;
-		pio.width = size;
-		pio.val = value;
-
-		(void)ioctl(iofd, IODEV_PIO, &pio);
+		/* TODO: Add support for IO BAR of PTDev */
+		warnx("Passthru: PIO write not supported, ignored\n");
```
dongyaozu: So this may happen, right? Can we gracefully handle it?
Also, the patch changed the semantics, make sure this is what we want.

devicemodel/hw/pci/passthrough.c
```diff
@@ -1096,24 +1088,16 @@ passthru_read(struct vmctx *ctx, int vcpu, struct pci_vdev *dev, int baridx,
...
-		pio.port = ptdev->bar[baridx].addr + offset;
-		pio.width = size;
-		pio.val = 0;
-
-		(void)ioctl(iofd, IODEV_PIO, &pio);
-
-		val = pio.val;
+		/* TODO: Add support for IO BAR of PTDev */
+		warnx("Passthru: PIO read not supported\n");
+		val = (uint64_t)(-1);
```
dongyaozu: ditto

ClaudZhang1995:
OK to verify

HuiHuangShi:

devicemodel/hw/pci/passthrough.c
binbinwu2: Normally, it will not happen, If this happens, it would be a bug in firmware/bios (assuem SOS doesn't reprogram the BAR register), it should be fixed in firmware/bios. I think just return error code is fine. What do you think?

devicemodel/hw/pci/passthrough.c
binbinwu2: Actually, PIO BAR access is not emulated, the ioctl is not implemented in sos kernel. For now, there is no device accesses PIO BAR.
But it can not be guaranteed on different platforms. Eventually, we probably need to add the emulation for PIO BAR. But for this patch, it should be fine.

Issue #3291, opened 2019-06-19
Update documentation and helper scripts to use newer `swupd` commands


gvancuts:
A number of `swupd` commands are being deprecated in favour of new ones (all existing functionalities have been preserved). The specific changes are:
- `swupd diagnose` (that replaces `verify`): https://github.com/clearlinux/swupd-client/commit/2cb0dd80b4ba98b63c754cb5562c297f0cebdc1a
- `--version` flag (instead of `-m`): https://github.com/clearlinux/swupd-client/commit/a3131db0feb70f4a7616d29e38742d8d7a81c984
- `os-install` (that replaces `verify --install`): https://github.com/clearlinux/swupd-client/commit/a84c60a500ca6efe76368919542a11ee01de6184

A few documents and helper scripts need to be updated accordingly:
* doc/getting-started/acrn_quick_setup.sh
* doc/getting-started/apl-nuc.rst
* doc/tutorials/create-up2-images.sh
* doc/tutorials/skl-nuc.rst
* doc/tutorials/agl-vms.rst

Note that we need to pay attention to the fact the new instructions/scripts will need to be used with a version of Clear Linux that has a `swupd` client with these new commands.

Mingyuan18:
[External_System_ID] ACRN-4151

Mingyuan18:
@gvancuts  Do you want to update another patch?

fuzhongl:
@gvancuts 
acrn_quick_setup.sh was updated.
Can this git issue be closed now?
Thanks!

gvancuts:
Yes, closing it.

PR #3293, opened 2019-06-19
doc: update acrn_quick_setup.sh helper script to use new swupd commands

gvancuts:
Clear Linux's swupd client has received a few updates that introduce new
commands which replace existing ones that will be deprecated in 6 months time.
This commit updates the acrn_quick_setup.sh script to use these new commands.

Note that this new script _may_ fail if the Clear Linux specified does *not*
provide a swupd client that implement these new commands yet.

Tracked-On: #3291

### DIFF
--- a/doc/getting-started/acrn_quick_setup.sh
+++ b/doc/getting-started/acrn_quick_setup.sh
@@ -72,7 +72,7 @@ function upgrade_sos()
         echo "Clear Linux version $sos_ver is already installed. Continuing to setup SOS..."
     else
         echo "Upgrading Clear Linux version from $VERSION_ID to $sos_ver ..."
-        swupd verify --fix --picky -m $sos_ver 2>/dev/null
+        swupd diagnose --fix --picky --version $sos_ver 2>/dev/null
     fi
 
     # Do the setups if previous process succeed.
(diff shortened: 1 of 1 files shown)


doc/getting-started/acrn_quick_setup.sh
```diff
@@ -72,7 +72,7 @@ function upgrade_sos()
         echo "Clear Linux version $sos_ver is already installed. Continuing to setup SOS..."
     else
         echo "Upgrading Clear Linux version from $VERSION_ID to $sos_ver ..."
-        swupd verify --fix --picky -m $sos_ver 2>/dev/null
+        swupd diagnose --fix --picky --version $sos_ver 2>/dev/null
```
dbkinder: Can/Should this check that the Clear Linux version supports this swupd command?  There should be error detection in this script to catch if the ``swupd diagnose`` command fails, right?  Similar comment for the changes below that depend on a newer version of Clear Linux to succeed.

gvancuts:
> What version of Clear Linux is required for this change to work? For now, I suspect we should leave it as is since the old commands will still work for months, right?

I believe they will keep the old commands in place for 6 months (in total).

dbkinder:
If we update these scripts to use the new commands we'll prevent folks from using the older version of Clear Linux that doesn't support the new commands. I'd recommend we hold off on this change for a month or two until we've got some distance from old versions of Clear Linux that only support the "old" commands.  How about we close this and create an issue to track it for a later release?

gvancuts:
OK, closing this PR and let's keep #3291 open to track this task.

Issue #3362, opened 2019-07-04
hv: fix external interrupt lost


lonzoc:
The first interrupt of  a  external device  connecting to IOAPIC would lost.

Mingyuan18:
[External_System_ID] ACRN-4559

lonzoc:
issue is fixed

PR #3363, opened 2019-07-04
hv: vioapic: fix interrupt lost of external device

lonzoc:
Need to set correct initial pin state of vioapic, and avoid generate
redundant interrupt into the guest:
1. set initial pin state accoring to rte attributes, ignore the value of
   entry->polarity because entry->polarity is uninitialized.
2. since vioapic_generate_intr relys on rte, we should build rte
   beforegenerating an interrput.

This patch fix the interrupt lost, typically the first interrput of
external device, and fix the redundant interrupts generated into guest.

Tracked-On: #3362
Signed-off-by: Cai Yulong <yulongc@hwtc.com.cn>

### DIFF
--- a/hypervisor/arch/x86/guest/assign.c
+++ b/hypervisor/arch/x86/guest/assign.c
@@ -391,6 +391,7 @@ static struct ptirq_remapping_info *add_intx_remapping(struct acrn_vm *vm, uint3
 			if (is_sos_vm(entry->vm)) {
 				entry->vm = vm;
 				entry->virt_sid.value = virt_sid.value;
+				entry->polarity = 0U;
 			} else {
 				pr_err("INTX pin%d already in vm%d with vpin%d, not able to add into vm%d with vpin%d",
 					phys_pin, entry->vm->vm_id, entry->virt_sid.intx_id.pin, vm->vm_id, virt_pin);
--- a/hypervisor/dm/vioapic.c
+++ b/hypervisor/dm/vioapic.c
@@ -326,6 +326,13 @@ static void vioapic_indirect_write(struct acrn_vioapic *vioapic, uint32_t addr, 
 			dev_dbg(ACRN_DBG_IOAPIC, "ioapic pin%hhu: redir table entry %#lx",
 				pin, vioapic->rtbl[pin].full);
 
+			/* remap for ptdev */
+			if ((new.bits.intr_mask == IOAPIC_RTE_MASK_CLR) || (last.bits.intr_mask  == IOAPIC_RTE_MASK_CLR)) {
+				/* VM enable intr */
+				/* NOTE: only support max 256 pin */
+				(void)ptirq_intx_pin_remap(vioapic->vm, pin, PTDEV_VPIN_IOAPIC);
+			}
+
 			/*
 			 * Generate an interrupt if the following conditions are met:
 			 * - pin is not masked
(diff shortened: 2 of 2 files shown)


HuiHuangShi:

hypervisor/arch/x86/guest/assign.c
```diff
@@ -187,14 +187,10 @@ ptirq_build_physical_rte(struct acrn_vm *vm, struct ptirq_remapping_info *entry)
 
 		/* init polarity & pin state */
 		if (rte.bits.intr_polarity == IOAPIC_RTE_INTPOL_ALO) {
-			if (entry->polarity == 0U) {
```
JasonChenCJ: every time when ptirq_alloc_entry, the entry will be memset to 0, so entry->polarity will be inited as 0. Not sure why there is redundant interrupt, could you help give more detail about the issue reproduce flow?

lonzoc: The code here is not to fix the redundant  interrupt issue , it is to fix external interrupt lost issue.

I have a touchscreen device using IOAPIC direct IRQ interrupt with edge trigger mode (polarity is active low). 

1. During SOS boot,  ptirq_intx_pin_remap get called for the touchscreen  interrupt pin,  and this interrupt pin enry is allocated and  then maped  to SOS domain (no driver request this IRQ,  sos kernel IOAPIC driver behavior?), at this point , enrty->polarity is set to 1U. 

2. Then, i launch Anroid uos, and assign the touchscreen IRQ to uos using hypercall in acrn-dm, and pass the ACPI Interrupt information for the touchscrren device.

3.  During android uos boot ,  it write IOAPCI REDTBL  register(touchscreen driver requsest irq) , hv traps this action and vioapic_indirect_write get called, and here i notice vioapic_need_intr return true and an interrput  is inject into the uos (redundant  interrupt), 
and ptirq_build_physical_rte get called again ,  because entry->polariry is set to 1U, then vioapic_set_irqline_nolock is not called and the intial pin state is not set to the right state.

4.  Touchscreen controller trigger one interput(pull down the interrupt pin to low),  and in the hv interrupt flow ,  because the wrong pin state,  vioapic_set_pinstate will ignore the  first interrupt(old level is equal to new level).

lonzoc: BTW,  even though entry->polarity is memset to zero when ptirq_alloc_entry, this initial value is meaningless, its value  shoud based on the rte.bits.intr_polarity, that is, after ` if (entry->polarity == 0U) {} and when entry->polarity = 1 or 0 ` (line 190~200 in assign.c),   the value become meaningful.

So we can't rely the value of entry->polarity  here.

hypervisor/dm/vioapic.c
```diff
@@ -326,6 +328,13 @@ static void vioapic_indirect_write(struct acrn_vioapic *vioapic, uint32_t addr,
 			dev_dbg(ACRN_DBG_IOAPIC, "ioapic pin%hhu: redir table entry %#lx",
 				pin, vioapic->rtbl[pin].full);
 
+			/* remap for ptdev */
+			if ((new.bits.intr_mask == IOAPIC_RTE_MASK_CLR) || (last.bits.intr_mask  == IOAPIC_RTE_MASK_CLR)) {
+				/* VM enable intr */
+				/* NOTE: only support max 256 pin */
+				(void)ptirq_intx_pin_remap(vioapic->vm, pin, PTDEV_VPIN_IOAPIC);
+			}
+
```
JasonChenCJ: I assume this is the actual fix, as vioapic_generate_intr may trigger interrupt without correct pin state config. Could you double check if only this code change fixed your problem?

hypervisor/arch/x86/guest/assign.c
JasonChenCJ: got it, it's because this ptirq entry first allocated to SOS then re-assign to UOS. The SOS set pin-state, and it become the init state for UOS - this is incorrect. How about just reset entry->polarity when we doing ptirq entry re-assignment? Like, in add_intx_remapping, add line 394?

390                 } else if (entry->vm != vm) {
391                         if (is_sos_vm(entry->vm)) {
392                                 entry->vm = vm;
393                                 entry->virt_sid.value = virt_sid.value;
394                                 entry->polarity = 0U;
395                         } else {

lonzoc: Agree， new patch has been  pushed.

HuiHuangShi:

Issue #3374, opened 2019-07-08
Potential interrupt info overwrite in acrn_handle_pending_request 


lirui34:
interrupt might be lost if we do acrn_handle_pending_request twice in one vmexit/vmentry of vcpu_thread. We need do it once per cycle of vmexit/vmentry

lirui34:
[External_System_ID] ACRN-4586

Mingyuan18:
No regression issue found in daily test:
ACRN TAG: acrn-2019w28.3-140000p
CL: 30270

PR #3377, opened 2019-07-08
Fix potential interrupt list issue

shuox:
One cycle of vmexit/vmentry might lost interrupts.
This is the scenario,
  1) vmexit, vmexit_handlers
  2) softirq & disable interrupt
  3) acrn_handle_pending_request
  4) schedule if needed, then back to 1) and loop again.
  5) vmentry

The step 3) might be executed twice. The problem is at the second
execution of acrn_handle_pending_request, we might overwrite
VMX_ENTRY_INT_INFO_FIELD of current vmcs, which cause guest lost
interrupts.

The fix is moving 4) prior to 3), then we will handle the pending
requests and vmentry directly.

Tracked-On: #3374


HuiHuangShi:

Issue #3390, opened 2019-07-11
Debug: Post-mortem debug analysis


Mingyuan18:
If a GuestOS crashes, the Hypervisor and ServiceOS shall be able to save the faulting VMs kernel-crashlogs and in-RAM-crash-dumps to files.
NOTE: See Informative for rationale and analysis use of the dumps implication.

Mingyuan18:
[External_System_ID] ACRN-3568

yaomichael:
This works only with AaaG. For LaaG, extra work need to be done inside Linux, which is out of this ticket's scope.

Mingyuan18:
Verified SOS is able to capture the SOS panic log on APL NUC with ACRN tag: acrn-2019w29.2-140000p

PR #3411, opened 2019-07-12
tools:acrn-crashlog: detect the panic event from all pstore files

lauxinw:
As all pstore files may include the kernel panic information, this patch
tries to find the clue from all messages under /sys/fs/pstore instead of
console-ramoops-0.
To avoid the constant growing of logs, it has to remove the pstore
files after panic being detected.

Tracked-On: #3390
Signed-off-by: Liu, Xinwu <xinwu.liu@intel.com>
Reviewed-by: Zhi Jin <zhi.jin@intel.com>
Acked-by: Chen, Gang <gang.c.chen@intel.com>


acrnsi:

acrnsi:
OK to verify

PR #3416, opened 2019-07-16
doc: remove vuart configuration in nuc and up2

conghuic23:
As the vuart config option is not support anymore, remove it.

### DIFF
--- a/doc/getting-started/apl-nuc.rst
+++ b/doc/getting-started/apl-nuc.rst
@@ -429,7 +429,7 @@ partition. Follow these steps:
       double-check the platform boot order using ``efibootmgr -v`` and
       modify it if needed.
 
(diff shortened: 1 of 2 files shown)


doc/getting-started/apl-nuc.rst
```diff
@@ -450,11 +450,6 @@ partition. Follow these steps:
          You should run ``dmesg |grep ttyS0`` to get port address from the output, and then
          add the ``uart`` parameter into the ``efibootmgr`` command.
 
-   #. ``vuart=ttySn@irqN``: this tells the hypervisor which virtual serial device SOS
```
dbkinder: Line 432 above mentions three options (and with this change there are only two)

conghuic23: Fixed, thanks.

PR #3417, opened 2019-07-16
Correct atomic API using

lifeix:
v5:
1) detail some comments why we need or not atomic operations
2) remove pcpu_id parameter for remove_from_cpu_runqueue

v4:
Correct atomic API using on vCPU layer
1) schedule fields which under lock protection don't need to operate atomically;
otherwises, schedule fields which has no lock protection need to operate atomically.
2) remove unnecessary atomic APIs, like atomic_load/store and atomic_set/clear.

v3:
1) revert secondary cpu start up sync logic, add a wmb to enforce sync_up visable to APs
2) add comments to detail io_req state update logic
3) add comments to detail which field in vLAPIC could be accessed simultaneously

v2:
1) remove atomic APIs in io_req. Instead add a write memory barrier to guarantee
io req_buf consistent
2) refine secondary cpu start up sync logic
3) clear up atomic using in vLPAIC

v1:
Now atomic APIs are used in some places where them don't needed or are used wrong
to plan to guarantte the access to the corresponding structure is atomic.
This patch is just a beginning, it tries to remove the unnecessary atomic using and
correct a wrong using case.

Tracked-On: #1842

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -154,7 +154,7 @@ void init_pcpu_pre(uint16_t pcpu_id_args)
 		}
 	}
 
-	bitmap_set_nolock(pcpu_id, &pcpu_active_bitmap);
+	bitmap_set_lock(pcpu_id, &pcpu_active_bitmap);
 
 	/* Set state for this CPU to initializing */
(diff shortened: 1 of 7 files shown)


HuiHuangShi:

PR #3422, opened 2019-07-17
doc: Add ACRN tag or Clear Linux version info for some tutorials

lirui34:
Neither ACRN tag nor Clear Linux version is specified in some tutorials.
Declare those information for those tutorials.


NanlinXie:
@lirui34 , Please also specify ACRN tag according to CL version, if has release tag v1.x is better, otherwise, ACRN daily tag is also OK for that.

Issue #3475, opened 2019-07-26
More vpci cleanup


donsheng:
Clean up vpci code introduced by vdev ops and enabling vhostbridge for sos:
Remove redundant function calling:
 The caller function has already done the checking to make sure the req is targeted
 for the called functions, so there is no need to do the same checking in called
 functions.

Make functions void because their return values are always 0 and not used.

lirui34:
[External_System_ID] ACRN-4609

Mingyuan18:
CI test pass in below tag: 
ACRN TAG: tag:acrn-2019w33.1-140000p

PR #3476, opened 2019-07-26
More vpci cleanup

donsheng:
Clean up vpci code introduced by vdev ops and enabling vhostbridge for sos:
Remove redundant function calling:
The caller function has already done the checking to make sure the req is targeted
for the called functions, so there is no need to do the same checking in called
functions.

Make functions void because their return values are always 0 and are not used.

Tracked-On: #3475
Signed-off-by: dongshen <dongsheng.x.zhang@intel.com>


HuiHuangShi:

Issue #3537, opened 2019-08-05
Kernel crashes while boot LaaG through the kernel image if we set Video IGD Aperture Size to 512


lirui34:
**HW/Board**
NUC7i7DNHE

**Image info**
E2E daily: 30500

**Preconditions**
Bios configuration to Advanced – Devices and Peripherals – Video – IGD Aperture Size to 512

**Steps**
1. Set up the SOS with Clear Linux version 30500
2. Set up the UOS with Clear Linux KVM 30500
3. Launch UOS through default kernel image: /usr/lib/kernel/default-iot-lts2018

**Actual result**
Lots of kernel crash info, refer to the attachment.

**Reproduce rate**
100 %

**Debugging info**
1. Able to boot LaaG using the OVMF.fd instead.
   OVMF: /usr/share/acrn/bios/OVMF.fd
2. Able to boot LaaG through kernel image with default bios configuration.

lirui34:
[External_System_ID] ACRN-4876

lirui34:
Close it as issue is verified on tag: ``acrn-2019w32.5-140000p``

PR #3538, opened 2019-08-05
dm: gvt: keep LaaG aperture size consistent with the host

liujunming:
Keep LaaG aperture size consistent with the host bios setting,
if we statically set LaaG aperture 256MB,
it may not have memory space for LaaG aperture.
We get host aperture size from io port 0x62.

Tracked-On: #3537
Signed-off-by: Junming Liu <junming.liu@intel.com>


acrnsi:

Issue #3542, opened 2019-08-06
can not generate acrn.bin in some case


Mingyuan18:
In hypervisor folder need to make twice to generate the acrn.bin,
for the first time it can generate the acrn.bin, it is only generate the lib modules for the second time

Mingyuan18:
[External_System_ID] ACRN-4933

PR #3554, opened 2019-08-09
hv:fix need to make twice in hypervisor folder

mingqiangchi:
now the dependency is like this in Makefile:
  acrn.bin << xxxx.a << xxxx.obj
if excute 'make' in hypervsior fold for the first time,
it can generate acrn.bin, if excute 'make' for second
time, it can not do the final link because there are
the same timestamp for acrn.bin and xxxx.a generated by
previous build, add PHONY to fix this issue.

Tracked-On: #3542

### DIFF
--- a/hypervisor/Makefile
+++ b/hypervisor/Makefile
@@ -317,6 +317,8 @@ MODULES += $(LIB_DEBUG)
 endif
 MODULES += $(SYS_INIT_MOD)
 
+.PHONY: $(MODULES)
+
 DISTCLEAN_OBJS := $(shell find $(BASEDIR) -name '*.o')
 VERSION := $(HV_OBJDIR)/include/version.h
 


HuiHuangShi:

ttzeng:
I doubt this is the right way to fix the issue. .PHONY should be used to define the pseudo build targets which are not actual files. Please take a look at its [definition](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html).

mingqiangchi:
@ttzeng in this case acrn.bin << PHONY $(MODULES) << xxxx.obj, that is the acrn.bin depends on xxx.a, and the xxx.a depends on the obj, if we don't add PHONY for MODULES when the obj updates, it  can not guarantee that we can get the acrn.bin for every make because of the stale check for acrn.bin and xxxx.a, if add PHONY, it means the PHONY target was run anyway for every make, then it can generate acrn.bin every time. add PHONY, it is one of methods to fix this issue,
if there are other better solutions are welcome, thanks

PR #3567, opened 2019-08-13
doc: update Build ACRN from Source

ClaudZhang1995:
Signed-off-by: yunzha4x <yunx.zhang@intel.com>

### DIFF
--- a/doc/getting-started/building-from-source.rst
+++ b/doc/getting-started/building-from-source.rst
@@ -275,7 +275,9 @@ the hypervisor using the updated ``.config``.
 
 .. code-block:: none
 
-   $ make menuconfig              # Modify the configurations per your needs
+   # Modify the configurations per your needs
+   $ cd ../         # Enter top-level folder of acrn-hypervisor source
+   $ make menuconfig -C hypervisor BOARD=kbl-nuc-i7   <select industry scenario>
 
 .. note::
    Menuconfig is python3 only.
@@ -293,7 +295,6 @@ Now you can build all these components in one go as follows:
 
 .. code-block:: none
 
-   $ cd ../                      # Enter top-level folder of acrn-hypervisor source
    $ make FIRMWARE=uefi          # Build the UEFI hypervisor with the new .config
 
 The build results are found in the ``build`` directory.  You can specify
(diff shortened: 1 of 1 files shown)


doc/getting-started/building-from-source.rst
```diff
@@ -238,53 +237,20 @@ corresponding VM configuration files in the corresponding
...
-are under the top-level directory of acrn-hypervisor, generate a default
-configuration file for UEFI, allow you to modify some configurations and build
-the hypervisor using the updated ``.config``.
+You can also change it directly from the command line.
 
 .. code-block:: none
 
-   $ make menuconfig              # Modify the configurations per your needs
+   $ cd  acrn-hypervisor/hypervisor
+   $ sed -i 's|default SDC|default INDUSTRY|' arch/x86/Kconfig
```
dbkinder: Is this really the "approved" way to select a scenario?  Why aren't we using ``make menuconfig``?

doc/getting-started/building-from-source.rst
```diff
@@ -238,53 +237,20 @@ corresponding VM configuration files in the corresponding
...
 
 .. note::
-   Menuconfig is python3 only.
-
-Refer to the help on menuconfig for a detailed guide on the interface.
-
-.. code-block:: none
+   The command will change ACRN Scenario ``SDC`` to ``INDUSTRY``.The default
```
dbkinder: Change to:
```
   This sed command will change the ACRN Scenario ``SDC`` to ``INDUSTRY``.
   The default is ``SDC``. You can also choose a different scenario this way.  
```

acrnsi:
OK to verify

doc/getting-started/building-from-source.rst
wenlingz: @ClaudZhang1995 , please don't delete make menuconfig, we can provide 2 ways to selectt one scenario. Thanks

junjiemao1: Specifying the scenario in menuconfig IS supported and we have been using that for long. Even you want to add commands to enable automated builds, it shall be the .config rather than Kconfig that you work on, and the recommended way is to create a brand new .config and use ''make oldconfig'' to generate the full configurations.

wenlingz:
@dbkinder please help review and merge it , thanks

Issue #3615, opened 2019-08-23
virtio-console not work for Debian


conghuic23:
when add the virtio-console to acrn-dm cmdline to boot Debian, there will be a segment fault in acrn-dm.

vtcon: read from bad offset/size 1/1
vtcon: write to bad offset/size 1/1
vtcon: read from bad offset/size 1/1
vtcon: write to bad offset/size 1/1
vtcon: read from bad offset/size 1/1
vtcon: write to bad offset/size 1/1
vtcon: read from bad offset/size 9/1
vtcon: write to bad offset/size 9/1
vtcon: read from bad offset/size 9/1
vtcon: write to bad offset/size 9/1
vtcon: read from bad offset/size 9/1
vtcon: write to bad offset/size 9/1
vtcon: read from bad offset/size 17/1
vtcon: write to bad offset/size 17/1
vtcon: read from bad offset/size 17/1
vtcon: write to bad offset/size 17/1
vtcon: read from bad offset/size 17/1
vtcon: write to bad offset/size 17/1

Thread 13 "vcpu 0" received signal SIGSEGV, Segmentation fault.
                                                               [Switching to LWP 748]
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x0000555555562536 in virtio_pci_legacy_write (ctx=<optimized out>, vcpu=<optimized out>, baridx=0, value=0, size=1, offset=25, dev=0x5555557e5750)
    at hw/pci/virtio/virtio.c:913
#2  virtio_pci_write (ctx=<optimized out>, vcpu=<optimized out>, dev=0x5555557e5750, baridx=0, offset=25, size=1, value=0) at hw/pci/virtio/virtio.c:1852
#3  0x00005555555802ed in pci_emul_io_handler (ctx=<optimized out>, vcpu=<optimized out>, in=<optimized out>, port=<optimized out>, bytes=1,
    eax=0x5555557dc058 <vhm_request_page+88>, arg=0x5555557e5750) at hw/pci/core.c:390
#4  0x00005555555a046c in emulate_inout (ctx=<optimized out>, pvcpu=<optimized out>, pio_request=<optimized out>) at core/inout.c:115
#5  0x00005555555a2475 in vmexit_inout (ctx=<optimized out>, vhm_req=0x5555557dc000 <vhm_request_page>, pvcpu=<optimized out>) at core/main.c:345
[...]

PR #3617, opened 2019-08-23
dm: virtio-console: add subclass

conghuic23:
Add subclass for virtio-console pci device. Without the subclass, the
device may be probed by 8250_pci serial driver,  and lead to segment
fault in acrn-dm. This issue appears when running debian as Guest OS.

But there is a potential issue, if a new serial driver, which use class
and subclass same as virtio-console, the problem may still trigger. Need
to fix it case by case.

Tracked-On: #3615


acrnsi:
OK to verify

HuiHuangShi:

Issue #3619, opened 2019-08-26
Make bzimage loader in DM follow spec for setup_sect detection.


Mingyuan18:
Use standard method to get the setup sector of bzimage.

Mingyuan18:
[External_System_ID] ACRN-5077

NanlinXie:
Patch merged and no regression. Close it.

PR #3620, opened 2019-08-26
dm: bzimage loader: get linux bzimage setup_sects from header

fyin1:
According to $(LINUX_SRC)/Documentation/x86/boot.txt, the header
of bzimage has setup sector number in offset 0x1f1. We don't
need to scan the SETUP_SIG and detect the setup sector number
actually which is not documented in x86 boot protocol.

Tracked-On: #3619


HuiHuangShi:

PR #3632, opened 2019-09-02
doc: update Language Extensions in coding guidelines

shiqingg:
- update Language Extensions in coding guidelines

### DIFF
--- a/doc/developer-guides/c_coding_guidelines.rst
+++ b/doc/developer-guides/c_coding_guidelines.rst
@@ -3640,3 +3640,10 @@ b) '__builtin_va_start', refers to section 6.20 in GCC 7.3 Manual.
 c) '__builtin_va_end', refers to section 6.20 in GCC 7.3 Manual.
 d) '__builtin_offsetof', refers to section 6.51 in GCC 7.3 Manual.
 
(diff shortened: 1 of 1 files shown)


deb-intel:
@shiqiangyu I noticed that your references are from GCC 7.3.   GCC does not list 7.3 as a valid release. Is there a reason we are not using a later version such as 9.2 or 8.3 (https://gcc.gnu.org/onlinedocs/)?

shiqingg:
> LGTM.
> 
> In a future PR, I will add the link to the GCC Manual (section 6) for reader ease: https://gcc.gnu.org/onlinedocs/gcc-7.3.0/gcc/C-Extensions.html#C-Extensions.
> 
> @shiqiangyu I noticed that your references are from GCC 7.3. I also noticed that GCC does not list 7.3 as a valid release (https://gcc.gnu.org/onlinedocs/). Is there a reason we are not using a later version such as 9.2 or 8.3 ?

@deb-intel GCC 7.3 is the first GCC version that contains the "retpoline" support needed to build the kernel (and perhaps other code) with resistance to the Spectre variant-2 vulnerability. ACRN hypervisor needs this feature, so, we decided to use this version at that time.

PR #3668, opened 2019-09-11
doc: Remove "or newer/higher" descriptions for other release notes an…

lirui34:
…d some tutorials.

ACRN tag is ONLY specific to one Clear Linux version.
For the tutorials, we just need to make sure the validated version is functional.


acrnsi:
OK to verify

Issue #3675, opened 2019-09-12
cbm length calculation,Extended model judge, print info error


cdliuqiang:
1. CPUID(EAX = 10H, ECX = ResID=1 or 2).EAX Bits 04 - 00: Length of the capacity bit mask for the corresponding ResID using minus-one notation
2. The Extended Model ID needs to be examined only when the Family ID is 06H or 0FH
3. check_name function, print info error

Mingyuan18:
[External_System_ID] ACRN-5175

PR #3676, opened 2019-09-12
fix cbm length calculation, Extended model judge,  print info error

cdliuqiang:
1. CPUID(EAX = 10H, ECX = ResID=1 or 2).EAX Bits 04 - 00: Length of the capacity bit mask for the corresponding ResID using minus-one notation
2. The Extended Model ID needs to be examined only when the Family ID is 06H or 0FH
3. check_name function, print info error

Tracked-On:#3675
Signed-off-by: Andy <andyx.liu@intel.com>

### DIFF
--- a/hypervisor/dm/vpci/vdev.c
+++ b/hypervisor/dm/vpci/vdev.c
@@ -30,6 +30,9 @@
 #include <vm.h>
 #include "vpci_priv.h"
 
+/**
+ * @pre vdev != NULL
+ */
 uint32_t pci_vdev_read_cfg(const struct pci_vdev *vdev, uint32_t offset, uint32_t bytes)
 {
 	uint32_t val;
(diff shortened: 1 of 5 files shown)


acrnsi:

dongyaozu:
This patch mixed many different thing, please split them...

cdliuqiang:
> This patch mixed many different thing, please split them...
@acrnsi  Do I have to wait until the first bug merges before submitting the last two bug codes?

dongyaozu:
Use separate PR for separate problem solving. Whether to make one depends on another is another story.  If they have natural dependency, one PR is fine.

dongyaozu:
you changed from 8 commits to 3 now. The 2nd one is unable to read yet.

dongyaozu:
Becomes 11 patches again. The series is unreadable. Please split.

cdliuqiang:
Because the branch change, the pull closes automatically , the new pull request is : fix cbm length calculation, Extended model judge, print info error #3693

Issue #3593, opened 2019-08-19
Makefile change which add isd build.


Mingyuan18:
Makefile is updated to generate both industry and sdc images.

Mingyuan18:
[External_System_ID] ACRN-5054

Mingyuan18:
The patch was merged.

PR #3690, opened 2019-09-17
Makefile: override .config with specified scenario

jsun26intel:
1. if SCENARIO=$(SCENARIO) is specified in make param, then override
the default CONFIG_$(SCENARIO) value in existed .config. This means
we don't need to modify the default SCENARIO config in Kconfig file
to build the hypervisor;

2. Some board name shortcuts like apl-nuc/kbl-nuc-i7 need to be converted
to official board name to make sure overrides working;

3. Previously make sbl-hypervisor did not load the defconfig so the
defconfig for boards are lost, this patch fix this issue;

4. remove SCENARIO param when make hypervisor, because SCENARIO could
be got from SCENARIO_NAME which parsed from .config;

Tracked-On: #3593


HuiHuangShi:

leilu:
OK to verify

PR #3693, opened 2019-09-18
fix cbm length calculation, Extended model judge,  print info error

cdliuqiang:
CPUID(EAX = 10H, ECX = ResID=1 or 2).EAX Bits 04 - 00: Length of the capacity bit mask for the corresponding ResID using minus-one notation
The Extended Model ID needs to be examined only when the Family ID is 06H or 0FH
check_name function, print info error
Tracked-On:#3675
Signed-off-by: Andy andyx.liu@intel.com


leilu:
OK to verify

HuiHuangShi:

PR #3695, opened 2019-09-18
hv: save/restore TSC in host's suspend/resume path

YadongQi:
TSC would be reset to 0 when enter suspend state on some platform.
This will fail the secure timer checking in secure world because
secure world leverage the TSC as source of secure timer which should
be increased monotonously.

This patch save/restore TSC in host suspend/resume path to guarantee
the mono increasing TSC.

Note: There should no timer setup before TSC resumed.

Tracked-On: #3698


ClaudZhang1995:
OK to verify

acrnsi:
OK to verify

PR #3699, opened 2019-09-18
doc: update gsg and acrn_quick_setup.sh

leilu:
1. From Clear Linux 31030, we could use kernel-iot-lts2018-sos kernel
   command line as Service OS kernel command line directly, so update
   the steps and codes to set the Service OS kernel command line.
2. Command "swupd verfiy" and the flag "-m" is superseded, use "swupd
   repair" and the flag "-V" instead
3. As UOS is launched by OVMF, the old steps and code to set up UOS does
   not work on Clear Linux 31030, update its related steps and code.
4. Update install Clear Linux steps.
5. Fix some small issues.

Signed-off-by: Lei, Lu <leix.lu@intel.com>

### DIFF
--- a/doc/getting-started/acrn_quick_setup.sh
+++ b/doc/getting-started/acrn_quick_setup.sh
@@ -41,8 +41,6 @@ skip_download_uos=0
 # switcher for disabling reboot device
 disable_reboot=0
 
-# acrn.conf path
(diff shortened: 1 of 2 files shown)


acrnsi:
OK to verify

Issue #3729, opened 2019-09-24
Cannot auto boot 2 VMs with acrnd


Mingyuan18:
To track the issue that 2 VMs cannot be made auto boot by using acrnd.

Mingyuan18:
[External_System_ID] ACRN-5304

Mingyuan18:
ClearLinux Info: 31230
Test build: weekly-2019ww39.4
Cannot reproduce this issue, so close it.

PR #3756, opened 2019-09-26
dm: hugetlb: add file lock to make sure huge page reserve atomic

fyin1:
Currently, DM only access /sys/kernel/mm/hugepages/hugepages-2048kB/
entries according to its own huge page requirement. So it could have
following race issue:

         DM1                                   DM2
      read nr pages
                                            read nr pages
                                            write DM2 nr pages
      write DM1 nr pages

Suppose we should write DM1 + DM2 nr page to kernel sysfs interface
to reserve enough huge page (DM1 + DM2). But actually only reserve
huge page requested by DM1.  Which could trigger one VM can't boot.
We can easily hit this issue if we enable multiple UOS auto boot
because more than one VM are started at almost same time.

We add file lock to make sure huge page reserving in DM atomic.

Tracked-On: #3729
Acked-by: Wang Yu <yu1.wang@intel.com>

### DIFF
--- a/devicemodel/core/hugetlb.c
+++ b/devicemodel/core/hugetlb.c
@@ -36,6 +36,7 @@
 #include <sys/types.h>
 #include <fcntl.h>
 #include <errno.h>
+#include <log.h>
 
 #include "vmmapi.h"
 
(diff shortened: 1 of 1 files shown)


HuiHuangShi:

PR #3774, opened 2019-10-03
doc: move the "Building ACRN in Docker" user guide

gvancuts:
Move the "Building ACRN in Docker" guide to the "Try using ACRN" section to be
more prominently visible. Reference it from the main "Build ACRN from Source"
user guide.


gvancuts:
One of my motivation behind this PR is also to move us closer to eliminating the instructions pertaining to Ubuntu, Debian, CentOS, Fedora, ... as those are poorly maintained. Having the instructions for a native Clear Linux installation **and** instructions on how to use a container should cover pretty much all Linux distributions out there. (and who knows, perhaps even Windows and Mac users... I should try that at some point).

gvancuts:
> (and who knows, perhaps even Windows and Mac users...

Building ACRN on MacOS using the container works fine :)

Issue #3758, opened 2019-09-26
Documentation: add a tutorial (or information) on how to change the Linux kernel parameters for User VMs


gvancuts:
There are different types of User VMs and different mechanisms for booting them up. We should enhance our existing documentation to explain how to change the Linux kernel command-line parameters for the various scenarios.

Mingyuan18:
[External_System_ID] ACRN-5326

PR #3775, opened 2019-10-03
doc: add a little info on how to configure the boot options

gvancuts:
Add a little information (at least pointers) as to how one can change the VM
(Pre-launched, Service or User) boot options.

Tracked-On: #3758

### DIFF
--- a/doc/introduction/index.rst
+++ b/doc/introduction/index.rst
@@ -378,6 +378,9 @@ and many other features.
 Boot Sequence
 *************
 
+.. _systemd-boot: https://www.freedesktop.org/software/systemd/man/systemd-boot.html
+.. _grub: https://www.gnu.org/software/grub/manual/grub/
+
 ACRN supports two kinds of boots: **De-privilege boot mode** and **Direct
 boot mode**.
 
@@ -408,6 +411,28 @@ The Boot process proceeds as follows:
    services shall not use IOMMU. In addition, currently we only support
    UEFI timer with HPET MSI.
 
+In this boot mode, both the Service and User VM boot options (e.g. Linux
+command-line parameters) are configured following the instructions for the EFI
+bootloader used by the Operating System (OS).
+
+* In the case of Clear Linux, this EFI bootloader is `systemd-boot`_ and Linux
+  kernel command-line parameters are defined in the ``.conf`` files.
+* Another popular EFI bootloader used by Linux distributions is `grub`_.
+  Distributions like Ubuntu/Debian, Fedora/CentOS use `grub`_
+
+.. note::
+
+   The `Slim Bootloader <https://www.intel.com/content/www/us/en/design/products-and-solutions/technologies/slim-bootloader/overview.html>`_
+   is an alternative boot firmware that can be used to boot ACRN. The `Boot ACRN Hyervisor <https://slimbootloader.github.io/how-tos/boot-acrn.html>`_
+   and :ref:`using-sbl-up2` tutorials have more information on how to use SBL with ACRN.
+
+.. note::
+
+   A virtual `Slim Bootloader <https://www.intel.com/content/www/us/en/design/products-and-solutions/technologies/slim-bootloader/overview.html>`_
+   can also be used to start User VMs, it is called ``vSBL``. The :ref:`acrn-dm_parameters` has more information
+   on how to boot a User VM using ``vSBL``. Note that in this case, the kernel command-line parameters are
+   defined by the combination of the ``cmdline.txt`` passed on to the ``iasimage`` script and in the launch script, via the ``-B`` option.
+
 Direct boot mode
 ================
 
@@ -428,6 +453,9 @@ The Boot process proceeds as follows:
 #. In the Service VM launch path, the Service VM kernel verifies and loads
    the ACRN Device Model and Virtual bootloader through dm-verity
 #. Virtual bootloader starts the User-side verified boot process
+
+In this boot mode, the boot options are defined via the ``VM{x}_CONFIG_OS_BOOTARGS``
+macro in the source code (replace ``{x}`` with the VM number).
 
 ACRN Hypervisor Architecture
 ****************************


doc/introduction/index.rst
```diff
@@ -429,6 +441,9 @@ The Boot process proceeds as follows:
    the ACRN Device Model and Virtual bootloader through dm-verity
 #. Virtual bootloader starts the User-side verified boot process
 
+In this boot mode, the boot options are defined via the ``VM{x}_CONFIG_OS_BOOTARGS``
+macro in the source code (replace ``{x}`` with the VM number).
+
```
jsun26intel: This depends, for SOS_VM and PRE_LAUNCHED_VM, the boot options are defined via the ``VM{x}_CONFIG_OS_BOOTARGS`` macro, but for POST_LAUNCHED_VM, the boot option is defined in UOS launch script if bootloader is VSBL, or in the .config file which embedded in disk image if bootloader is OVMF.

doc/introduction/index.rst
```diff
@@ -408,6 +411,15 @@ The Boot process proceeds as follows:
    services shall not use IOMMU. In addition, currently we only support
    UEFI timer with HPET MSI.
 
+In this boot mode, both the Service and User VM boot options (e.g. Linux
+command-line parameters) are configured following the instructions for the EFI
+bootloader used by the Operating System (OS).
+
```
jsun26intel: Do we need to care about VSBL support for UOS? The boot option is defined in UOS launch script in that case.

doc/introduction/index.rst
NanlinXie: Suggest we only mention Service VM boot here , we need have other words for User VM in other PRs.

doc/introduction/index.rst
NanlinXie: Suggest we only mention Service VM boot here.

NanlinXie:
@gvancuts , could you update with deleting the words for User VM?

gvancuts:
> @gvancuts , could you update with deleting the words for User VM?

Hi @NanlinXie , I don't want to do that. The main part of the request was specifically on how to change the kernel command-line parameters for User VMs.

Let me see how I can address @jsun26intel 's comments and I will push an updated PR.

doc/introduction/index.rst
gvancuts: > for POST_LAUNCHED_VM, the boot option is defined in UOS launch script if bootloader is VSBL

Good point, I forgot about VSBL. Let me see how I can add some text to cover this (at least mention it)

> or in the .config file which embedded in disk image if bootloader is OVMF.

I guess that's one way someone can change the kernel. But I did not see any specific kernel parameter set in there, can you point these out to me?

We also already provide instructions on how to modify the kernel command-line parameters in various User VMs by modifying the User OS bootloader (it being grub, systemd-boot) directly - see https://projectacrn.github.io/latest/tutorials/running_ubun_as_user_vm.html#enable-the-ubuntu-console-instead-of-the-user-interface for example. I personally like it better because it's more flexible and easier than having to change the kernel config, rebuild it and update the image with these new kernel and modules. So I'd suggest we only focus on that method here, is that OK with you?

jsun26intel: For Ubuntu/Debian User VMs, we just modify the kernel cmdline in grub, but for ClearImage, the cmdline is located in /loader/entries/acrn.conf of EFI partition (need to mount the img to a loop device).
Or shall we mention it in https://projectacrn.github.io/latest/tutorials/building_uos_from_clearlinux.html#build-uos-image-in-clear-linux-os ?

acrnsi:
OK to verify

doc/introduction/index.rst
gvancuts: > For Ubuntu/Debian User VMs, we just modify the kernel cmdline in grub, but for ClearImage, the cmdline is located in /loader/entries/acrn.conf of EFI partition (need to mount the img to a loop device).

Yes. That config file is how  one can configure `systemd-bootd`, which is the EFI bootloader used by Clear Linux. So here as well, the method boils down to knowing what EFI bootloader your OS uses, and configure it according to your liking.

> Or shall we mention it in https://projectacrn.github.io/latest/tutorials/building_uos_from_clearlinux.html#build-uos-image-in-clear-linux-os ?

It would be good to mention it in there, but I still want to have the overall mechanism described in a central place.

doc/introduction/index.rst
gvancuts: I think we need to cover both (eventually). But there is no mention at all about (v)SBL in this document at this stage.

Does booting the Service VM from SBL follow the same sequence as described in the [De-privilege boot mode](https://projectacrn.github.io/latest/introduction/index.html#de-privilege-boot-mode) section?

acrnsi:
OK to verify

gvancuts:
I have updated the PR with more notes/text. @NanlinXie @jsun26intel , PTAL

NanlinXie:
@jsun26intel , Could you help to review? Thanks!

doc/introduction/index.rst
```diff
@@ -408,6 +411,28 @@ The Boot process proceeds as follows:
...
+   The `Slim Bootloader <https://www.intel.com/content/www/us/en/design/products-and-solutions/technologies/slim-bootloader/overview.html>`_
+   is an alternative boot firmware that can be used to boot ACRN. The `Boot ACRN Hyervisor <https://slimbootloader.github.io/how-tos/boot-acrn.html>`_
+   and :ref:`using-sbl-up2` tutorials have more information on how to use SBL with ACRN.
+
+.. note::
+
+   A virtual `Slim Bootloader <https://www.intel.com/content/www/us/en/design/products-and-solutions/technologies/slim-bootloader/overview.html>`_
+   can also be used to start User VMs, it is called ``vSBL``. The :ref:`acrn-dm_parameters` has more information
+   on how to boot a User VM using ``vSBL``. Note that in this case, the kernel command-line parameters are
+   defined directly in the launch script, via the ``-B`` option.
```
jsun26intel: @gvancuts 
sorry for my mistake, if the bootloader is vSBL, the cmdline is also defined in .conf file, i.e. it is same as OVMF. So just remove the phrase "Note that in this case, the kernel command-line parameters are defined directly in the launch script, via the ``-B`` option.".

The "-B" option is needed only when "-k" parameter specified.

gvancuts: Thanks for the review @jsun26intel . I will remove this reference to the `-B` option (or perhaps I should really also add it as a third method one can use to boot a User VM, or is being deprecated?). If I understand it correctly, there are three different methods for launching a post-launched User VM:
1. Using `OVMF.fd` (virtual EFI firmware - cmdline defined in the User VM OS bootloader)
2. Using `vSBL` (cmdline defined in the `cmdline.txt` file passed on to the `iasimage` script)
3. Using a kernel directly (with the `-k` option, in which case the cmdline is defined using the `-B` option)
Can you confirm that this is correct and if so, that the third option is here to stay?

I think your statement is slightly incorrect in that if one uses the `vSBL` firmware, the kernel command-line parameters are defined in the `cmdline.txt` file passed on to the `iasimage` script (and not a conf file). Can you confirm?

doc/introduction/index.rst
binbinwu1: Part of the platform specific UOS cmdline (such as acrn-gt) from the launch script when the guest boot loader is vSBL.
But for SBL to boot SOS, the cmdline is stitched to the boot image.

doc/introduction/index.rst
jsun26intel: @binbinwu1 
please confirm UOS cmdline on vSBL path.

binbinwu1: @jsun26intel Yes, for vSBL, launch script should add -B option to pass paramters to UOS.

jsun26intel: @gvancuts 
Confirmed with binbin,  the cmdline should be combination of cmdline.txt and -B option for vSBL path, and actually the cmdline.txt is null with current implementation.
Synced with Michael, the vSBL support is deprecated, do we need to mention them in doc?

NanlinXie:
@gvancuts , do you have any update for this?

gvancuts:
> @gvancuts , do you have any update for this?

@NanlinXie just pushed an update. PTAL. Thanks!

jsun26intel:
I am fine with this PR.

Issue #3771, opened 2019-10-02
Warning when building the ACRN hypervisor (EFI): `boot.c: In function ‘switch_to_guest_mode’`


gvancuts:
The following warning is triggered when building the ACRN hypervisor (`acrn-2019w38.1-140000p-95-g28b50463`) using the default option (i.e. `make clean && make > /dev/null`):
```
boot.c: In function ‘switch_to_guest_mode’:
boot.c:314:24: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
  314 |  mbi->mi_loader_name = (UINT32)uefi_boot_loader_name;
      |            
```

Mingyuan18:
[External_System_ID] ACRN-5340

PR #3785, opened 2019-10-08
hv: use 'uintptr_t' to properly cast a pointer to unsigned int

gvancuts:
Use 'uintptr_t' to properly cast a pointer to a UINT32.

Tracked-On: #3771

### DIFF
--- a/misc/efi-stub/boot.c
+++ b/misc/efi-stub/boot.c
@@ -311,7 +311,7 @@ switch_to_guest_mode(EFI_HANDLE image, EFI_PHYSICAL_ADDRESS hv_hpa)
 	 * The host physical start address of boot loader name is stored in multiboot header.
 	 */
 	mbi->mi_flags |= MULTIBOOT_INFO_HAS_LOADER_NAME;
-	mbi->mi_loader_name = (UINT32)uefi_boot_loader_name;
+	mbi->mi_loader_name = (uintptr_t)uefi_boot_loader_name;
 
 	asm volatile ("pushf\n\t"
 		      "pop %0\n\t"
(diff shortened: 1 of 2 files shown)


gvancuts:
Note: I have not tested this on real HW but that at least gets rid of the compiler warning... I'm also not an expert on these so if there is a better or more preferred way, I'm all ears.

gvancuts:
Thanks for the comment @jren1 , do you know how this can be addressed or we should just live with it?

It sounds like the definition of the multiboot protocol will *always* cause this if they mandate it must be a 32-bit pointer but some architectures would use a different length?

HuiHuangShi:

jren1:
> Thanks for the comment @jren1 , do you know how this can be addressed or we should just live with it?
> 
> It sounds like the definition of the multiboot protocol will _always_ cause this if they mandate it must be a 32-bit pointer but some architectures would use a different length?

The pointer could be 32bit or 64bit, depends on the underneath architecture. But the protocol always defines the bootloader name filed as as 32bit.
For example: 
https://www.gnu.org/software/grub/manual/multiboot/multiboot.html#multiboot_002eh 
  /* Boot Loader Name */
  multiboot_uint32_t boot_loader_name;

So my proposal still keep the original definition (uint32_t). That means the location of bootloadername should be below 32bit memory.

HuiHuangShi:

HuiHuangShi:

acrnsi:
OK to verify

acrnsi:
OK to verify

HuiHuangShi:

Issue #3801, opened 2019-10-12
hv: error debug message in  hcall_set_callback_vector


yonghuah:
Error debug message in this function, there is no string variable for %s output in below statement:
pr_err("%s: Invalid passed vector\n");

int32_t hcall_set_callback_vector(const struct acrn_vm *vm, uint64_t param)
{
	int32_t ret;

	if (!is_sos_vm(vm)) {
		pr_err("%s: Targeting to service vm", __func__);
	        ret = -EPERM;
	} else if ((param > NR_MAX_VECTOR) || (param < VECTOR_DYNAMIC_START)) {
		pr_err("%s: Invalid passed vector\n");

Mingyuan18:
[External_System_ID] ACRN-4886

Mingyuan18:
The bug cannot reproduced with below build
Installed version: 31380
ACRN tag: acrn-2019w43.5-140000p

PR #3803, opened 2019-10-12
hv: fix error debug message in hcall_set_callback_vector

yonghuah:
this patch is to fix error debug message
 for invalid 'param' case, there is no string
 variable for '%s' output, which will potenially
 trigger hypervisor crash as it may access random
 memroy address and trigger SMAP violation.

Tracked-On: #3801


HuiHuangShi:

PR #3856, opened 2019-10-17
doc: hld-security memory management enhancement update

yonghuah:
Remove some obsolete descriptions and update content
  that is not consistent with the design.


wenlingz:
OK to verify

PR #3862, opened 2019-10-18
Doc: update MMU & Timer HLD

lifeix:
1. update memory management hld
2. update timer hld

### DIFF
--- a/doc/developer-guides/hld/hv-memmgt.rst
+++ b/doc/developer-guides/hld/hv-memmgt.rst
@@ -9,7 +9,7 @@ Overview
 ********
 
 The hypervisor (HV) virtualizes real physical memory so an unmodified OS
-(such as Linux or Android) running in a virtual machine, has the view of
+(such as Linux or Android), running in a virtual machine, has the view of
 managing its own contiguous physical memory.  HV uses virtual-processor
 identifiers (VPIDs) and the extended page-table mechanism (EPT) to
 translate guest-physical address into host-physical address. HV enables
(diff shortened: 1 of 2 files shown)


doc/developer-guides/hld/hv-memmgt.rst
```diff
@@ -62,7 +61,10 @@ Hypervisor Physical Memory Management
 In the ACRN, the HV initializes MMU page tables to manage all physical
 memory and then switches to the new MMU page tables. After MMU page
 tables are initialized at the platform initialization stage, no updates
-are made for MMU page tables.
+are made for MMU page tables except hv_access_memory_region_update is called.
+However, the memory region which hv_access_memory_region_update will update
+must have not been accessed by ACRN hypervisor because there's no flush
```
fyin1: Should we say?
hv_access_memory_region could only be accessed by BSP and must not be accessed after AP is started.

lifeix: No. Actually, hv_access_memory_region_update  would be called after AP is started. Even after SOS has booted. But there's only one case which was NPK for debug use. Other cases were all called on BSP before AP was started. Thanks.

fyin1: Does ACRN access NPK which call hv_access_memory_region_update? I suppose yes. Then
is "However, the memory region which hv_access_memory_region_update will update
must have not been accessed by ACRN hypervisor because there's no flush" conflict with this behavior?

acrnsi:
OK to verify

acrnsi:
OK to verify

Issue #3852, opened 2019-10-17
RTVM cannot launched successfully after have launched vxworks or waag.


Mingyuan18:
1.Environment
[Board]:KBL I7 NUC
SOS +Win + Vxwork +RTVM
Installed version: 31230
root@sos-dom0~ # uname -a
Linux sos-dom0 4.19.73-PKT-190919T160721Z #1 SMP PREEMPT Thu Sep 26 07:07:20 UTC 2019 x86_64 GNU/Linux
root@sos-dom0~ # acrn-dm -v
DM version is: 1.3-unstable-187fa97e (daily tag:acrn-2019w38.1-140000p), build by root@2019-09-26 07:06:56
2.Reproduce step:
1). Select passthru-device sata or nvme in Config_app UI, then genereate related launch_uos_script.sh
Notes: Selecting SATA or nvme depends on which disk RTVM is installed on.
click "ImportBoardInfo", and select "Scenario Setting" & "Launch Setting"
Cover:
Boardinfo: nuc7i7dnb
Scenario Setting:industry
Launch Setting: industry_launch_3uos
2). Check launch script which add sata or nvme device passthru, fock on vpid&bdf meet your board, hide the PCI device, add the passthrough configuration in acrn-dm command line.
../xmls/config-xmls/nuc7i7dnb/output/launch_uos_id2.sh
such as:
declare -A passthru_vpid
declare -A passthru_bdf

passthru_vpid=(
["nvme"]="8086 f1a6"
)
passthru_bdf=(
["nvme"]="0000:02:00.0"
)
modprobe pci_stub

Passthrough NVME
 echo ${passthru_vpid["nvme"]} > /sys/bus/pci/drivers/pci-stub/new_id
 echo ${passthru_bdf["nvme"]} > /sys/bus/pci/devices/${passthru_bdf["nvme"]}/driver/unbind
 echo ${passthru_bdf["nvme"]} > /sys/bus/pci/drivers/pci-stub/bind
......
  -s 16,passthru,02/00/0 \
   hard_rtvm
3). Launch SOS
4). ./launch_uos_id1.sh to boot waag or ./launch_uos_id3.sh to boot vxworks.
5). ./launch_uos_id2.sh to boot RTVM.
6).
3. Current result:
RTVM cannot launched successfully after have launched vxworks or waag.
4. Exected result:
RTVM can be launched successfully after have launched vxworks or waag.
6.Logs:
[...]

PR #3872, opened 2019-10-21
acrn-config: define vm name while generating launch script for preempt RT

liuwei987:
1. acrn-config: define vm name for Preempt-RT Linux in launch script
2. acrn-config: refine board name with undline_name api
3. acrn-config: acrn-config: add white list to skip item check

    Tracked-On: #3852
    Acked-by: Victor Sun <victor.sun@intel.com>

### DIFF
--- a/misc/acrn-config/library/launch_cfg_lib.py
+++ b/misc/acrn-config/library/launch_cfg_lib.py
@@ -585,3 +585,12 @@ def get_pt_dev():
     cap_pt = PASSTHRU_DEVS
 
     return cap_pt
+
+
+def undline_name(name):
+    """
+    This convert name which has contain '-' to '_'
+    :param name: name which contain '-' and ' '
+    :return: name_str which contain'_'
+    """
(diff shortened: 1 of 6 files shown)


HuiHuangShi:

misc/acrn-config/library/launch_cfg_lib.py
```diff
@@ -462,6 +462,10 @@ def args_aval_check(arg_list, item, avl_list):
     """
     # args should be set into launch xml from webUI
     i_cnt = 1
+    skip_check_list = ['gvt_args', 'rootfs_dev']
+    if item in skip_check_list:
+        return
```
binbinwu1: If these two options are needed, it also will be skipped?

liuwei987: yes, the two arguments are not must for some launch vm script. 
The two arguments can only be chosen from webUI which always showing valid options. So we don't need to check these items.

Issue #3882, opened 2019-10-22
Update HLD document.


Mingyuan18:
Update HLD document.

Mingyuan18:
[External_System_ID] ACRN-5496

NanlinXie:
Patch merged and no regression. Close it.

PR #3884, opened 2019-10-22
document: update HLD for hypervisor startup

JasonChenCJ:
updated this chapter based on latest master

Tracked-On: #3882


doc/developer-guides/hld/hv-startup.rst
```diff
@@ -36,11 +45,13 @@ description for the flow:
...
    as the base physical console for HV and Service OS.
 
--  **Shell Init:** Start a command shell for HV accessible via the UART.
-
 -  **Memory Init:** Initialize memory type and cache policy, and creates
    MMU page table mapping for HV.
 
+-  **Scheduler Init:** Initialize scheduler framework, which provide the
+   capability to switch different threads(like vcpu vs. idle thread) on a
+   physical CPU, and further to support CPU sharing.
```
fyin1: v1.4 has CPU sharing support. Suppose we could remove "further" here

doc/developer-guides/hld/hv-startup.rst
```diff
@@ -98,40 +111,62 @@ Memory
...
+   ``vm config`` for this VM.
+
+-  **Build vACPI:** For Service VM, the hypervisor will customize a virtual ACPI
+   table based on native ACPI table (this is in the TODO).
+   For pre-launched VM, the hypervisor will build a simple ACPI table with necessary
+   information like MADT.
+   For post-launched User VM, DM will build its ACPI table dynamically.
+
+-  **SW Load:** Prepares for each VM's SW configuration including kernel entry
+   address, ramdisk address, bootargs, or zero page for launching bzImage etc.
```
fyin1: What about change this to:
Prepare for each VM's SW configuration according to guest OS requirement?

The reason is that we are not only for linux kernel but also for pre-launched OS which may have different configuration info.

acrnsi:
OK to verify

PR #3909, opened 2019-10-23
document: update HLD for vm management

JasonChenCJ:
add VM state transition, and make this chapter into
part of hypervisor hld, the user level vm management
should be introduced in ACRN tools.

Tracked-On: #3882


acrnsi:
OK to verify

Issue #3779, opened 2019-10-08
Modify the Make file.


Mingyuan18:
Modify the Make file.

Mingyuan18:
[External_System_ID] ACRN-5338

Mingyuan18:
The patch was merged,

PR #3921, opened 2019-10-24
makefile: add dash support to build efi

jsun26intel:
Usually We use '==' or '=' to compare strings under sh environment,
but '==' is not supported by dash which is the default sh environment
on Ubuntu, this leads efi build failure with current makefile.

Change the '==' to '=' to support both bash and dash.

Tracked-On: #3779


HuiHuangShi:

PR #3934, opened 2019-10-25
document: update HLD for hypervisor overview

JasonChenCJ:
updated this chapter based on latest master
some part still need update:
- vSBL need be replaced by OVMF after removed all vSBL stuff

Tracked-On: #3882


acrnsi:
OK to verify

Issue #3947, opened 2019-10-29
WaaG auto boot failed after systemctl enable acrnd during S5 testing


Mingyuan18:
1.Environment
[Board]:KBL I7 NUC
SOS +Win + RTVM
Installed version: 31230
Version URL: [http://linux-ftp.sh.intel.com/pub/mirrors/clearlinux/update]
Content URL: [http://linux-ftp.sh.intel.com/pub/mirrors/clearlinux/update]
SOS info:
root@clr-31805682d8d84b0496ee12c2a3733ed5~ # uname -r
4.19.73-PKT-190919T160721Z
root@clr-31805682d8d84b0496ee12c2a3733ed5~ # acrn-dm -v
DM version is: 1.3-unstable-187fa97e (daily tag:acrn-2019w38.1-140000p), build by root@2019-09-26 07:06:56
[http://10.239.147.147/acrn_daily_build/2019ww39.4/1//isd/]
Branch: master
Summary: WaaG auto boot failed after systemctl enable acrnd during S5 testing
Steps:
boot up SOS
modify launch_win.sh for S5 test:**launch_win_s5.sh:
#for pm setting
pm_channel="--pm_notify_channel uart "
pm_by_vuart="--pm_by_vuart pty,/run/acrn/life_mngr_"$vm_name
pm_vuart_node=" -s 1:0,lpc -l com2,/run/acrn/life_mngr_"$vm_name
#for memsize setting
mem_size=4G
acrn-dm -A -m $mem_size -c $2 -s 0:0,hostbridge -U 38158821-5208-4005-b72a-8a609e4190d0 {color}
-s 2,pci-gvt -G "$3" \
-s 3,virtio-blk,/home/img/win10-ltsc-virtio.img \
-s 4,virtio-net,tap0,mac=00:16:3E:dd:14:47 \
--ovmf ./OVMF.fd \
$pm_channel $pm_by_vuart $pm_vuart_node {color}
$vm_nam
Set WaaG/rtvm auto boot in SOS:
acrnctl add ./launch_win_s5.sh
acrnctl add ./launch_hard_rt_vm_s5.sh
systemctl unmask acrnd.service
systemctl enable acrnd.service
poweroff and poweron
after poweron, sos will boot up
Waag and rtvm will auto boot after sos boot up
repeat step 5-7
Expect result:
WaaG and RTVM should auto boot up normally
Actual result:
WaaG auto boot failed after systemctl enable acrnd during S5 testing

[...]

PR #3952, opened 2019-10-29
dm: file lock should be held till all mmap is done

chejianj:
Lock should be held till all the mmap operations are done. This is to
avoid the mmap failure when multiple guests are created concurrently.
For example consider the following case in which vm1 and vm2 are
created by acrnd concurrently:
- vm1 is created with 4G+2M memory.
- 4G+2M memory is reserved in hugetlb now and vm1 continues to
  allocate memory for the lowmem without lock held.
- 2G memory is allocated by vm1 for its lowmem, and 2G+2M memory
  is available in hugetlb.
- At this time vm2 is created with 1G+2M memory. It finds that enough
  memory is reserved (2G+2M), so it does not try to reserve more
  memory.
- vm2 allocates some memory for its lowmem/highmem/ovmf.
- vm1 tries to allocate memory for its highmem/ovmf, the allocation
  will fail. vm1 creation failed in this case.

Tracked-On: #3947

### DIFF
--- a/devicemodel/core/hugetlb.c
+++ b/devicemodel/core/hugetlb.c
@@ -740,15 +740,13 @@ int hugetlb_setup_memory(struct vmctx *ctx)
 		}
 	}
 
+	lock_acrn_hugetlb();
+
 	/* it will check each level memory need */
 	has_gap = hugetlb_check_memgap();
 	if (has_gap) {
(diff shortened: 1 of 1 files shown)


HuiHuangShi:

PR #3998, opened 2019-11-05
doc:  correct some mistakes add add cores and threads for CPU of supported hardware

coneypo:
KBL NUC NUC7i5DNH has serial port;

The default binary will not support 2 core / 2 thread any more, so the minimun system requirements for ACRN should be 2 core/4 thread or 4 core+;

Add the cores and threads for each CPU.

Signed-off-by: Xie Zhengtian <zhengtian.xie@intel.com>


acrnsi:

Issue #4012, opened 2019-11-06
Error formatting flag for hypcall_id


Mingyuan18:
Error formatting flag for hypcall_id

Mingyuan18:
[External_System_ID] ACRN-5607

Mingyuan18:
No regression issue found, so close this issue.
ACRN tag: acrn-2019w46.2-140000p
CL: 31630

PR #4013, opened 2019-11-06
HV: correct the formatting flag of hypcall_id

KaigeFu:
hypcall_id has a type of uint64_t and should use 'llx' as
formatting flag instead of '%d'. Otherwise, we will get a
confusing error log when not-allowed hypercall occurs.

Without this patch:
[96707209us][cpu=1][sev=3][seq=2386]:hypercall -2147483548 is only allowed from SOS_VM!

With this patch:
[84613395us][cpu=1][sev=3][seq=2136]:hypercall 0x80000064 is only allowed from SOS_VM!

So, we can figure out which not-allowed hypercall has been triggered more conveniently.

BTW, this patch adds hypcall_id which triggered from non-ring0 into error log.

Tracked-On: #4012


HuiHuangShi:

Issue #2793, opened 2019-03-15
Hard to override EFI loader


rossburton:
In acrn-hypervisor up to 0.6 it was possible to override the EFI loader path in the `.config`, made easy for distro packages by writing a `.config` with just that value and then running `make oldconfig`.

master removed that and now the loader path is a `#define` that needs to be patched.  Please consider either reintroducing the Kconfig, or if Kconfig is being phased out then let me pass `-DDEFAULT_UEFI_OS_LOADER_NAME=/some/path` in CPPFLAGS by guarding the `#define` with `#ifndef`.

Mingyuan18:
@rossburton  , We will solve this issue with #2745, and please give suggestions about configurations below #2745, thank you!

gvancuts:
hi @Mingyuan18 , I fail to see how #2745 is relevant here, can you clarify?

Mingyuan18:
@gvancuts Sorry, It should be #2794.

rossburton:
The EFI loader configuration *was in the kconfig* in 0.7 but has been removed since (commit 334382f9ef1f0564598cd2f96ad22d1a7f8b4c1c).  If you're planning on adding it back then fine, #2794 needs to be fixed, but with current master it isn't relevant.

I'm currently using this patch so I can at least pass the value in `$CPPFLAGS`:

```
--- a/efi-stub/boot.c
+++ b/efi-stub/boot.c
@@ -336,7 +336,10 @@ static inline EFI_STATUS isspace(CHAR8 ch)
     return ((uint8_t)ch <= ' ');
 }

+#ifndef DEFAULT_UEFI_OS_LOADER_NAME
 #define DEFAULT_UEFI_OS_LOADER_NAME "\\EFI\\org.clearlinux\\bootloaderx64.efi"
+#endif
```

Mingyuan18:
@rossburton , Can you close this issue? Because issue  #2794 has already merged the patch

yifanddd:
[External_System_ID] ACRN-3372

yifanddd:
[External_System_ID] ACRN-3373

fuzhongl:
@rossburton 
Any more concern, please update the comments. If no more feedback within two weeks, this ticket will be closed by default. 
Appreciate for your understanding.

rossburton:
Still difficult.  See https://github.com/projectacrn/acrn-hypervisor/blob/master/efi-stub/boot.c#L350

[...]

PR #4019, opened 2019-11-07
hv: add Kconfig parameter to define the Service VM EFI bootloader

gvancuts:
Add a Kconfig parameter called UEFI_OS_LOADER_NAME to hold the Service VM EFI
bootloader to be run by the ACRN hypervisor. A new string manipulation function
to convert from (char *) to (CHAR16 *) has been added to facilitate the
implementation.

The default value is set to systemd-boot (bootloaderx64.efi)

Tracked-On: #2793

### DIFF
--- a/hypervisor/arch/x86/Kconfig
+++ b/hypervisor/arch/x86/Kconfig
@@ -303,3 +303,11 @@ config MAX_KATA_VM_NUM
 	int "Maximum number of Kata Containers in SOS"
 	range 0 1
 	default 0
+
+config UEFI_OS_LOADER_NAME
+	string "UEFI OS loader name"
+	default "\\EFI\\org.clearlinux\\bootloaderx64.efi"
+	help
+	  Name of the UEFI bootloader that starts the Service VM. This is
+	  typically the systemd-boot or Grub bootloader but could be any other
+	  UEFI executable.
--- a/misc/efi-stub/boot.c
+++ b/misc/efi-stub/boot.c
@@ -394,7 +394,6 @@ fail:
 
 }
 
-#define DEFAULT_UEFI_OS_LOADER_NAME L"\\EFI\\org.clearlinux\\bootloaderx64.efi"
 /**
  * efi_main - The entry point for the OS loader image.
  * @image: firmware-allocated handle that identifies the image
@@ -460,7 +459,7 @@ efi_main(EFI_HANDLE image, EFI_SYSTEM_TABLE *_table)
 		 * bootloader name to be used. Fall back to the default bootloader
 		 * as specified in config.h
 		 */
-		bootloader_name = DEFAULT_UEFI_OS_LOADER_NAME;
+		bootloader_name = ch8_2_ch16(CONFIG_UEFI_OS_LOADER_NAME, strlen(CONFIG_UEFI_OS_LOADER_NAME));
 	}
 
 	section = ".hv";
--- a/misc/efi-stub/stdlib.h
+++ b/misc/efi-stub/stdlib.h
@@ -108,4 +108,19 @@ static inline char *ch16_2_ch8(CHAR16 *str16, UINTN len)
 	return str8;
 }
 
+static inline CHAR16 *ch8_2_ch16(char *str8, UINTN len)
+{
+	UINTN i;
+	CHAR16 *str16;
+
+	str16 = AllocatePool((len + 1) * sizeof(CHAR16));
+
+	for (i = 0; i < len; i++)
+		str16[i] = str8[i];
+
+	str16[len] = 0;
+
+	return str16;
+}
+
 #endif /* __STDLIB_H__ */


HuiHuangShi:

jsun26intel:
@gvancuts
This patch has been reverted for a long time because we thought that the config is not related to hypervisor. Do we really need to add it back?

The setting could be configured in efibootmgr command parameter, like:

# For SATA
sudo efibootmgr -c -l "\EFI\acrn\acrn.efi" -d /dev/sda -p 1 \
       -L "ACRN Hypervisor" -u "bootloader=\EFI\ubuntu\grubx64.efi"

gvancuts:
> @gvancuts
> This patch has been reverted for a long time because we thought that the config is not related to hypervisor. Do we really need to add it back?

Yes, this is the whole point of the lengthy discussion in the thread in #2793. Please review the various exchanges and let me know if anything is unclear.

jsun26intel:
For build time support, how about write script like acrn_quick_setup.sh to initialize efibootmgr?

gvancuts:
> For build time support, how about write script like acrn_quick_setup.sh to initialize efibootmgr?

That's installation time, not build time.

jsun26intel:
The problem is, if someone specified bootload name in efibootmgr, then the Kconfig will not work.
And in build time we could modify the 


> For build time support, how about write script like acrn_quick_setup.sh to initialize efibootmgr?
> 
> That's installation time, not build time.

The problem is, if someone specified bootload name in efibootmgr, then the Kconfig will not work.
And in build time we could modify the script for acrn installation, right? And if Acrn is already installed, means user had configured efibootmgr correctly.

Anyway, I leave this question to Eddie/Jack/Jason. If Kconfig is needed, then the patch need to be modified to avoid overriding from efibootmgr.

gvancuts:
> > For build time support, how about write script like acrn_quick_setup.sh to initialize efibootmgr?
> > That's installation time, not build time.
> 
> The problem is, if someone specified bootload name in efibootmgr, then the Kconfig will not work.
> And in build time we could modify the script for acrn installation, right? And if Acrn is already installed, means user had configured efibootmgr correctly.

I don't believe many will necessarily use the script to perform a full system installation (*). And I think it's perfectly fine to have the user being able to overwrite the bootloader name using the `-u` parameter during the installation phase, this is in fact available today. The real problem is that this is today the *only* option. There is no option today to generate a build of ACRN that has a custom bootloader built-in without patching the code.

*: in my opinion, the script is just a nice help to improve the out-of-the-box experience. A proper integration would mean being able to install and configure ACRN automatically via the distro preferred mechanism (such as `swupd` for Clear Linux, `dnf` for Fedora, `apt` for Debian/Ubuntu)

> 
> Anyway, I leave this question to Eddie/Jack/Jason. If Kconfig is needed, then the patch need to be modified to avoid overriding from efibootmgr.

I don't see the reason why we would need this mechanism?

This is very similar to the UART parameter. It's set via `Kconfig`, but can be overwritten by setting it to a different value (or disabling it) via `efibootmgr`

jsun26intel:
> For build time support, how about write script like acrn_quick_setup.sh to initialize efibootmgr?
> That's installation time, not build time.
> 
> The problem is, if someone specified bootload name in efibootmgr, then the Kconfig will not work.
> And in build time we could modify the script for acrn installation, right? And if Acrn is already installed, means user had configured efibootmgr correctly.
> 
> I don't believe many will necessarily use the script to perform a full system installation (*). And I think it's perfectly fine to have the user being able to overwrite the bootloader name using the -u parameter during the installation phase, this is in fact available today. The real problem is that this is today the only option. There is no option today to generate a build of ACRN that has a custom bootloader built-in without patching the code.
> *: in my opinion, the script is just a nice help to improve the out-of-the-box experience. A proper integration would mean being able to install and configure ACRN automatically via the distro preferred mechanism (such as swupd for Clear Linux, dnf for Fedora, apt for Debian/Ubuntu)
> 
> Anyway, I leave this question to Eddie/Jack/Jason. If Kconfig is needed, then the patch need to be modified to avoid overriding from efibootmgr.
> 
> I don't see the reason why we would need this mechanism?
> This is very similar to the UART parameter. It's set via Kconfig, but can be overwritten by setting it to a different value (or disabling it) via efibootmgr



> For build time support, how about write script like acrn_quick_setup.sh to initialize efibootmgr?
> That's installation time, not build time.
> 
> The problem is, if someone specified bootload name in efibootmgr, then the Kconfig will not work.
> And in build time we could modify the script for acrn installation, right? And if Acrn is already installed, means user had configured efibootmgr correctly.
> 
> I don't believe many will necessarily use the script to perform a full system installation (*). And I think it's perfectly fine to have the user being able to overwrite the bootloader name using the -u parameter during the installation phase, this is in fact available today. The real problem is that this is today the only option. There is no option today to generate a build of ACRN that has a custom bootloader built-in without patching the code.
> *: in my opinion, the script is just a nice help to improve the out-of-the-box experience. A proper integration would mean being able to install and configure ACRN automatically via the distro preferred mechanism (such as swupd for Clear Linux, dnf for Fedora, apt for Debian/Ubuntu)
> 
> Anyway, I leave this question to Eddie/Jack/Jason. If Kconfig is needed, then the patch need to be modified to avoid overriding from efibootmgr.
> 
> I don't see the reason why we would need this mechanism?
> This is very similar to the UART parameter. It's set via Kconfig, but can be overwritten by setting it to a different value (or disabling it) via efibootmgr
good point, I have no concern now.

gvancuts:
> good point, I have no concern now.

Thanks @jsun26intel

wenlingz:
@dongyaozu please help approve this PR thanks

PR #4011, opened 2019-11-06
doc: Add Advisory notes

lirui34:
Signed-off-by: lirui34 <ruix.li@intel.com>

PR #4050, opened 2019-11-08
doc: Release notes v1.4

lirui34:

### DIFF
--- a/doc/release_notes/release_notes_1.4.rst
+++ b/doc/release_notes/release_notes_1.4.rst
@@ -1,0 +1,374 @@
+.. _release_notes_1.4:
+
+ACRN v1.4 (Oct 2019)
+####################
+
+We are pleased to announce the release of ACRN version 1.4.
+
+ACRN is a flexible, lightweight reference hypervisor, built with
+real-time and safety-criticality in mind, optimized to streamline embedded
+development through an open source platform. Check out the :ref:`introduction` for more information.
+All project ACRN source code is maintained in the https://github.com/projectacrn/acrn-hypervisor
+repository and includes folders for the ACRN hypervisor, the ACRN device
+model, tools, and documentation. You can either download this source code as
+a zip or tar.gz file (see the `ACRN v1.4 GitHub release page
+<https://github.com/projectacrn/acrn-hypervisor/releases/tag/v1.4>`_)
+or use Git clone and checkout commands::
+
+   git clone https://github.com/projectacrn/acrn-hypervisor
+   cd acrn-hypervisor
+   git checkout v1.4
+
+The project's online technical documentation is also tagged to correspond
+with a specific release: generated v1.4 documents can be found at https://projectacrn.github.io/1.4/.
+Documentation for the latest (master) branch is found at https://projectacrn.github.io/latest/.
+ACRN v1.4 requires Clear Linux* OS version 31520. Please follow the
+instructions in the :ref:`rt_industry_setup`.
+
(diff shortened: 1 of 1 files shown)


doc/release_notes/release_notes_1.4.rst
```diff
@@ -0,0 +1,335 @@
...
+* ACRN is now conformant to Microsoft(r) Hypervisor Top Level Functional Specification(TLFS).
+* Added basic CPU sharing capabilities.
+* Improved WaaG (Windows as a guest) stability and performance.
+* Improved Realtime performance of RTVM (preempt-RT kernel based).
+
+Document updates
+================
+We have many new `reference documents available <https://projectacrn.github.io>`_, including:
+
+* [to be update once pr merge]:enable-s5
```
dbkinder: Need to update these placeholder notes.

doc/release_notes/release_notes_1.4.rst
```diff
@@ -0,0 +1,335 @@
...
+* [to be update once pr merge]:acrn-sanity-test
+* :ref:`enable_laag_secure_boot`
+* :ref:`How-to-enable-secure-boot-for-windows`
+* [to be update once pr merge]:asa
+
```
dbkinder: Recommend adding a new section before the New Features Details, for material from PR #4011 
```
Security Vulnerabilities 
************************

We recommend all developers upgrade to this v1.4 release, which addresses these security
issues discovered in earlier releases:

AP Trampoline Is Accessible to Service VM
   This vulnerability is triggered when validating the memory isolation between
   VM and hypervisor. AP Trampoline code exists in LOW_RAM region in hypervisor but is
   potentially accessible to service VM. This could be used by an attacker to mount DoS
   attacks on the hypervisor if service VM is compromised.

Improper Usage Of ``LIST_FOREACH()`` macro
   Testing discovered that the MACRO ``LIST_FOREACH()`` was incorrectly used for some cases
   which may induce a "wild pointer" and cause ACRN Device Model crash. An attacker
   could use this issue to cause a denial of service (DoS). 

Failure to Create User VM Hangs a User VM that is Already Running
   This vulnerability was discovered while validating the isolation between VMs. An attacker
   could mount DoS attack on any already running user VM with a crafted launch script to start
   a new user VM and make it fail.

Hypervisor Crashed When Fuzzing HC_SET_CALLBACK_VECTOR
   This vulnerability was reported by Fuzzing tool for debug version of ACRN. When software fails
   to validate input properly, an attacker is able to craft the input in a form that is
   not expected by the rest of the application. This can lead to parts of the system
   receiving unintended input, which may result in altered control flow, arbitrary control
   of a resource, or arbitrary code execution.

FILE Pointer Is Not Closed After Using
   This vulnerability was reported by Fuzzing tool. Leaving the file unclosed will cause
   leaking file descriptor and may cause unexpected errors in Device Model program.

Descriptor of Directory Stream Is Referenced After Release
   This vulnerability was reported by Fuzzing tool. A successful call to ``closedir(DIR *dirp)``
   also closes the underlying file descriptor associated with ``dirp``. Access to the released
   descriptor may point to some arbitrary memory location or cause undefined behavior.

Mutex Is Potentially Kept in Locked State Forever
   This vulnerability was reported by Fuzzing tool. pthread_mutex_lock/unlock pairing was not
   always done. Leaving a mutex in a locked state forever can cause program deadlock,
   depending on the usage scenario.

We recommend all developers upgrade to ACRN release v1.4. 
```

Issue #4064, opened 2019-11-10
hv: instr_emul: use cs segment when fetching instructions


peterfang:
In non-64-bit mode, CS segment base address should be considered when determining the linear address of the vcpu's instruction pointer.

peterfang:
[External_System_ID] ACRN-5592

Mingyuan18:
No regression issue found, so close this issue.
ACRN tag: acrn-2019w46.2-140000p
CL: 31630

PR #4065, opened 2019-11-10
hv: instr_emul: use cs segment when fetching instructions

peterfang:
In non-64-bit mode, CS segment base address should be considered when
determining the linear address of the vcpu's instruction pointer. Use
vie_calculate_gla() for instruction address translation which also takes
care of 64-bit mode.

Tracked-On: #4064


HuiHuangShi:

PR #4131, opened 2019-11-19
Makefile: do not rm board acpi info header

jsun26intel:
The $(BOARD)_acpi_info.h is generated by acrn-config tool, remove this
header in make clean would cause failure when user finish configuring
in webUI and start to make acrn-hypervisor by the command
"make hypervisor BOARD=xxx SCENARIO=yyy" because we mandatory do make
clean before making hypervisor.

The patch replace the file removal with a warning string to hint user
to check the file validity.

Tracked-On: #3779

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ DOC_OUT := $(ROOT_OUT)/doc
 BUILD_VERSION ?=
 BUILD_TAG ?=
 DEFAULT_MENU_CONFIG_FILE ?= $(T)/hypervisor/build/.config
+GENED_ACPI_INFO_HEADER = $(T)/hypervisor/arch/x86/configs/$(BOARD)/$(BOARD)_acpi_info.h
 
 ifneq ($(BOARD_FILE),)
     override BOARD_FILE := $(shell if [ -f $(BOARD_FILE) ]; then realpath $(BOARD_FILE); fi)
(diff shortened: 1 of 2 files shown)


HuiHuangShi:

gvancuts:
Have we evaluated how to *not* enforce a `make clean` when running `make`? This would be a better solution to this problem in my opinion.

I'm also wondering how this patch can potentially affect tooling such as Yocto (`meta-acrn`), @rossburton can you chime in?

rossburton:
I can't see any good reason to enforce a `make clean` before a build.  It makes it harder to drop in configuration files that get removed in clean, and when iterating development means you can't just rebuild the pieces that changed.

*Any* problems that enforcing a clean before a build solves are bugs in the makefiles and should be fixed there.

jsun26intel:
This is a good discussion, looks we need more comments from others. If make clean is decoupled from make hypervisor, then CI process and documents need to be changed also.

jsun26intel:
@NanlinXie @fyin1 @JasonChenCJ  @binbinwu1 @wenlingz @fuzhongl @terryzouhao

mgcao:
@rossburton @gvancuts  "enforce a make clean before a build", that is at the beginning, we just use a simple way to keep each time to do a "clean" build hypervisor and device model;  then we have not handled the depend files ( each C file depend many header file).  If it is better to improve that, we can use another ticket to track/discuss, it is not related with this PR.

Makefile
```diff
@@ -120,7 +121,10 @@ hypervisor:
 			echo "CONFIG_ENFORCE_VALIDATED_ACPI_INFO=y" >> $(HV_OUT)/.config;	\
 		elif [ "$(SCENARIO)" ]; then \
 			echo "CONFIG_$(shell echo $(SCENARIO) | tr a-z A-Z)=y" >> $(HV_OUT)/.config;	\
-		fi \
+		fi; \
+		if [ -f $(GENED_ACPI_INFO_HEADER) ] && [ "$(CONFIG_XML_ENABLED)" != "true" ]; then \
```
binbinwu1: I am fine with the patch, but such kind of warning is easily to be ignored. Can we add same statistics about the # of warning/error at the end of the build?

gvancuts:
> @rossburton @gvancuts "enforce a make clean before a build", that is at the beginning, we just use a simple way to keep each time to do a "clean" build hypervisor and device model;

I think the point is that if you do want to make sure you start from a clean state, you run `make clean` yourself, on purpose. The objection here is that it is done unconditionally by the build infrastructure.

> If it is better to improve that, we can use another ticket to track/discuss, it is not related with this PR.

This was raised in January by @rossburton in https://github.com/projectacrn/acrn-hypervisor/issues/2412. Can we use this to track this (and if you can help prioritize it too that would be great)

PR #4171, opened 2019-11-29
doc: add windows 10 activation

liuyuan-GitHub:
doc: add windows 10 activation


huyuan3x:
This windows has 90 days free trial after install windows according to GSG. And this windows is activation state.you can buy windows product key If you want to fully activate the system.

PR #4305, opened 2019-12-25
doc: remove no support OS parts

wenlingz:
Signed-off-by: wenlingz <wenling.zhang@intel.com>


deb-intel:
@wenlingz I merged this per request of @NanlinXie but please note that @gvancuts added requests that might need to be addressed in a new PR.

PR #4321, opened 2019-12-27
doc: add more details to the Kata Containers with ACRN tutorial

gvancuts:
* Add more details about Kata Containers and ACRN in the introduction
* Adjust and correct some of the instructions

### DIFF
--- a/doc/tutorials/run_kata_containers.rst
+++ b/doc/tutorials/run_kata_containers.rst
@@ -1,10 +1,15 @@
 .. _run-kata-containers:
 
-Run Kata containers on a Service VM
+Run Kata Containers on a Service VM
 ###################################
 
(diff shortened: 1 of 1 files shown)


deb-intel:
@gvancuts Do we want users to look at the old commit (which you added at line 23) or can we simply state that it was changed/updated in v1.x (or the correct version). Just curious. No biggie.

gvancuts:
> @gvancuts Do we want users to look at the old commit (which you added at line 23) or can we simply state that it was changed/updated in v1.x (or the correct version). Just curious. No biggie.

I usually prefer to give a minimum version number too but opted to keep a link to the commit because it shows the user what the "fix" is... and it's really not a fix, rather a couple of kernel options that need to be turned on.

I actually think that we should even get rid of the comment in a month or two.

PR #4323, opened 2019-12-30
doc: Update some of the wrong path in acrn configuration tool doc.

lirui34:


deb-intel:
@gvancuts Do we want users to look at the old commit (which you added at line 23) or can we simply state that it was changed/updated in v1.x (or the correct version). Just curious. No biggie.

PR #4351, opened 2020-01-09
Doc: Style and grammar edits to GSG for ACRN Industry Scenario.

deb-intel:
Signed-off-by: Deb Taylor <deb.taylor@intel.com>

### DIFF
--- a/doc/getting-started/rt_industry.rst
+++ b/doc/getting-started/rt_industry.rst
@@ -20,10 +20,8 @@ for the RTVM.
 - Intel Kaby Lake (aka KBL) NUC platform with two disks inside
   (refer to :ref:`the tables <hardware_setup>` for detailed information).
 - If you need to enable the serial port on KBL NUC, navigate to the
-  :ref:`troubleshooting <connect_serial_port>` to prepare the cable.
-- Follow below steps to install Clear Linux OS (Ver: 31670) onto the SATA disk
-  of the KBL NUC(Here we install Clear linux with version 31670 and then the
-  subsequent ACRN quick setup script will upgrade Clear linux to version 32030):
+  :ref:`troubleshooting section <connect_serial_port>` that discusses how to prepare the cable.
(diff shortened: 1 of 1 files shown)


doc/getting-started/rt_industry.rst
```diff
@@ -20,10 +20,8 @@ for the RTVM.
 - Intel Kaby Lake (aka KBL) NUC platform with two disks inside
   (refer to :ref:`the tables <hardware_setup>` for detailed information).
 - If you need to enable the serial port on KBL NUC, navigate to the
-  :ref:`troubleshooting <connect_serial_port>` to prepare the cable.
-- Follow below steps to install Clear Linux OS (Ver: 31670) onto the SATA disk
-  of the KBL NUC(Here we install Clear linux with version 31670 and then the
-  subsequent ACRN quick setup script will upgrade Clear linux to version 32030):
+  :ref:`troubleshooting section <connect_serial_port>` that discusses how to prepare the cable.
+- Follow the below steps to install Clear Linux OS (version 31670) onto the SATA disk of the KBL NUC. In the example below, we will install Clear Linux with version 31670 and then follow the subsequent ACRN quick setup script which will upgrade Clear Linux to version 32030:
```
deb-intel: I made this change (and cleaned up the language a bit) but I think these steps are confusing. Why are we telling the user to install v 31670 first and then upgrade to 32030? Is there a better way to say this?

gvancuts: I think we should point people at the 32030 image right from the start. IIRC, the clr-installer will automatically pull the latest version, and it's the sccript that will revert that back to 32030. The going back and forth is confusing but at least we'd be consistent with our user in terms of what we make them install and what the target is. That should make things a little less confusing hopefully

deb-intel: Ok. Should I change this to the following and leave out the rest (....in the example below......)?
Follow the below steps to install Clear Linux OS 32030 onto the SATA disk of the KBL NUC.

PR #4370, opened 2020-01-19
Delete pass-through audio to WaaG in default

shenfang2019:
Delete pass-through audio to WaaG in default because WHL cannot support audio reset.

Tracked-On: #4397
Signed-off-by: fangfang.shen <fangfang.shen@intel.com>

### DIFF
--- a/devicemodel/samples/nuc/launch_win.sh
+++ b/devicemodel/samples/nuc/launch_win.sh
@@ -14,10 +14,6 @@ if [[ "$result" != "" ]]; then
   exit
 fi
 
-#add audio passthrough device
-echo "8086:9d71" > /sys/bus/pci/drivers/pci-stub/new_id
-echo "0000:00:1f.3" > /sys/bus/pci/drivers/pci-stub/bind
-
 #for memsize setting
 mem_size=4096M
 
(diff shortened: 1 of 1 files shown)


shenfang2019:
> What reset are you mean? S3? Reboot? What is the side effect?
It means PCI HW device reset function indicated in config space. For "Split Device Model", pass-through PCI will be implemented in HV, "reset" param is added into DM cmdline to ensure PCI device's status before launching VM. WaaG will fail to boot if adding pass-through audio in default, because current audio card's reset function isn't supported on WHL platform.

HuiHuangShi:

lifeix:
> What reset are you mean? S3? Reboot? What is the side effect?

@ywan170  There's another PCI(e) device (USB controller or device) doesn't support PCI(e) reset which could not pass through to Waag too.

ywan170:
@lifeix So we are forcing the PCI devices can't passthrough if they are not support FLR? I remembered it is only impact the guest reboot?

wenlingz:
@shenfang2019 please add Tracked-On info, thanks

shenfang2019:
Tracked-On: #4397

leilu:
OK to verify

Issue #2462, opened 2019-01-31
Enable cache QOS with CAT


taoyuhong:
The Hypervisor and ServiceOS shall enable cache QOS with CAT (Cache Allocation Tech), CDP (Cache Data Prioritization) and pseudo-locking across GuestVMs. This feature availability shall be a build time or boot time configuration. The cache configuration shall be accessible via runtime interfaces accessible to both GuestVMs and ServiceOS processes, through some trusted mechanism.

wenlingz:
[External_System_ID] ACRN-2408

Mingyuan18:
this issue is ok on build:
UP2 build:
Clear Linux 28260,
acrn hypervisor tag: 2019w11.1.140000p

PR #4394, opened 2020-02-07
acrn-config: a few changes on vm_config[] clos generation

ZideChen0:
- python script crashes if the desired 'clos' tag doesn't exist in the
  scenario xml file. In the following example, tag '0' doesn't exist:
      clos_set: {1: '0', 2: '0'}
  referring to tag 0 directly ("vm_info.clos_set[0] == None") causes crash.

  To fix it, this patch uses "if i in vm_info.clos_set:" where 'i' is the
  VM id and serves as the tag in the clos_set dictionary, to sanity
  check the dictionary.

- if the hardware doesn't support CAT, don't need to generate clos entry
  in the vm_config[], which could be confusing.

- to reduce redundant code, wrap the clos generation code as a function.

Tracked-On: #2462

### DIFF
--- a/misc/acrn-config/scenario_config/vm_configurations_c.py
+++ b/misc/acrn-config/scenario_config/vm_configurations_c.py
@@ -5,6 +5,7 @@
 
 import sys
 import scenario_cfg_lib
+import board_cfg_lib
 import common
 
 C_HEADER = scenario_cfg_lib.HEADER_LICENSE + r"""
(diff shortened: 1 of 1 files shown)


binbinwu1:
LGTM, @jsun26intel Could you have a review too?

HuiHuangShi:

jsun26intel:
looks good to me

PR #4409, opened 2020-02-21
hv: support BVT scheduler

conghuic23:
As the current IORR scheduler has below defects:
1) unfair
2) cannot prevent I/O attack: Use I/O to maliciously preempt pCPU resources
BVT is introduced. It can provide fair share of pCPU resource to vCPU,
and prevent I/O attack. BVT also has competitive I/O performance as IORR.

### DIFF
--- a/hypervisor/arch/x86/Kconfig
+++ b/hypervisor/arch/x86/Kconfig
@@ -55,6 +55,15 @@ config SCHED_IORR
           IORR (IO sensitive Round Robin) scheduler supports multipule vCPUs running on
 	  on one pCPU, and they will be scheduled by a IO sensitive round robin policy.
 
+config SCHED_BVT
+        bool "BVT scheduler"
+        help
+	  BVT (Borrowed Virtual time) is virtual time based scheduling algorithm, it
+	  dispatching the runnable thread with the earliest effective virtual time.
+	  TODO: BVT scheduler will be built on top of prioritized scheduling mechanism,
+	  i.e. higher priority threads get scheduled first, and same priority tasks are
+	  scheduled per BVT.
+
 endchoice
 
 
--- a/hypervisor/common/sched_bvt.c
+++ b/hypervisor/common/sched_bvt.c
@@ -113,8 +113,33 @@ static int64_t get_svt(struct thread_object *obj)
 	return svt;
 }
 
-static void sched_tick_handler(__unused void *param)
+static void sched_tick_handler(void *param)
 {
+	struct sched_control  *ctl = (struct sched_control *)param;
+	struct sched_bvt_control *bvt_ctl = (struct sched_bvt_control *)ctl->priv;
+	struct sched_bvt_data *data;
+	struct thread_object *current;
+	uint16_t pcpu_id = get_pcpu_id();
+	uint64_t rflags;
+
+	obtain_schedule_lock(pcpu_id, &rflags);
+	current = ctl->curr_obj;
+
+	if (current != NULL ) {
+		data = (struct sched_bvt_data *)current->data;
+		/* only non-idle thread need to consume run_countdown */
+		if (!is_idle_thread(current)) {
+			data->run_countdown -= 1U;
+			if (data->run_countdown == 0U) {
+				make_reschedule_request(pcpu_id, DEL_MODE_IPI);
+			}
+		} else {
+			if (!list_empty(&bvt_ctl->runqueue)) {
+				make_reschedule_request(pcpu_id, DEL_MODE_IPI);
+			}
+		}
+	}
+	release_schedule_lock(pcpu_id, rflags);
 }
 
 /*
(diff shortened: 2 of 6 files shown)


hypervisor/arch/x86/Kconfig
```diff
@@ -55,6 +55,12 @@ config SCHED_IORR
           IORR (IO sensitive Round Robin) scheduler supports multipule vCPUs running on
 	  on one pCPU, and they will be scheduled by a IO sensitive round robin policy.
 
+config SCHED_BVT
+        bool "BVT scheduler"
+        help
+          BVT (Borrowed Virtual Time) scheduler supports multipule vCPUs running on
+	  on one pCPU, and they will be scheduled based on virtual time.
```
dongyaozu: why on on?

dongyaozu: Add comments this is for better real time behavior?

dongyaozu: Add explicit statement to say BVT scheduler is built on top of prioritized scheduling mechanism, i.e. higher priority threads get scheduled first, and same priority tasks are scheduled per BVT.

hypervisor/common/sched_bvt.c
```diff
@@ -30,17 +30,49 @@ struct sched_bvt_data {
...
 }
 
-static void sched_bvt_deinit(__unused struct sched_control *ctl)
+static int sched_bvt_init(struct sched_control *ctl)
 {
+	struct sched_bvt_control *bvt_ctl = &per_cpu(sched_bvt_ctl, ctl->pcpu_id);
+	uint64_t tick_period = BVT_MCU_MS * CYCLES_PER_MS;
+	int ret = 0;
+
+	ASSERT(get_pcpu_id() == ctl->pcpu_id, "Init scheduler on wrong CPU!");
```
dongyaozu: add @pre to make sure ASSERT never hits

hypervisor/common/sched_bvt.c
```diff
@@ -154,12 +155,85 @@ static void sched_bvt_init_data(struct thread_object *obj)
...
+		first_obj = list_entry(first, struct thread_object, data);
+		first_data = (struct sched_bvt_data *)first_obj->data;
+
+		if (sec != NULL) {
+			second_obj = list_entry(sec, struct thread_object, data);
+			second_data = (struct sched_bvt_data *)second_obj->data;
+			delta_mcu = second_data->evt - first_data->evt;
+			/* run_countdown is the real time the thread can run */
+			first_data->run_countdown = v2p(delta_mcu, first_data->vt_ratio)
+				+ BVT_CSA_MCU;
```
dongyaozu: can be in one line

hypervisor/common/sched_bvt.c
```diff
@@ -154,12 +155,85 @@ static void sched_bvt_init_data(struct thread_object *obj)
...
+		} else {
+			/* there is only one object in runqueue, it can run for a long
+			 * time before reschedule */
```
dongyaozu: Didn't understand the comment.

hypervisor/arch/x86/Kconfig
conghuic23: Will add more detail explanation, thanks.

hypervisor/common/sched_bvt.c
conghuic23: Will change to @pre.

dongyaozu: both @pre and ASSERT is preferred. The former one is for FuSa, and the later for debug

hypervisor/arch/x86/Kconfig
conghuic23: BVT is a general purpose scheduler, we may not say it is better for real time. It supports priority-based scheduling, so can work for real-time thread. Besides, the priority is achieved by 'warp' back virtual time. That means, a thread with larger warp value will have earlier virtual time, so can be dispatched firstly.

So, can I change it to :
BVT (Borrowed Virtual Time) is a virtual time based scheduling algorithm,
dispatching the runnable thread with the earliest effective virtual time. With
"warp" back in virtual time, it can provide prioritized scheduling mechanism,
and support real-time scenario.

hypervisor/common/sched_bvt.c
conghuic23: The run_countdown is used to store how may mcu the next thread can run for.  It is set in pick_next handler, and decreases in tick handler.  Normally, the next thread can run until its AVT is ahead of the next runnable thread for one CSA(context switch allowance). But when there is only one object in runqueue,  it can run forever. so, set a very very large number to it so that it can run for a long time.  Here, UINT64_MAX can make it run for 292,471,208 years before rescheduled.

hypervisor/arch/x86/Kconfig
dongyaozu: I thought you mis-understood. In general, ACRN uses prioritized scheduling algorithm. The thread with higher priority got scheduled earlier than lower priority thread. This is an hard requirement for RT usage.

However, if 2 threads have same priority, we use BVT to schedule between them.

HuiHuangShi:

PR #4414, opened 2020-02-24
disable iommu snoop control and add dummy lpc when enable gvt-d

liujunming:
GPU IOMMU doesn’t support snoop control, to enable gvt-d, 
need to disable iommu snoop control for all VMs.

When enabling gvt-d, there is a restriction that 
Windows guest needs a lpc bridge device
located in 00:1f.0 PCI slot.
So add an extra lpc bridge.

### DIFF
--- a/hypervisor/acpi_parser/dmar_parse.c
+++ b/hypervisor/acpi_parser/dmar_parse.c
@@ -8,6 +8,7 @@
 #include <logmsg.h>
 #include <io.h>
 #include <spinlock.h>
+#include <cpu_caps.h>
 #include "pci.h"
 #include "vtd.h"
 #include "acpi.h"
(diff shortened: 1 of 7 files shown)


hypervisor/acpi_parser/dmar_parse.c
```diff
@@ -204,11 +205,14 @@ handle_one_drhd(struct acpi_dmar_hardware_unit *acpi_drhd,
...
-		     (dev_scope->bus << 8U) |
-		     dev_scope->devfun) == CONFIG_GPU_SBDF) {
-			ASSERT(dev_count == 1, "no dedicated iommu for gpu");
-			drhd->ignore = true;
+		/* Disable GPU IOMMU due to gvt-d hasn’t been enabled on APL yet. */
+		if (is_apl_platform()) {
+			if (((drhd->segment << 16U) |
+		     	     (dev_scope->bus << 8U) |
+		     	     dev_scope->devfun) == CONFIG_GPU_SBDF) {
+				ASSERT(dev_count == 1, "no dedicated iommu for gpu");
```
dongyaozu: If this is never to happen case, we need to use @pre  to make sure it never happen for FuSa.
If this may happen, we need gracefully handle it. ASSERT is not graceful.
In this case, this is the old code, and I believe we never hit here, can you check with Binbin to fix it after this patch

HuiHuangShi:

hypervisor/acpi_parser/dmar_parse.c
liujunming: OK. I will check this with binbin.

PR #4540, opened 2020-03-27
Clean-up some codes

qianwang-intel:
Clean up codes.

hypervisor/acpi_parser/dmar_parse.c:
1.removed multiple returns

hypervisor/arch/x86/vtd.c:
1.corrected the return type of some functions,
2.simplified the logic of dmar_wait_completion() and iommu_read64(),
3.renamed some static functions
4.remove some code branches about iommu_domain.is_host and iommu_domain.is_tt_ept
since their values are determined

/hypervisor/include/arch/x86/vtd.h
1.removed the "is_host" and "is_tt_ept" members of struct iommu_domain
since their values are always determined

hypervisor/hw/pci.c
1.renamed some data structures to make them more understandable

### DIFF
--- a/hypervisor/arch/x86/vtd.c
+++ b/hypervisor/arch/x86/vtd.c
@@ -1054,7 +1054,7 @@ static int32_t iommu_attach_device(struct iommu_domain *domain, uint8_t bus, uin
 	struct dmar_entry *context;
 	struct dmar_entry *root_entry;
 	struct dmar_entry *context_entry;
-	uint64_t hi_64;
+	uint64_t hi_64 = 0UL;
 	uint64_t lo_64 = 0UL;
 	int32_t ret = 0;
 	/* source id */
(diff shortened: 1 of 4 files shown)


acrnsi:

hypervisor/arch/x86/vtd.c
```diff
@@ -255,29 +255,13 @@ static inline void dmar_wait_completion(const struct dmar_drhd_rt *dmar_unit, ui
...
-		condition = (temp_condition == pre_condition) ? true : false;
-
-		if (condition) {
-			break;
-		}
 		ASSERT(((rdtsc() - start) < CYCLES_PER_MS),
 			"DMAR OP Timeout!");
 		asm_pause();
-	}
+	} while( ((bool)(*status & mask)) == pre_condition);
```
dongyaozu: This is with assumption that the caller side will use BIT_MASK too.

dongyaozu: So we donn;t need to do type conversion.

Issue #4550, opened 2020-03-30
code styling refine


lifeix:

lifeix:
[External_System_ID]  ACRN-6175

NanlinXie:
Patch merged and no regression. Close it.

PR #4556, opened 2020-03-31
hv: vlapic: minor fix about dereference vcpu from vlapic

lifeix:
Since vcpu if remove from vlapic, we could not dereference it directly.

Tracked-On: #4550


wenlingz:

Issue #4688, opened 2020-04-22
RELEASE=n does not take effect while using xml to make hypervisor


jsun26intel:
Steps
1. Launch web ui app and choose whl-ipc-i5 board.
2. Select industry scenario.
3. just make sure RELEASE=n
4. Generate configuration files.
5. close web ui app.
6. run command to make hypervisor:
make hypervisor BOARD_FILE=$PWD/misc/acrn-config/xmls/board-xmls/whl-ipc-i5.xml SCENARIO_FILE=$PWD/misc/acrn-config/xmls/config-xmls/whl-ipc-i5/user_defined/industry.xml
Expected result
RELEASE is N
Actual result
RELEASE is Y

jsun26intel:
[External_System_ID]ACRN-6223]

PR #4690, opened 2020-04-22
Makefile: do not override RELEASE when build with XML

jsun26intel:
SCENARIO XML file has included RELEASE or DEBUG info already, so if RELEASE is not specified in make command, Makefile should not override RELEASE info in SCENARIO XML. If RELEASE is specified in make command, then RELEASE info in SCENARIO XML could be overridden by make command.

The patch also fixed a issue that get correct board defconfig when build
hypervisor from TARGET_DIR;

Tracked-On: #4688

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,6 @@ else
 	FIRMWARE ?= uefi
 endif
 
-RELEASE ?= 1
 SCENARIO ?= sdc
 
 O ?= build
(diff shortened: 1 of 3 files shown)


binbinwu1:
@jsun26intel  "If RELEASE is specifiled in make command, then it could be overridden." you mean the variable specified in make command will be override by the value in XML file?

jsun26intel:
> @jsun26intel  "If RELEASE is specifiled in make command, then it could be overridden." you mean the variable specified in make command will be override by the value in XML file?

@binbinwu1 I've updated the commit info as below:
SCENARIO XML file has included RELEASE or DEBUG info already, so if RELEASE
    is not specified in make command, Makefile should not override RELEASE info
    in SCENARIO XML. If RELEASE is specified in make command, then RELEASE info
    in SCENARIO XML could be overridden by make command.

HuiHuangShi:
No New Name Conflict

Issue #4719, opened 2020-04-27
Garbage display after shutting down Linux GVT-d guest


yakuizhao:
The Linux GVT-d guest can be booted.
But after Linux guest is closed, it will display the garbage.
So the GPU passthrough device needs to be reset after guest is closed.

yakuizhao:
[External_System_ID]ACRN-6210

PR #4720, opened 2020-04-27
ACRN/DM: Reset the passthrough device to fix garbage display issue

yakuizhao:
Currently the DM will return the passthrough device to SOS when the DM
needs to be terminated. But the device is not reset. In such case it
will cause the garbage display issue when the GPU passthrough device
is used.

Tracked-On: #4719

### DIFF
--- a/devicemodel/hw/pci/passthrough.c
+++ b/devicemodel/hw/pci/passthrough.c
@@ -569,6 +569,9 @@ passthru_deinit(struct vmctx *ctx, struct pci_vdev *dev, char *opts)
 	struct passthru_dev *ptdev;
 	uint16_t virt_bdf = PCI_BDF(dev->bus, dev->slot, dev->func);
 	struct acrn_assign_pcidev pcidev = {};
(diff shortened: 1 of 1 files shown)


HuiHuangShi:

devicemodel/hw/pci/passthrough.c
```diff
@@ -599,6 +605,22 @@ passthru_deinit(struct vmctx *ctx, struct pci_vdev *dev, char *opts)
 		 */
 		vm_deassign_pcidev(ctx, &pcidev);
 	}
+	if (phys_bdf) {
```
fyin1, line 608: We should move this reset to !rtvm path. We don't deassign pcidev if the guest is RTVM. Otherwise, we break the feature that acrn-dm exit shouldn't impact RTVM.

yakuizhao: Make sense. It will be updated.

HuiHuangShi:

Issue #4616, opened 2020-04-13
ACRN needs to support dynamic CPU affinity configuration


ZideChen0:
The statically configured vCPU affinity for post-launched VM in vm_configs[] could be used in different ways:

1. legacy approach: acrn-dm launch post-launched VM without specifying vCPU affinity and hypervisor launches the VM on all the pCPUs that are defined in vm_configs[].
2. the new dynamic configuration: acrn-dm can choose to specify vCPU affinity in the CREATE_VM hypercall, but the new vCPU affinity should be the entire or a subset of the pre-defined pcpu_affinity_bitmap.

ZideChen0:
[External_System_ID] ACRN-6203

NanlinXie:
Patch merged and no regression. Close it.

PR #4742, opened 2020-04-29
hv: don't overwrite the statically configured vm_configs[] in hypercall

ZideChen0:
The statically configured vm_configs[].cpu_affinity_bitmap should remain
intact during the life cycle of the SOS, otherwise user can't destroy
and create the same VM with different CPU affinity. For example:

- Initially vm_configs[1].cpu_affinity_bitmap is set to 0xF: pCPU 0/1/2/3.
- VM1 is created on pCPU1 and pCPU2 and vm_configs[1].cpu_affinity_bitmap
  is overwritten as 0x6.
- VM1 is destroyed.
- Now VM1 can't be launched again on pCPU0 or pCPU3.

This patch fixes this by saving the static VM configuration before the
create_vm hypercall and restore it when the post-launched VM is shutting
down.

Tracked-On: #4616

### DIFF
--- a/hypervisor/arch/x86/configs/vm_config.c
+++ b/hypervisor/arch/x86/configs/vm_config.c
@@ -15,12 +15,35 @@ static uint8_t rtvm_uuid1[16] = POST_RTVM_UUID1;
 static uint8_t safety_vm_uuid1[16] = SAFETY_VM_UUID1;
 
 /*
+ * To be used for post-launched VMS only.
+ *
+ * acrn-dm could modify post-launched VM configuration through command line arguments.
+ * We make use of this additional array to make sure that the dynamic configuration
+ * from acrn-dm won't overwrite the static vm_configs[].
+ */
+static struct acrn_vm_config vm_configs_saved[CONFIG_MAX_VM_NUM];
+
+/*
  * @pre vm_id < CONFIG_MAX_VM_NUM
  * @post return != NULL
  */
 struct acrn_vm_config *get_vm_config(uint16_t vm_id)
 {
 	return &vm_configs[vm_id];
+}
+
+/*
+ * @pre vm_id < CONFIG_MAX_VM_NUM
+ */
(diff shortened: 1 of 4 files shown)


HuiHuangShi:
2 new potential violations to the coding guideline detected. Please help check. Thanks!<br><1> config:part_config filename:/hypervisor/arch/x86/configs/vm_config.c func: offset:24<br> reason:Variable should be declared static. : vm_configs_saved<br><2> config:sharing_config filename:/hypervisor/arch/x86/configs/vm_config.c func: offset:24<br> reason:Variable should be declared static. : vm_configs_saved<br><br>No New Name Conflict<br><br>No New Function Declaration/Definition Mismatch<br>

HuiHuangShi:

JasonChenCJ:
> 1: DM_OWNED_GUEST_FLAG_MASK doesn't contain the bits you mentioned. Mismatch?
> 2: For the cpu_affinity_bitmap, it does have the problem as your commit message says, but I didn't know why we allow guest to change the set. This is super_set list, we should never change,

Yes, we should never change cpu_affinity_bitmap, I think zide's change here is because we are actually using vm_config->cpu_affinity_bitmap to create vcpu every time we create a new VM.

@zide, could we add a new param in create_vm function to pass cv.cpu_affinity?

ZideChen0:
> > 1: DM_OWNED_GUEST_FLAG_MASK doesn't contain the bits you mentioned. Mismatch?
> > 2: For the cpu_affinity_bitmap, it does have the problem as your commit message says, but I didn't know why we allow guest to change the set. This is super_set list, we should never change,
> 
> Yes, we should never change cpu_affinity_bitmap, I think zide's change here is because we are actually using vm_config->cpu_affinity_bitmap to create vcpu every time we create a new VM.
> 
> @zide, could we add a new param in create_vm function to pass cv.cpu_affinity?

Yes, will do it in master branch.

PR #4747, opened 2020-04-29
doc: add note about capricious UP2 EFI firmware

gvancuts:
The UP2 EFI firmware is a bit capricious and does not consistantly keep the
boot order after adding one using `efibootmgr`. There is no magic recipe (or
known reliable sequence) and hence we add a note warning the user that this
can happen, and when it does the only solution is to try modifying some more
the list of boot entries (inc. re-ordering them).


HuiHuangShi:

Issue #4828, opened 2020-05-20
fix a bug in ptirq_release_entry


binbinwu1:
The mask valuei 0x3F was added to prevent out of range in array access.
However, it should not be hardcoded.
Also, there could be the chance that more than 1 core to access a same 64bit var.

binbinwu1:
[External_System_ID] ACRN-6371

PR #4829, opened 2020-05-20
hv: ptirq : fix a bug in ptirq_release_entry

binbinwu1:
The mask valuei 0x3F was added to prevent out of range in array access.
However, it should not be hardcoded.
Since in ptirq_alloc_entry_id, the valid allocated id is no greater
than CONFIG_MAX_PT_IRQ_ENTRIES, it will not cause out of range array
access without mask.
So this patch removes the mask.

Also, use bitmap_clear_lock instead of bitmap_clear_nolock becuase there
could be the chance that more than 1 core to access a same 64bit var.

Tracked-On: #4828

### DIFF
--- a/hypervisor/common/ptdev.c
+++ b/hypervisor/common/ptdev.c
@@ -112,8 +112,7 @@ void ptirq_release_entry(struct ptirq_remapping_info *entry)
 	del_timer(&entry->intr_delay_timer);
 	CPU_INT_ALL_RESTORE(rflags);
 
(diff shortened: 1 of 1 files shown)


HuiHuangShi:
1 new potential violations to the coding guideline detected. Please help check. Thanks!<br><1> config:common filename:/hypervisor/include/arch/x86/lib/bits.h func:bitmap_clear_nolock offset:216<br> reason:Static procedure is not explicitly called in code analysed. : bitmap_clear_nolock<br><br>No New Name Conflict<br><br>No New Function Declaration/Definition Mismatch<br>

Issue #4840, opened 2020-05-22
dm: bzimage: dynamically adapt to ramdisk size


tttech-industrial-buchsbaum:
Using images with ramdisks > 4MB with `acrn-dm -r <ramdisk_file> ...` is not possible. This is required e.g. when using ramdisk based systems, as I used for RT-VM tests to avoid using virtio-blk device thus ruling out any influence on RT behavior. These ramdisks contain an entire Linux system and tend to be rather large, up to 128MB and above.
Other than making the size configurable, which still does not avoid the need to rebuild the devicemodel when ramdisks change, we can solve this by adapting to the ramdisk size dynamically. Thus the supported ramdisk size only depends on the memory available.
This is also one of the patches used in #4833.

fuzhongl:
[External_System_ID] ACRN-6381

PR #4841, opened 2020-05-22
dm: bzimage: dynamically adapt to ramdisk size

tttech-industrial-buchsbaum:
To remove the limit of 4MB ramdisk size simply adjust layout
dynamically according to ramdisk size rounded up to a 4K boundary.
This enables ramdisk based virtual machines which tend to have
large ramdisk memory requirements (128MB and above).

Tracked-On: #4840
Signed-off-by: Helmut Buchsbaum <helmut.buchsbaum@opensource.tttech-industrial.com>

### DIFF
--- a/devicemodel/core/sw_load_bzimage.c
+++ b/devicemodel/core/sw_load_bzimage.c
@@ -50,9 +50,9 @@
  * +-----------------------------------------------------+
  * | ...                                                 |
  * +-----------------------------------------------------+
- * | offset: lowmem - 4MB - 2K (kernel gdt)              |
+ * | offset: lowmem - RAMDISK_LOAD_SIZE - 2K (kernel gdt)|
  * +-----------------------------------------------------+
- * | offset: lowmem - 4MB (ramdisk image)                |
+ * | offset: lowmem - RAMDISK_LOAD_SIZE (ramdisk image)  |
  * +-----------------------------------------------------+
  * | offset: lowmem - 8K (bootargs)                      |
  * +-----------------------------------------------------+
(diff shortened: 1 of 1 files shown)


acrnsi:

HuiHuangShi:

devicemodel/core/sw_load_bzimage.c
```diff
@@ -183,14 +185,6 @@ acrn_prepare_ramdisk(struct vmctx *ctx)
...
 	}
 
-	if (len > (BOOTARGS_LOAD_OFF(ctx) - RAMDISK_LOAD_OFF(ctx))) {
-		pr_err("SW_LOAD ERR: the size of ramdisk file is too big"
-				" file len=0x%lx, limit is 0x%lx\n", len,
-				BOOTARGS_LOAD_OFF(ctx) - RAMDISK_LOAD_OFF(ctx));
-		fclose(fp);
-		return -1;
-	}
-
```
ywan170, line 193: We need to limit the max size of ramdisk, otherwise it may override the first two sub-regions("ACPI table" and "kernel image"). Also it may met issue when the size exceed the ctx->lowmem.

tttech-industrial-buchsbaum: We only have the kernel image and the ACPI table in the memory below RAMDISK_LOAD_OFF. Everything else should be taken care off by the changes in the respective macros. So testing if the kernel image fits into the `RAMDISK_LOAD_OFF - 16*MB` region is sufficient, and this is already tested in `acrn_prepare_kernel()`
https://github.com/projectacrn/acrn-hypervisor/blob/6c02a8d5ed97b248202775244313ec6f53e97293/devicemodel/core/sw_load_bzimage.c#L234-L238
and thus loading with a ramdisk file too large would fail there. This is implicitly the required limitiation of the ramdisk size. Nevertheless we should add a check to verify ramdisk size is small enough to avoid overflows int the calculation macros, i.e. `RAMDISK_LOAD_SIZE < ctx-lowmem - 2*KB`. If you agree, I will prepare an updated patch that contains such a test.

HuiHuangShi:

Issue #4848, opened 2020-05-25
Failed to launch RTVM again if shutdown RTVM


mgcao:
Reproduce steps:
-------------
1.Boot up into Yocto SOS
2.In Yocto SOS, execute launch_hard_rt_vm.sh to launch RTVM
3.After RTVM boot up, poweroff RTVM
4.Execute launch_hard_rt_vm.sh to launch RTVM again

mgcao:
[External_System_ID] ACRN-6333

PR #4851, opened 2020-05-27
HV: fix bug to clear guest flags after it not used

mgcao:
in shutdown_vm, it uses guest flags when handling the phyiscal
CPUs whose LAPIC is pass-through. So if it is cleared first,
the related vCPUs and pCPUs can not be switched to correct state.

so move the clear action after the flags used.

Tracked-On: #4848


HuiHuangShi:

Issue #4853, opened 2020-05-28
Inter-VM communication for post-launched VMs


liuyuan-GitHub-zz:
The hypervisor and device model shall provide inter-VM communication method for specific post-launched VMs, which shall expose to guest as a PCI-based device, and the communication shall be based on shared memory.

The device model should help to organize the PCI device resource which include its share memory allocation.

[External_System_ID]ACRN-6392

wenlingz:
[External_System_ID]ACRN-6392

NanlinXie:
Closed with no regression issue.

PR #4854, opened 2020-05-28
 Support shared memory based inter-vm communication for ACRN post-launched VMs

liuyuan-GitHub:
The ivshmem mediator emulates one standard PCI device which expose the shared memory region through its BAR2. The shared memory region is allocated through Linux shm mechanism and post-launched VM needs to specify the shm name and size as the acrn-dm parameter, the VMs have same shm name parameter can communicate over the shared memory.

For first stage, only support shared memory. Consider to support
notification(interrupt) in future.

Tracked-On: #4853


HuiHuangShi:

PR #4858, opened 2020-05-29
use variables for installation directories.

saininav:
Don't hardcode install paths. Instead of hardcoding where binaries are
installed, add variables that installer can override.

Signed-off-by: Chee Yang Lee <chee.yang.lee@intel.com>
Signed-off-by: Naveen Saini <naveen.kumar.saini@intel.com>

### DIFF
--- a/devicemodel/Makefile
+++ b/devicemodel/Makefile
@@ -2,6 +2,8 @@
 # ACRN-DM
 #
 include ../VERSION
+include ../paths.make
+
 FULL_VERSION=$(MAJOR_VERSION).$(MINOR_VERSION)$(EXTRA_VERSION)
 BASEDIR := $(shell pwd)
 DM_OBJDIR ?= $(CURDIR)/build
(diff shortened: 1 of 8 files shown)


acrnsi:

HuiHuangShi:

NanlinXie:
@fuzhongl and @wenlingz , Is there anything we need change for our current CI and document guide with this PR?

fuzhongl:
No any change needed.

HuiHuangShi:

gvancuts:
I've not tested this but I suspect it will break the code in a number of ways (so not just the doc):
* ACRN config tool unconditionally uses `/usr/share` (in `launch_config`)
* The launch scripts we provide will be broken (wrong path to `OVMF.fd`)
* Don't we need to also update he `misc/efi-stub/Makefile`?
* The `acrn-crashlog` and `acrn_mngr` use some `/usr/share` path in the code (controlled by variables but I'm not seeing the logic to keep those in sync)

The documentation will also need to be updated in various places (adding a note to indicate the pth may be different depending on how it was compiled)

wenlingz:
OK to verify

HuiHuangShi:

NanlinXie:
@jsun26intel , could you check the impact from configuration tool view?

wenlingz:
OK to verify

HuiHuangShi:

dongyaozu:
@terryzouhao can u review?

terryzouhao:
@saininav variable directory(bindir) rather than hardcode(/usr/bin) is good way to support cross compilation environment and more flexible for system integration. But as @gvancuts mentioned, there are dependency of some acrn tools (acrn-crash, acrnlog, acrntrace,arn bridge, acrn-config) with this directory also. 
So the variable(bindir) can only be set as default value(/usr/bin), just passed basic CI check. If change/override to other value, some tool maybe impacted. 
If variable directory is necessary/must-have to support next integration, i would suggest accept this patch, and add one known issue "#variable directory(bindir) could only be set as default : /usr/bin". We will have followup design discuss/patch to fix this limitation. thanks.

saininav:
> @saininav variable directory(bindir) rather than hardcode(/usr/bin) is good way to support cross compilation environment and more flexible for system integration. But as @gvancuts mentioned, there are dependency of some acrn tools (acrn-crash, acrnlog, acrntrace,arn bridge, acrn-config) with this directory also.
> So the variable(bindir) can only be set as default value(/usr/bin), just passed basic CI check. If change/override to other value, some tool maybe impacted.
> If variable directory is necessary/must-have to support next integration, i would suggest accept this patch, and add one known issue "#variable directory(bindir) could only be set as default : /usr/bin". We will have followup design discuss/patch to fix this limitation. thanks.

Noted and yes, it is required in Yocto cross compilation environment.

gvancuts:
> Noted and yes, it is required in Yocto cross compilation environment.

Completely agree on the requirement. Is this patch already useful despite some of the brokenness it carries?

If we merge this as is, I'd like to make sure we also open a bug report with what's known to be broken. My list above is a starting point but not a thorough analysis of the situation.

terryzouhao:
Yes, @gvancuts we have reviewed and acknowledged this requirement from design/arch perspective. Really valuable to improve ACRN flexibility and usability.  
Good news is this PR is already useful if don't change default value e.g., bindir as 'usr/bin', but we do noticed there will be limitation/change needed to fully support directory change, e.g., config tool.  So we are doing testing(acrn log, trace...) to evaluate the limitation (open bug for detailed known broken parts) and also trying if we can add some fix with this PR together. 
Anyway, we target decision of this PR this week, keep you updated.

wenlingz:
@zhouji3x @ywan170 please help approve this PR, thanks

PR #5059, opened 2020-07-20
HV: inject physical PCIEXBAR into vhostbridge, and refine LoC

qianwang-intel:
hv: vpci: inject physical PCIEXBAR into vhostbridge, and refine LoC

To solve ACRN-6464

### DIFF
--- a/hypervisor/dm/vpci/vhostbridge.c
+++ b/hypervisor/dm/vpci/vhostbridge.c
@@ -48,43 +48,26 @@ static void init_vhostbridge(struct pci_vdev *vdev)
 	/* PCI config space */
 	pci_vdev_write_vcfg(vdev, PCIR_VENDOR, 2U, 0x8086U);
 	pci_vdev_write_vcfg(vdev, PCIR_DEVICE, 2U, 0x5af0U);
-
 	pci_vdev_write_vcfg(vdev, PCIR_REVID, 1U, 0xbU);
-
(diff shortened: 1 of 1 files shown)


hypervisor/dm/vpci/vhostbridge.c
```diff
@@ -43,8 +43,51 @@
  * @pre vdev != NULL
  * @pre vdev->vpci != NULL
  */
+/* The chart below shows the hostbridge DID high-byte of the platform later than broadwell, whose PCIEXBAR are always
```
dongyaozu, line 46: The commit message is wrong. What we want to do is to implement a full host bridge following HW spec. All the rest is a side effect or side gain.

hypervisor/dm/vpci/vhostbridge.c
```diff
@@ -48,43 +48,26 @@ static void init_vhostbridge(struct pci_vdev *vdev)
 	/* PCI config space */
 	pci_vdev_write_vcfg(vdev, PCIR_VENDOR, 2U, 0x8086U);
 	pci_vdev_write_vcfg(vdev, PCIR_DEVICE, 2U, 0x5af0U);
-
```
dongyaozu, line 51: refine the init_vhostbridge to use DWORD operations to simplify the initialization operation. LOC is an additional gain.

Issue #5115, opened 2020-07-30
WaaG can not reboot or poweroff when cpu-sharing is enabled.


conghuic23:
On WHL, when Waag share CPU with other VMs, there would be problem with reboot or poweroff.

conghuic23:
[External_System_ID]ACRN-6587

PR #5116, opened 2020-07-30
hv: fix deadloop in sleep_thread_sync

conghuic23:
As we only set BLOCKED status in context switch_out, which means, only
running thread can be changed to BLOCKED, but runnable thread can not.
This lead to the deadloop in sleep_thread_sync.

To solve the problem, in sleep_thread, we set the status to BLOCKED
directly when the original thread status is RUNNABLE.

Tracked-On: #5115


HuiHuangShi:

Issue #5148, opened 2020-08-07
 dm: support to provide ACPI SSDT for UOS


lonzoc:
SSDT is a continuation of the DSDT, those device belonging to sepcific board should better been add to SSDT.
    
We can describe devices in a SSDT ASL file, compile the SSDT asl file using ACPI compiler(e.g. iasl) before passing it to acrn-dm by --acpi_ssdt parameter, then dm will combine it with other ACPI tables.

fuzhongl:
[External_System_ID] ACRN-5149

PR #5149, opened 2020-08-07
dm: support to provide ACPI SSDT for UOS

lonzoc:
SSDT is a continuation of the DSDT, those device belonging to sepcific
board should better been add to SSDT.

We can describe devices in a SSDT ASL file, compile the SSDT asl file
using ACPI compiler(e.g. iasl) before passing it to acrn-dm by
--acpi_ssdt parameter, then dm will combine it with other ACPI tables.

Tracked-On: #5148
Signed-off-by: Cai Yulong <yulongc@hwtc.com.cn>

### DIFF
--- a/devicemodel/core/main.c
+++ b/devicemodel/core/main.c
@@ -144,7 +144,7 @@ usage(int code)
 		"       %*s [--vtpm2 sock_path] [--virtio_poll interval] [--mac_seed seed_string]\n"
 		"       %*s [--vmcfg sub_options] [--dump vm_idx] [--debugexit] \n"
 		"       %*s [--logger-setting param_setting] [--pm_notify_channel]\n"
-		"       %*s [--pm_by_vuart vuart_node] <vm>\n"
+		"       %*s [--pm_by_vuart vuart_node] [--acpi_ssdt ssdt_aml_path] <vm>\n"
 		"       -A: create ACPI tables\n"
 		"       -B: bootargs for kernel\n"
 		"       -E: elf image path\n"
@@ -168,6 +168,7 @@ usage(int code)
 		"       --vsbl: vsbl file path\n"
 		"       --ovmf: ovmf file path\n"
 		"       --cpu_affinity: list of pCPUs assigned to this VM\n"
+		"       --acpi_ssdt: acpi SSDT aml file path\n"
 		"       --part_info: guest partition info file path\n"
 		"       --enable_trusty: enable trusty for guest\n"
 		"       --debugexit: enable debug exit function\n"
@@ -736,6 +737,7 @@ enum {
 	CMD_OPT_VSBL = 1000,
 	CMD_OPT_OVMF,
 	CMD_OPT_CPU_AFFINITY,
+	CMD_OPT_ACPI_SSDT,
 	CMD_OPT_PART_INFO,
 	CMD_OPT_TRUSTY_ENABLE,
 	CMD_OPT_VIRTIO_POLL_ENABLE,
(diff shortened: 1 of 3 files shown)


HuiHuangShi:

wenlingz:

lonzoc:
@ywan170  We put board specific device which doesnt belong to the soc platform into vSSDT,   like tochscreen,  sensors, video codec, etc.

I think acrn-dm should be generic and doesn't need be recompiled when used at different board.

ywan170:
> @ywan170 We put board specific device which doesnt belong to the soc platform into vSSDT, like tochscreen, sensors, video codec, etc.
> 
> I think acrn-dm should be generic and doesn't need be recompiled when used at different board.

You want to pass the SOS's physical SSDT to guest through acrn-dm?

lonzoc:
No, not the SOS's physical SSDT table,   the devices described in vSSDT is for guest OS.
This vSSDT table provide a way for adding new device into guest os without recompiling acrn-dm.

lonzoc:
For example, we  passthrough one I2C bus to the guest os,  and there is a audio codec under this I2C bus,    if current acrn-dm design, we should add one ACPI node into passthrough.c and recompile the acrn-dm,    but with --acpi-ssdt option, we can define the audio codec in one SSDT asl file and compile it using iasl and then pass it to acrn-dm.

ywan170:
> This vSSDT table provide a way for adding new device into guest os without recompiling acrn-dm.

What device do you want to expose as this way? The device should has its own hardware resources(like mmio/pio/interrupt). Only expose the vSSDT table is not enough. 

By our original design, the related ACPI DSDT table is insert during virtual devices initialization. W/o re-compile the acrn-dm, you can just add command line parameter to indicate which virtual devices you want to expose to the guest, and the mediator of virtual device should inject its own ACPI device entity if necessary. Even you want to passthrough some PCI devices, the related ACPI table should hard-code in hw/pci/passthrough.c. You can refer to :

static void
write_dsdt_xdci(struct pci_vdev *dev)
{
    pr_info("write virt-%x:%x.%x in dsdt for XDCI @ 00:15.1\n",
           dev->bus,
           dev->slot,
           dev->func);

    dsdt_line("");
    dsdt_line("Device (XDCI)");
    dsdt_line("{");
    dsdt_line("    Name (_ADR, 0x%04X%04X)", dev->slot, dev->func);
    dsdt_line("    Name (_DDN, \"Broxton XDCI controller\")");
    dsdt_line("    Name (_STR, Unicode (\"Broxton XDCI controller\"))");
    dsdt_line("}");
    dsdt_line("");
}

static void
passthru_write_dsdt(struct pci_vdev *dev)
{
 ...
    /* Provides ACPI extra info */
    if (device == 0x5aaa)
        /* XDCI @ 00:15.1 to enable ADB */
        write_dsdt_xdci(dev);
...
}

lonzoc:
Yes, your description is fit for PCI devices.
For, non-PCI devices( non-discovery devices), like i2c/spi  client device,   we need modify it for every different board and recompile acrn-dm.

ywan170:
> For example, we passthrough one I2C bus to the guest os, and there is a audio codec under this I2C bus, if current acrn-dm design, we should add one ACPI node into passthrough.c and recompile the acrn-dm, but with --acpi-ssdt option, we can define the audio codec in one SSDT asl file and compile it using iasl and then pass it to acrn-dm.

I see you point. But actually this is one big topic, not just one dynamic insert vSSDT table.
As you known, the ACPI device objects are usually describe the platform/board specific resources. Likes I2C Addr, I/O APIO pin, GPIO, etc. Lots of command line parameters are dependent each other, all of them need to be set otherwise the device will not work... Likes you expose one I2C device w/ your dynamic vSSDT table but hasn't map ioapic or gpio..

The simple way is hard code everything in the acrn-dm and distinguish devices via their IDs which specified in command line just like passthrough case. Otherwise, it is hard to check the correctness if everything are passed through command line parameter. Likes your way, we need to parse what resources in your vSSDT file and if all of them are set correctly in the command line...

lonzoc:
I agree, my solution heavily depends on comline been correctly set.

ywan170:
> I agree, my solution heavily depends on comline been correctly set.

And some resources haven't provide command line parameters interface like ioapic/pic...
The acrn is designed for embedded products, it is not like could hypervisor. So needn't provide too much flexibility guest configurations. To my understanding, the acrn-dm have to modify per products..

Issue #5411, opened 2020-10-22
Supporting power off for pre-launched VMs


donsheng:
Currently powering off pre-launched VMs is not supported.
Need to add this support for release 2.3

[External_System_ID] ACRN-6585

donsheng:
[External_System_ID] ACRN-6585

PR #5457, opened 2020-11-03
backport "Add support for pre-launched VM shutdown" from master branch

donsheng:
More than one VM may request shutdown on the same pCPU before
shutdown_vm_from_idle() is called in the idle thread when pCPUs are shared among VMs.

Currently, ACRN only support shutdown when triple fault happens, because ACRN
doesn't present/emulate a virtual HW, i.e. port IO, to support shutdown. This
patch emulate a virtual shutdown component, and the vACPI method for guest OS
to use.

Also, the Reset Register Supported flag is not set in ACPI FADT table, so pre-
launched VMs will not use the ACPI reset register to reboot.

This patch series is therefore an attempt to address the above-mentioned gap.

Changes:

Use a per-pCPU bitmap to store all the VMIDs requesting shutdown

Specify valid ACPI reset register address and value in FADT table so that
pre-launched VMs can recognize and use ACPI reset register to reboot

Change ACPI version from V3 to V5 to support hardware-reduced ACPI, which is
a V5 feature. Using hardware-reduced mode allows us to use a much simpler form
of ACPI that does not require supporting the legacy of previous versions of the
specification such as SCI IRQ

When hardware-reduced ACPI is used, we need to specify the kernel boot argument
'reboot=acpi'. This is because if hardware reduced ACPI is detected, Linux will
set the reboot type to use EFI for rebooting. In this patch series, we specify
the "reboot=acpi" kernel parameter to force Linux to use ACPI for rebooting.
Another alternative to fix this issue is to set CONFIG_EFI to 'n' in kernel
config file when building the kernel for pre-launched VMs. We'd like to hear
from you about which one is better.

Tracked-On: #5411


HuiHuangShi:

dongyaozu (approved):

Issue #5461, opened 2020-11-03
DM: gvt: Identical mapping for GPU DSM


psun3x:
**Describe the bug**

**Platform**
Describe the platform information, 

**Codebase**
Describe which codebase is used.

**Scenario**
Describe the scenario: Hybrid, Industry or logical_partition

**To Reproduce**

**Expected behavior**

**Additional context**
Attach the following logs if available:

- native Linux boot kernel message log
- ACRN serial port log (when issues happens)
- Service VM kernel message when issue happens
- acrn-dm log: all files in /var/log/acrn-tm
- console log when running script to launch guest VM

psun3x:
[External_System_ID] ACRN-6780

PR #5465, opened 2020-11-03
DM: gvt: Identical mapping for GPU DSM

psun3x:
Windows graphic driver obtains DSM address from in-BAR mmio register
which has passthroughed. Not like the other platforms obtained from
pci configure space register which has virtualized. So TGL has to
keep identical mapping to avoid trap mmio BAR to do the emulation.

To keep simple, this patch hardcode the TGL DSM region in vE820
table, this will cause memory waste here. In the near future, we
need refine the entire vE820 logic as it is hard to maintained
due to many reserved regions have introduced in recently.

Signed-off-by: Sun Peng <peng.p.sun@intel.com>
Tracked-On: #5461


ywan170 (approved):

HuiHuangShi:

PR #5477, opened 2020-11-06
doc: fix breathe config for C vs CPP

dbkinder:
Breathe and doxygen work best for C++, but we're using C.  This shows up
as API documentaiton having a C++ flavor (modules and classes) instead
of the expected C flavor.  We really need to upgrade the versions of
doxygen and breathe to newer versions, and this configuration tweak
prepares for this.  (It will need CI coordination to update these tools,
but the changes in this PR are compatible with the current older tools.)


HuiHuangShi:

PR #5499, opened 2020-11-11
HV: hide CONFIG_SERIAL_PIO_BASE IO port for SOS

taoyuhong:
Always trap for accessing IO port at CONFIG_SERIAL_PIO_BASE(0x3f8).

SOS will trap at IO 0x3f8 access because vuart[0]. When HW platform
has this serial IO port, and HV use it as debug shell, If vuart[0]
is not set, SOS & HV console input/output are mixed up.

So SOS should always trap at CONFIG_SERIAL_PIO_BASE

Tracked-On: #5498


ywan170:
@fyin1 @yaomichael  please help approve

fyin1 (approved):

ywan170:
For this patch, the functionality is working then we can merge for 2.3 release.
For master, we need to review in the mailing list.
@taoyuhong , I think the logic needs to move to uart16550_init.

PR #5503, opened 2020-11-11
Fix: do not need to hide IO 0x3f8 for SOS on EHL

taoyuhong:
when HW platform like EHL use mmio type serial, do not need to hide
serial IO port for SOS

Tracked-On: #5498


wenlingz:

fyin1 (approved):

PR #5659, opened 2021-01-26
doc: use 'apt' to install GCC for Windows

gvancuts:
Replace the original Clear Linux (`swupd`) command to add the GCC
compiler for Windows by the equivalent in Ubuntu (i.e. using Ubuntu
package manager 'apt')


dbkinder (approved):
Good catch, thanks!

PR #5663, opened 2021-01-27
doc: add script to build acrn ovmf with GOP driver

chejianj:
Add a script to build acrn ovmf with GOP driver. The build is
using a ubuntu 16.04 based docker image.

### DIFF
--- a/doc/static/downloads/build_acrn_ovmf.sh
+++ b/doc/static/downloads/build_acrn_ovmf.sh
@@ -1,0 +1,220 @@
+#!/bin/bash
+# Copyright (C) 2021 Intel Corporation.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# PREREQUISITES:
+# 1) Get your specific "IntelGopDriver.efi" and "Vbt.bin"
+#    from your BIOS vender
+# 2) Install Docker on your host machine and allow non-root users
+#    For Ubuntu: https://docs.docker.com/engine/install/ubuntu/
+#    To enable non-root users: https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user
+# 3) If you are working behind proxy, create a file named
+#    "proxy.conf" in ${your_working_directory} with
+#    configurations like below:
+#    Acquire::http::Proxy "http://x.y.z:port1";
+#    Acquire::https::Proxy "https://x.y.z:port2";
+#    Acquire::ftp::Proxy "ftp://x.y.z:port3";
+#
+# HOWTO:
+# 1) mkdir ${your_working_directory}
+# 2) cd ${your_working_directory}
+# 2) mkdir gop
+# 3) cp /path/to/IntelGopDriver.efi /path/to/Vbt.bin gop
+# 4) cp /path/to/build_acrn_ovmf.sh ${your_working_directory}
+# 5) ./build_acrn_ovmf.sh
+#
+# OUTPUT: ${your_working_directory}/acrn-edk2/Build/OvmfX64/DEBUG_GCC5/FV/OVMF.fd
+#
+# For more information, ./build_acrn_ovmf.sh -h
+#
+
+gop_bin_dir="./gop"
+docker_image_name="ubuntu:ovmf.16.04"
+proxy_conf="proxy.conf"
+acrn_ver="latest"
+
+if [ ! -x "$(command -v docker)" ]; then
+    echo "Install Docker first:"
+    echo "If you are using Ubuntu, you can refer to: https://docs.docker.com/engine/install/ubuntu/"
+    exit
+fi
+
+if [ ! -d "${gop_bin_dir}" ]; then
+    mkdir ${gop_bin_dir}
+    echo "Copy IntelGopDriver.efi and Vbt.bin to ${gop_bin_dir}"
+    exit
+fi
+
+if [ ! -f "${gop_bin_dir}/IntelGopDriver.efi" ]; then
+    echo "Copy IntelGopDriver.efi to ${gop_bin_dir}"
+    exit
+fi
+
+if [ ! -f "${gop_bin_dir}/Vbt.bin" ]; then
+    echo "Copy Vbt.bin to ${gop_bin_dir}"
+    exit
+fi
+
+if [ ! -f "${proxy_conf}" ]; then
+    touch "${proxy_conf}"
+fi
+
+usage()
+{
+    echo "$0 [-v ver] [-i] [-s] [-h]"
+    echo "  -v ver: The release version of ACRN, e.g. 2.3"
+    echo "  -i:     Delete the existing docker image ${docker_image_name} and re-create it"
+    echo "  -s:     Delete the existing acrn-edk2 source code and re-download/re-patch it"
+    echo "  -h:     Show this help"
+    exit
+}
+
+re_download=0
+re_create_image=0
+
+while getopts "hisv:" opt
+do
+    case "${opt}" in
+        h)
+            usage
+            ;;
+        i)
+            re_create_image=1
+            ;;
+        s)
+            re_download=1
+            ;;
+        v)
+            acrn_ver=${OPTARG}
+            ;;
+        ?)
+            echo "${OPTARG}"
+            ;;
+    esac
+done
+shift $((OPTIND-1))
+
+if [[ "${re_create_image}" -eq 1 ]]; then
+    if [[ "$(docker images -q ${docker_image_name} 2> /dev/null)" != "" ]]; then
+        echo "===================================================================="
+        echo "Deleting the old Docker image ${docker_image_name}  ..."
+        echo "===================================================================="
+        docker image rm -f "${docker_image_name}"
+    fi
+fi
+
+if [[ "${re_download}" -eq 1 ]]; then
+    echo "===================================================================="
+    echo "Deleting the old acrn-edk2 source code ..."
+    echo "===================================================================="
+    sudo rm -rf acrn-edk2
+fi
+
+create_acrn_edk2_workspace()
+{
+    echo "===================================================================="
+    echo "Downloading & patching acrn_edk2 source code ..."
+    echo "===================================================================="
+
+    [ -d acrn-edk2 ] && sudo rm -rf acrn-edk2
+
+    git clone https://github.com/projectacrn/acrn-edk2.git
+    if [ $? -ne 0 ]; then
+        echo "git clone acrn-edk2 failed"
+        return 1
+    fi
+
+    cd acrn-edk2
+    git submodule update --init CryptoPkg/Library/OpensslLib/openssl
+    if [ $? -ne 0 ]; then
+        echo "git submodule acrn-edk2 failed"
+        return 1
+    fi
+
+    if [ "${acrn_ver}" != "latest" ]; then
+        git checkout --recurse-submodules -b "v${acrn_ver}" "ovmf-acrn-v${acrn_ver}"
+        if [ $? -ne 0 ]; then
+            echo "git checkout --recurse-submodules -b v${acrn_ver} ovmf-acrn-v${acrn_ver} failed"
+            return 1
+        fi
+    fi
+
+    wget -q https://projectacrn.github.io/${acrn_ver}/_static/downloads/Use-the-default-vbt-released-with-GOP-driver.patch
+    if [ $? -ne 0 ]; then
+        echo "Downloading Use-the-default-vbt-released-with-GOP-driver.patch failed"
+        return 1
+    fi
+
+    wget -q https://projectacrn.github.io/${acrn_ver}/_static/downloads/Integrate-IntelGopDriver-into-OVMF.patch
+    if [ $? -ne 0 ]; then
+        echo "Downloading Integrate-IntelGopDriver-into-OVMF.patch failed"
+        return 1
+    fi
+
+    git am --keep-cr Use-the-default-vbt-released-with-GOP-driver.patch
+    if [ $? -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


doc/static/downloads/build_acrn_ovmf.sh
```diff
@@ -0,0 +1,206 @@
+#!/bin/bash
+# Copyright (C) 2021 Intel Corporation.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# PREREQUISITES:
+# 1) Get your specific "IntelGopDriver.efi" and "Vbt.bin"
+#    from your BIOS vender
+# 2) Install docker on your host machine WITHOUT sudo.
+#    For ubuntu: https://docs.docker.com/engine/install/ubuntu/
```
gvancuts, line 9:
```suggestion
#    For Ubuntu: https://docs.docker.com/engine/install/ubuntu/
#    To enable non-root users: https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user
```

doc/static/downloads/build_acrn_ovmf.sh
```diff
@@ -0,0 +1,206 @@
+#!/bin/bash
+# Copyright (C) 2021 Intel Corporation.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# PREREQUISITES:
+# 1) Get your specific "IntelGopDriver.efi" and "Vbt.bin"
+#    from your BIOS vender
+# 2) Install docker on your host machine WITHOUT sudo.
```
gvancuts, line 8:
```suggestion
# 2) Install Docker on your host machine and allow non-root users
```

doc/static/downloads/build_acrn_ovmf.sh
```diff
@@ -0,0 +1,206 @@
...
+#
+# For more information, ./build_acrn_ovmf.sh -h
+#
+
+gop_bin_dir="./gop"
+docker_image_name="ubuntu:ovmf.16.04"
+proxy_conf="proxy.conf"
+
+if [ ! -x "$(command -v docker)" ]; then
+    echo "Please install docker first:"
```
gvancuts, line 35:
```suggestion
    echo "Please install Docker first:"
```

doc/static/downloads/build_acrn_ovmf.sh
```diff
@@ -0,0 +1,206 @@
...
+    echo "If you are using ubuntu, you can refer to: https://docs.docker.com/engine/install/ubuntu/"
```
gvancuts, line 36:
```suggestion
    echo "If you are using Ubuntu, you can refer to: https://docs.docker.com/engine/install/ubuntu/"
```

doc/static/downloads/build_acrn_ovmf.sh
```diff
@@ -0,0 +1,206 @@
...
+fi
+
+if [ ! -f "${proxy_conf}" ]; then
+    touch "${proxy_conf}"
+fi
+
+usage()
+{
+    echo "$0 [-i] [-s] [-h]"
+    echo "  -i:    Delete the existing docker image ${docker_image_name} and re-create it"
```
gvancuts, line 63:
```suggestion
    echo "  -i:    Delete the existing Docker image ${docker_image_name} and re-create it"
```

doc/tutorials/gpu-passthru.rst
```diff
@@ -165,3 +165,11 @@ Keep in mind the following:
       ``Build/OvmfX64/DEBUG_GCC5/FV/OVMF.fd``. Transfer the binary to
       your target machine.
    -  Modify the launch script to specify the OVMF you built just now.
+
+Script
+=====
```
gvancuts, line 170:
```suggestion
======
```

doc/tutorials/gpu-passthru.rst
```diff
@@ -165,3 +165,11 @@ Keep in mind the following:
...
+
+Here is a `Script <../_static/downloads/build_acrn_ovmf.sh>`_ to do all the
+steps above automatically. Before you can use the script to build ACRN OVMF
+with GOP driver enabled please install docker environment first. For details
```
gvancuts, line 174:
```suggestion
with GOP driver enabled please install Docker environment first. For details
```

doc/tutorials/gpu-passthru.rst
```diff
@@ -165,3 +165,11 @@ Keep in mind the following:
...
+about the script usage please check the script itself.
```
gvancuts, line 175:
```suggestion
about the script usage please check the script itself (``build_acrn_ovmf.sh -h``).
```

gvancuts (approved):

doc/tutorials/gpu-passthru.rst
```diff
@@ -165,3 +165,11 @@ Keep in mind the following:
       ``Build/OvmfX64/DEBUG_GCC5/FV/OVMF.fd``. Transfer the binary to
       your target machine.
    -  Modify the launch script to specify the OVMF you built just now.
+
+Script
+======
+
+Here is a `Script <../_static/downloads/build_acrn_ovmf.sh>`_ to do all the
+steps above automatically. Before you can use the script to build ACRN OVMF
+with GOP driver enabled please install Docker environment first. For details
```
dbkinder, line 174: Change this to read better (and without please):
```
Once you've installed the Docker environment, you can use this
script to build ACRN OVMF with the GOP driver enabled. For more details
about the script usage, run ``build_acrn_ovmf.sh -h``.
```

dbkinder (changes requested):

doc/static/downloads/build_acrn_ovmf.sh
```diff
@@ -0,0 +1,207 @@
...
+    fi
+
+    cd acrn-edk2
+    git submodule update --init CryptoPkg/Library/OpensslLib/openssl
+    if [ $? -ne 0 ]; then
+        echo "git submodule acrn-edk2 failed"
+        return 1
+    fi
+
+    wget -q https://projectacrn.github.io/latest/_static/downloads/Use-the-default-vbt-released-with-GOP-driver.patch
```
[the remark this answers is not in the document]
gvancuts: > 
> 
> This creates a dependency on the `/latest/` folder from an (older) tagged version of the release. This wget needs to fetch from the appropriate tagged release's documentation tree (/2.4/ for example). We should talk about how to make this work (maybe add a configuration variable with the release name (e.g., "latest" or "2.4" corresponding to the release version)? @gvancuts @NanlinXie any good ideas here?

No, I wanted to put a comment in there too but refrained from doing so since I did not have any better idea :-)

> 
> Here's one way I figured out that automatically figures this out by using `git describe` to find what tagged branch we're on and figures out the folder name in the documentation to use (either `latest` or the branch name without the leading `v`), but might need some robustness testing:
> 
> ```
> tag=`git describe --exact-match 2> /dev/null`
> folder=`[ -z "$tag" ] && echo latest || echo ${tag:1}`
> 
> wget -q https://projectacrn.github.io/${folder}/_static/downloads/.....
> ```

This could _nearly_ work but not quite because we're in the `acrn-edk2` clone at that point and we're always using the latest version... so this will invariably resolve to `latest`. **If** we introduce a mechanism to specify the ACRN version we want to use, then this could be used to **both** check out the corresponding version of `acrn-edk2` **and** the right version of the patch. This seems like the best option to me, i.e. add an argument to this script to build for a specific version, defaulting to `latest` if none is specified.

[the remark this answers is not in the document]
dbkinder: I didn't notice we're in a clone of a different repo, sigh.  In that case, yes, we should add an argument for what version of the acrn-hypervisor (documentation) we should pull from or pull the script from a particular branch name in the git repo.  We're trying to solve an issue we have where scripts are referenced in tagged version documentation (e.g., 2.3) but aren't pointing to a script version specific to that tagged release (they're pointing to /latest).

doc/static/downloads/build_acrn_ovmf.sh
```diff
@@ -0,0 +1,220 @@
...
+    fi
+
+    cd acrn-edk2
+    git submodule update --init CryptoPkg/Library/OpensslLib/openssl
+    if [ $? -ne 0 ]; then
+        echo "git submodule acrn-edk2 failed"
+        return 1
+    fi
+
+    if [ "${acrn_ver}" != "latest" ]; then
```
gvancuts, line 134: I think it would be better to put this section before the submodule initialization. It does not matter too much at the moment as we always point at master in the submodule but that may change in the future and so it would be better to make sure the submodule update picks up the version specified in tagged `acrn-edk2` source code.

dbkinder (approved):
I think this should work, thanks.

NanlinXie (approved):

doc/static/downloads/build_acrn_ovmf.sh
chejianj: My understanding is that when we change to a tag in the acrn-edk2 repo, the sub-module will be checked out to the according commit id of that tag.

gvancuts: As far as I know, this is not the default behaviour. You need to add `--recurse-submodules` to the `checkout` command if you want that. It's actually a very good option, even better than moving this block.

https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---recurse-submodules

dbkinder:
@gvancuts Checking if we need to have that ``--recurse-submodules`` option added before merging this PR?  (You've approved the PR as is.)

doc/static/downloads/build_acrn_ovmf.sh
chejianj: https://git-scm.com/book/en/v2/Git-Tools-Submodules

The DbConnector directory is there, but empty. You must run two commands: git submodule init to initialize your local configuration file, and git submodule update to fetch all the data from that project and check out the appropriate commit listed in your superproject:

There is another way to do this which is a little simpler, however. If you pass --recurse-submodules to the git clone command, it will automatically initialize and update each submodule in the repository, including nested submodules if any of the submodules in the repository have submodules themselves.

gvancuts: I'm not sure I'm following you @chejianj , can  you explain further what you mean? As I understand it, this is what you do at the moment:
* Git clone (not using the `--recurse-submodule` option
* You populate the submodule in the next step (`git submodule update --init`), **note** you're on master at this point
* Conditionally checkout a different tag of the repo. But since you're not using `--recurse-module` here, your submodules will stay at the version that was checked out while you were on master, and that may be different than the one we should use for the tag being checked out.

gvancuts (changes requested):

gvancuts:
> 
> 
> @gvancuts Checking if we need to have that `--recurse-submodules` option added before merging this PR? (You've approved the PR as is.)

Let's close on that question before we merge this.

doc/static/downloads/build_acrn_ovmf.sh
gvancuts: Hi @chejianj , here is a concrete example of what I mean.
* With the current script (not checking out any particular version):
```
$ git describe HEAD --long && git submodule status
ovmf-acrn-v2.3-2-g1c8bef7363
 c3656cc594daac8167721dde7220f0e59ae146fc CryptoPkg/Library/OpensslLib/openssl (OpenSSL_1_1_1d-2-gc3656cc594)
```
* With the current script (checking out 1.1):
```
$ git describe HEAD --long && git submodule status
ovmf-acrn-v1.1-0-gfa06a760f4
+c3656cc594daac8167721dde7220f0e59ae146fc CryptoPkg/Library/OpensslLib/openssl (OpenSSL_1_1_1d-2-gc3656cc594)
```
* With what I am proposing to use (see diff below [1])
```
$ git describe HEAD --long && git submodule status
ovmf-acrn-v1.1-0-gfa06a760f4
 d4e4bd2a8163f355fa8a3884077eaec7adc75ff7 CryptoPkg/Library/OpensslLib/openssl (OpenSSL_1_1_0h)
```

The difference is that in the current version of the script, the version of the submodule stays at the version used in master, although we are checking out an older tag. Does that help understand what I mean?

[1] Proposed changes to the script:
```
diff --git a/doc/static/downloads/build_acrn_ovmf.sh b/doc/static/downloads/build_acrn_ovmf.sh
index e4a5b27d..1a56d504 100644
--- a/doc/static/downloads/build_acrn_ovmf.sh
+++ b/doc/static/downloads/build_acrn_ovmf.sh
@@ -118,23 +118,17 @@ create_acrn_edk2_workspace()
 
     [ -d acrn-edk2 ] && sudo rm -rf acrn-edk2
 
-    git clone https://github.com/projectacrn/acrn-edk2.git
+    git clone --recurse-submodules=CryptoPkg/Library/OpensslLib/openssl https://github.com/projectacrn/acrn-edk2.git
     if [ $? -ne 0 ]; then
         echo "git clone acrn-edk2 failed"
         return 1
     fi
 
     cd acrn-edk2
-    git submodule update --init CryptoPkg/Library/OpensslLib/openssl
-    if [ $? -ne 0 ]; then
-        echo "git submodule acrn-edk2 failed"
-        return 1
-    fi
-
     if [ "${acrn_ver}" != "latest" ]; then
-        git checkout -b "v${acrn_ver}" "ovmf-acrn-v${acrn_ver}"
+        git checkout --recurse-submodules "ovmf-acrn-v${acrn_ver}"
         if [ $? -ne 0 ]; then
-            echo "git checkout -b v${acrn_ver} ovmf-acrn-v${acrn_ver} failed"
+            echo "git checkout --recurse-submodules ovmf-acrn-v${acrn_ver} failed"
             return 1
         fi
     fi
```

gvancuts (approved):
LGTM, thanks!

Issue #5669, opened 2021-01-28
Top level "make clean" 


gvancuts:
A top-level `make clean` (or `make clean` in the `doc/` folder) will produce the following output:
```
...
make -C /home/gvancuts/work/acrn-hypervisor/doc BUILDDIR=/home/gvancuts/work/acrn-hypervisor/build/doc clean
make[1]: Entering directory '/home/gvancuts/work/acrn-hypervisor/doc'
rm -fr /home/gvancuts/work/acrn-hypervisor/build/doc
# Keeping these temporarily, but no longer strictly needed.
rm -fr doxygen
rm -fr misc
rm -fr reference/kconfig/*.rst
make[1]: Leaving directory '/home/gvancuts/work/acrn-hypervisor/doc'
rm -rf /home/gvancuts/work/acrn-hypervisor/build
```

The `# Keeping these temporarily, but no longer strictly needed.` line is intended to be a comment internal to the `Makefile` and while it's completely harmless it could leave the user a bit puzzled. It is preferable, IMHO, to make this silent.

gvancuts:
Patch was merged, closing.

PR #5670, opened 2021-01-28
Makefile: make internal comment in doc Makefile silent

gvancuts:
There is a comment in the doc/Makefile that is beinhg spit out
when calling 'make clean'. This is harmless but can be confusing
to users so let's make it silent.

Tracked-On: #5669


dbkinder (approved):
sure. That was your comment anyway :)

PR #5691, opened 2021-02-02
doc: remove Kconfig reference documentation

dbkinder:
With the new ACRN configuration architecture, we no longer use Kconfig
files.  Remove the Kconfig option documentation scripting (genrest.py)
Python dependencies, and Makefile commands, and change references in the
documentation from the Kconfig option (such as
:option:`CONFIG_MEM_LOGLEVEL`) to the new schema definition-based option
documentation (:option:`hv.DEBUG_OPTION.MEM_LOGLEVEL`).


gvancuts:
There are other references to the `Kconfig` mechanism via things like `make menuconfig`, etc. Should these be part of this PR or are we planning to remove/update those at a later date?

dbkinder:
@gvancuts This PR focuses on removing the generated kconfig documents and fixing docs that referenced these now deleted docs.  More PRs will be needed to clean up the rest of the configuration documentation that talks about ``make menuconfig`` and such.

PR #5697, opened 2021-02-03
Keylocker support

shuox:
KeyLocker is a new security feature available in new Intel CPUs that
protects data-encryption keys for the Advanced Encryption Standard (AES)
algorithm. These keys are more valuable than what they guard. If stolen
once, the key can be repeatedly used even on another system and even
after vulnerability closed.

It also introduces a CPU-internal wrapping key (IWKey), which is a key-
encryption key to wrap AES keys into handles. While the IWKey is
inaccessible to software, randomizing the value during the boot-time
helps its value unpredictable.

Keylocker usage:
 - New “ENCODEKEY” instructions take original key input and returns HANDLE
   crypted by an internal wrap key (IWKey, init by “LOADIWKEY” instruction)
 - Software can then delete the original key from memory
 - Early in boot/software, less likely to have vulnerability that allows
   stealing original key
 - Later encrypt/decrypt can use the HANDLE through new AES KeyLocker
   instructions
 - Note:
      * Software can use original key without knowing it (use HANDLE)
      * HANDLE cannot be used on other systems or after warm/cold reset
      * IWKey cannot be read from CPU after it's loaded (this is the
        nature of this feature) and only 1 copy of IWKey inside CPU.

The virtualization implementation of Key Locker on ACRN is:
 - Each vCPU has a 'struct iwkey' to store its IWKey in struct
   acrn_vcpu_arch.
 - At initilization, every vCPU is created with a random IWKey.
 - Hypervisor traps the execution of LOADIWKEY (by 'LOADIWKEY exiting'
   VM-exectuion control) of vCPU to capture and save the IWKey if guest
   set a new IWKey. Don't support randomization (emulate CPUID to
   disable) of the LOADIWKEY as hypervisor cannot capture and save the
   random IWKey.
 - In context_switch_in() of each vCPU, hypervisor loads that vCPU's
   IWKey into pCPU by LOADIWKEY instruction.
 - There is an assumption that ACRN hypervisor will never use the
   KeyLocker feature itself.

v4:
  - Enabled host CR4.KL bit at boot time if CPU support keylocker feature.
  - Introduced a check_vmx_ctrl_64() for new 64bit MSR_IA32_VMX_PROCBASED_CTLS
  - Set per_cpu whose_iwkey to NULL at associated vcpu reset unconditionally.
  - Added some doc around MSRs
  - Other small refinements

v3:
  - Refined changelog
  - Moved cached CR4_KL flag logic into the first patch and used in following patches.
  - Refined some code logic.
  - Used pcpu_has_cap(X86_FEATURE_KEYLOCKER) directly instead of a wrapper of X86_FEATURE_KEYLOCKER bit.
  - Used local variables in C code in load_iwkey() to save/restore xmm0/xmm1/xmm2.
  - Assumed RDRAND CPU capability and check it at boot time.
  - Introduced per_cpu variable whose_iwkey to avoid the reload IWKey of same vCPU.
  - Wrap CR4 operation into loadiwkey().
  - Add overall design in patch 4/6

v2:
  - Rebased to latest codebase.
  - Support keylocker backup MSRs
  - Support ramdom initialized IWKey for each vCPU.
  - Optimized logic in context switch. Only do load IWKey action while vCPU CR4_KL bit enabled.


HuiHuangShi:

wenlingz:
@dongyaozu  please help approve this PR, which is for 2.4 release, thanks

dongyaozu (approved):

Issue #5689, opened 2021-02-02
[EHL][SBL] copy GPA error when booting zephyr as pre-launched VM


szhen11:
HW/Board
EHL SBL

Steps
1. build zephyr 2.5 from master refer to: https://docs.zephyrproject.org/latest/getting_started/index.html
https://docs.zephyrproject.org/latest/boards/x86/ehl_crb/doc/index.html
west build -p auto -b acrn samples/hello_world
2. build acrn with hybrid scenario from master + patch
http://acrn-dms-02.sh.intel.com/report/resource/zephyr/victor/0001-acrn-config-update-zephyr-entry-point-to-0x1000.patch测试hybrid
3. boot acrn and pre-launched zephyr

Expected result
pre-launched zephyr boot successfully without error

Actual result
pre-launched zephyr boot successfully with copy GPA error messages.
[20229639us][cpu=0][(null)][sev=2][seq=1]:HV version 2.4-unstable-2021-02-01 07:42:57-fb78029f0-dirty DBG (daily tag:acrn-2021w05.5-180000p) hybrid@ehl-crb-b build by zwl, start time 20221255us
[20249064us][cpu=0][(null)][sev=2][seq=2]:API version 1.0
[20255012us][cpu=0][(null)][sev=2][seq=3]:Detect processor: Genuine Intel(R) CPU 0000 @ 1.90GHz
[20473291us][cpu=3][(null)][sev=3[seq=19]:local_copy_gpa,vm[0] gpa 0xfffffffffffff010,GPA is unmapping
[20482841us][cpu=0][(null)][sev=2][seq=29]:Start VM id: 1 name: ACRN SOS VM
*[20483626us][cpu=3][(null)][sev=3][seq=30]:Unable to copy HPA 0x11052de0 to GPA 0xfffffffffffff010 in VM0

[20502933us][cpu=3][(null)][sev=3][seq=33]:local_copy_gpa,vm[0] gpa 0xfffffffffffff000,GPA is unmappingACRN:\>
[20514061us][cpu=3][(null)][sev=3][seq=34]:Unable to copy HPA 0x11050cac to GPA 0xfffffffffffff000 in VM0
*
[20525287us][cpu=3][(null)][sev=2][seq=36]:Start VM id: 0 name: ACRN PRE-LAUNCHED VM0
[20985728us][cpu=0][vm1:vcpu0][sev=3][seq=39]:vlapic: Start Secondary VCPU1 for VM[1]...
[...]

PR #5722, opened 2021-02-08
update entry point address for pre-launched zephyr on ehl-crb-b

szhen11:
config_tools: update zephyr entry point address on ehl-crb-b to work with zephyr 2.5
config_tools: update condition for bootargs error check only when kernel type is KERNEL_BZIMAGE.
HV: init VM bootargs only for LaaG
Currently the VM bootargs load address is hard-coded at 8KB right before
kernel load address, this should work for Linux kernel only since Linux
kernel is guaranteed to be loadered high than GPA 8K so its load address
would never be overflowed, other OS like Zephyr has no such assumption.

Tracked-On: #5689


HuiHuangShi:

terryzouhao (approved):

PR #5747, opened 2021-02-13
doc: fix all headings to use title case

dbkinder:
While we hoped to make the headings consistent over time while doing
other edits, we should instead just make the squirrels happy and do them
all at once or they'll likely never be made consistent.

A python script was used to find the headings, and then a call to
https://pypi.org/project/titlecase to transform the title.  A visual
inspection was used to tweak a few unexpected resulting titles.


gvancuts:
Yeah! 👍

PR #5753, opened 2021-02-22
doc: update build-from-source instructions

junjiemao1:
Users no longer need to explicitly generate configuration source code as
they are now generated at build time. This patch updates the relevant
instructions in the documentation.


junjiemao1:
@dbkinder May I know where I can find the source file of doc/tutorials/images/offline_tools_workflow.png? I would like to simplify the image as users do not need to invoke code generation explicitly now.

doc/getting-started/building-from-source.rst
```diff
@@ -176,99 +182,51 @@ Assuming that you are at the top level of the acrn-hypervisor directory, perform
 
   .. code-block:: none
 
-    $ make all BOARD_FILE=$PWD/misc/vm_configs/xmls/board-xmls/nuc6cayh.xml \
-    SCENARIO_FILE=$PWD/misc/vm_configs/xmls/config-xmls/nuc6cayh/sdc.xml
+    $ make all BOARD=$PWD/misc/vm_configs/xmls/board-xmls/nuc6cayh.xml \
```
gvancuts, line 185: The paths used here are no longer valid. This is not a very good example actually as this can also more simply be done using `make all BOARD=nuc6cayh SCENARIO=sdc`.

We should add a more concrete example showing how people can use that with XML files stored **outside** of the repo. I.e. explain that if someone keeps their custom XML files (`BOARD` and `SCENARIO`) in a folder called `my-own-board` that's outside of the `acrn-hypervisor` repo. The example becomes: `make BOARD=../my-own-board/my-board.xml SCENARIO=../my-own-board/my-scenario.xml`

doc/getting-started/building-from-source.rst
```diff
@@ -82,9 +82,10 @@ Install the necessary tools for the following systems:
           libnuma-dev \
           liblz4-tool \
           flex \
-          bison
+          bison \
+          xsltproc
 
-     $ sudo pip3 install kconfiglib
+     $ sudo pip3 install xmlschema
```
gvancuts, line 88: During my testing, it's `lxml` that one needs to install, not `xmlschema`.

I'm actually not sure if `xmlschema` is needed at all, I've never installed it and could build all different configurations.

doc/getting-started/building-from-source.rst
```diff
@@ -176,99 +182,51 @@ Assuming that you are at the top level of the acrn-hypervisor directory, perform
...
-    SCENARIO_FILE=$PWD/misc/vm_configs/xmls/config-xmls/nuc6cayh/sdc.xml
+    $ make all BOARD=$PWD/misc/vm_configs/xmls/board-xmls/nuc6cayh.xml \
+      SCENARIO=$PWD/misc/vm_configs/xmls/config-xmls/nuc6cayh/sdc.xml
 
+.. note::
+   ACRN uses XML files to summarize board characteristics and scenario
+   settings. The ``BOARD`` and ``SCENARIO`` variables accept board/scenario
+   names as well as paths to XML files. When board/scenario names are given, the
+   build system searches for XML files with the same names under
+   ``misc/config_tools/data/``. When relative files paths are given, the build
```
gvancuts, line 193: I would like that we enhance this explanation a bit, how about: `When paths (absolute or relative) to the XML files are given, the build system uses the files pointed at. If relative paths are used, they are considered relative to the root acrn-hypervisor` folder.

gvancuts (changes requested):
Thanks for updating our documentation @junjiemao1 !

Will you also update the Getting Started Guides or use a different PR for those?

junjiemao1:
> Thanks for updating our documentation @junjiemao1 !
> 
> Will you also update the Getting Started Guides or use a different PR for those?

The `Getting Started Guide` series all refers to specific ACRN versions. Updating these guides require not only changes related to configuration but also software versions and (possibly) other stuff, I would like to keep such changes in a separate PR for clarity.

doc/getting-started/building-from-source.rst
junjiemao1: Just refined the steps a bit by using arbitrary paths (assumed to be `/path/to/xxx`).

doc/getting-started/building-from-source.rst
junjiemao1: Thanks for pointing this out. I've added `lxml` to the list.

About `xmlschema`, it is recommended to be installed as it allows further checking to the configuration XML files at configuration time. That said, we still allow building without it due to Yocto integration issues.

doc/getting-started/building-from-source.rst
junjiemao1: Sure. I've integrated your suggestion in the refined version. BTW, relative paths are considered relative to the current working directory.

junjiemao1:
Compared with the previous version, the following changes are added:

1. The `ACRN Configuration Tool` page is refactored for clearer logic flow and updated per the new build steps.
2. References to `BOARD_FILE/SCENARIO_FILE` and `misc/vm_configs/xmls/...` are updated according to the new build steps and directory layout.

gvancuts:
> > Thanks for updating our documentation @junjiemao1 !
> > Will you also update the Getting Started Guides or use a different PR for those?
> 
> The `Getting Started Guide` series all refers to specific ACRN versions. Updating these guides require not only changes related to configuration but also software versions and (possibly) other stuff, I would like to keep such changes in a separate PR for clarity.

That makes sense, thanks :+1:

doc/tutorials/vuart_configuration.rst
```diff
@@ -334,13 +334,12 @@ The ACRN vUART related XML fields:
...
 
-  make BOARD_FILE=$PWD/misc/acrn-config/xmls/board-xmls/<board>.xml \
-     SCENARIO_FILE=$PWD/misc/acrn-config/xmls/config-xmls/<board>/<scenario>.xml
+  make BOARD=<board> SCENARIO=<scenario>
 
 The configuration tools will test your settings, and check :ref:`vUART
 Rules <index-of-vuart>` for compilation issue. After compiling, you can find
-``./misc/vm_configs/scenarios/<scenario>/<board>/pci_dev.c`` has been
-changed by the configuration tools based on the XML settings, something like:
+the generated sources under ``build/configs/<scenario>/<board>/pci_dev.c``
```
gvancuts, line 341:
```suggestion
the generated sources under ``build/configs/scenarios/<scenario>/pci_dev.c``
```

There is also an instance of `misc/vm_configs/scenarios/<xxx>/vm_configuration.c` in this file (above) that should be updated.

doc/tutorials/vuart_configuration.rst
```diff
@@ -267,7 +267,7 @@ and :ref:`How to Configure a Communication Port
 <how-to-configure-a-communication-port>`.
 
 You can configure both Legacy vUART and PCI-vUART in
-``./misc/vm_configs/xmls/config-xmls/<board>/<scenario>.xml``. For
+``./misc/config_tools/data/<board>/<scenario>.xml``. For
```
gvancuts, line 270: A more generic comment here. We should really advice people not to make direct changes to these files. Instead, they should create a copy, make modifications to the copy and then use it with the `SCENARIO` parameter.

gvancuts (changes requested):

doc/tutorials/acrn_configuration_tool.rst
```diff
@@ -4,59 +4,133 @@ ACRN Configuration Tool
...
+The board configuration stores hardware-specific information extracted on the
+target platform. It describes the capacity of hardware resources (such as
+processors and memory), platform power states, available devices and BIOS
+versions. This information is used by ACRN configuration tool to check feature
+availability and allocate resources among VMs, as well as by ACRN hypervisor to
+initialize and manage the platform at runtime.
 
-The hypervisor configuration uses the ``Kconfig`` mechanism.  The configuration
-file is located at ``acrn-hypervisor/hypervisor/arch/x86/Kconfig``.
+The board configuration is scenario-irrelevant by nature. Thus, multiple
```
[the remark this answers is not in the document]
dbkinder: ... or scenario-neutral

doc/tutorials/vuart_configuration.rst
dbkinder: yes indeed.  We don't want to have folks editing files that are under git control unless they intend to submit a PR

doc/getting-started/building-from-source.rst
```diff
@@ -149,7 +151,11 @@ HYBRID_RT:
...
 
-Assuming that you are at the top level of the acrn-hypervisor directory, perform the following:
+XML configuration files for those scenarios on supported boards are available
+under the ``misc/config_tool/data`` directory.
+
+Assuming that you are at the top level of the acrn-hypervisor directory, perform
+the following to build the hypervisor, device model and tools:
 
 .. note::
    The release version is built by default, ``RELEASE=0`` builds the debug version.
```
dbkinder, line 161: The ``whl-ipc-i7/hybrid_rt.xml`` explicitly sets ``RELEASE=n``  and the default value for ``RELEASE`` in the config.xsd file is also ``n``, so it would appear the default is now to build a debug image.    Also, does the make command need to say ``RELEASE=n`` now instead of ``RELEASE=0``?

doc/getting-started/building-from-source.rst
```diff
@@ -172,103 +178,68 @@ Assuming that you are at the top level of the acrn-hypervisor directory, perform
...
+scenario XML.
 
+The following commands show how to customize the scenario XML based on the
+predefined ``INDUSTRY`` scenario for ``nuc7i7dnb`` and rebuild the hypervisor.
 
 .. code-block:: none
 
-   $ make all BOARD_FILE=$PWD/misc/vm_configs/xmls/board-xmls/nuc7i7dnb.xml \
-   SCENARIO_FILE=$PWD/misc/vm_configs/xmls/config-xmls/nuc7i7dnb/industry.xml TARGET_DIR=xxx
+   $ make all BOARD=nuc7i7dnb SCENARIO=industry RELEASE=0
```
dbkinder, line 235: The default now is to build a debug version.  If you do want to leave the RELEASE setting on this make command, does it need to be ``RELEASE=n``?

dbkinder (changes requested):

doc/getting-started/building-from-source.rst
junjiemao1: All the predefined scenario XMLs specify `RELEASE=n`. There is another default setting in the Makefile (which is `y`), but considering the fact that we require a scenario XML to kick off the build and `RELEASE` is mandatory in that XML file, I don't think the one in Makefile ever takes effect. I'll update the doc accordingly.

I think `RELEASE=y|n` is our preference now while we have kept `RELEASE=0|1` for backward compatibility. I'll update the doc to use our latest preference as well.

junjiemao1: In this PR I'll only update the occurences of the items above in `building-from-source.rst`. There are also other places using `RELEASE=0|1` but they are all tutorials using a specific version of ACRN.

doc/getting-started/building-from-source.rst
junjiemao1: I'll remove the `RELEASE=0` setting for simplicity. The debug/release setting does not make any difference here as we want to introduce the steps to customize configurations.

doc/tutorials/acrn_configuration_tool.rst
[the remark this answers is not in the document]
junjiemao1: Will change that to `scenario-neutral`. BTW, the term `scenario-relevant` is kept untouched.

doc/tutorials/acrn_configuration_tool.rst
```diff
@@ -4,59 +4,133 @@ ACRN Configuration Tool
...
+   - VM attributes, such as VM names
+   - Maximum number of VMs supported
+   - Resources allocated to each VM, such as number of vCPUs, amount of guest
+     memory and pass-through devices.
+   - Guest OS settings, such as boot protocol and guest kernel parameters
+
+For pre-launched VMs, the VM attributes and resources are exactly the amount of
+resource allocated to them. For post-launched VMs, they define the upper limit
+that the Service VM can allocate to them, except the pass-through devices which
+are under the control of the Service VM.
```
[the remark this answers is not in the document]
junjiemao1: The intention of this sentence is to explain that devices are under the control of the Service VM and can be dynamically passed through to post-launched VMs when the Service VM creates them (`user VM` is not precise here as a `user VM` can be either pre-launched or post-launched). Also I also notice some other kinds of configurations after a closer look to the predefined scenarios. I'll update this section to make it clearer and more precise.

doc/tutorials/acrn_configuration_tool.rst
```diff
@@ -4,59 +4,133 @@ ACRN Configuration Tool
...
+#. Get the board info.
+
+   a. Set up a native Linux environment on the target board.
+   #. Copy the ``target`` folder into the target file system and then run the
+      ``sudo python3 board_parser.py $(BOARD)`` command.
+   #. A $(BOARD).xml that includes all needed hardware-specific information
+      is generated in the ``./out/`` folder. Here, ``$(BOARD)`` is the
+      specified board name.
+
+      | **Native Linux requirement:**
```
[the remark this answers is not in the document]
junjiemao1: I'll move these requirements to step a and call out that these requirements shall be met before the following steps are conducted.

doc/tutorials/acrn_configuration_tool.rst
```diff
@@ -4,59 +4,133 @@ ACRN Configuration Tool
...
+   #. Run the ``misc/acrn-config/config_app/app.py`` tool on the host
+      machine and import the $(BOARD).xml. Select your working scenario under
+      **Scenario Setting** and input the desired scenario settings. The tool
+      will do a sanity check on the input based on the $(BOARD).xml. The
+      customized settings can be exported to your own $(SCENARIO).xml.
+   #. In the configuration tool UI, input the launch script parameters
+      for the post-launched User VM under **Launch Setting**. The tool will
+      sanity check the input based on both the $(BOARD).xml and
+      $(SCENARIO).xml and then export settings to your $(LAUNCH).xml.
+   #. The user defined XMLs can be imported by acrn-config for modification.
```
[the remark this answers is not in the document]
junjiemao1: I just moved these sections around currently. I think it is referring to the configuration app (which is not referred to consistently throughout the doc yet). I'll update these sections in the next version of this PR.

doc/tutorials/acrn_configuration_tool.rst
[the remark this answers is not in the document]
gvancuts: Ah, now I understand better what you intended to say. Thanks for the clarification.

doc/getting-started/building-from-source.rst
gvancuts: > I think `RELEASE=y|n` is our preference now while we have kept `RELEASE=0|1` for backward compatibility. I'll update the doc to use our latest preference as well. The logic seems to work as expected here. I think it's a good idea to indicate [0|1] is our preference nowadays.

There is however a problem with what gets printed out at the end of the build (and recorded in the `build/hypervisor/cfg.log` file). It **nearly always** reports a `BUILD type = RELEASE` **except** when using `make RELEASE=0`. The problem is in [this line](https://github.com/projectacrn/acrn-hypervisor/blob/master/Makefile#L102) and the value of `RELEASE` is not propagated correctly.

Here is a little experiment I ran today
* Using default values:
```
make
...
ACRN Configuration Summary:
BOARD = nuc7i7dnb        SCENARIO = industry
BUILD type = RELEASE
VM configuration is based on:
        Source code at:                 /home/gvancuts/work/acrn-hypervisor/build/hypervisor/configs

$ grep RELEASE build/ -r
build/hypervisor/cfg.log:BUILD type = RELEASE
build/hypervisor/configs/config.mk:CONFIG_RELEASE=n
build/hypervisor/.scenario.xml:        <RELEASE>n</RELEASE>
```
* Setting `RELEASE` to `y` and `1`:
```
--------------
make RELEASE=y
...
ACRN Configuration Summary:
BOARD = nuc7i7dnb        SCENARIO = industry
BUILD type = RELEASE
VM configuration is based on:
        Source code at:                 /home/gvancuts/work/acrn-hypervisor/build/hypervisor/configs

$ grep RELEASE build/ -r
build/hypervisor/cfg.log:BUILD type = RELEASE
build/hypervisor/configs/config.mk:CONFIG_RELEASE=y
build/hypervisor/configs/scenarios/industry/nuc7i7dnb.config:CONFIG_RELEASE=y
build/hypervisor/include/config.h:#define CONFIG_RELEASE 1
build/hypervisor/.scenario.xml:        <RELEASE>y</RELEASE>

--------------
make RELEASE=1
...
ACRN Configuration Summary:
BOARD = nuc7i7dnb        SCENARIO = industry
BUILD type = RELEASE
VM configuration is based on:
        Source code at:                 /home/gvancuts/work/acrn-hypervisor/build/hypervisor/configs

$ grep RELEASE build/ -r
build/hypervisor/cfg.log:BUILD type = RELEASE
build/hypervisor/configs/config.mk:CONFIG_RELEASE=y
build/hypervisor/configs/scenarios/industry/nuc7i7dnb.config:CONFIG_RELEASE=y
build/hypervisor/include/config.h:#define CONFIG_RELEASE 1
build/hypervisor/.scenario.xml:        <RELEASE>y</RELEASE>
```
* Setting `RELEASE` to `n` and `0`:
```
--------------
make RELEASE=n
...
ACRN Configuration Summary:
BOARD = nuc7i7dnb        SCENARIO = industry
BUILD type = RELEASE
VM configuration is based on:
        Source code at:                 /home/gvancuts/work/acrn-hypervisor/build/hypervisor/configs

$ grep RELEASE build/ -r
build/hypervisor/cfg.log:BUILD type = RELEASE
build/hypervisor/configs/config.mk:CONFIG_RELEASE=n
build/hypervisor/.scenario.xml:        <RELEASE>n</RELEASE>

--------------
make RELEASE=0
...
ACRN Configuration Summary:
BOARD = nuc7i7dnb        SCENARIO = industry
BUILD type = DEBUG
VM configuration is based on:
        Source code at:                 /home/gvancuts/work/acrn-hypervisor/build/hypervisor/configs

$ grep RELEASE build/ -r
build/hypervisor/configs/config.mk:CONFIG_RELEASE=n
build/hypervisor/.scenario.xml:        <RELEASE>n</RELEASE>

```

doc/getting-started/building-from-source.rst
gvancuts: I believe `all` is also the target by default and I would therefore recommend that we leave it out and simply use `make BOARD=xxx SCENARIO=xxx` throughout the documents.

We could indicate a particular component (e.g. `hypervisor`, `devicemodel`, `doc`) if and where we'd want them to build just a particular component. **Although** I am actually wondering if the `RELEASE` value would be properly handled if one wants to build the `devicemodel` or `tools` only for example (as the logic for assigning the value to `RELEASE` lies in `hypervisor/scripts/makefile/`

doc/getting-started/building-from-source.rst
gvancuts: > > I think `RELEASE=y|n` is our preference now while we have kept `RELEASE=0|1` for backward compatibility. I'll update the doc to use our latest preference as well. The logic seems to work as expected here.
> 
> I think it's a good idea to indicate [n|y] is our preference nowadays.

On second thought, I don't think this is quite ready yet - see #5772 . Unless you think we can fix this on time for v2.4, my recommendation is that we stick to using `[0|1]`. And we might even want to include a note to explain that the `RELEASE` value from the scenario XML file is not properly taken into account for all components...

gvancuts (commented):
Just noticed that there is also a note that states we use `menuconfig`, this is in the "Install Build Tools and Dependencies" section. Can you remove it?

I would suggest that we also remove the one about the min version of the compiler and `binutils`. I expect all distros use recent enough versions of those these days.

doc/getting-started/building-from-source.rst
junjiemao1: I'll remove the unnecessary `all` throughout the guide.

I think the inconsistency between the top-level Makefile and the component-wise Makefiles has already made it subtle. I'll try creating a PR to fix this first.

junjiemao1:
> Just noticed that there is also a note that states we use `menuconfig`, this is in the "Install Build Tools and Dependencies" section. Can you remove it?

Sure. Will do in the next version.

> 
> I would suggest that we also remove the one about the min version of the compiler and `binutils`. I expect all distros use recent enough versions of those these days.

How about adding a constraint on the version of Ubuntu instead, i.e. `Ubuntu development system (18.04 or later)`?

Also I'll refine the first paragraph of `Install Build Tools and Dependencies` section as we do not introduce steps to build ACRN on different distributions now.

doc/getting-started/building-from-source.rst
gvancuts: Thanks @junjiemao1 👍

gvancuts:
> > Just noticed that there is also a note that states we use `menuconfig`, this is in the "Install Build Tools and Dependencies" section. Can you remove it?
> 
> Sure. Will do in the next version.

Thanks 👍 

> 
> > I would suggest that we also remove the one about the min version of the compiler and `binutils`. I expect all distros use recent enough versions of those these days.
> 
> How about adding a constraint on the version of Ubuntu instead, i.e. `Ubuntu development system (18.04 or newer)`?

Yes, that works for me 👍 

> 
> Also I'll refine the first paragraph of `Install Build Tools and Dependencies` section as we do not introduce steps to build ACRN on different distributions now.

Yes, good catch :-)

dbkinder:
@junjiemao1 I pushed a fix for the CI doc build heading underline error

doc/getting-started/building-from-source.rst
junjiemao1: With #5791 merged, I'll update the example build commands here to use ``RELEASE=y``.

doc/tutorials/vuart_configuration.rst
junjiemao1: In the third patch of this PR, I've updated the instructions in ``vuart_configuration.rst`` and ``enable_ivshmem.rst`` a bit so that they no longer instruct users to modify predefined scenario XMLs.

junjiemao1:
> @junjiemao1 I pushed a fix for the CI doc build heading underline error

Thanks, David! I'm editing the doc offline as well and will combine the fix in the next version.

HuiHuangShi:

junjiemao1:
The third version of the changes have been submitted with the following main changes.

     * RELEASE now defaults to n.
     * Cleanup the unnecessary target `all` in the `make` commands
     * Remove menuconfig related stuff.
     * Refine the introductory paragraph as only steps on Ubuntu is introduced
     * Also introduce the targets introduced by PR #5791
     * Refactor the sections to introduce the configuration concepts, workflow and details in order.
     * Align names of the components for ACRN configuration.
     * Always use "scenario/launch configuration" rather than "scenario/launch setting".
     * discourage modifications to predefined scenario XMLs

doc/tutorials/acrn_configuration_tool.rst
```diff
@@ -366,38 +344,37 @@ Instructions
...
-   ``misc/vm_configs/xmls/config-xmls/[board]/``.
-   We can edit the scenario name when creating or loading a scenario. If the
-   current scenario name is duplicated with an existing scenario setting
-   name, rename the current scenario name or overwrite the existing one
-   after the confirmation message.
+   The default scenario XMLs are located at
+   ``misc/config_tools/data/[board]/``. You can edit the scenario name when
+   creating or loading a scenario. If the current scenario name is duplicated
+   with an existing scenario setting name, rename the current scenario name or
+   overwrite the existing one after the confirmation message.
```
gvancuts, line 371: Where does the tool save this new scenario? Could it be overwriting the files we have in the repo if the user selects 'overwrite'?

doc/tutorials/vuart_configuration.rst
```diff
@@ -334,13 +319,12 @@ The ACRN vUART related XML fields:
...
 
-  make BOARD_FILE=$PWD/misc/acrn-config/xmls/board-xmls/<board>.xml \
-     SCENARIO_FILE=$PWD/misc/acrn-config/xmls/config-xmls/<board>/<scenario>.xml
+  make BOARD=<board> SCENARIO=<scenario>
 
 The configuration tools will test your settings, and check :ref:`vUART
 Rules <index-of-vuart>` for compilation issue. After compiling, you can find
-``./misc/vm_configs/scenarios/<scenario>/<board>/pci_dev.c`` has been
-changed by the configuration tools based on the XML settings, something like:
+the generated sources under ``build/configs/scenarios/<scenario>/pci_dev.c``
```
gvancuts, line 326:
```suggestion
the generated sources under ``build/hypervisor/configs/scenarios/<scenario>/pci_dev.c``
```

doc/introduction/index.rst
```diff
@@ -423,7 +421,8 @@ The Boot process proceeds as follows:
 
 In this boot mode, the boot options of pre-launched VM and service VM are defined
 in the variable of ``bootargs`` of struct ``vm_configs[vm id].os_config``
-in the source code ``misc/vm_configs/$(SCENARIO)/vm_configurations.c`` by default.
+in the source code ``configs/scenario/$(SCENARIO)/vm_configurations.c`` (which
```
gvancuts, line 424:
```suggestion
in the source code ``configs/scenarios/$(SCENARIO)/vm_configurations.c`` (which
```

doc/tutorials/acrn_configuration_tool.rst
```diff
@@ -432,62 +410,50 @@ Instructions
 
    .. note::
       All customized scenario XMLs will be in user-defined groups, which are
-      located in ``misc/vm_configs/xmls/config-xmls/[board]/user_defined/``.
+      located in ``misc/config_tools/data/[board]/user_defined/``.
```
gvancuts, line 413: I think this is something we should change. We shouldn't add files to the source code that's under the git version control. We should always ask the user where to put these. If we want to use a default value, let's use the user's home directory.

We do not need to fix this in this PR obviously :-)

gvancuts (approved):
Just a few minor things but this looks really good!

The only piece that's missing still in my opinion is a section that explains all `hv*config` options available (i.e. `hvdefconfig`, `hvshowconfig`, `hvapplydiffconfig`, etc.). We touch on some of them but I think it's be useful to have a section/paragraph that presents all of them in a bullet list. If this can be added to thir PR, then absolutely great! Else, let's do this in a subsequent PR.

doc/tutorials/acrn_configuration_tool.rst
junjiemao1: Agree. As we have previously mentioned, in v2.5 we plan to tweak the configuration editor a bit so that it can be invoked by Makefile to edit the configuration in the build directory, just like how ``menuconfig`` works in Linux.

doc/introduction/index.rst
junjiemao1: Will update in v4.

doc/tutorials/acrn_configuration_tool.rst
junjiemao1: @szhen11 who can help clarify. Technically we have not made any change on the usage of the configuration editor in v2.4. The workflow may not be ideal now, and we do plan to refine it in v2.5.

doc/tutorials/vuart_configuration.rst
junjiemao1: Will fix in v4.

doc/developer-guides/hld/vuart-virt-hld.rst
```diff
@@ -95,9 +95,9 @@ Usage
 
 -  For console vUART
 
-   To enable the console port for a VM, change the
-   port_base and IRQ in ``misc/vm_configs/scenarios/<scenario
-   name>/vm_configurations.c``. If the IRQ number has been used in your
+   To enable the console port for a VM, change the port_base and IRQ in
+   ``configs/scenarios/<scenario name>/vm_configurations.c`` under the
```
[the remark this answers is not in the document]
junjiemao1: I guess the preference is to use the combinations listed right after this paragraph. Will update a bit in v4.

In fact I don't think it is a good practice to instruct users to modify generated files directly; the preference is to modify the scenario XML. We may need another PR to update such instructions throughout the guides.

junjiemao1:
> Just a few minor things but this looks really good!
> 
> The only piece that's missing still in my opinion is a section that explains all `hv*config` options available (i.e. `hvdefconfig`, `hvshowconfig`, `hvapplydiffconfig`, etc.). We touch on some of them but I think it's be useful to have a section/paragraph that presents all of them in a bullet list. If this can be added to thir PR, then absolutely great! Else, let's do this in a subsequent PR.

I'll introduce these targets in v4 of this PR. The `hvdefconfig` and `hvshowconfig` will be added both to the ``building-from-source.rst`` and ``acrn_configuration_tool.rst`` The ``hvdiffconfig`` and ``hvapplydiffconfig`` will only be introduced in ``acrn_configuration_tool.rst`` because we do not want to encourage users to modify the generated configuration files; these targets are there mainly for backward compatibility in case any user has his/her own patches that cannot be done by modifying the scenario XML.

junjiemao1:
The 4th version has been pushed with the following changes.

 * fix typos in paths
 * explain on the candidate values of ``port_base`` and ``irq`` fields
 * add descriptions to hypervisor configuration targets

dbkinder:
@junjiemao1 @gvancuts Thanks for all the work and reviews on updating these docs!  I'm liking what I'm reading now. I went through and forced pushed some updates for spelling, grammar, and clarity.  It may be time to merge this and give it a good read-through in its new form...

Issue #5788, opened 2021-02-26
Need support of high MMIO addresses


lifeix:

lifeix:
[External_System_ID]https://jira.devtools.intel.com/projects/ACRN/issues/ACRN-6046

NanlinXie:
Closed with no regression issue.

PR #5790, opened 2021-02-26
hv: use dynamic page allocation for pagetable mapping

lifeix:
For FuSa's case, we remove all dynamic memory allocation use in ACRN HV. Instead,
we use static memory allocation or embedded data structure. For pagetable page,
we prefer to use an index (hva for MMU, gpa for EPT) to get a page from a special
page pool. The special page pool should be big enougn for each possible index.
This is not a big problem when we don't support 64 bits MMIO. Without 64 bits MMIO
support, we could control the index not larger than DRAM_SIZE + 4G.

However, if ACRN plan to support 64 bits MMIO in SOS, we could not use the static
memory alocation any more. This is because there's a very huge hole between the
top DRAM address and the bottom 64 bits MMIO address. We could not reserve such
many pages for pagetable mapping as the CPU physical address bits may very large.

This patch will use dynamic page allocation for pagetable mapping. We also need
reserve a big enough page pool at first.

v4:
1. Use large pages (not small pages) to do EPT mapping for MMIO for RTVM which suffers PSC-MCE.
2. Assume we only support platform which maximum PA wodth is less than MAXIMUM_PA_WIDTH (39 now)
and do the check in detect_hardware_support on early boot.
3. Reserve the page pools for HV MMU and guest EPT by MAXIMUM_PA_WIDTH and CONFIG_PLATFORM_RAM_SIZE
4. Refine the comments how much pages we should reserve
5. Detail the commit messages by Eddie's comments.

v3:
1. remove 4K pagetable mapping for HV MMU
2. reserve EPT pagetable pages according the CONFIG_PLATFORM_RAM_SIZE and CONFIG_MAX_PCI_DEV_NUM
for per VM

v2:
1. refine the commit message.
2. wrap alloc_ept_page; rename try_to_free_page to try_to_free_pgtable_page
3. fix may allocate a page out of page pool since the index large than the pool->bitmap_size

Tracked-On: #5788

### DIFF
--- a/hypervisor/Makefile
+++ b/hypervisor/Makefile
@@ -336,7 +336,6 @@ SYS_INIT_C_OBJS := $(patsubst %.c,$(HV_OBJDIR)/%.o,$(SYS_INIT_C_SRCS))
 ifneq ($(CONFIG_RELEASE),y)
 CFLAGS += -DHV_DEBUG -DPROFILING_ON -fno-omit-frame-pointer
 endif
-CFLAGS += -DCONFIG_LAST_LEVEL_EPT_AT_BOOT
 
 MODULES += $(LIB_MOD)
 MODULES += $(BOOT_MOD)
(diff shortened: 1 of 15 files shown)


HuiHuangShi:

dongyaozu (approved):

PR #5797, opened 2021-03-02
doc: start v2.4 release notes draft

dbkinder:

### DIFF
--- a/doc/release_notes/release_notes_2.4.rst
+++ b/doc/release_notes/release_notes_2.4.rst
@@ -1,0 +1,91 @@
+.. _release_notes_2.4:
+
+ACRN v2.4 (Mar 2021) - DRAFT
+############################
+
+We are pleased to announce the release of the Project ACRN
+hypervisor version 2.4.
+
+ACRN is a flexible, lightweight reference hypervisor that is built with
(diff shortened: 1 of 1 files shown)


doc/release_notes/release_notes_2.4.rst
```diff
@@ -0,0 +1,87 @@
...
+
+
+Document Updates
+****************
+
+New and updated reference documents are available, including:
+
+.. rst-class:: rst-columns2
+
+* :ref:`GVT-g_api`
```
gvancuts, line 46: GVT-g really isn't used much and I think these updates were mostly about deleting i915 driver parameters that became obsolete. Can we put those at the bottom of the list, I do not want people to get the impression the GVT-g stuff is up-to-date.

dbkinder: Sure, this list is generated by a script that looks for docs with more than 20 lines changed since the v2.3 release (as computed by ``git diff --stat `git rev-parse v2.3` `git rev-parse master` | grep \.rst``).  After that, we can edit which ones we actually include in the release notes and their order.

junjiemao1:
Since we have changed the Makefile and file layout a bit, shall we add something like an "upgrade guide" to help users moving from prior versions to v2.4?

Issue #5644, opened 2021-01-14
Refinement of ACRN offline supporting tools


NanlinXie:
Configuration Tool Arch need to be optimized.

NanlinXie:
[External_System_ID]ACRN-6699

NanlinXie:
Closed with no regression issue.

PR #5817, opened 2021-03-05
doc: update compile-time configuration HLD

junjiemao1:
Starting from v2.4, ACRN configuration uses solely XML files to store
configuration data and customized scripts to manipulate
configurations. This patch updates the HLD of compile-time configuration to
reflect this properly.

As the refinement to the configuration toolset itself is still ongoing,
this patch only adds brief introduction to the key generated files involved
in ACRN configuration. More details will be added after the refinement
completes.

Tracked-On: #5644


doc/developer-guides/hld/hv-config.rst
```diff
@@ -3,54 +3,44 @@
...
-
-   c. **menuconfig** presents a terminal UI to navigate and modify the
-      configurations in an interactive manner.
-
-2. The target oldconfig is also used to create a .config if a .config
-   file does not exist when building the source directly.
-
-3. The other two files for makefiles and C sources are regenerated after
-   .config changes.
+Whenever ``.board.xml`` or ``.scenario.xml`` is modified, the hypervisor will be
```
gvancuts, line 41: Is this the way to modify an existing build?

What happens if someone modifies other files such as `configs/config.mk` or others?

gvancuts (approved):
LGTM!

fitchbe (approved):
Pushed a few copy edits.

doc/developer-guides/hld/hv-config.rst
junjiemao1: Yes, modifying `.board.xml` and `.scenario.xml` is the only supported way to modify the configurations of an existing build. Changing other generated files (e.g. `config.h`, `config.mk` or else) will lead to unpredictable behavior, possibly due to inconsistent data in these files.

In v2.5 the preferred way will be using the configuration editor (via the `menuconfig` target) We still need to add a few command line options to the editor to enable this.

PR #5873, opened 2021-03-25
doc: add summary of config changes and upgrading guides

junjiemao1:
This patch adds a summary of the significant changes to ACRN configuration and build in v2.4 and gives some instructions on how to upgrade from a previous release.

### DIFF
--- a/doc/release_notes/release_notes_2.4.rst
+++ b/doc/release_notes/release_notes_2.4.rst
@@ -3,12 +3,12 @@
 ACRN v2.4 (Mar 2021) - DRAFT
 ############################
 
-We are pleased to announce the release of the Project ACRN
-hypervisor version 2.4.
+We are pleased to announce the release of the Project ACRN hypervisor
+version 2.4.
 
 ACRN is a flexible, lightweight reference hypervisor that is built with
 real-time and safety-criticality in mind. It is optimized to streamline
-embedded development through an open source platform. Check out the
+embedded development through an open-source platform. See the
 :ref:`introduction` introduction for more information.  All project ACRN
 source code is maintained in the
 https://github.com/projectacrn/acrn-hypervisor repository and includes
@@ -16,7 +16,7 @@ folders for the ACRN hypervisor, the ACRN device model, tools, and
 documentation. You can either download this source code as a zip or
 tar.gz file (see the `ACRN v2.4 GitHub release page
 <https://github.com/projectacrn/acrn-hypervisor/releases/tag/v2.4>`_) or
-use Git clone and checkout commands::
+use the Git ``clone`` and ``checkout`` commands::
 
    git clone https://github.com/projectacrn/acrn-hypervisor
    cd acrn-hypervisor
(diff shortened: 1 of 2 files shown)


fitchbe (approved):
Pushed copy edits.

doc/release_notes/release_notes_2.4.rst
```diff
@@ -38,9 +38,124 @@ What's New in v2.4
...
+****************************************
+
+Users are highly recommended to follow the instructions below in order to
+upgrade to v2.4 from previous ACRN releases.
+
+Additional Dependencies
+=======================
+
+Python 3.6 or above is required to build ACRN v2.4. If you are using an old
+version of Ubuntu (e.g. 16.04), you can use PPA (e.g. `New Python Versions
```
dbkinder, line 73: I don't think we should recommend using unofficial releases of Python via deadsnakes.  Instead, let's ask them to check their python version (using ``python --version``) and if it's not version 3.6 or higher, use ``apt`` to check for an update.

dbkinder (changes requested):

doc/release_notes/release_notes_2.4.rst
junjiemao1: Unfortunately Ubuntu 16.04 only ships Python 3.5.2 in the official package repo.

junjiemao1:
Added another section `New Configuration Options` which lists the added options in v2.4 and provides a script to upgrade a v2.3-compliant scenario XML.

doc/release_notes/release_notes_2.4.rst
```diff
@@ -38,9 +38,160 @@ What's New in v2.4
...
+<https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa>`_ from deadsnakes
+team) for recent Python versions.
+
+In addition, the following tools and packages are needed to build ACRN:
```
gvancuts, line 77: Can we make it clear that this is not the complete list but rather new tools that need to be installed compared to v2.3

I also ran into an issue whereby I had to install `libxml2-utils` which is currently not mentioned in our documentation (see PR #5862). This should also be mentioned here.

doc/release_notes/release_notes_2.4.rst
```diff
@@ -38,9 +38,160 @@ What's New in v2.4
...
+     $ sudo pip3 install lxml xmlschema
+
+Configuration File Format
+=========================
+
+Starting from v2.4, Kconfig is no longer used, and the contents of scenario
+XMLs have been simplified. You need to upgrade your own Kconfig-format files
+or scenario XMLs if you maintain any.
+
+For Kconfig-format file, you must translate your configurations to a scenario
```
gvancuts, line 91:
```suggestion
For Kconfig-format file, you must translate your configuration to a scenario
```
?

gvancuts (approved):
Just some minor comments but otherwise LGTM!

dbkinder (changes requested):

doc/release_notes/release_notes_2.4.rst
junjiemao1: Thanks for the detailed instructions. I've updated the doc accordingly.

doc/release_notes/release_notes_2.4.rst
junjiemao1: Typo fixed.

doc/release_notes/release_notes_2.4.rst
junjiemao1: Thanks for pointing this out. I've added a note to call out that this is not the complete list. Also added libxml2-utils here.

gvancuts (approved):

dbkinder (approved):
LGTM, thanks!

PR #5881, opened 2021-03-26
DM: gvt: Identical mapping for GPU DSM refine to support EHL/TGL

psun3x:
Windows graphic driver obtains DSM address from in-BAR mmio register
which has passthroughed. Not like the other platforms obtained from
pci configure space register which has virtualized. GPU GuC must use
WOPCM in DSM, besides, Windows OS wants to manage DSM also. These two
reason force acrn has to keep identical mapping to avoid trap mmio
BAR to do the emulation.

To keep simple, this patch reserve a big address range hole for DSM
& opregion of gpu in vE820 table, this will cause memory waste here.
In the near future, we need refine the entire vE820 logic as it is
hard to maintained due to many reserved regions have introduced in
recently.

Tracked-On: #5880
Signed-off-by: Peng Sun <peng.p.sun@intel.com>

### DIFF
--- a/devicemodel/core/sw_load_common.c
+++ b/devicemodel/core/sw_load_common.c
@@ -39,9 +39,12 @@
 int with_bootargs;
 static char bootargs[BOOT_ARG_LEN];
 
+extern uint32_t gpu_opregion_gpa;
+
 /*
  * Default e820 mem map:
  *
+ * there is reserved memory hole for device memory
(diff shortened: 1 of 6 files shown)


HuiHuangShi:

ywan170:
Please hold this patch. We found several issues that need to be fixed.

wenlingz:
> Please hold this patch. We found several issues that need to be fixed.

OK

PR #5891, opened 2021-03-30
doc: update 'enable secure boot in windows'

yonghuah:
- use one command to generate x509 cert file,
   remove the intermediate file.

 - remove the "Keycontainer" field in INF file,
   which is not mandatory.


gvancuts (approved):
Thanks!

dbkinder (approved):
LGTM

PR #5926, opened 2021-04-12
doc: update doc build instructions

dbkinder:
We've validated doc build tool versions, so let's make sure those are
the versions the instructions say to install.  The version of doxygen
you get when you use ``sudo apt install doxygen`` may get a newer
version that may still work so let's tell them that.

Also, we no longer use kconfig files in the document build process, so
remove mentioning that in the build documentation.

### DIFF
--- a/doc/tutorials/docbuild.rst
+++ b/doc/tutorials/docbuild.rst
@@ -131,11 +131,12 @@ and these other tools:
 * sphinx                    version: 3.2.1
 * docutils                  version: 0.16
 * sphinx-rtd-theme          version: 0.5.0
-* kconfiglib                version: 14.1.0
 * sphinx-tabs               version: 1.3.0
 * doxygen                   version: 1.8.13
 
-Depending on your Linux version, install the needed tools.
+Depending on your Linux version, install the needed tools. You may get a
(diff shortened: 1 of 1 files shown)


doc/tutorials/docbuild.rst
```diff
@@ -140,7 +139,7 @@ For Ubuntu use:
 
   .. code-block:: bash
 
-     sudo apt install doxygen python3-pip \
+     sudo apt install doxygen=1.8.13-10 python3-pip \
```
gvancuts, line 142: The problem with this approach is that it only works on Ubuntu 18.04. That version is not available on Ubuntu 20.04 for example. That may be the case too for other distros (e.g. Fedora, OpenSUSE Tumbleweed).

I think we need to be more explicit that it is indeed the version that is known to work, *but* that this exact version is only available on Ubuntu 18.04 and we recommend that people try first with the `doxygen` version that comes with their distro, even if the version does not exactly match.

dbkinder: OK, sigh.  Thanks for catching that.

dbkinder:
@gvancuts ... that better?

gvancuts (approved):
Thanks!

PR #5949, opened 2021-04-21
doc: tweaks for latexpdf build

dbkinder:
Update missing captions on figures to remove remaining broken references
during latexpdf building.  Also, require doing a "make html" before
doing a "make latexpdf" to build all the artifacts needed for running
the latexpdf build.  (We might change that later if needed.)

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -90,5 +90,5 @@ endif
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(OPTS) is meant as a shortcut for $(SPHINXOPTS).
-%: Makefile doxy
+%: Makefile
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(OPTS)
--- a/doc/developer-guides/hld/hv-partitionmode.rst
+++ b/doc/developer-guides/hld/hv-partitionmode.rst
@@ -78,6 +78,8 @@ events in chronological order.
    :align: center
    :name: partBSPsetup
 
+   Event Order for Processor Set Up
+
 Memory
 ======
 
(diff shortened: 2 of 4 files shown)


dbkinder:
@NanlinXie @wenlingz looks like this doc-only PR is stuck with a YOCTO BUILD check...

wenlingz:
> @NanlinXie @wenlingz looks like this doc-only PR is stuck with a YOCTO BUILD check...

Fixed, exceed token led to

doc/Makefile
```diff
@@ -90,5 +90,5 @@ endif
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(OPTS) is meant as a shortcut for $(SPHINXOPTS).
-%: Makefile doxy
+%: Makefile
```
gvancuts, line 93: I do not understand this change, can you explain?

Also, since one needs to run `make html` first, wouldn't it be better to define a new phony target and make it depend on the html target? I.e.:
```
.PHONY: latexpdf
latexpdf: html
```
```

gvancuts: How about also adding `xindy` to ``doc/tutorials/docbuild.rst` directly?

dbkinder: As for removing the doxy dependency here, doxy wasn't sufficient for building with the unmentioned builders, so I was being a bit lazy and just say build html first and then build latexpdf.  (I'm still figuring out how to get rid of the PDF-building errors and was running it multiple times, so no need to repeat the html step every time I wanted to rerun the latexpdf step.  Once we get the latexpdf more cleanly working we can consolidate this into one step that creates the input needed to do PDF generation that's done now for the html generation.

dbkinder:
I'll merge this for now and follow up with changes to the Makefile and docbuild documentation for PDF generation in another PR.

doc/Makefile
gvancuts: OK, makes sense. Thanks!

PR #5945, opened 2021-04-20
doc/conf.py: enable formatting docs in a PDF file

junjiemao1:
This patch tweaks the settings in doc/conf.py to allow formatting the
documentation to a PDF file by Sphinx. The changes include:

 - Use `xelatex` rather than the default `pdflatex` as the LaTeX engine, as
   `pdflatex` is not that good at formatting non-ascii characters out of
   the box.
 - Use DejaVu fonts (which are available in common Linux distributions) in
   the generated PDF.
 - Restrict the depths of the table of contents to 3.

Signed-off-by: Junjie Mao <junjie.mao@intel.com>

PR #5967, opened 2021-04-23
doc: clean up PDF generation for ACRN docs

dbkinder:
PRs #5945 and #5949 introduced fixes to the doc building process to
support PDF generation of the documentation set.  This PR refines the
doc build process, cleaning up the Makefile, adding display of tool
version information, and updates the doc building documentation to
include additional dependencies needed for building the PDF and
instructions for how to build the PDF.  The latexpdf make target is
provided to just run the latex and PDF producing process that depends on
the HTML artifacts from a make html run.  A new make pdf target is
provided that combines the two steps into one.

A new know-issues pattern file is added that verifies the expected
output from the latexpdf process is returned, as it can't be completely
eliminated without losing potential error messages that need to be
resolved.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -12,7 +12,8 @@ SPHINXOPTS    ?= -q
 SPHINXBUILD   = sphinx-build
 SPHINXPROJ    = "Project ACRN"
 BUILDDIR      ?= _build
-SOURCEDIR     = $(BUILDDIR)/rst/
+SOURCEDIR     = $(BUILDDIR)/rst
+LATEXMKOPTS   = -silent
 
 # document publication assumes the folder structure is setup
 # with the acrn-hypervisor and projectacrn.github.io repos as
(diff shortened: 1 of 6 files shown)


doc/Makefile
```diff
@@ -37,32 +38,45 @@ help:
...
 
+# Used to pull the acrn kernel source (for API docs)
+
 pullsource:
 	$(Q)scripts/pullsource.sh
 
-html: copy-to-sourcedir doxy content
+html: content doxy
+	@echo making HTML content
+	@./scripts/show-versions.py
```
gvancuts, line 66: Shouldn't we use
```suggestion
	$(Q)./scripts/show-versions.py
```
to be consistent with the rest of the Makefile?

doc/scripts/pullsource.sh
```diff
@@ -2,7 +2,8 @@
 # Copyright (C) 2019 Intel Corporation.
 # SPDX-License-Identifier: BSD-3-Clause
 
-q="--quiet"
+#q="--quiet"
+q=""
```
gvancuts, line 6: I'm assuming you changed and kept this modification on purpose?

doc/Makefile
```diff
@@ -37,32 +38,45 @@ help:
...
 	$(Q)./scripts/filter-doc-log.sh $(BUILDDIR)/doc.log
 
-singlehtml: doxy content
+singlehtml: content doxy
 	-$(Q)$(SPHINXBUILD) -t $(DOC_TAG) -b singlehtml -d $(BUILDDIR)/doctrees $(SOURCEDIR) $(BUILDDIR)/html $(SPHINXOPTS) $(OPTS) >> $(BUILDDIR)/doc.log 2>&1
 	$(Q)./scripts/filter-doc-log.sh $(BUILDDIR)/doc.log
 
+pdf: html
+	@echo now making $(BUILDDIR)/latex/acrn.pdf
+	$(Q)make --silent latexpdf LATEXMKOPTS=$(LATEXMKOPTS) >> $(BUILDDIR)/doc.log 2>&1
```
gvancuts, line 76: Do we also need to make `make` silent here?

gvancuts (approved):
This looks good to me! Just a few minor comments.

Have you also thought about creating a new target for this from the top-level Makefile (e.f. `make doc-pdf` that would run `$(MAKE) -C $(T)/doc pdf BUILDDIR=$(DOC_OUT)`

doc/Makefile
dbkinder: yep to both

doc/scripts/pullsource.sh
dbkinder: yes.  This script can take a l-o-n-g time on WSL, so it's best it not run quietely.

doc/Makefile
dbkinder: It stops it from writing out the entering directory messages, so I thought so, yes.

Issue #5830, opened 2021-03-10
[Modularization] paging/memory part


lifeix:

lifeix:
[External_System_ID] ACRN-6908

PR #5971, opened 2021-04-27
hv: mmu: fix a minor bug

lifeix:
We should only map [low32_max_ram, 4G) MMIO region as UC attribute,
not map [low32_max_ram, low32_max_ram + 4G) region as UC attribute.
Otherwise, the HV will complain [4G, low32_max_ram + 4G) region has
already mapped.

Tracked-On: #5830


wenlingz:

dongyaozu (approved):

Issue #5973, opened 2021-04-27
[mainline] Fail to launch RTVM with --psram


yonghuah:
*Environment*

 HV repo: [https://github.com/projectacrn/acrn-hypervisor.git]

Branch： Master

 kernel:  ACRN  latest master [image|http://sit-image.sh.intel.com/acrn_daily_build/YOCTO-2021ww17.1-1/], SOS kernel has updated to 5.4.106.
  HW: TGL

*Steps*
 # Config to enable PSRAM in TGL scenario xml (misc/config_tools/data/tgl-rvp/industry.xml)                                                                         <PSRAM>
 <PSRAM_ENABLED>n</PSRAM_ENABLED>
 </PSRAM>
 # build hypervisor w: make BOARD=tgl-rvp SCENARIO=industry release=0
 # replace the build on test BOARD and boot HV and SOS
 # launch RTVM with --psram in acrn-dm cmd line

*Expect Result*

launch RTVM with --psram successfully 

*Actual result*

 Launch RTVM with --psram failed：

yonghuah:
[External_System_ID] ACRN-7011

PR #5974, opened 2021-04-27
dm: bugfix in create_and_inject_vrtct

yonghuah:
GPA of software SRAM is available only after
 build_vrtct() function is called and the return
 value of it is valid(Not NULL).

 This patch fix bug in create_and_inject_vrtct()
 function which violate above pre-condition when
 calling get_software_sram_base_gpa().

Tracked-On: #5973

### DIFF
--- a/devicemodel/hw/platform/acpi/acpi.c
+++ b/devicemodel/hw/platform/acpi/acpi.c
@@ -1101,7 +1101,6 @@ int create_and_inject_vrtct(struct vmctx *ctx)
 		.prot = PROT_ALL
 	};
 
-	memmap.gpa = get_software_sram_base_gpa();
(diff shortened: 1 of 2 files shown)


wenlingz:

ywan170 (approved):
I am curious how it is work before?..

yonghuah:
@ywan170 
With below commit, SW SRAM does not work for post-RTVM,  mainline or v2.4 branch:

From 2a96c567b80b21869a16011a3f5bce46483f8ee9 Mon Sep 17 00:00:00 2001
From: ppsun <peng.p.sun@intel.com>
Date: Mon, 29 Mar 2021 08:53:29 +0000
Subject: [PATCH 1/2] DM: gvt: Identical mapping for GPU DSM refine to support

ywan170:
@yonghuah get it. Thanks for clarify

PR #5997, opened 2021-05-10
Nested virtualization initial commit: part 1

ZideChen0:
The first part of nested virtualization.


dongyaozu:
HV side is fine. Can @terryzouhao review the config tool part?

wenlingz:
OK to verify

misc/hv_prebuild/vm_cfg_checks.c
```diff
@@ -140,9 +140,21 @@ bool sanitize_vm_config(void)
 			}
 			break;
 		case SOS_VM:
-			if ((vm_config->severity != (uint8_t)SEVERITY_SOS) || ((vm_config->guest_flags & GUEST_FLAG_LAPIC_PASSTHROUGH) != 0U)) {
+			if (vm_config->severity != (uint8_t)SEVERITY_SOS) {
```
terryzouhao, line 143: @ZideChen0 , yes we are switching to XML-based config tool, and feature/FLAG check could  be handled by XML schema, just like you did for SCHEDULER:  <xs:assert test="not (NVMX_ENABLED = 'y' and not (SCHEDULER = 'SCHED_NOOP'))"/> So could you move this sanity check into schema from vm_cfg_checks.c ?

misc/config_tools/data/tgl-rvp/hybrid.xml
```diff
@@ -20,6 +20,7 @@
             <RDT_ENABLED>n</RDT_ENABLED>
             <CDP_ENABLED>n</CDP_ENABLED>
         </RDT>
+        <NVMX_ENABLED>n</NVMX_ENABLED>
```
terryzouhao, line 23: nested-virtualization is target enabled on tgl-hybird scenario first right, so should we set NVMX_enabled as 'y' by default?

terryzouhao (commented):
Others are fine, only 2 comments for config tool:  prefer to do all sanity check in XML schema rather than in .c file, and do we need set 'NVMX_ENABLED' as 'y' for tgl-hybrid scenario.

misc/config_tools/data/tgl-rvp/hybrid.xml
dongyaozu: Nested VMX feature is still an POC feature. So, it won't be turn on by default.

ZideChen0:
Updated patch "misc: sanity check VM config for nested virtualization" to move the sanity check from vm_cfg_checks.c to schema.

misc/hv_prebuild/vm_cfg_checks.c
terryzouhao: Refined sanity check to XML schema, looks good to me now, thanks.

terryzouhao (approved):

PR #6004, opened 2021-05-12
Paging/Memory Modularization - Cache/TLB Part

lifeix:
Wrap common APIs for Cache/TLB Part

For Cache, ACRN provides three APIs:
    - flush_invalidate_all_cache
    - flush_cacheline
    - flush_cache_range

For TLB, ACRN provides two APIs:
    - flush_tlb
    - flush_tlb_range

This patch also remove ppt_set/clear_(attribute) to set_paging_(attribute)

Tracked-On: #5830

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -31,6 +31,7 @@
 #include <vpci.h>
 #include <ivshmem.h>
 #include <asm/rtcm.h>
+#include <reloc.h>
 
 #define CPU_UP_TIMEOUT		100U /* millisecond */
 #define CPU_DOWN_TIMEOUT	100U /* millisecond */
@@ -445,8 +446,8 @@ void cpu_dead(void)
 	if (bitmap_test(pcpu_id, &pcpu_active_bitmap)) {
 		/* clean up native stuff */
 		vmx_off();
-		/* TODO: a cpu dead can't effect the RTVM which use Software SRAM */
-		cache_flush_invalidate_all();
+
+		flush_cache_range((void *)get_hv_image_base(), CONFIG_HV_RAM_SIZE);
 
 		/* Set state to show CPU is dead */
 		pcpu_set_current_state(pcpu_id, PCPU_STATE_DEAD);
--- a/hypervisor/arch/x86/guest/vmexit.c
+++ b/hypervisor/arch/x86/guest/vmexit.c
@@ -409,7 +409,7 @@ static int32_t wbinvd_vmexit_handler(struct acrn_vcpu *vcpu)
 
 	/* GUEST_FLAG_RT has not set in post-launched RTVM before it has been created */
 	if ((!is_software_sram_enabled()) && (!has_rt_vm())) {
-		cache_flush_invalidate_all();
+		flush_invalidate_all_cache();
 	} else {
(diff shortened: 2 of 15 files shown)


hypervisor/arch/x86/cpu.c
```diff
@@ -446,7 +446,7 @@ void cpu_dead(void)
 		/* clean up native stuff */
 		vmx_off();
 		/* TODO: a cpu dead can't effect the RTVM which use Software SRAM */
```
dongyaozu, line 448: This TODO should be completed now.

hypervisor/arch/x86/guest/vmexit.c
```diff
@@ -409,7 +409,7 @@ static int32_t wbinvd_vmexit_handler(struct acrn_vcpu *vcpu)
 
 	/* GUEST_FLAG_RT has not set in post-launched RTVM before it has been created */
 	if ((!is_software_sram_enabled()) && (!has_rt_vm())) {
-		cache_flush_invalidate_all();
+		wbinvd();
 	} else {
 		if (is_rt_vm(vcpu->vm)) {
 			walk_ept_table(vcpu->vm, ept_flush_leaf_page);
```
dongyaozu, line 415: This function (to flush all) for both SSRAM case or not, can be a seperate wrap API.

hypervisor/arch/x86/cpu.c
```diff
@@ -446,7 +446,7 @@ void cpu_dead(void)
...
-		cache_flush_invalidate_all();
+		wbinvd();
```
dongyaozu, line 449: why not "flush_invalidate_all_cache"?

dongyaozu: BTW, given that flush_tlb actually means flush and invalidate, I think we can use flush_cache_all here too.

lifeix: > why not "flush_invalidate_all_cache"?

IMHO, we should use wbinvd for x86 special code and use flush_invalidate_all_cache for non-arch related code

lifeix: > BTW, given that flush_tlb actually means flush and invalidate, I think we can use flush_cache_all here too.

Do you mean flush_cache_range(0, all address space) ?

hypervisor/arch/x86/guest/vmexit.c
lifeix: I think this function would have large logic change when we support multi-RTVM. In that time, we could hide SSRAM from vmexit module.

hypervisor/arch/x86/cpu.c
lifeix: OK.

hypervisor/arch/x86/cpu.c
dongyaozu: Yes

dongyaozu: x86 code call general code should be fine too.

hypervisor/arch/x86/guest/vmexit.c
dongyaozu: OK. But call flush_cache_all is better IMO

dongyaozu (approved):

PR #6008, opened 2021-05-13
Refinement to the board inspector

junjiemao1:
This patch series introduce a refined board inspector aiming at the
following objectives:

  1. Encode the characteristics of the target platform more systematically,
  2. Make it easier for the other components of the configuration toolset
     to refer to and manipulate board data.

To fulfill the objectives above, this series collects board information by
parsing well-defined interfaces and hierarchically places the information
in a tree-like structure. The interfaces from which the inspector collect
board information includes:

  - CPUID leaves
  - E820 memory maps
  - ACPI tables, including DSDT/SSDT
  - PCIe configuration space
  - SMBIOS tables

With the refined layout of board XML, detection of assumption violations
can be managed as XSD assertions and applied to the board and scenario XMLs
at compile time. This patch series also adds a framework and some examples
how such assumptions can be written.

The followings are out of the scope of this series.

  - Add a complete list of assumptions we currently have.
  - Design a better way to present assumption violations to users.


wenlingz:

terryzouhao (approved):

Issue #5923, opened 2021-04-09
Nested Virtualization


NanlinXie:
The hypervisor shall support to run a hypervisor inside a partitioned VM, and such feature shall be configurable. The hypervisor shall support to run a hypervisor inside a partitioned VM, and such feature shall be configurable.

NanlinXie:
[External_System_ID] ACRN-6965

gvancuts:
@NanlinXie , this one will definitely require some documentation ;-)

NanlinXie:
Doc Impact Analysis: 

- [x] New feature -> new HLD and usage guide
- [ ] [HLDs](https://projectacrn.github.io/2.3/developer-guides/hld/index.html)
- [ ] [Configuration data](https://projectacrn.github.io/2.3/tutorials/acrn_configuration_tool.html)
- [ ] [Kernel module parameters](https://projectacrn.github.io/2.3/user-guides/kernel-parameters.html)
- [ ] [Device Model (acrn-dm) parameters](https://projectacrn.github.io/2.3/user-guides/acrn-dm-parameters.html)
- [ ] [Hypervisor shell commands](https://projectacrn.github.io/2.3/user-guides/acrn-shell.html)
- [ ] [ACRN tools (debug and/or service)](https://projectacrn.github.io/2.3/develop.html#configuration-and-tools)
- [ ] [Build system](https://projectacrn.github.io/latest/getting-started/building-from-source.html)
- [ ] [Other documents (e.g. tutorials)](https://projectacrn.github.io/2.3/develop.html#enable-acrn-features)

wenlingz:
[External_System_ID] ACRN-6965

NanlinXie:
doc PR: https://github.com/projectacrn/acrn-hypervisor/pull/6083
@shenfang2019 , please help to verify and review.

PR #6023, opened 2021-05-17
config-tools: guest_flag must be assigned with a valid value

ZideChen0:
commit 873ed752d ("misc: sanity check VM config for nested virtualization")
requires that the guest_flag tag can't be empty, or it fails to build.

This patch changes all instances of "<guest_flag></guest_flag>"
to "<guest_flag>0</guest_flag>".

Tracked-On: #5923


szhen11:
Changing the guest_flag values in xml is just the workaround solution, right? We need a real fix to let the value None or "" assigned to guest_flag in the schema sanity check.

wenlingz:
try verify ldra

wenlingz:

NanlinXie (approved):

Issue #6020, opened 2021-05-17
support RTCTv2


yonghuah:
The hypervisor and device module shall be able to support both RTCTv1 and RTCTv2 for pre or post-launched RTVM.

yonghuah:
[External_System_ID] ACRN-7144

NanlinXie:
Closed since patch merged without regression.

PR #6049, opened 2021-05-21
HV: cache: Fix page fault by flushing cache for VM trusty RAM in HV

taoyuhong:
The accrss right of HV RAM can be changed to PAGE_USER (eg. trusty RAM
of post-launched VM). So before using clflush(or clflushopt) to flush
HV RAM cache, must allow explicit supervisor-mode data accesses to
user-mode pages. Otherwise, it may trigger page fault.

Tracked-On: #6020

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -447,7 +447,9 @@ void cpu_dead(void)
 		/* clean up native stuff */
 		vmx_off();
 
+		stac();
 		flush_cache_range((void *)get_hv_image_base(), CONFIG_HV_RAM_SIZE);
+		clac();
 
 		/* Set state to show CPU is dead */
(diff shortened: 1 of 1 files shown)


hypervisor/arch/x86/cpu.c
```diff
@@ -447,7 +447,9 @@ void cpu_dead(void)
 		/* clean up native stuff */
 		vmx_off();
 
+		clac();
 		flush_cache_range((void *)get_hv_image_base(), CONFIG_HV_RAM_SIZE);
```
lifeix, line 451: This call sequence of clac and stac is wrong.

lifeix (changes requested):

hypervisor/arch/x86/cpu.c
taoyuhong: Fixed

fyin1 (approved):

PR #6052, opened 2021-05-21
vlapic: extend vlapic_x2apic_pt_icr_access to support more destination modes and destination shorthand

lifeix:
Guest would use "All Including Self" or "All Excluding Self" Destination Shorthand
to broadcast IPIs if there're more than one target vCPUs. However, ACRN didn't support
this case if the current vCPU was working in x2APIC mode and ACRN pass through this
vCPU to Guest.

This patch fix this by extending vlapic_x2apic_pt_icr_access to support more
destination modes and destination shorthand.

v2:
vlapic_x2apic_pt_icr_access only works when the guest was in VM_VLAPIC_X2APIC mode.

Tracked-On: #5923


wenlingz:

dongyaozu (approved):

PR #6055, opened 2021-05-21
nested virtualization - part 3

ZideChen0:
The delta between this PR and the v8 patch series:

- in  [patch 2/8], check EFER.LMA and CS.L, and inject #UD if L1 hypervisor is not in long mode
- in [patch 8/8], remove is_long_mode() code in both vmread and vmwrite() handler
- remove VMX_MSR_BITMAP_FULL from vmcs_shadowing_fields[], and set gpa_field_dirty in vmwrite handler if L1 VMWRITE to it
- in [patch 4/8], in the "nested->current_vmcs12_ptr == vmcs12_gpa", add "nested_vmx_result(VMsucceed, 0)" so that VMX_GUEST_RFLAGS is properly set in all possible paths.

Updated some comments and commit messages, trying to correctly differentiate vmcs02, shadow vmcs02, and vmcs12. Also removed some "ordinary VMCS" text, because as per ISDM, the VMCS shadow indicator is the only way to determine a VMCS is an ordinary or shadow VMCS. So if a VMCS whose shadow indicator is set and loaded by VMPTRLD, it's still a shadow VMCS, and we can't call it ordinary.


ZideChen0:
Forgot to mention one more delta. In [patch 7/8], in the vmclear handler, I added "clear_va_vmcs(nested->vmcs02)" after  flush_current_vmcs12(), to complete the vmclear of vmcs12.

wenlingz:
@ZideChen0 please also create PR on release_2.5 branch, thanks

dongyaozu (approved):

Issue #6056, opened 2021-05-24
dm: a minor bug fix of unregister_mem_int


lifeix:
[External_System_ID] ACRN-7104

wenlingz:
[External_System_ID] ACRN-7104

PR #6058, opened 2021-05-24
release_2.5: dm: rb: only free rb_entry when we remove this entry from the rb tree

lifeix:
Only free rb_entry when we remove this entry from the rb tree, otherwise, a
page fault would trigger when next rb itreation would access the freed rb_entry.

Tracked-On: #6056


ywan170:
@fyin1

fyin1 (approved):

PR #6060, opened 2021-05-24
RFC: efi-stub: add acrn efi application for secure boot use cases

tnishiok:
RFC: efi-stub: add acrn efi application for secure boot use cases

ACRN EFI application based on the legacy efi-stub code provides booting method
of HV on UEFI-BIOS without using the Grub Bootloader. It is supposed to be used
for the Secure Boot use cases. By means of that users can boot HV, Service VM
kernel, pre-launched VM kernel and its ACPI table binary packed in the Slim
Bootloader container boot image file format. By default ACRN EFI application
does not get compiled since the final image has dependencies to kernel images.
It is supposed to be compiled by by the meta-acrn layer.

Tracked-On: To be determined
Signed-off-by: Toshiki Nishioka<toshiki.nishioka@intel.com>
Co-developed-by: Yifan Liu <yifan1.liu@intel.com>

This is a PR for Request for Comment. All commits shall be squashed when we remove the RFC tag.


wenlingz:
@tnishiok please add Track-On and signed-off in your commit, thanks

Makefile
```diff
@@ -66,6 +66,7 @@ endif
 O ?= build
 ROOT_OUT := $(shell mkdir -p $(O);cd $(O);pwd)
 HV_OUT := $(ROOT_OUT)/hypervisor
+EFI_OUT := misc/efi-stub
```
tnishiok, line 69: to be removed, no longer required if we don't compile efi-stub by default

misc/efi-stub/Makefile
```diff
@@ -0,0 +1,132 @@
...
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
```
tnishiok, line 31: RFC: Shall we add the "SPDX-License-Identifier" to every efi-stub files as the coding guideline specified?

C-CS-19: Legal entity shall be documented in every file
https://projectacrn.github.io/latest/developer-guides/c_coding_guidelines.html#id134

hypervisor/arch/x86/boot/cpu_primary.S
```diff
@@ -216,7 +216,15 @@ jmpbuf_32:
...
-.code64
+    .org 0x200
+    .code64
+    .global cpu_primary_start_64
+
+cpu_primary_start_64:
+    lea boot_regs(%rip), %rax
+    movl %edi, (%rax)
+    movl %esi, 4(%rax)
+
```
tnishiok, line 227: RFC: At the end I would like to squash all commits into single commit. But still is it better to separate a commit for HV from the one for efi-stub? I mean, we could have two commits at the end.

misc/efi-stub/container.c
```diff
@@ -0,0 +1,488 @@
...
+		}
+	}
+	/* exclude hypervisor and SBL signature files. e.g.)
+	 *    lzh_count = 9 (hv_cmdline, acrn.32.out, vm0_tag, vm0_kernel, vm1_tag, vm1_kernel, vm0_acpi_tag, vm0_acpi, sig)
+	 *    mod_count = 3 (vm0_tag + vm0_kernel, vm1_tag + vm1_kernel, vm0_acpi_tag + vm0_acpi)
+	 */
+	ctr->mod_count = (ctr->lzh_count - 3) / 2;
+	
+	/* allocate single memory region to store all binary files to avoid mmap fragmentation */
+	err = emalloc_reserved_aligned(&(ctr->mod_hpa), ctr->total_modsize,
```
tnishiok, line 229: RFC: Is it redundant? Currently I allocated memory to copy Kernel & ACPI table files from container to. Since all files are embedded into the ACRN EFI image which is already loaded by BIOS as part of the EfiLoaderCode, technically it's should be possible to directly pass the file locations to HV w/o data copying. Do you see any concerns to do so? For example, it's not a good manner that HV access Bootloaders's any code section.

misc/efi-stub/container.c
```diff
@@ -0,0 +1,488 @@
...
+	/* allocate single memory region to store all binary files to avoid mmap fragmentation */
+	err = emalloc_reserved_aligned(&(ctr->mod_hpa), ctr->total_modsize, 
+										EFI_PAGE_SIZE, 256U * MEM_ADDR_1MB, MEM_ADDR_4GB);
+	if (err != EFI_SUCCESS) {
+		Print(L"Failed to allocate memory for modules %r\n", err);
+		goto out;
+	}
+
+	p = (UINT8 *)ctr->mod_hpa;
+	for (i = LZH_BOOT_IMG + 2; i < ctr->lzh_count - 1; i = i + 2) {
```
tnishiok, line 237: for (i = LZH_BOOT_IMG + 2; i < ctr->lzh_count - 1; i = i + 2) {

shall be replaced with

for (i = LZH_BOOT_IMG + LZH_MOD0_CMD; i < ctr->lzh_count - 1; i = i + 2) {

hypervisor/arch/x86/boot/cpu_primary.S
gvancuts: I'll let the maintainers chime in but in my opinion, it's preferable to have small commits that are organized somewhat logically (by components, etc) as opposed to one big commit that gets squashed.

misc/efi-stub/Makefile
```diff
@@ -0,0 +1,132 @@
...
+# Copyright (c) 2011, Intel Corporation
```
gvancuts, line 2: Is this a file that's coming from somewhere else and that you are copying as-is? I'm asking because I see the year in the header is 2011 (from 10 years ago).

misc/efi-stub/Makefile
```diff
@@ -0,0 +1,132 @@
...
+INCLUDE_PATH += $(INCDIR)/efi
+INCLUDE_PATH += $(INCDIR)/efi/$(ARCH)
+INCLUDE_PATH += $(HV_OBJDIR)/include
+INCLUDE_PATH += $(HV_SRC)/include/arch/x86
+INCLUDE_PATH += $(HV_SRC)/include/arch/x86/guest
+INCLUDE_PATH += $(HV_SRC)/include/public
+INCLUDE_PATH += $(HV_SRC)/include/lib
+INCLUDE_PATH += $(HV_SRC)/boot/include/guest
+
+SCENARIO ?= sdc
```
gvancuts, line 47: I'm wondering what would happen if we end up here with `SCENARIO` not defined yet. It's probably not going to happen, but for the sake of consistency I would use `industry` as the default (which is what we use these days).

misc/efi-stub/Makefile
```diff
@@ -0,0 +1,132 @@
...
+	CFLAGS += -mno-red-zone
+endif
+
+LDFLAGS=-T $(LDSCRIPT) -Bsymbolic -shared -nostdlib -znocombreloc \
+		-L$(LIBDIR) $(CRT0)
+EFIBIN=$(HV_OBJDIR)/$(HV_FILE).efi
+BOOT=$(EFI_OBJDIR)/boot.efi
+HV_BIN=$(HV_OBJDIR)/$(HV_FILE).bin
+
+CONF_FILE=$(CURDIR)/clearlinux/acrn.conf
```
gvancuts, line 95: I don't think we want/need this anymore. This was a `systemd-boot` configuration file. But you're not using this here, are you?

misc/efi-stub/Makefile
```diff
@@ -0,0 +1,132 @@
...
+
+$(EFIBIN): $(BOOT) $(HV_BIN)
+	$(OBJCOPY) --add-section .hv="$(HV_BIN)"  --change-section-vma .hv=0x6e000 --set-section-flags .hv=alloc,data,contents,load  --section-alignment 0x1000 $(BOOT) $(EFIBIN)
+
+$(BOOT): $(EFI_OBJDIR)/boot.so
+
+$(EFI_OBJDIR)/boot.so: $(ACRN_OBJS)
+	$(LD) $(LDFLAGS) -o $@ $^  -lgnuefi -lefi $(shell $(CC) $(CFLAGS) -print-libgcc-file-name)
+
+install-conf: $(CONF_FILE)
```
gvancuts, line 114: You can probably remove this too.

gvancuts (commented):
I'm not seeing how you use this, i.e. how do you build this component?

We will also need to document somewhere that the `gnu-efi` package is needed (again) on the development machine to build this.

misc/efi-stub/Makefile
tnishiok: Thanks. I am thinking of adding readme to meta-acrn side, but I will add readme to under misc/efi-stub/. At least I will split it into two commits one for hv and another for efi-stub. The files derived from ACRN release_2.2. I will update the year in the headers. https://github.com/projectacrn/acrn-hypervisor/blob/release_2.2/misc/efi-stub/Makefile

misc/efi-stub/Makefile
tnishiok: Let me set the default values as below since the feature is required for certain use case

SCENARIO ?= hybrid_rt
BOARD ?= tgl-rvp

misc/efi-stub/Makefile
tnishiok: Yes I will remove two lines and the clearlinux directory.

HV_BIN=$(HV_OBJDIR)/$(HV_FILE).bin
CONF_FILE=$(CURDIR)/clearlinux/acrn.conf

misc/efi-stub/Makefile
tnishiok: Yes I will remove it.

tnishiok:
Hi Geoffroy, thank you for your comments. I cleaned up patch sets and rebased

Issue #6061, opened 2021-05-24
Clear user VM's memory before allocation


lifeix:
[External_System_ID] ACRN-7145

wenlingz:
[External_System_ID] ACRN-7145

NanlinXie:
Closed since patch merged without regression.

PR #6062, opened 2021-05-24
dm: mitigate reset attack

lifeix:
When a platform reboots or shuts down, the contents of RAM are not immediately
lost but begins to decay. During this period, there is a short timeframe during
which an attacker can turn the platform back on to boot into a program that
dumps the contents of memory (e.g., cold boot attacks). Encryption keys and
other secrets can be easily compromised through this method.

We already erasing the guest memory data when the guest is shut down normally.
However, if the guest is shut down abnormally, the contents of RAM may still
there. This patch mitigate this kind reset attack for a DM launched VM by
erasing the guest memory data by the guest has been created.

Tracked-On: #6061

### DIFF
--- a/devicemodel/core/vmmapi.c
+++ b/devicemodel/core/vmmapi.c
@@ -404,6 +404,8 @@ vm_map_memseg_vma(struct vmctx *ctx, size_t len, vm_paddr_t gpa,
 int
 vm_setup_memory(struct vmctx *ctx, size_t memsize)
 {
+	int ret;
+
 	/*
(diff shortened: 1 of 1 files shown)


wenlingz:
OK to verify

wenlingz:

wenlingz:
OK to verify

wenlingz:
try verify EHL

wenlingz:
try verify EHL

wenlingz:
OK to verify

PR #6066, opened 2021-05-24
config-tools: enable PTM through config-tools

yuchuyang:
Configure PTM in post-launched VM using <PTM> element. If the //vm/PTM
sets to 'y', pci_dev.c.xsl appends the virtual root port to
corresponding struct acrn_vm_pci_dev_config of that VM. Currently it
supports only post-launched VMs.

Configure enable_ptm for dm argument. If a uos/enable_ptm with uos id
= 'vm_id 'sets to 'y' and the vm/PTM with the same vm_id sets to 'y',
append an "enable_ptm" flag to the end of passthrough ethernet devices.
Currently there is only ethernet card can support the "enable_ptm"flag.

For the schema validation, the <PTM> can only be ['y', 'n'].

For the launched script validation, the <enable_ptm> can only be ['y',
'n']. If the <enable_ptm> sets to 'y' but the corresponding <PTM> sets
to 'n', the launch script will fail to generate.

### DIFF
--- a/misc/config_tools/data/adl-rvp/industry.xml
+++ b/misc/config_tools/data/adl-rvp/industry.xml
@@ -125,7 +125,6 @@
         <base>0</base>
         <size>0</size>
     </epc_section>
-
     <legacy_vuart id="0">
         <type>VUART_LEGACY_PIO</type>
         <base>COM1_BASE</base>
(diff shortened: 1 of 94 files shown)


wenlingz:

terryzouhao (approved):

PR #6069, opened 2021-05-25
Release 2.5:Refine vbar sizing code

taoyuhong:
Simplify the vBAR sizeing code.
1.when update vBAR, always store (value & size_mask | type_bit) to vfcg.
2.when update vBAR base, check if the vBAR value is in address window,
so sizing vBAR will not cause re-mapping base.
3.Remove pci_vdev_read_vbar().


hypervisor/include/dm/vacpi.h
```diff
@@ -26,6 +26,10 @@
 #define UOS_VIRT_PCI_MMCFG_BASE		0xE0000000UL
 #define UOS_VIRT_PCI_MMCFG_START_BUS	0x0U
 #define UOS_VIRT_PCI_MMCFG_END_BUS	0xFFU
+#define UOS_VIRT_PCI_MEMBASE32      0x80000000UL    /* 2GB */
+#define UOS_VIRT_PCI_MEMLIMIT32     0xC0000000UL    /* 3GB */
```
lifeix, line 30: I think this should change to 0xE0000000UL /**3.5 GB/. Same with the value which is defined in acrn-dm.

lifeix (changes requested):

hypervisor/include/dm/vacpi.h
taoyuhong: Fixed, change UOS_VIRT_PCI_MEMLIMIT32 to 3.5GB

fyin1 (approved):

PR #6082, opened 2021-05-26
Use physical APIC IDs as vLAPIC IDs for pre-launched and post-launched VMs

donsheng:
This pull request introduces the support to use physical APIC IDs as vLAPIC IDs for
pre-launched and post-launched VMs, as an attempt to reproduce the host CPU/cache
topology in pre-launched and post-launched VMs.

Tracked-On: #6020
Signed-off-by: dongshen dongsheng.x.zhang@intel.com


donsheng:
Retrieve physical APIC IDs from board xml file and use them to fill in the ACPI MADT table
for pre-Launched VMs.

Note that the config-tool will throw an error if the processors/die/core/thread tags are absent.
User needs to run board_inspector.py to regenerate the board xml file when this commit is merged,
if the processors/die/core/thread tags are missing in the board xml file.

wenlingz:
OK to verify

dongyaozu (approved):

terryzouhao (approved):
Fine for config tool part.

PR #6086, opened 2021-05-26
release_2.5: hv: vlapic: a minor refine about vlapic_x2apic_pt_icr_access

lifeix:
In physical destination mode, the destination processor is specified by its
local APIC ID. When a CPU switch xAPIC Mode to x2APIC Mode or vice versa,
the local APIC ID is not changed. So a vcpu in x2APIC Mode could use physical
Destination Mode to send an IPI to another vcpu in xAPIC Mode by writing ICR.

This patch adds support for a vCPU A could write ICR to send IPI to another
vCPU B which is in different APIC mode.

Tracked-On: #5923


dongyaozu (approved):

wenlingz:
OK to verify

wenlingz:
OK to verify

PR #6088, opened 2021-05-26
Make ACRN HV with hybrid_rt bootable without GRUB on UEFI BIOS

tnishiok:
This PR has 2 commits:

hv: restore the hypervisor 64-bit entry address for efi-stub
efi-stub: add acrn efi application as an alternative bootloader for secure boot

This PR introduces a new ACRN EFI application based on the legacy efi-stub code which provides booting method of HV on UEFI-BIOS without using the GRUB Bootloader. It is supposed to be used for secure booting on certain platform. By means of that users can boot HV, Service VM kernel, pre-launched VM kernel and its ACPI table binary packed in the Slim Bootloader container boot image file format. ACRN EFI application has additional dependencies to compile which are not listed in the existing ACRN GSG doc. Since this is an optional feature but not all users need, it does not get compiled by default to avoid causing any confusion for existing users. README for how to use the feature will come later in a separated PR.

### DIFF
--- a/misc/efi-stub/Makefile
+++ b/misc/efi-stub/Makefile
@@ -1,0 +1,96 @@
+#
+# Copyright (c) 2011 - 2021, Intel Corporation
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#    * Redistributions of source code must retain the above copyright
+#      notice, this list of conditions and the following disclaimer.
+#    * Redistributions in binary form must reproduce the above
+#      copyright notice, this list of conditions and the following
+#      disclaimer in the documentation and/or other materials provided
+#      with the distribution.
+#    * Neither the name of Intel Corporation nor the names of its
+#      contributors may be used to endorse or promote products derived
+#      from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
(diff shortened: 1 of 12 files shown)


fyin1 (approved):

wenlingz:
@tnishiok make doc fail, this doc part must  go with code ? could you create doc PR on master separately, usually doc PR and code  PR are created separately.
 
New errors/warnings found, please fix them:
==============================================

/home/acrn/workspace/make_doc/pme-yocto-build-make-doc-01/acrn-hypervisor/build/doc/rst/misc/efi-stub/README.rst: WARNING: document isn't included in any toctree

Makefile:65: recipe for target 'html' failed
make[1]: *** [html] Error 1
make[1]: Leaving directory '/home/acrn/workspace/make_doc/pme-yocto-build-make-doc-01/acrn-hypervisor/doc'
Makefile:129: recipe for target 'doc' failed
make: *** [doc] Error 2
+ error make doc

wenlingz:
try verify EHL

tnishiok:
I will remove README.rst from this PR. I will create another PR for doc and add orphan to README.rst. It can work with EHL as well.

misc/efi-stub/Makefile
```diff
@@ -0,0 +1,115 @@
...
+BOARD ?= tgl-rvp
+SCENARIO ?= hybrid_rt
+
+OBJCOPY=objcopy
+
+HOST  = $(shell $(CC) -dumpmachine | sed "s/\(-\).*$$//")
+ARCH := $(shell $(CC) -dumpmachine | sed "s/\(-\).*$$//")
+
+ifeq ($(ARCH),x86_64)
+	FORMAT=efi-app-x86-64
```
gvancuts, line 55: A couple of questions regarding this logic:
* If I understand it correctly, we are essentially deriving the target architecture format based on the output from `gcc -dumpmachine`. I believe our objective here is to actually generate a 32-bit EFI app or 64-bit EFI app based on whether the EFI firmware is itself 32-bit or 64-bit. Is it guaranteed that `gcc -dumpmachine` output will match the EFI firmward format? Or is it possible that it returns that we're running a 32-bit OS on a 64-bit EFI firmward (or vice-versa)?
* Regardless of the answer above, if we assume that the user may be building on a dev system which is different than the target system, it's likely better to not 'guess' the architecture but rather set it via a compile-time variable.

Thoughts?

gvancuts (commented):
I'm still not clear how you build this, but I guess this will be clarified with the doc PR you will send later ;-)

misc/efi-stub/Makefile
```diff
@@ -0,0 +1,115 @@
...
+#
+
+HV_OBJDIR:=build
+HV_SRC:=../../hypervisor
+C_SRCS = boot.c pe.c malloc.c container.c
+ACRN_OBJS := $(patsubst %.c,$(EFI_OBJDIR)/%.o,$(C_SRCS))
+INCLUDE_PATH += $(INCDIR)/efi
+INCLUDE_PATH += $(INCDIR)/efi/$(ARCH)
+INCLUDE_PATH += $(HV_OBJDIR)/include
+INCLUDE_PATH += $(HV_SRC)/include/arch/x86
```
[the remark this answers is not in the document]
tnishiok: I decoupled readme from this PR following the suggestion from wenlingz. Another PR https://github.com/projectacrn/acrn-hypervisor/pull/6089 provides the readme. I will delete unused INCLUDE_PATH.

misc/efi-stub/Makefile
tnishiok: Thank you. I will remove the if condition. The expectation is to build a 64-bit EFI app.

gvancuts: Thanks for the confirmation.

misc/efi-stub/Makefile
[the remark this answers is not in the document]
gvancuts: Thanks (I'll take a look at the doc PR right away).

misc/efi-stub/Makefile
```diff
@@ -0,0 +1,96 @@
+#
+# Copyright (c) 2011 - 2021, Intel Corporation
```
dbkinder, line 2: Shouldn't all these files use the approved ACRN license/copyright header? @NanlinXie

Issue #5915, opened 2021-04-09
Hypervisor PTM Support


NanlinXie:
The Hypervisor shall provide the hypervisor specific functionality needed for PTM(Precision Time Measurement) to function on PTM enabled PCIE devices that are passed through or mediated.

NanlinXie:
[External_System_ID] ACRN-5896

rliu88:
For post launch:
To use, to "enable-ptm" when pass through device, example in launch script:
/home/labuser/RL/bin/acrn-dm -A -m $mem_size -s 0:0,hostbridge \
  -s 3,virtio-blk,/home/labuser/RL/ubuntu_images/uos-test.img \
  -s 4,virtio-net,tap0 \
  -s 5,virtio-console,@stdio:stdio_port \
  -s 6,passthru,a9/00/0,**enable_ptm** \
  --ovmf /usr/share/acrn/bios/OVMF.fd \
}


Sample output from guest os when ptm is enabled (check using dmesg):
uos@uos-Standard-PC-i440FX-PIIX-1996:~$ dmesg | grep -i ptm
[    1.555284] <PTM-init>pci_ptm_init: 00:00.00, ispcie=1, type=0x4
[    1.555306] <PTM-init>no upstream bridge
[    1.555356] <PTM-init> Cannot find PTM ext cap.
[    1.561311] <PTM-init>pci_ptm_init: 00:03.00, ispcie=0, type=0x0
[    1.567146] <PTM-init>pci_ptm_init: 00:04.00, ispcie=0, type=0x0
[    1.572983] <PTM-init>pci_ptm_init: 00:05.00, ispcie=0, type=0x0
[    1.718038] <PTM-init>pci_ptm_init: 00:06.00, ispcie=1, type=0x4
[    1.718056] <PTM-init>no upstream bridge
[    1.722034] <PTM-init> ptm is ptm_root.
[    1.723033] <PTM-init> Condition-2: ptm is enabled.
[    1.723052] pci 0000:00:06.0: PTM enabled (root), 4ns granularity
[    1.766438] <PTM-init>pci_ptm_init: a9:00.00, ispcie=1, type=0x0
[    1.766456] <PTM-init>Ethernet controller upstream bridge port: 00:06.00
[    5.715000] igc_probe enable ptm.
[    5.715068] <PTM-en>pci_enable_ptm: a9:00.00, ispcie=1, type=0x0
[...]

PR #6093, opened 2021-05-26
dm: return error if PTM is not enabled on physical root port

rliu88:
Cherry-pick the changes from master to release 2.5 since my last patch
is merged already.

hv is responsible to ensure PTM is always enabled on hw root port if
that root port is PTM root-capable.  If PTM root is not enabled already in physical
root port before guest launch, guest OS can only enable it in root port's virtual
config space and PTM may not function as desired so we would rather not
to allow user to enable PTM on pass-thru device in this case.
Also revisit a few comments to add assumption that acrn only support a
simple PTM hierarch emulation i.e., EP (ptm requestor) is directly connected
to root port (ptm root), or ptm requestor is rcie).

V4:
 - Change behavior from V3: When users tries to enable PTM while PTM root is not enabled on
physical root port, allow user to launch VM and pass thru the device
with no PTM support.
V3:
 - When users tries to enable PTM while PTM root is not enabled on
physical root port, allow user to launch VM and enable PTM in virtual
root port.  This is not going to enable PTM on physical root port.

Tracked-On: #5915
Signed-off-by: Rong Liu <rong.l.liu@intel.com>
Acked-by: Jason Chen <jason.cj.chen@intel.com>

### DIFF
--- a/devicemodel/hw/pci/ptm.c
+++ b/devicemodel/hw/pci/ptm.c
@@ -181,12 +181,15 @@ int ptm_probe(struct vmctx *ctx, struct passthru_dev *ptdev, int *vrp_sec_bus)
 		/* hv is responsible to ensure that PTM is enabled on hw root port if
 		 * root port is PTM root-capable.  If PTM root is not enabled already in physical
 		 * root port before guest launch, guest OS can only enable it in root port's virtual
-		 * config space and PTM may not function as desired.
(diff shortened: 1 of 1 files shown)


wenlingz:

ywan170:
@fyin1

JasonChenCJ (approved):

fyin1 (approved):

PR #6107, opened 2021-05-27
Config_tools: Update XML and save board xml when loading from config  tool UI

Kunhui-Li:
1. For nuc11tnbi5 board,
   1) Add hybrid.xml, industry.xml, hybrid_rt.xml, logical_partition.xml and industry_launch_2uos.xml.
   2) Update nuc11tnbi5.xml for GSG.
2. Update tgl-rvp board xml.
3. save the board xml into misc/config_tools/data/[board]/ when loading a new board from config tool UI.


wenlingz:
OK to verify

gvancuts (changes requested):
>     3. save the board xml into misc/config_tools/data/[board]/ when loading a new board from config tool UI.

This is not the correct behaviour. The tool shouldn't modify the `acrn-hypervisor` repository. This can be put in the build folder or ask the user where he wants to keep this.

I'm also not completely understanding why the `$(BOARD).xml` would be changed by the tool if you import it, can you explain what the tool does to it?

szhen11:
> > ```
> > 3. save the board xml into misc/config_tools/data/[board]/ when loading a new board from config tool UI.
> > ```
> 
> This is not the correct behaviour. The tool shouldn't modify the `acrn-hypervisor` repository. This can be put in the build folder or ask the user where he wants to keep this.
> 
> I'm also not completely understanding why the `$(BOARD).xml` would be changed by the tool if you import it, can you explain what the tool does to it?



> > ```
> > 3. save the board xml into misc/config_tools/data/[board]/ when loading a new board from config tool UI.
> > ```
> 
> This is not the correct behaviour. The tool shouldn't modify the `acrn-hypervisor` repository. This can be put in the build folder or ask the user where he wants to keep this.
> 
> I'm also not completely understanding why the `$(BOARD).xml` would be changed by the tool if you import it, can you explain what the tool does to it?

Hi, @gvancuts ,
When users generate a new board xml from a new board and want to build ACRN with the new board, users can first import the new board xml from the config tool UI, then config tool will automatically generate a folder in misc/config_tools/data/[board_type] if not exist and copy the new board xml and scenario xmls(from misc/config_tools/data/generic_board to misc/config_tools/data/[board_type], thus user can use [the standard way](https://projectacrn.github.io/latest/getting-started/building-from-source.html#build-with-the-acrn-scenario) "make BOARD=[board type] SCENARIO=[scenario]" to build ACRN, users may also modify the scenario xml for the current board easily. The build folder should be used to put configs from build steps, it may be erased when make clean and is not suitable to put the new board/scenario xmls user edited; I suppose we will use separated repo to put config xmls in the future, after that we can improve the board/scenario xml saving process and let user to put them in where he wants to keep them.
If we import the board xml, it would be put into misc/config_tools/data/[board_type]/ and its content should not be changed.

gvancuts:
> Hi, @gvancuts ,
> When users generate a new board xml from a new board and want to build ACRN with the new board, users can first import the new board xml from the config tool UI, then config tool will automatically generate a folder in misc/config_tools/data/[board_type] if not exist and copy the new board xml and scenario xmls(from misc/config_tools/data/generic_board to misc/config_tools/data/[board_type], thus user can use [the standard way](https://projectacrn.github.io/latest/getting-started/building-from-source.html#build-with-the-acrn-scenario) "make BOARD=[board type] SCENARIO=[scenario]" to build ACRN, users may also modify the scenario xml for the current board easily. The build folder should be used to put configs from build steps, it may be erased when make clean and is not suitable to put the new board/scenario xmls user edited; I suppose we will use separated repo to put config xmls in the future, after that we can improve the board/scenario xml saving process and let user to put them in where he wants to keep them.
> If we import the board xml, it would be put into misc/config_tools/data/[board_type]/ and its content should not be changed.

We cannot change the content of the repo. Custom board and scenario XMLs files should be kept elsewhere. The user will point at those when building ACRN, this is why we changed the build parameters `BOARD` and `SCENARIO` so that it can also take path and not just a filename.

NanlinXie:
@gvancuts , let's revert the UI tool modification, just add xml files in this PR.
For the UI tool flow, I suggest we need have more clear flow for UI tool flow.

gvancuts:
> @gvancuts , let's revert the UI tool modification, just add xml files in this PR.
> For the UI tool flow, I suggest we need have more clear flow for UI tool flow.

Sounds good, thanks.

NanlinXie (approved):

Issue #5928, opened 2021-04-13
tutorial error in building acrn over qemu


iaGuoZhi:
**Describe the bug**
When I want to boot acrn through qemu, tutorial [**Enable ACRN Over QEMU/KVM**](https://projectacrn.github.io/latest/tutorials/acrn_on_qemu.html) tells me build acrn hypervisor using  tag: acrn-2020w19.5-140000p. however there is no qemu.xml  in tag: acrn-2020w19.5-140000p which is used as BOARD_FILE. 

**Codebase**

https://github.com/projectacrn/acrn-hypervisor/tree/acrn-2020w19.5-140000p/misc/acrn-config/xmls/board-xmls

gvancuts:
Good catch! (I wonder how this even passed our Q&A testing... @NanlinXie ?)

We should really take the opportunity to refresh this to be based on the latest release, i.e. `v2.4`.

iaGuoZhi:
Is there any arch-hypervisor release can build acrn over qemu? I'm eager to try ACRN over qemu.

gvancuts:
I'm in the process of trying to update this document to be based on ACRN v2.4 (the latest release as of today).

Unfortunately, ACRN v2.4 does not start. It throws the following error on the console:
```
Loading ACRN hypervisor with SDC scenario ...
error: Can't get controller info..
Multiboot Bootloader: GRUB 2.02-2ubuntu8.21
[0us][cpu=0][(null)][sev=3][seq=1]:vlapic: Basic apicv ops is used!
detect_hardware_support, physical-address width (40) over maximum physical-address width (39)
[0us][cpu=0][(null)][sev=1][seq=2]:PANIC: init_pcpu_pre line: 132

[0us][cpu=0][(null)][sev=1][seq=3]:hardware not support!
```

@NanlinXie can you confirm what is the latest version that we have validated? Can you also ask one of the engineer to chime in as to what the problem may be with this ACRN v2.4 error on QEMU?

gvancuts:
For information, I have tried this on both Skylake and Kaby Lake processors with the same results.

iaGuoZhi:
Hi gvancuts!
[...]

PR #6109, opened 2021-05-28
doc: update ACRN version QEMU tutorial

gvancuts:
We missed a section during our last update to the QEMU tutorial
that references which version (tag) of ACRN to use.

Tracked-On: #5928


dbkinder (approved):
LGTM, thanks.

Issue #6113, opened 2021-05-31
[config_tools][ADL-S]generated board xml parse error on ADL-S


Kunhui-Li:
Environment
setup native on ADL-S
# must include but not limited to the items below

HW/Board
ADL-S
# D0 MRB, APLNUC, KBLNUC, UP2 ,etc.

Build link
acrn commit: 736a03222f5f359aadac9119062ee8255160bad2
# daily/weekly build link

Image info

# including IOC version, SOS version, SOS kernel version, UOS Type, UOS version, UOS kernel version.

Steps
1. setup native OS on ADL-S
2. generate adl-s-crb.xml by
cd misc/config_tools/target
python3 board_parser.py adl-s-crb
adl-s-crb.xml generated in output/
3. use adl-s-crb.xml, import into config editor or build acrn

# clarify the reproduce steps clearly .

Expected result
no errors.

# clarify the expected result clearly.

Actual result
xml parse error:
>>> import xml.etree.ElementTree as ElementTree
>>> board_root = ElementTree.parse('adl-s-crb.xml').getroot()
Traceback (most recent call last):
File "<stdin>", line 1, in <module>
File "/usr/lib/python3.7/xml/etree/ElementTree.py", line 1197, in parse
tree.parse(source, parser)
File "/usr/lib/python3.7/xml/etree/ElementTree.py", line 598, in parse
self._root = parser._parse_whole(source)
xml.etree.ElementTree.ParseError: mismatched tag: line 70, column 3

# clarify the actual result clearly.

Reproduce rate

# percentage is not acceptable, please use the actually records like 1/20, 2/20 ,etc.

Debugging info
illegal characters in board xml

00:00.0 Host bridge: Intel Corporation Device 4660 (rev 01)
Region 3: Memory at <ignored> (32-bit, non-prefetchable)
00:02.0 VGA compatible controller: Intel Corporation Device 4680 (rev 04)
Region 0: Memory at 6000000000 (64-bit, non-prefetchable) [size=16M]

# debug logs could be put here or attached in an attachment.

Kunhui-Li:
[External_System_ID] ACRN-7129

PR #6114, opened 2021-05-31
config_tools: replace illegal character with escaped character

Kunhui-Li:
For illegal characters, replace original characters with escaped characters in board.xml.

Tracked-On: #6113
Signed-off-by: Kunhui Li <kunhuix.li@intel.com>


wenlingz:
OK to verify

terryzouhao (approved):

Issue #6147, opened 2021-06-04
ASAN reports UAF + SEGV when fuzzing exposed PIO with Hypercube guest VM.


shuox:
==468==ERROR: AddressSanitizer: heap-use-after-free on address 0x6180000200e0 at pc 0x559e3c1f0895 bp 0x7ffdd5f07de0 sp 0x7ffdd5f07dd0
READ of size 4 at 0x6180000200e0 thread T0 (mevent)
    #0 0x559e3c1f0894 in timer_handler core/timer.c:45
    #1 0x559e3c1c42bb in mevent_handle core/mevent.c:203
    #2 0x559e3c1c754e in mevent_dispatch core/mevent.c:441
    #3 0x559e3c1e527f in main core/main.c:1091
    #4 0x7f49155a609a in __libc_start_main ../csu/libc-start.c:308
    #5 0x559e3bebbdfd in _start (/usr/bin/acrn-dm+0x1eadfd)0x6180000200e0 is located 96 bytes inside of 784-byte region [0x618000020080,0x618000020390)
freed by thread T81 (vcpu 0) here:
    #0 0x559e3bfa6a4f in __interceptor_free (/usr/bin/acrn-dm+0x2d5a4f)
    #1 0x559e3c15ac40 in virtio_rnd_deinit hw/pci/virtio/virtio_rnd.c:514
    #2 0x559e3c0ea7a5 in pci_emul_deinit hw/pci/core.c:1048
    #3 0x559e3c0eb241 in deinit_pci hw/pci/core.c:1620
    #4 0x559e3c1e117d in vm_reset_vdevs core/main.c:581
    #5 0x559e3c1e13dc in vm_system_reset core/main.c:635
    #6 0x559e3c1e2545 in vm_loop core/main.c:712
    #7 0x559e3c1e3562 in start_thread core/main.c:249
    #8 0x7f4916b25fa2 in start_thread /build/glibc-vjB4T1/glibc-2.28/nptl/pthread_create.c:486previously allocated by thread T81 (vcpu 0) here:
    #0 0x559e3bfa7046 in calloc (/usr/bin/acrn-dm+0x2d6046)
    #1 0x559e3c15c7de in virtio_rnd_init hw/pci/virtio/virtio_rnd.c:394
    #2 0x559e3c0d8c54 in pci_emul_init hw/pci/core.c:1034
    #3 0x559e3c0ebb25 in init_pci hw/pci/core.c:1419
[...]

PR #6148, opened 2021-06-04
dm: Reset virtio device before release

shuox:
With virtio polling mode enabled, a timer is running in the virtio
backend service. And the timer will also be triggered if its frondend
driver didn't do the device reset in shutdown. A freed virtio device
will be accessed in the polling timer handler.

Do the virtio reset() callback specifically to clear the polling timer
before the free.

Tracked-On: #6147


ywan170 (approved):
I just feel that it will cause some regressions.
The deinit callbacks will be involved in VM reset/s3/s5. Please verify the important devices like net, block before merge.

Issue #6157, opened 2021-06-07
coding style fix on v2.5 branch 


yonghuah:
Suspicious dereference of pointer 'entry' by passing argument 2 to function 'mmio_rb_tree_RB_REMOVE' at line 246 before NULL check at line 252
./acrn-hypervisor/devicemodel/core/mem.c:246 | unregister_mem_int()

 Possible memory leak. Dynamic memory stored in 'buf' allocated through function 'malloc' at line 1191 can be lost at line 1201
./acrn-hypervisor/devicemodel/hw/platform/acpi/acpi.c:1201 | create_and_inject_vrtct()

yonghuah:
[External_System_ID] ACRN-7196

PR #6158, opened 2021-06-07
dm: remove redundant null check in unregister_mem_init

yonghuah:
return value 'err' of mmio_rb_lookup() being 0 ensures
  'entry' is not NULL, hence checking it before 'free(entry)'
  is unnecessary.

Tracked-On: #6157


wenlingz:

ywan170 (approved):

PR #6188, opened 2021-06-10
update scenario xml

Kunhui-Li:
1.  Update the value of the tag MAX_PT_IRQ_ENTRIES from 64 to 128 in TGL-RVP scenario xml.
2.   Update the value of the tag MAX_MSIX_TABLE_NUM from 64 to empty to fix the issue that MAX_MSIX_TABLE_NUM value is auto set as 64 when generate an new scenario xml for all scenario xml.


NanlinXie (approved):

wenlingz:
OK to verify

PR #6190, opened 2021-06-10
Fix the bug that pci passthru ethernet failed to get ip address

rliu88:
Root cause:  vrp accidentally changed passthru device's max payload which doesn't match its backend's max payload.  This match causes tx/rv to fail on ethernet card.

Fix: Set vrp's max payload to match hardware value


wenlingz:

fyin1 (approved):

PR #6191, opened 2021-06-10
Fix the bug that pci passthru ethernet failed to get ip address

rliu88:
Root cause: vrp accidentally changed passthru device's max payload which doesn't match its backend's max payload. This match causes tx/rv to fail on ethernet card.

Fix: Set vrp's max payload to match hardware value


wenlingz:

ywan170:
Is it 2.5 blocking bugs? If not, please submit it to the mailing list to review.

rliu88:
Yes.  It is a fix of 2.5 blocking bug.  They are in the mailing list and Jason reviewed these.

ywan170 (approved):

Issue #6134, opened 2021-06-02
[acrn-configuration-tool] lxml module not found when get board xml following doc


Kunhui-Li:
Environment

# must include but not limited to the items below

HW/Board

# D0 MRB, APLNUC, KBLNUC, UP2 ,etc.

Build link

# daily/weekly build link

Image info

# including IOC version, SOS version, SOS kernel version, UOS Type, UOS version, UOS kernel version.

Steps
doc: https://projectacrn.github.io/latest/tutorials/acrn_configuration_tool.html#using-acrn-configuration-toolset
get board xml from native environment.
# clarify the reproduce steps clearly .

Expected result
get board xml successfully
# clarify the expected result clearly.

Actual result
1. ./board_inspector/README: need to update
usage: python3 run.py <board_name> [--out board_info_file]
should be usage: python3 cli.py <board_name> [--out board_info_file]

2. python3 cli.py adl-s-crb
Traceback (most recent call last):
File "cli.py", line 11, in <module>
import lxml.etree
ModuleNotFoundError: No module named 'lxml'

# clarify the actual result clearly.

Reproduce rate

# percentage is not acceptable, please use the actually records like 1/20, 2/20 ,etc.

Debugging info

# debug logs could be put here or attached in an attachment.

Kunhui-Li:
[External_System_ID] ACRN-7157

PR #6213, opened 2021-06-16
config_tools: modify README file

Kunhui-Li:
1. Update python script name that is executed to generate board xml file
from run.py to cli.py in README file.
2. Update the OS requirement under native Linux in
misc/config_tools/board_inspector/README file.

Tracked-On: #6134

### DIFF
--- a/misc/config_tools/board_inspector/README
+++ b/misc/config_tools/board_inspector/README
@@ -1,6 +1,6 @@
 board_parser.py will collect all board related info and then generate a board info file for acrn-config host tool usage.
 
-usage: python3 run.py <board_name> [--out board_info_file]
+usage: python3 cli.py <board_name> [--out board_info_file]
 
 board_name : the name of board that run ACRN hypervisor, like apl-up2/nuc7i7dnb. It will be used as name of the board configurations folder which created by acrn-config host tool.
 board_info_file : (optional) the name of board info file. if it is not specified, a name of <board_name>.xml will be generated under the current working directory by default.
@@ -8,6 +8,6 @@ board_info_file : (optional) the name of board info file. if it is not specified
 Please run this script under native Linux environment with root privilege.
 
 OS requirement:
-	Release:	Ubuntu 18.04+ or ClearLinux 30210+
-	Tools:		cpuid, rdmsr, lspci, dmidecode
+	Release:	Ubuntu 18.04+
+	Tools:		cpuid, rdmsr, lspci, lxml, dmidecode (optional)
 	kernel cmdline: "idle=nomwait intel_idle.max_cstate=0 intel_pstate=disable"


NanlinXie (approved):

gvancuts:
Could we use a better name than `cli.py` for the tool? Not sure what it's supposed to stand for but to me I read it as "command-line interface", which does not give any hint as to what the tool actually does.

junjiemao1:
> Could we use a better name than `cli.py` for the tool? Not sure what it's supposed to stand for but to me I read it as "command-line interface", which does not give any hint as to what the tool actually does.

It is named so because the package name (i.e. ``board_inspector``) already indicates what scripts under this directory is aimed at. Naming it ``board_inspector.py`` seems to be redundant.

It is quite common to name the module that is supposed to be used as the command line interface as ``cli.py`` in python packages, e.g. gitlab, xmlschema and flask.

gvancuts:
> It is named so because the package name (i.e. `board_inspector`) already indicates what scripts under this directory is aimed at. Naming it `board_inspector.py` seems to be redundant.
> 
> It is quite common to name the module that is supposed to be used as the command line interface as `cli.py` in python packages, e.g. gitlab, xmlschema and flask.

Really? Is this the gitlab Python utility you're referring to: https://python-gitlab.readthedocs.io/en/stable/cli-usage.html? If so, it seems that it may be the name of the `.py` file but that it's wrapped up in a command you can run from a terminal that uses a much more meaningful name.

junjiemao1:
> > It is named so because the package name (i.e. `board_inspector`) already indicates what scripts under this directory is aimed at. Naming it `board_inspector.py` seems to be redundant.
> > It is quite common to name the module that is supposed to be used as the command line interface as `cli.py` in python packages, e.g. gitlab, xmlschema and flask.
> 
> Really? Is this the gitlab Python utility you're referring to: https://python-gitlab.readthedocs.io/en/stable/cli-usage.html? If so, it seems that it may be the name of the `.py` file but that it's wrapped up in a command you can run from a terminal that uses a much more meaningful name.

Yes, that's right. I only looked at the source code of these packages which do not contain the wrapper scripts (which are created by pip upon installation). I should have noticed that earlier.

There are two options for improvement:

1. Rename ``cli.py`` or create another wrapper script with a meaningful name.
2. Create a ``setup.py`` for the board inspector so that users (or Yocto builds) can install the tool along with all dependencies using ``pip``. The wrapper script can then be created by ``pip`` instead.

Any preference?

gvancuts:
> Yes, that's right. I only looked at the source code of these packages which do not contain the wrapper scripts (which are created by pip upon installation). I should have noticed that earlier.

Thanks for the confirmation! And no worries for not noticing this before.

> 
> There are two options for improvement:
> 
>     1. Rename `cli.py` or create another wrapper script with a meaningful name.
> 
>     2. Create a `setup.py` for the board inspector so that users (or Yocto builds) can install the tool along with all dependencies using `pip`. The wrapper script can then be created by `pip` instead.

I like that!! So we would install the tool and dependencies in a single command, e.g. `sudo pip install acrn_board_inspector` for example?

> 
> 
> Any preference?

Love the idea of having a the board inspector that can be installed using `pip`!

junjiemao1:
> I like that!! So we would install the tool and dependencies in a single command, e.g. `sudo pip install acrn_board_inspector` for example?

Using a single command means we need to publish the board inspector on pypi. At least we can reduce the deployment of the board inspector to two steps:

1. Copy the board_inspector directory to target, and
2. ``sudo pip3 install <path to board inspector>``

> 
> > Any preference?
> 
> Love the idea of having a the board inspector that can be installed using `pip`!

Got it. I'll take a look how this can be done when I have time.

Issue #5741, opened 2021-02-10
doc: Supported HW doc is obsolete and misleading


dbkinder:
The supported HW document (https://projectacrn.github.io/latest/reference/hardware.html) describes platforms as "verified".  We know that older hardware platforms, such as Apollo Lake, cannot fully support newer versions of ACRN, so the hardware list is misleading.  Platforms were verified for a specific version of ACRN, but are longer verified as new versions of ACRN are released.

We at least need to add a notation indicating the ACRN version used to verify a platform (e.g., ``V1.0``).  We should also list which ACRN features are verified on particular (newer) platforms and won't run on older platforms.  (That way, someone with an older HW platform can know what they can and cannot do with ACRN.)  @NanlinXie Can the QA or support team provide such information?

Alternatively, we may need to just remove older platforms that don't fully support the latest release of ACRN.

dbkinder:
@junjiemao1 can we use the board.xml information to automatically figure out what ACRN features a board can or cannot support as part of the board/scenario validation?

junjiemao1:
@dbkinder We are about to refine the structure of board.xml in the v2.5 release to make the XML file more informative and machine-friendly. We also plan to introduce some examples how unavailable features due to hardware limitations can be recorded as XSD assertions and reported in the board configuration phase. More thorough checks will be introduced in the release after the next, after we verified the infrastructure.

gvancuts:
To add to this:
* Can we confirm whether the table indicating what is validated or not is still accurate? (note: we've discovered many issues when running on Apollo Lake which make me suspicious about the SDC validation on that platform...)
* The list of usage scenarios does not match what we have in the code today... we need to update the list **and** make sure we check the ones we **actually** validate ;-)

![image](https://user-images.githubusercontent.com/2192801/109822212-0989bf80-7c37-11eb-85b0-ea3e6aa84dc7.png)

dbkinder:
@NanlinXie it is OK to tag this as a documentation task for v2.5?

PR #6176, opened 2021-06-08
config_tools: clean up the board folders

Kunhui-Li:
1. Remove apl-up2, apl-up2-n3350, whl-ipc-i5, apl-mrb, nuc6cayh board
   folders from the latest code base.
2. Copy tgl-rvp.xml to generic_board.xml.

Tracked-On: #6175
Signed-off-by: Kunhui-Li <kunhuix.li@intel.com>

PR #6217, opened 2021-06-16
doc: update supported hardware document

dbkinder:
Fixes: #5741

### DIFF
--- a/doc/reference/hardware.rst
+++ b/doc/reference/hardware.rst
@@ -25,35 +25,28 @@ Minimum Requirements for Processor
 
 Known Limitations
 *****************
-Platforms with multiple PCI segments
 
-ACRN assumes the following conditions are satisfied from the Platform BIOS
+Platforms with multiple PCI segments are not supported.
 
-* All the PCI device BARs should be assigned resources, including SR-IOV VF BARs if a device supports.
+ACRN assumes the following conditions are satisfied from the Platform BIOS:
 
-* Bridge windows for PCI bridge devices and the resources for root bus, should be programmed with values
+* All the PCI device BARs must be assigned resources, including SR-IOV VF BARs if a device supports it.
+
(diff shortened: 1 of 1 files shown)


dbkinder:
@nanlin I've submitted this PR to update the supported hardware document with the information you sent me.  I do note that some of the boards on your list don't have a corresponding XML file so are they truly supported?   Also I see PR #6176 is removing some board XML files that I'm referencing in this update.  Should we just delete those rows, or just indicate ``N/A`` for those boards?

fuzhongl:
> @nanlin I've submitted this PR to update the supported hardware document with the information you sent me. I do note that some of the boards on your list don't have a corresponding XML file so are they truly supported? Also I see PR #6176 is removing some board XML files that I'm referencing in this update. Should we just delete those rows, or just indicate `N/A` for those boards?

Echo to @dbkinder 👍 
Prefer to remove all info about the boards which don't have a corresponding XML file.

NanlinXie:
> @nanlin I've submitted this PR to update the supported hardware document with the information you sent me. I do note that some of the boards on your list don't have a corresponding XML file so are they truly supported? Also I see PR #6176 is removing some board XML files that I'm referencing in this update. Should we just delete those rows, or just indicate `N/A` for those boards?

For the board that we don't have predefined XML file , we could add note to let them try configtool to generate the board.xml on their own board.

gvancuts:
The boards that do not have a corresponding XML are variations of other boards. I'm starting to think we should do the following:
* Specify which board/system is supported and has a pre-defined XML file in the repo
* Add a note that similar systems (e.g. running i5 instead of i7) should be easy to bring up provided an adequate `$(BOARD).xml` is generated and used
* There is also actually a board that's not listed but *does* have a XML set of files: `apl-up2`

dbkinder:
I'm going to merge this now.  We can submit further edits as needed.

Issue #6208, opened 2021-06-14
config-tools: allow board_inspector to run in nested environment


yuchuyang:
While running in a nested environment, such as qemu, parse the board information should be allowed even it is not in a native environment.

yuchuyang:
[External_System_ID] ACRN-7228

PR #6219, opened 2021-06-17
config-tools: do not exit when the board inspector runs in hypervisor

yuchuyang:
While running in a nested environment, such as qemu, parse the board information should be allowed even it is not in a native environment.
Replace the error with warning message and does not exit the program.

Tracked-On: #6208

### DIFF
--- a/misc/config_tools/board_inspector/cli.py
+++ b/misc/config_tools/board_inspector/cli.py
@@ -11,11 +11,23 @@ import subprocess
 import lxml.etree
 import argparse
 from importlib import import_module
+from cpuparser import parse_cpuid, get_online_cpu_ids
 
 script_dir = os.path.dirname(os.path.realpath(__file__))
 sys.path.append(os.path.join(script_dir))
 
+def native_check():
+    cpu_ids = get_online_cpu_ids()
+    cpu_id = cpu_ids.pop(0)
+    leaf_1 = parse_cpuid(1, 0, cpu_id)
+    if leaf_1.hypervisor != 0:
+        logging.warning(f"Board inspector is running inside a Virtual Machine (VM). Running ACRN inside a VM is only" \
+        "supported under KVM/QEMU. Unexpected results may occur when deviating from that combination.")
+
 def main(board_name, board_xml, args):
+    # Check if this is native os
+    native_check()
+
     try:
         # First invoke the legacy board parser to create the board XML ...
         legacy_parser = os.path.join(script_dir, "legacy", "board_parser.py")
--- a/misc/config_tools/board_inspector/cpuparser/__init__.py
+++ b/misc/config_tools/board_inspector/cpuparser/__init__.py
@@ -59,3 +59,15 @@ def parse_cpuid(leaf, subleaf, cpu_id):
             return None
     else:
         return None
+
+def get_online_cpu_ids():
+    acc = list()
+    with open("/sys/devices/system/cpu/online", "r") as f:
+        line = f.read().strip()
+        for r in line.split(","):
+            if r.find("-") > 0:
+                first, last = tuple(map(int, r.split("-")))
+                acc.extend(range(first, last + 1))
+            else:
+                acc.append(int(r))
+    return acc
--- a/misc/config_tools/board_inspector/cpuparser/cpuids.py
+++ b/misc/config_tools/board_inspector/cpuparser/cpuids.py
@@ -75,6 +75,7 @@ class LEAF_1(CPUID):
     avx = cpuidfield(ECX, 28, 28)
     f16c = cpuidfield(ECX, 29, 29)
     rdrand = cpuidfield(ECX, 30, 30)
+    hypervisor = cpuidfield(ECX, 31, 31)
 
     fpu = cpuidfield(EDX, 0, 0)
     vme = cpuidfield(EDX, 1, 1)
--- a/misc/config_tools/board_inspector/extractors/10-processors.py
+++ b/misc/config_tools/board_inspector/extractors/10-processors.py
@@ -6,7 +6,7 @@
 import logging
 import lxml.etree
 
-from cpuparser import parse_cpuid
+from cpuparser import parse_cpuid, get_online_cpu_ids
 from extractors.helpers import add_child, get_node
 
 level_types = {
@@ -16,18 +16,6 @@ level_types = {
     4: "tile",
     5: "die",
 }
-
-def get_online_cpu_ids():
-    acc = list()
-    with open("/sys/devices/system/cpu/online", "r") as f:
-        line = f.read().strip()
-        for r in line.split(","):
-            if r.find("-") > 0:
-                first, last = tuple(map(int, r.split("-")))
-                acc.extend(range(first, last + 1))
-            else:
-                acc.append(int(r))
-    return acc
 
 def get_parent(processors_node, topo_level, topo_id):
     n = get_node(processors_node, f"//{topo_level}[@id='{topo_id}']")
--- a/misc/config_tools/board_inspector/legacy/board_parser.py
+++ b/misc/config_tools/board_inspector/legacy/board_parser.py
@@ -30,25 +30,6 @@ def check_permission():
         parser_lib.print_red("You need run this tool with root privileges (sudo)!")
         sys.exit(1)
 
-
-def native_check():
-    """Check if this is natvie os"""
-    cmd = "cpuid -r -l 0x01"
-    res = parser_lib.cmd_execute(cmd)
-    while True:
-        line = parser_lib.decode_stdout(res)
-
-        if line:
-
-            if len(line.split()) <= 2:
-                continue
-
-            reg_value = line.split()[4].split('=')[1]
-            break
-
-    return int(reg_value, 16) & 0x80000000 == 0
-
-
 def vendor_check():
     """Check the CPU vendor"""
     with open("/proc/cpuinfo", 'r') as f_node:
@@ -90,10 +71,6 @@ def check_env():
             if int(version) < 20170122:
                 parser_lib.print_yel("Need CPUID version >= 20170122")
                 sys.exit(1)
-
-    if not native_check():
-        parser_lib.print_red("Please run this tools in a native OS environment!")
-        sys.exit(1)
 
     if os.path.exists(OUTPUT):
         shutil.rmtree(OUTPUT)


yuchuyang:
@wenlingz  Please hold this PR, do not merge yet. Thanks.

NanlinXie:
@fuzhongl , could you help to verify this?

misc/config_tools/board_inspector/cli.py
```diff
@@ -11,11 +11,24 @@
...
 
 script_dir = os.path.dirname(os.path.realpath(__file__))
 sys.path.append(os.path.join(script_dir))
 
+def native_check():
+    cpu_ids = get_online_cpu_ids()
+    cpu_id = cpu_ids.pop(0)
+    leaf_1 = parse_cpuid(1, 0, cpu_id)
+    if leaf_1.hypervisor != 0:
+        logging.error(f"Please run this tools in a native OS environment!")
```
dbkinder, line 24: We avoid saying "please".  Also you might include saying it looks like you're running the tool in a VM.  Perhaps:
"The board inspector must be run in a native OS environment and not from a VM."

misc/config_tools/board_inspector/cli.py
```diff
@@ -11,11 +11,24 @@
...
+        sys.exit(1)
+
 def main(board_name, board_xml, args):
+    if not args.no_native_check:
+        # Check if this is natvie os
```
dbkinder, line 29:
```suggestion
        # Check if this is native os
```

dbkinder (changes requested):
Does  ``--no-native-check`` allow the tool to work in a VM? Does the tool still work as expected then?  Why do we need this option?

Also, we have no documentation about the board inspector options :(

yuchuyang:
@dbkinder This option by far is particular for qemu (it maybe for other nested env in the future) like the commit message stats "While running in a nested environment, such as qemu, parse the board information should be allowed even it is not in a native environment."

qemu may configure the env unlike how it does without qemu. It is needed to get the board information which is configured by qemu, that's why this flag is implemented.

fuzhongl:
> @fuzhongl , could you help to verify this?
Verified! Please check following log:

```
root@ubuntu:/home/acrn/acrn-hypervisor# python3 misc/config_tools/board_inspector/cli.py --no-native-check qemu
Warning:The Cx data for ACRN relies on acpi_idle driver but it is not found, Warning:please make sure ACPI Cstate is enabled in BIOS.
Warning:No scaling_driver found.
Resource Allocation is not supported!
qemu.xml has been generated successfully!
root@ubuntu:/home/acrn/acrn-hypervisor# ls misc/config_tools/board_inspector
acpiparser  cli.py  cpuparser  extractors  legacy  lib  memmapparser  pcieparser  qemu.xml  README  smbiosparser
```

gvancuts:
How about we do this the other way around, i.e.:
1. Check whether we are running in KVM (Python equivalent of `cpuid -l  0x40000000 | grep -c KVMKVMVKM`
2. If we are, allow the tool to run but display a warning to the user that they are creating a $(BOARD).xml for a virtual machine

That way we do not need a new special flag.

yuchuyang:
@gvancuts tool may not run inside a VM, for example, virtual box would fail. This flag is recommended only if user knows what exactly she/he is doing.
I would think if we shall explicitly specify the flag is for qemu only(by far) to prevent out-of-box usage...

gvancuts:
> @gvancuts tool may not run inside a VM, for example, virtual box would fail. This flag is recommended only if user knows what exactly she/he is doing.
> I would think if we shall explicitly specify the flag is for qemu only(by far) to prevent out-of-box usage...

Yes, this is my proposal. Check for `cpuid -l 0x40000000 | grep -c KVMKVMVKM` to be equal or greater than 1 and if so, it means we are running in QEMU/KVM. In that case, allow the script to proceed **and** display a warning so that the user is aware of the special case.

yuchuyang:
> > @gvancuts tool may not run inside a VM, for example, virtual box would fail. This flag is recommended only if user knows what exactly she/he is doing.
> > I would think if we shall explicitly specify the flag is for qemu only(by far) to prevent out-of-box usage...
> 
> Yes, this is my proposal. Check for `cpuid -l 0x40000000 | grep -c KVMKVMVKM` to be equal or greater than 1 and if so, it means we are running in QEMU/KVM. In that case, allow the script to proceed **and** display a warning so that the user is aware of the special case.

There are 2 ways to remove the "--no-native-check" flag: 
1. if we don't care about which hypervisor is now, based on this patch, we could remove the --no-native-check flag and replace the logging.error(f"The board inspector must be run in a native OS environment and not from a VM.")
 with logging.warning(f"The board inspector is running on top of hypervisor, you should running this on top of QEMU only.") and do not exit. If it fails, it gets sufficient information.
2. as Geoffroy proposed, If we want to be precise, cpuid with 0x40000000 would work too.. just extra work to implement a class here. Btw, @gvancuts do you know if in the 0x40000000 would return "" KVMKVMKVM  " or "TCGTCGTCGTCG"? 

@junjiemao1 any comments?

gvancuts:
>     2. as Geoffroy proposed, If we want to be precise, cpuid with 0x40000000 would work too.. just extra work to implement a class here. Btw, @gvancuts do you know if in QEMU, cpuid with 0x40000000 would return "" KVMKVMKVM  " or "TCGTCGTCGTCG"?

It will return `KVMKVMKVM`.

Out of curiosity, where did you get this `TCGTCGTCGTCG` from? I have never seen it, I have seen the CPUID for ACRN, Xen, Jailhouse and QEMU but not this one.

junjiemao1:
> > > @gvancuts tool may not run inside a VM, for example, virtual box would fail. This flag is recommended only if user knows what exactly she/he is doing.
> > > I would think if we shall explicitly specify the flag is for qemu only(by far) to prevent out-of-box usage...
> > 
> > 
> > Yes, this is my proposal. Check for `cpuid -l 0x40000000 | grep -c KVMKVMVKM` to be equal or greater than 1 and if so, it means we are running in QEMU/KVM. In that case, allow the script to proceed **and** display a warning so that the user is aware of the special case.
> 
> There are 2 ways to remove the "--no-native-check" flag:
> 
> 1. if we don't care about which hypervisor is now, based on this patch, we could remove the --no-native-check flag and replace the logging.error(f"The board inspector must be run in a native OS environment and not from a VM.")
>    with logging.warning(f"The board inspector is running on top of hypervisor, you should running this on top of QEMU only.") and do not exit. If it fails, it gets sufficient information.
> 2. as Geoffroy proposed, If we want to be precise, cpuid with 0x40000000 would work too.. just extra work to implement a class here. Btw, @gvancuts do you know if in QEMU, cpuid with 0x40000000 would return "" KVMKVMKVM  " or "TCGTCGTCGTCG"?
> 
> @junjiemao1 any comments?

It depends on our expected usage. If KVM (with nested support) is the only target, we can check `` KVMKVMKVM ``. ``TCGTCGTCGTCG`` is the leaf reported by Qemu with KVM disabled (where ``TCG``, i.e. ``Tiny Code Generator``, is the intermediate representation of binary in Qemu for binary translation).

yuchuyang:
@gvancuts  I have removed the flag and replace the error message with warning instead. Now we only support few nested environment, but could be expand in the future. I would like keep it simple, no flag, no exit but a warning message and do not block running board inspector in any condition.

wenlingz:
@yuchuyang please notify me once this PR can be merged, thanks

gvancuts:
> @gvancuts I have removed the flag and replace the error message with warning instead.

Thanks.

> Now we only support few nested environment, but could be expand in the future.

Which ones do we support (beyond KVM)?

> I would like keep it simple, no flag, no exit but a warning message and do not block running board inspector in any condition.

Yes, this is a good approach. But we need to catch the fact we're running in KVM since this is supported, and hence no "unexpected results" should occur.

yuchuyang:
> > @gvancuts I have removed the flag and replace the error message with warning instead.
> 
> Thanks.
> 
> > Now we only support few nested environment, but could be expand in the future.
> 
> Which ones do we support (beyond KVM)?
> 
Not for now except the KVM. I keep the positive thought :)

> > I would like keep it simple, no flag, no exit but a warning message and do not block running board inspector in any condition.
> 
> Yes, this is a good approach. But we need to catch the fact we're running in KVM since this is supported, and hence no "unexpected results" should occur.

KVM is widely used for different hypervisor, for Acrn + qemu/KVM case, it's sufficient but may not always true since qemu/KVM is one of the possibility. To ensure that acrn can work with the board.xml, it should based on its capability, for example: vmx. Based on this statement, I would think do the schema validation while compiling may be sufficient.

yuchuyang:
@wenlingz  Please helps to merge this PR, thanks.

wenlingz:

NanlinXie:
@gvancuts , are you ok for this patch?

gvancuts:
> > > Now we only support few nested environment, but could be expand in the future.
> > 
> > 
> > Which ones do we support (beyond KVM)?
> 
> Not for now except the KVM. I keep the positive thought :)
> 
> > > I would like keep it simple, no flag, no exit but a warning message and do not block running board inspector in any condition.
> > 
> > 
> > Yes, this is a good approach. But we need to catch the fact we're running in KVM since this is supported, and hence no "unexpected results" should occur.
> 
> KVM is widely used for different hypervisor, for Acrn + qemu/KVM case, it's sufficient but may not always true since qemu/KVM is one of the possibility. To ensure that acrn can work with the board.xml, it should based on its capability, for example: vmx. Based on this statement, I would think do the schema validation while compiling may be sufficient.

I am not following you here. What I think we should do is, check if running in a native env:
* if true just generate the `$(BAORD).xml`
* if running in QEMU/KVM, print an indication the tool is collecting the `$(BOARD).xml` for a QEMU/KVM Virtual Machine (CM)
* if running in a different hypervisor, print a warning that this may lead to unpredictable results

gvancuts:
> @gvancuts , are you ok for this patch?

Not yet

yuchuyang:
> > KVM is widely used for different hypervisor, for Acrn + qemu/KVM case, it's sufficient but may not always true since qemu/KVM is one of the possibility. To ensure that acrn can work with the board.xml, it should based on its capability, for example: vmx. Based on this statement, I would think do the schema validation while compiling may be sufficient.
> 
> I am not following you here. What I think we should do is, check if running in a native env:
> 
> * if true just generate the `$(BAORD).xml`
> * if running in QEMU/KVM, print an indication the tool is collecting the `$(BOARD).xml` for a QEMU/KVM Virtual Machine (CM)
> * if running in a different hypervisor, print a warning that this may lead to unpredictable results

If tool run inside the virtualbox(this is not the case we will do using ACRN, but anyone can think out of box), it is running  on top of hypervisor and it is using KVM. However, it doesn't mean Acrn will work in such environment.

gvancuts:
> 
> If tool runs inside the virtualbox(this is not the case we will do using ACRN, but anyone can think out of box), it is running on top of hypervisor and it is using KVM. However, it doesn't mean Acrn will work in such environment.

Ah, I see!! I did not know that running in VirtualBox would return that `KVMKVMKVM` string too.

I would suggest we return a warning then and state in there that only QEMU/KVM is validated.

misc/config_tools/board_inspector/cli.py
```diff
@@ -11,11 +11,23 @@
...
 sys.path.append(os.path.join(script_dir))
 
+def native_check():
+    cpu_ids = get_online_cpu_ids()
+    cpu_id = cpu_ids.pop(0)
+    leaf_1 = parse_cpuid(1, 0, cpu_id)
+    if leaf_1.hypervisor != 0:
+        logging.warning(f"The board inspector is running inside hypervisor. Unexpected result may occur.")
+        logging.warning(f"Only QEMU/KVM environment is valid.")
+
```
yuchuyang, line 26: @gvancuts Any suggestion of improving logging message?

pengfujx:
OK to verify doc

misc/config_tools/board_inspector/cli.py
gvancuts: How about the following text:
```
Board inspector is running inside a Virtual Machine (VM). Running ACRN inside a VM is only supported under KVM/QEMU.
Unexpected results may occur when deviating from that combination.
```

gvancuts (approved):

yuchuyang:
@wenlingz  This PR is ready to merge. Thanks!

wenlingz:
@NanlinXie @terryzouhao please help approve this PR, thanks

wenlingz:

terryzouhao (approved):

PR #6222, opened 2021-06-21
Update Release Note 2.5 with Upgrade Recommendations

junjiemao1:
This PR adds scripts for automatically upgrade scenario XMLs and recommendations to upgrade to ACRN v2.5 from the previous version.


doc/release_notes/release_notes_2.5.rst
```diff
@@ -57,93 +57,67 @@ Upgrading to v2.5 From Previous Releases
...
-.. code-block:: bash
-
-   $ python3 --version
-   Python 3.5.2
-
-Only when the reported version is less than 3.6 (as is the case in the example above) do
-you need an upgrade. The first (and preferred) choice is to install the latest
-Python 3 from the official package repository:
+To use the new board inspector, ensure you have Python >= 3.6 on the target
+board with the package lxml installed. Refer to :ref:`upgrade_python` for
```
gvancuts, line 70:
```suggestion
board with the package ``lxml`` installed. Refer to :ref:`upgrade_python` for
```

doc/release_notes/release_notes_2.5.rst
```diff
@@ -57,93 +57,67 @@ Upgrading to v2.5 From Previous Releases
...
+detailed steps to check and upgrade Python version. The lxml package can be
```
gvancuts, line 71:
```suggestion
detailed steps to check and upgrade Python version. The ``lxml`` package can be
```

doc/release_notes/release_notes_2.5.rst
```diff
@@ -57,93 +57,67 @@ Upgrading to v2.5 From Previous Releases
...
+installed by executing the following command:
 
 .. code-block:: bash
 
-   $ sudo apt install python3
-   ...
-   $ python --version
-   Python 3.8.8
+   $ sudo pip3 install lxml
```
gvancuts, line 76:
```suggestion
   sudo pip3 install lxml
```

doc/release_notes/release_notes_2.5.rst
```diff
@@ -57,93 +57,67 @@ Upgrading to v2.5 From Previous Releases
...
-
-We recommend you update the usage of variables ``BOARD_FILE`` and
-``SCENARIO_FILE``, which are being deprecated,  and ``RELEASE``:
-
- - ``BOARD_FILE`` should be replaced with ``BOARD``. You should not specify
-   ``BOARD`` and ``BOARD_FILE`` at the same time.
- - Similarly, ``SCENARIO_FILE`` should be replaced with ``SCENARIO``.
- - The value of ``RELEASE`` should be either ``y`` (previously was ``1``) or
-   ``n`` (previously was ``0``).
+   $ misc/config_tools/upgrade_scenario_xml <path to scenario XML> v2.4
```
gvancuts, line 109:
```suggestion
   misc/config_tools/upgrade_scenario_xml <path to scenario XML> v2.4
```

doc/release_notes/release_notes_2.5.rst
```diff
@@ -57,93 +57,67 @@ Upgrading to v2.5 From Previous Releases
...
 
-``BOARD_FILE`` and ``SCENARIO_FILE`` can still be used but will take effect
-only if ``BOARD`` and ``SCENARIO`` are not defined. They will be deprecated in
-a future release.
+The script will generate a v2.5 compliant scneario XML by applying the changes
```
gvancuts, line 111:
```suggestion
The script will generate a v2.5 compliant scenario XML by applying the changes
```

doc/release_notes/release_notes_2.5.rst
```diff
@@ -57,93 +57,67 @@ Upgrading to v2.5 From Previous Releases
...
 
-Patches on Generated Sources
-============================
+.. code-block:: bash
 
-The C files generated from board and scenario XML files were removed from the
-repository in v2.4. Instead they will be generated in the build output when
-building the hypervisor. See the instructions in the :ref:`release_notes_2.4`
-release notes for more information.
+   $ misc/config_tools/upgrade_scenario_xml <path to scenario XML> v2.3
```
gvancuts, line 118:
```suggestion
   misc/config_tools/upgrade_scenario_xml <path to scenario XML> v2.3
```

misc/config_tools/upgrade_scenario_xml
```diff
@@ -0,0 +1,65 @@
...
+fi
+
+if [[ ! -f "${file}" ]]; then
+	echo "${file}: No such file"
+	exit 1
+fi
+
+if [[ ! "${ver}" =~ v[0-9]+.[0-9]+ ]]; then
+	echo "${ver}: Invalid version number"
+	echo "Example of valid version number: v2.0"
```
gvancuts, line 26: I think it would be better to indicate a supported version here, i.e. v2.3 or v2.4. I'd go with `v2.3` since this is also what's used in the `usage()` function.
```suggestion
	echo "Example of valid version number: v2.3"
```

misc/config_tools/upgrade_scenario_xml
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+function usage() {
+	echo "Usage: $0 <scenario XML> [<version of the given scenario XML>]"
+	echo "Upgrade the given scenario XML to align with the current version of ACRN."
+	echo ""
+	echo "Example: $0 logical_partition.xml v2.3"
+	echo "    The output file is logical_partition-v2.3.xml"
```
gvancuts, line 8: Does that mean that this new file with extension `-v2.3` is in fact the new file for v2.5?

misc/config_tools/upgrade_scenario_xml
```diff
@@ -0,0 +1,65 @@
...
+file=$1
+ver=$2
+
+if [[ "${file}" = "" ]]; then
+	usage
+	exit 1
+fi
+
+if [[ ! -f "${file}" ]]; then
+	echo "${file}: No such file"
```
gvancuts, line 20: Probably no harm in displaying here also the `usage()`

misc/config_tools/upgrade_scenario_xml
```diff
@@ -0,0 +1,65 @@
...
+	exit 1
+fi
+
+if [[ ! "${ver}" =~ v[0-9]+.[0-9]+ ]]; then
```
gvancuts, line 24: Should we restrict this to 2.3 and above?
```suggestion
if [[ ! "${ver}" =~ v[2-9]+.[3-9]+ ]]; then
```

misc/config_tools/upgrade_scenario_xml
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+function usage() {
+	echo "Usage: $0 <scenario XML> [<version of the given scenario XML>]"
+	echo "Upgrade the given scenario XML to align with the current version of ACRN."
+	echo ""
+	echo "Example: $0 logical_partition.xml v2.3"
```
gvancuts, line 7: Is there any existing version info on the XML file that can help the user determine what version his XML file is from?

misc/config_tools/upgrade_scenario_xml
```diff
@@ -0,0 +1,65 @@
...
+IFS='.' read -ra current_version <<<"${ver:1}"
+xforms_dir=$(dirname ${BASH_SOURCE})/xforms
+find ${xforms_dir} -type d -and -name 'v*.*' | sort -V | while read dir; do
+	target_version_string="${dir##*/v}"
+	IFS='.' read -ra target_version <<<"${target_version_string}"
+	if ((${current_version[0]} < ${target_version[0]} || (${current_version[0]} == ${target_version[0]} && ${current_version[1]} < ${target_version[1]}))); then
+		echo "Upgrade to ${target_version_string}"
+		find ${dir} -name *.xsl | sort | while read xform; do
+			echo -e "\tApplying ${xform}"
+			xsltproc -o ${outfile} ${xform} ${outfile}
```
gvancuts, line 52: I would add a check at the beginning of the script to check that both `xsltproc` and `xmllint` are installed. (`if [[ -x`)

gvancuts: Should we check for any potential error while running this?

misc/config_tools/upgrade_scenario_xml
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
```
gvancuts, line 2: @dbkinder do we need a license header here?

gvancuts (commented):
Any reason why the bash script does not have the `.sh` extension?

doc/release_notes/release_notes_2.5.rst
dbkinder: How about instead of @gvancuts suggestion, change to:
```
board and install the ``lxml`` PyPI package. Refer to :ref:`upgrade_python` for
```

(I did see we've already got ``lxml`` in the list of PyPi packages in our setup instructions in the GSG.)

doc/release_notes/release_notes_2.5.rst
```diff
@@ -57,93 +57,67 @@ Upgrading to v2.5 From Previous Releases
...
-   $ sudo apt install python3
-   ...
-   $ python --version
-   Python 3.8.8
+   $ sudo pip3 install lxml
 
-If this does not get you an appropriate version, you may use the deadsnakes PPA
-(using the instructions below) or build from source yourself.
+.. note::
+   This is not the complete list of tools required to run the board
```
dbkinder, line 79: Let's merge these two sentences into one:
```
Refer to :ref:`acrn_config_workflow` for a complete list of tools required to run the board inspector.
```

doc/release_notes/release_notes_2.5.rst
```diff
@@ -57,93 +57,67 @@ Upgrading to v2.5 From Previous Releases
...
+   inspector. Refer to :ref:`acrn_config_workflow` for a complete guide.
 
-.. code-block:: bash
+With the prerequisites, copying the board inspector (avaliable at
```
dbkinder, line 82: There are lots of files in the board_inspector folder.  Do we need to copy the whole folder to the target?
Also should we tell the user to set a value for the BOARD variable so the instructions work as written, or tweak the instructions to say ``my_board_name`` or something?

Assuming yes on copying the entire folder, change to:
```
With the prerequisites ready, copy the entire board inspector folder from
``misc/config_tools/board_inspector`` to the target board, ``cd`` to that directory
on the target, and run the tool using::

   sudo python3 cli.py $(BOARD) 

This will generate a ``$(BOARD).xml`` in the current working directory.
```

I doesn't seem intuitive that the board inspector tool is called ``cli.py`` instead of something like ``board_inspector.py``.

misc/config_tools/upgrade_scenario_xml
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
dbkinder, line 1: yep
```
#!/bin/bash
# Copyright (C) 2021 Intel Corporation.
```

misc/config_tools/upgrade_scenario_xml
dbkinder: calling the updated XML file ``-v2.3`` is not intuitive.  Also, the code in the script seems to use a computed "latest_ver_path" for the extension, not the name given on the command line?

misc/config_tools/upgrade_scenario_xml
dbkinder: Why do you need to know the old XML file ACRN version? And no, there's no indication in the XML file which version of ACRN (or the tool) was used to create it.

misc/config_tools/upgrade_scenario_xml
dbkinder: Nice try.  This pattern change though would not allow v3.1
Have you verified the pattern match works?

doc/release_notes/release_notes_2.5.rst
```diff
@@ -57,93 +57,67 @@ Upgrading to v2.5 From Previous Releases
...
-building the hypervisor. See the instructions in the :ref:`release_notes_2.4`
-release notes for more information.
+   $ misc/config_tools/upgrade_scenario_xml <path to scenario XML> v2.3
 
-Modifying generated files is not a recommended practice.
-If you find a configuration that is not flexible enough to meet your
-needs, please let us know by sending mail to `the acrn-dev mailing
-list <https://lists.projectacrn.org/g/acrn-dev>`_ or submitting a
-`GitHub issue <https://github.com/projectacrn/acrn-hypervisor/issues>`_.
+Upgrades from scenario XMLs prior to v2.3 is not supported.
```
dbkinder, line 120: This seems a rather important note, and should be put at the front of this section not at the end.

dbkinder (changes requested):

dbkinder (changes requested):
You should separate the release notes document change, from the addition of all the XML transformations and script.  The latter isn't part of the release notes.

wenlingz:
@junjiemao1 this PR also includes code part, could you help add the Tracked-On ? Thanks

doc/release_notes/release_notes_2.5.rst
junjiemao1: Will follow both suggestions in v2.

doc/release_notes/release_notes_2.5.rst
junjiemao1: Will update in v2.

doc/release_notes/release_notes_2.5.rst
junjiemao1: Typically I use ``$`` to indicate that the line is a command to be executed. May I know the overall guideline on writing such snippets?

doc/release_notes/release_notes_2.5.rst
junjiemao1: Will fix in v2.

misc/config_tools/upgrade_scenario_xml
junjiemao1: Will update in v2.

misc/config_tools/upgrade_scenario_xml
junjiemao1: Yes, the generated file will have the latest version as the postfix. Will fix in v2.

misc/config_tools/upgrade_scenario_xml
junjiemao1: Sure

misc/config_tools/upgrade_scenario_xml
junjiemao1: I'll keep this regex for generic versions and use another check to ensure the given version is at least v2.3.

misc/config_tools/upgrade_scenario_xml
junjiemao1: The version of the old XML is essential as it decides the set of XML transformations to be applied.

For example, assume in v2.4 we added `hv/FEATURES/PSRAM` and in v2.5 we rename it to `hv/FEATURES/SSRAM`.  There will be two transformations (in sequence):

1. Adding a `PSRAM` node under `hv/FEATURES` (if it does not exist in the old XML).
1. Renaming `hv/FEATURES/PSRAM` to `hv/FEATURES/SSRAM`.

If a given XML is already v2.5-compliant, reapplying the transformations will introduce another `hv/FEATURE/SSRAM` node, leading to schema validation failure.

misc/config_tools/upgrade_scenario_xml
junjiemao1: We should. Will add in v2.

misc/config_tools/upgrade_scenario_xml
junjiemao1: Will add both checks in v2.

doc/release_notes/release_notes_2.5.rst
```diff
@@ -57,93 +57,67 @@ Upgrading to v2.5 From Previous Releases
...
-=======================
+Generate New Board XML
+======================
 
-Python version 3.6 or higher is required to build ACRN v2.5. You can check the version of
-Python you are using by:
+Board XML files, generated by ACRN board inspector, contain board information
+that is essential to build ACRN. Compared to previous versions, ACRN v2.5
+extends the schema of board XMLs to summarize board information more
+systematically. It is thus required to regenerate the board XML using the new
```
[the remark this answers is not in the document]
junjiemao1: Will update in v2.

doc/release_notes/release_notes_2.5.rst
junjiemao1: Will update in v2.

doc/release_notes/release_notes_2.5.rst
junjiemao1: Yes, the whole folder needs to be copied. Will update the instructions in v2.

As for the name of ``cli.py``, Geoffroy and I had some discussions at https://github.com/projectacrn/acrn-hypervisor/pull/6213#issuecomment-862125944

misc/config_tools/upgrade_scenario_xml
junjiemao1: Will add in v2.

doc/release_notes/release_notes_2.5.rst
junjiemao1: I tried to shorten the description of the usage of the script a bit in v2, but find no way to put this note at the front where the script itself is not introduced. I used a note to emphasis it from normal paragraphs instead.

junjiemao1:
@gvancuts @dbkinder I've updated the PR to integrate your suggestions.

As for whether we want to keep the scripts in the PR, let's discuss offline.

doc/release_notes/release_notes_2.5.rst
gvancuts: > Typically I use `$` to indicate that the line is a command to be executed. May I know the overall guideline on writing such snippets?

This is also what we've done for a long time. The problem with this approach is that people cannot copy-paste the instructions as it also copies the `$` symbol. I do not believe we have documented this anywhere but we tend to only write the command without the leading prompt (it being `$` or `#`).

I believe that in the Zephyr Project, the HTML content that gets generated gets such prompt added to those code blocks automatically, and in a way that they do not get copied when people copy-paste them.

@dbkinder can correct me or provide more details on that I guess ;-)

dbkinder (approved):
Thanks! There are a few formatting/layout errors I'll fix in a subsequent PR.

Issue #6225, opened 2021-06-21
Updates (tweaks) to the GSG


gvancuts:
There are a number of tweaks to the GSG that would be welcome:
* Simplify some instruction sequences
* Rename the doc to simply be "Getting Started Guide"

PR #6226, opened 2021-06-21
doc: updates to the Getting Started Guide

gvancuts:
Updates to the Getting Started Guide:
* Update title to simply be "Getting Started Guide"
* Simplify and remove instructions that are redundant
* Add a note explaining the difference between 'nuc11tnbi5' and
  'nuc11tnhi5'

Tracked-On: #6225

### DIFF
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -1,11 +1,22 @@
 .. _rt_industry_ubuntu_setup:
 
-Getting Started Guide for ACRN Industry Scenario With Ubuntu Service VM
-#######################################################################
+Getting Started Guide
+#####################
 
 .. contents::
    :local:
    :depth: 1
+
+Introduction
+************
+
+This document describes the various steps to set up a system based on the following components:
+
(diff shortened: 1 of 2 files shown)


doc/getting-started/getting-started.rst
```diff
@@ -1,12 +1,23 @@
 .. _rt_industry_ubuntu_setup:
 
-Getting Started Guide for ACRN Industry Scenario With Ubuntu Service VM
-#######################################################################
+Getting Started Guide
```
dbkinder, line 3: While changing the title is a good idea, changing the filename introduces a problem for external links to the ACRN documentation.  It takes a while for the google search indexer to catch up with this change, and until then, searches will link to the old file name.  Similarly, links from external websites to a specific page will break if we change the filename.  We do have a workaround though, you can add a line in the conf.py to build a redirect from the old filename to the new one.  I'll do that.

dbkinder (approved):
OK.  I'll add the conf.py setting to create a redirect from the old filename to the new...

Issue #5692, opened 2021-02-02
Keep Open: Used when updating config option documentation in .xsd files


dbkinder:
Documentation for ACRN configuration options is now maintained in ``misc/config_tools/schema/*.xsd`` files.  This Jira issue (ACRN-6954) and GitHub issue (5692) are open for reference in GitHub for making edits to the documentation within these files so they're not caught by CI checks as an undocumented code change.

Use ``Tagged-On: #5692`` in PRs that are doc-only changes but also touch one of the ``.xsd`` files to satisfy the CI check.

dbkinder:
[External_System_ID] ACRN-6954

NanlinXie:
Closed with no regression issue.

PR #6243, opened 2021-06-23
doc: fix documentation's issue

Kunhui-Li:
1. Update the necessary libraries to consistent with the "Build ACRN From Source"
document in the "Getting Started Guide" document.
2. Delete the related introduction with acrngt.conf and launch_uos_id1.sh files in
"Getting Started Guide" document.
3. Update WHL-IPC-I7 board's processor in Supported HW document.
4. Add cpu_affinity element's description in ACRN Configuration Data.
5. Update the description for shm_region in Launch XML format.
6. Update configurable/readonly attributes values.
7. Update the description for hv.CAPACITIES.MAX_MSIX_TABLE_NUM in schema/config.xsd.


dbkinder (commented):
LGTM, but needs a ``Tracked-On`` tag to satisfy CI check (because of the "code" change to the ``.xsd`` file.

misc/config_tools/schema/config.xsd
```diff
@@ -273,7 +273,8 @@ devices.</xs:documentation>
     </xs:element>
     <xs:element name="MAX_MSIX_TABLE_NUM" default="64">
         <xs:annotation>
-          <xs:documentation>Maximum number of MSI-X tables per device.</xs:documentation>
+          <xs:documentation>Maximum number of MSI-X tables per device,
+if this value is empty, then the default value would be calculated from board xml file.</xs:documentation>
```
dbkinder, line 277: If this statement is correct, the ``default="64"`` must be removed from the line above.  (That's what's showing up in the generated documentation.)

dbkinder (changes requested):
Thanks for the doc update.  One comment needs addressing and this needs a "Tracked-On" tag because CI thinks you're editing code (the ``.xsd`` file).  You can use ``Tracked-On: #5692`` that I keep around for this purpose.

misc/config_tools/schema/config.xsd
NanlinXie: We still suggest the default value is 64, but user could choose to let config tool calculate based on board.xml.

dbkinder: The doc process adds the value of ``default=value`` to the generated documentation. This will be confusing if the documentation has two default value statements for this option.  With the proposed change, the documentation will read:

- Maximum number of MSI-X tables per device, if this value is empty, then the default value
   would be calculated from board xml file. (Default value is 64)

fitchbe: "Maximum number of MSI-X tables per device. If this value is empty, the default value **will** be calculated from the board XML file. (The default value is 64.)"

misc/config_tools/schema/config.xsd
```diff
@@ -273,7 +273,8 @@ devices.</xs:documentation>
     </xs:element>
     <xs:element name="MAX_MSIX_TABLE_NUM" default="64">
         <xs:annotation>
-          <xs:documentation>Maximum number of MSI-X tables per device.</xs:documentation>
+          <xs:documentation>Maximum number of MSI-X tables per device,
```
dbkinder, line 276:
```suggestion
          <xs:documentation>Maximum number of MSI-X tables per device.
```

misc/config_tools/schema/config.xsd
dbkinder:
```suggestion
If this value is empty, then the default value will be calculated from the board XML file.</xs:documentation>
```

dbkinder (changes requested):

misc/config_tools/schema/config.xsd
Kunhui-Li: Updated.

misc/config_tools/schema/config.xsd
Kunhui-Li: Updated.

dbkinder (approved):
LGTM, thanks!

Issue #6312, opened 2021-07-21
hv: debug: enable hv console when lapic-pt is enabled on BSP


yfliuuu:
In some scenarios (e.g., nested) where lapic-pt is enabled for a vcpu running on a pcpu hosting console timer, the hv console will be inaccessible. The console can be re-enabled by hooking the callback into every VM-exit event.

yfliuuu:
[External_System_ID] ACRN-7305

NanlinXie:
Closed since patch merged without regression.

PR #6313, opened 2021-07-21
hv: debug: Add hv console callback to VM-exit event

yfliuuu:
In some scenarios (e.g., nested) where lapic-pt is enabled for a vcpu
running on a pcpu hosting console timer, the hv console will be
inaccessible.

This patch adds the console callback to every VM-exit event so that the
console can still be somewhat functional under such circumstance.

Since this is VM-exit driven, the VM-exit/second can be low in certain
cases (e.g., idle or running stress workload). In extreme cases where
the guest panics/hangs, there will be no VM-exits at all.

In most cases, the shell is laggy but functional (probably enough for
debugging purpose).

Tracked-On: #6312

### DIFF
--- a/hypervisor/arch/x86/guest/vmexit.c
+++ b/hypervisor/arch/x86/guest/vmexit.c
@@ -22,6 +22,7 @@
 #include <asm/guest/vcpuid.h>
 #include <trace.h>
 #include <asm/rtcm.h>
+#include <debug/console.h>
 
 /*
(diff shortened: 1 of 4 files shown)


dongyaozu (commented):
Did u compile release version?

yfliuuu:
@dongyaozu Yes. This patch contains an empty function console_vmexit_callback in release/console.c

dongyaozu (approved):

PR #6317, opened 2021-07-23
Revert uc lock

taoyuhong:


wenlingz:
Because CI hasn't add ADL-S platform, so didn't catch https://jira.devtools.intel.com/browse/ACRN-7306 this issue---ADL-S SOS bootup fail. So need to revert patches now. We will online ADL-S next week. Thanks for your understanding.

Issue #6334, opened 2021-07-30
hv console does not support 64bit bar PCI uart with 32bit bar space


fuzhongl:
**Describe the bug**
Currently, ACRN console does not support 64bit bar PCI uart with 32bit bar space

**Platform**
EHL

**Codebase**
V2.5

**Scenario**
Industry

fuzhongl:
[External_System_ID] ACRN-7318

PR #6336, opened 2021-07-30
HV: debug: support 64bit BAR pci uart with 32bit space

jsun26intel:
Currently the HV console does not support PCI UART with 64bit BAR, but in the
case that the BAR is in 64bit and the BAR space is below 4GB (i.e. the high
32bit address of the 64bit BAR is zero), HV should be able to support it.

Tracked-On: #6334

### DIFF
--- a/hypervisor/debug/uart16550.c
+++ b/hypervisor/debug/uart16550.c
@@ -155,11 +155,15 @@ void uart16550_init(bool early_boot)
 				uart.reg_width = 1;
 				pci_pdev_write_cfg(uart.bdf, PCIR_COMMAND, 2U, cmd | PCIM_CMD_PORTEN);
 			} else {
-				if ((bar0 & 0xfU) == 0U) { /* 32 bits MMIO Space */
+				uint32_t bar_hi = pci_pdev_read_cfg(uart.bdf, pci_bar_offset(1), 4U);
+
+				/* Enable the PCI UART if the BAR is 32bit, or 64bit with 4GB- mmio space. */
+				if (((bar0 & 0x7U) == 0U) || (((bar0 & 0x7U) == 4U) && (bar_hi == 0U))) {
 					uart.type = MMIO;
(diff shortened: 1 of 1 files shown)


wenlingz:
@JasonChenCJ please help approve this PR, thanks

hypervisor/debug/uart16550.c
```diff
@@ -155,7 +155,9 @@ void uart16550_init(bool early_boot)
 				uart.reg_width = 1;
 				pci_pdev_write_cfg(uart.bdf, PCIR_COMMAND, 2U, cmd | PCIM_CMD_PORTEN);
 			} else {
-				if ((bar0 & 0xfU) == 0U) { /* 32 bits MMIO Space */
+				uint32_t bar_hi = pci_pdev_read_cfg(uart.bdf, pci_bar_offset(1), 4U);
+
+				if (((bar0 & 0xfU) == 0U) || (((bar0 & 0xfU) == 4U) && (bar_hi == 0U))) {
```
JasonChenCJ, line 160: do we really need consider the whole low 4 bits, should we just consider 3 bits (remove bit for prefetchable)?

dongyaozu: +1. No need to checkk bit 3.

dongyaozu: BTW, why we cannot support pure 64 bits MMIO serial card?

jsun26intel: @dongyaozu This is because we didn't do memory remap with 4G+ memory space in uart initialization. I will add a "TODO" in updated version and implement in the future. This PR is a quick fix for the latest EHL SBL update.

dongyaozu: Need to add comments in this place too. Otherwise, people will be confused why we do so.

dongyaozu (approved):

Issue #6340, opened 2021-07-30
[EF]Invalid LPC entry prevents GOP driver from working properly in WaaG for DP3


Kunhui-Li:
In recent WaaG launch scripts the LPC bus PCI device is configured incorrectly as "-s 31:0,lpc". (original value was "-s 1:0,lpc")
This causes the OVMF to fails running the GOP driver properly for GVT-d.

Kunhui-Li:
[External_System_ID] ACRN-7313

PR #6341, opened 2021-07-30
config_tools: update lpc slot number

Kunhui-Li:
Update lpc slot to origin value 1 from 31 because GOP driver has assumption
to config space layout of the device on 00:1f.0.

Tracked-On: #6340


NanlinXie:
@Kunhui-Li , could you confirm your validation result?

NanlinXie (approved):

Issue #6345, opened 2021-08-02
hv: SMBIOS passthrough: UEFI environment checks should be added before probing


yfliuuu:
Previously SMBIOS passthrough assumes UEFI environment but this may cause problems on non-UEFI environment. Checks should be added before probing SMBIOS table.

yfliuuu:
[External_System_ID] ACRN-6527

NanlinXie:
Closed since patch merged without regression.

PR #6346, opened 2021-08-02
hv: SMBIOS: add check before SMBIOS probing

yfliuuu:
The SMBIOS passthrough is supposed to work under UEFI environment.
Previous patches assumes UEFI environment but it may cause problems on
some platforms. This patch adds checking before probing SMBIOS table to
avoid this problem.

Tracked-On: #6345


wenlingz:

fyin1 (approved):

PR #6352, opened 2021-08-05
Manage PCI INTx allocation to pre-launched VMs in general

junjiemao1:
Today we have seen a couple of requirements on passing through non-MSI-capable PCI devices to pre-launched VMs:

  * The GPIO controller, which is not a PCI device at all but identified in
    the ACPI namespace

  * The SPI controller, which is a non-MSI-capable PCI device

  * The eMMC controller, which is a non-MSI-capable PCI device

Now ACRN is capable of assigning specific interrupt lines to a pre-launched VM, but with the following shortcomings:

  * Users are totally on their own to make sure that the assigned interrupt
    line is not shared among devices passed through to different
    VMs. Assigning an interrupt line which can be triggered by another
    device in another VM allows other VMs to impact the timing of the
    pre-launched VM.

  * The VM has no way to identify the interrupt lines connected to INTx
    (i.e. INTA# to INTD#) of a vPCI device. This prevents modern drivers
    from working properly.

This patch series aims at resolving the above two issues by adding the following steps in the current build process.

  1. When collecting board information, interpret the PCI routing tables in
     DSDT/SSDT to collect:
       a. the mapping from PCI INTx to interrupt lines
       b. the INTx used by each PCI device.

  2. During compilation, figure out the interrupt lines used by
     non-MSI-capable PCI devices that are passed through to a pre-launched
     VMs and report failure when any interrupt line is shared between two
     pre-launched VMs or between a pre-launched VM and the service
     VM. Interrupt lines explicitly assigned in scenario XMLs are
     considered in this stage, too.

  3. If there is no conflict in interrupt line allocation, the
     configuration tools generate a virtual DSDT that includes the
     following information:

       - A device object representing the PCI root bus, encoding its device
         identification information, current resources descriptors (which
         contains the PCI hole ranges) and PCI routing table of virtual PCI
         devices.

       - One device object for each pass-through PCI device, if the device
         has an accompanied device object in the physical DSDT/SSDT. This
         may include device identification and current resource
         descriptors.

     The pt_intx.c is updated accordingly to include the interrupt line
     mappings for vPCI INTx as well.

  4. Required device objects in the virtual ACPI (vACPI) namespace are now
     automatically identified and generated based on the physical ACPI
     namespace. Only a selected set of standard device objects will be
     included in the vACPI namespace, mainly for device identification.

This patch series consists of three main parts:

  1. Patch 01 - 13 are fixes to existing bugs in the board inspector,
     mainly in parsing and interpreting AML code.

  2. Patch 14 - 33 adds facilities to the board inspector for manipulating
     ACPI AML code and PCI routing tables. They also updates the generation
     of board XML files to include the information above.

  3. Patch 34 - 37 modifies the configuration tool to implement the
     interrupt line conflict detection and allocation logic and the
     generation of vDSDT.

Opens:

  1. The interrupt line conflict detection does not take into account the
     interrupt pins of MSI-capable devices, assuming that the guest OSes
     always use MSI as the preferred interrupt mechanism.

v1 -> v2:

  * Add a tree visitor which collect dependencies among devices due to
    operation region accesses. Such dependencies will be generated to board
    XMLs and used by the configuration toolset to extract the ACPI device
    objects that should be filled into the vACPI tables of a pre-launched
    VM.

  * The static allocator will now also complain if any interrupt line is
    allocated to a VM with LAPIC_PASSTHROUGH.

  * In addition to the accompanying device objects of PCI devices, also add
    TPM device objects to vACPI tables if TPM is assigned to the
    corresponding VM.

  * The old vDSDT generation routine and the static DSDT sources are
    replaced with the new vDSDT generation logic.

  * The layout of vACPI tables is modified a bit to allow larger vDSDT.

  * A couple of fixes and minor cleanups to the previous series, including:

    - Resume a not-in-D0 PCI device before parsing its configuration space.
    - Allow a ConstObj term to be used as a TermObj.
    - Export the PATH environment variable to the legacy board parser.


wenlingz:
try verify EHL

wenlingz:
OK to verify

wenlingz:
OK to verify

wenlingz:

NanlinXie (approved):

Issue #6355, opened 2021-08-05
ACRN build improvement


liuhang-bit:
User could use make targz-pkg command to generate tar package in
build directory,which could help user simplify the process
of installing acrn hypervisor in target board. user need to copy the
tarball package to target board,and extract it to "/" directory.

liuhang-bit:
[External_System_ID] ACRN-7322

NanlinXie:
Closed since patch merged without regression.

PR #6356, opened 2021-08-05
Makefile:add targz-pkg entry in Makefile

liuhang-bit:
User could use make targz-pkg command to generate tar package in
build directory,which could help user simplify the process
of installing acrn hypervisor in target board. user need to copy the
tarball package to target board,and extract it to "/" directory.

Tracked-On: #6355
Signed-off-by: liu hang1 <hang1.liu@intel.com>
Reviewed-by: VanCutsem, Geoffroy <geoffroy.vancutsem@intel.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,10 @@ SHELL := /bin/bash
 # global helper variables
 T := $(CURDIR)
 
+# ACRN Version Information
+include VERSION
+export FULL_VERSION=$(MAJOR_VERSION).$(MINOR_VERSION)$(EXTRA_VERSION)
+
 ifdef TARGET_DIR
   $(warning TARGET_DIR is obsoleted because generated configuration files are now stored in the build directory)
 endif
@@ -69,6 +73,7 @@ HV_OUT := $(ROOT_OUT)/hypervisor
 DM_OUT := $(ROOT_OUT)/devicemodel
 TOOLS_OUT := $(ROOT_OUT)/misc
 DOC_OUT := $(ROOT_OUT)/doc
+TARBALL_OUT := $(ROOT_OUT)/tarball
 BUILD_VERSION ?=
 BUILD_TAG ?=
 HV_CFG_LOG = $(HV_OUT)/cfg.log
@@ -175,3 +180,9 @@ tools-install:
 
 life_mngr-install:
 	$(MAKE) -C $(T)/misc OUT_DIR=$(TOOLS_OUT) RELEASE=$(RELEASE) acrn-life-mngr-install
+
+.PHONY: targz-pkg
+targz-pkg:
+	$(MAKE) install DESTDIR=$(TARBALL_OUT)
+	cd $(TARBALL_OUT) && \
+	tar -zcvf $(ROOT_OUT)/acrn-$(FULL_VERSION).tar.gz *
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -6,6 +6,11 @@ SHELL := /bin/bash
 # global helper variables
 T := $(CURDIR)
 
+# ACRN Version Information
+include VERSION
+export
+FULL_VERSION=$(MAJOR_VERSION).$(MINOR_VERSION)$(EXTRA_VERSION)
```
gvancuts, line 12: This should be on the same line as `export`.

gvancuts (changes requested):

wenlingz:
OK to verify

wenlingz:
try verify EHL

wenlingz:
OK to verify doc

wenlingz:
try verify ldra

gvancuts (approved):

Makefile
```diff
@@ -175,3 +180,9 @@ tools-install:
 
 life_mngr-install:
 	$(MAKE) -C $(T)/misc OUT_DIR=$(TOOLS_OUT) RELEASE=$(RELEASE) acrn-life-mngr-install
+
+.PHONY: targz-pkg
```
gvancuts, line 184:
```suggestion
.PHONY: targz-pkg
```

Makefile
```diff
@@ -175,3 +180,9 @@ tools-install:
 
 life_mngr-install:
 	$(MAKE) -C $(T)/misc OUT_DIR=$(TOOLS_OUT) RELEASE=$(RELEASE) acrn-life-mngr-install
+
+.PHONY: targz-pkg 
+targz-pkg:
+	$(MAKE) install DESTDIR=$(TARBALL_OUT)
+	cd $(TARBALL_OUT) && \
+	tar -zcvf $(ROOT_OUT)/acrn-$(FULL_VERSION).tar.gz *
```
gvancuts, line 188:
```suggestion
	tar -zcvf $(ROOT_OUT)/acrn-$(FULL_VERSION).tar.gz *
```

gvancuts:
One thing I've noticed that seems odd to me, is that `make install` seems to re-trigger a lot of build operations. Is this expected?

Take the following sequence:
```
make BOARD=nuc7i7dnb SCENARIO=industry
make targz-pkg
```
I would expect that running `make targz-pkg` would just create the tarball out of the build artefacts that were created in the previous step. But instead, it recompiles a bunch of stuff again, is it normal?

/cc @junjiemao1

gvancuts (changes requested):

NanlinXie (approved):

NanlinXie:
> One thing I've noticed that seems odd to me, is that `make install` seems to re-trigger a lot of build operations. Is this expected?
> 
> Take the following sequence:
> 
> ```
> make BOARD=nuc7i7dnb SCENARIO=industry
> make targz-pkg
> ```
> 
> I would expect that running `make targz-pkg` would just create the tarball out of the build artefacts that were created in the previous step. But instead, it recompiles a bunch of stuff again, is it normal?
> 
> /cc @junjiemao1

We root caused, but we need do more validation to limit the impact. @liuhang-bit  will file PR later.
Created https://github.com/projectacrn/acrn-hypervisor/issues/6365 to tracked.
Let's merge this change firstly.

PR #6359, opened 2021-08-05
doc: update Kconfig related doc

Kunhui-Li:
Remove Kconfig related scheduler configuration in tutorials because the
Kconfig related files have been removed in the other PR 6358.

### DIFF
--- a/doc/tutorials/cpu_sharing.rst
+++ b/doc/tutorials/cpu_sharing.rst
@@ -128,26 +128,11 @@ and BVT (Borrowed Virtual Time) scheduler.
 
 Scheduler configuration
 
+* The scheduler used at runtime is defined in the scenario XML file
+  via the :option:`hv.FEATURES.SCHEDULER` option. The default scheduler
+  is **SCHED_BVT**. Use the :ref:`ACRN configuration tool <acrn_configuration_tool>` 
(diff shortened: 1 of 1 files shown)


doc/tutorials/cpu_sharing.rst
```diff
@@ -126,29 +126,6 @@ and BVT (Borrowed Virtual Time) scheduler.
     timesharing.
 
 
-Scheduler configuration
```
dbkinder, line 129: Need to add how to configure the scheduler instead of using Kconfig, something like:
```
The scheduler used at runtime is defined in the scenario XML file
via the :option:`hv.FEATURES.SCHEDULER` option.
The default scheduler is **SCHED_BVT**.
```

dbkinder (changes requested):
Ok to remove the use of kconfig, but need to add how to configure the scheduler option.

doc/tutorials/cpu_sharing.rst
dbkinder: @Kunhui-Li  Can you make this documentation change?

Kunhui-Li: Sorry for my later reply, updated it, thanks.

dbkinder (approved):

PR #6372, opened 2021-08-10
HV: two minor refine to support running ACRN on qemu

lifeix:
Before supporting to run ACRN on qemu, ACRN always runs on the hardware platforms.
In order to achieve better performance, there're some constrains about the hardware
platforms, like the hardware platforms must support XSAVE Feature or the VTD mmio read
must be quickly and so on. However, these constrains are not met on most of the virtual
platforms which QEMU emulated. This patch set does two minor refine to support running
ACRN on qemu:
1. Check whether condition is met before check whether time is out after iommu_read32.
    This is because iommu_read32 would cause time out on some virtual platform in
    spite of the current DMAR status meets the pre_condition.
2. ACRN could run without XSAVE Capability. So remove XSAVE dependence to support
    more (hardware or virtual) platforms.

V2:
  add X86_FEATURE_COMPACTION_EXT dependence back.

V3:
  1. Inject UD when guest try to call xsetbv when guest doesn't support XSAVE.
  2. Assume the platform must both support XSAVE Capability on CPU side on XSAVE_CTL
     on VMX side or not, in this case, we could pass through CR4.OSXSAVE bit.

v4
 1. revert vCPUID change
 2. for vsetbv: (a) inject GP when the current privilege level is not 0.
                (b) injecy UD when CR4.OSXSAVE[bit 18] = 0
 3. rename check_xsave_cap to check_xsave_combined_cap_valid

Tracked-On: #6287

### DIFF
--- a/hypervisor/arch/x86/cpu_caps.c
+++ b/hypervisor/arch/x86/cpu_caps.c
@@ -415,6 +415,26 @@ static inline bool pcpu_has_vmx_unrestricted_guest_cap(void)
 	return ((msr_read(MSR_IA32_VMX_MISC) & MSR_IA32_MISC_UNRESTRICTED_GUEST) != 0UL);
 }
 
+static bool is_valid_xsave_combination(void)
+{
+	uint64_t value64 = msr_read(MSR_IA32_VMX_PROCBASED_CTLS2);
+	uint32_t high = (uint32_t)(value64 >> 32U);	/* allowed 1-settings */
+	bool ret;
+
+	/* Now we only assume the platform must support XSAVE on CPU side and XSVE_XRSTR on VMX side or not,
+	 * in this case, we could pass through CR4.OSXSAVE bit.
+	 */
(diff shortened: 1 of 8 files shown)


hypervisor/arch/x86/cpu_caps.c
```diff
@@ -415,6 +415,26 @@ static inline bool pcpu_has_vmx_unrestricted_guest_cap(void)
 	return ((msr_read(MSR_IA32_VMX_MISC) & MSR_IA32_MISC_UNRESTRICTED_GUEST) != 0UL);
 }
 
+static bool check_xsave_combined_cap_valid(void)
```
dongyaozu, line 418: Change the name. too long

dongyaozu (changes requested):

hypervisor/arch/x86/cpu_caps.c
lifeix: done

dongyaozu (approved):

PR #6394, opened 2021-08-12
doc:add more explanation on parameter --windows

liuhang-bit:
acrn-dm --windows parameter is used for specify
the Windows ORACLE virtio device driver,which is
used when enable secure boot in Windows, otherwise
the default REDHAT virtio device driver will be
used which is not certified by Microsoft

Signed-off-by: Liu, Hang1 <hang1.liu@intel.com>


fitchbe (changes requested):

dbkinder (approved):
@fitchbe I directly updated @liuhang-bit  PR with your review comments. LGTM

Issue #6387, opened 2021-08-12
enable GVT-d for pre-launched linux guest


liujunming:
Pass-thru GPU to pre-launched Linux guest in logical partition mode

liujunming:
[External_System_ID] ACRN-7337

PR #6395, opened 2021-08-12
hv:enable GVT-d for pre-launched linux guest in logical partion mode

liujunming:
When pass-thru GPU to pre-launched Linux guest,
need to pass GPU OpRegion to the guest.
Here's the detailed steps:
1. reserve a memory region in ve820 table for GPU OpRegion
2. build EPT mapping for GPU OpRegion to pass-thru OpRegion to guest
3. emulate the pci config register for OpRegion
For the third step, here's detailed description:
The address of OpRegion locates on PCI config space offset 0xFC,
Normal Linux guest won't write this register,
so we can regard this register as read-only.
When guest reads this register, return the emulated value.
When guest writes this register, ignore the operation.

Tracked-On: #6387

Signed-off-by: Liu,Junming <junming.liu@intel.com>

### DIFF
--- a/hypervisor/include/common/ptdev.h
+++ b/hypervisor/include/common/ptdev.h
@@ -19,6 +19,11 @@ enum intx_ctlr {
 #define PTDEV_INTR_MSI		(1U << 0U)
 #define PTDEV_INTR_INTX		(1U << 1U)
 
+#define GPU_OPREGION_SIZE	0x5000U
+#define GPU_OPREGION_GPA	0x40880000U
+#define PCIR_ASLS_CTL		0xfcU /* register offset in PCIe configuration space for Opregion base address */
+#define PCIM_ASLS_OPREGION_MASK	0xfffff000U /* opregion need 4KB aligned */
+
 #define INVALID_PTDEV_ENTRY_ID 0xffffU
 
 #define DEFINE_MSI_SID(name, a, b)	\
--- a/hypervisor/arch/x86/guest/ve820.c
+++ b/hypervisor/arch/x86/guest/ve820.c
@@ -12,10 +12,11 @@
 #include <vacpi.h>
 #include <logmsg.h>
 #include <asm/rtcm.h>
+#include <ptdev.h>
 
 #define ENTRY_HPA1_LOW_PART1	2U
-#define ENTRY_HPA1_LOW_PART2	4U
-#define ENTRY_HPA1_HI		8U
+#define ENTRY_HPA1_LOW_PART2	5U
+#define ENTRY_HPA1_HI		9U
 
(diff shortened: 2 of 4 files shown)


JasonChenCJ (approved):

hypervisor/include/common/ptdev.h
```diff
@@ -19,6 +19,11 @@ enum intx_ctlr {
 #define PTDEV_INTR_MSI		(1U << 0U)
 #define PTDEV_INTR_INTX		(1U << 1U)
 
+#define GPU_OPREGION_SIZE	0x5000U
+#define GPU_OPREGION_GPA	0x40880000U
+#define PCIR_ASLS_CTL		0xfcU /* Opregion start addr register */
```
dongyaozu, line 24: register offset in PCIe configuration space for Opregion base address.

Will this be better?
And what does ASLS mean?

hypervisor/include/common/ptdev.h
```diff
@@ -19,6 +19,11 @@ enum intx_ctlr {
...
+#define PCIM_ASLS_OPREGION_MASK	0xfffff000U /* opregion need 4KB aligned */
```
dongyaozu, line 25: Why this is related to PCIM_ ?

hypervisor/include/common/ptdev.h
liujunming: Hi, Eddie. The comment `register offset in PCIe configuration space for Opregion base address.` is better, I will refine the comments.

ASLS stands for "ASL Storage" from Bspec. https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/i915/i915_reg.h#L347 Linux uses `ASLS` too.

hypervisor/include/common/ptdev.h
liujunming: Hi, Eddie. I think it's memory-related, so I use `PCIM_`, Maybe I use the wrong prefix, any suggestion for the name? It stands for the page offset for Opregion base address, I will refine the comment too.

hypervisor/include/common/ptdev.h
dongyaozu: Make an explicit comment for readers to easily understand. We can assume ASL is a known term.

hypervisor/include/common/ptdev.h
dongyaozu: By checking all the existing PCIM_XXX macros, It seems to be confusing... But I am fine to take for now.

hypervisor/include/common/ptdev.h
liujunming: OK, will refine the comment.

dongyaozu (approved):

PR #6404, opened 2021-08-16
doc: update hld-devicemodel with current e820 mem map

liuhang-bit:
Default e820 mem map in hld-devicemodel.rst is
mismatched with sw_load_common.c,update the doc
to align with code.

Signed-off-by: Liu, Hang1 <hang1.liu@intel.com>


dbkinder (approved):
While this does update the documentation (thanks), the correct way to include this would be via doxygen comments that would appear in the API documentation.  Alternatively, we could use an include to pull this comment from the source file into the documentation.  Either of these would keep the documentation in sync with the code vs. duplicating the comment in the docs (that's how it got out of sync in the first place).

dbkinder:
@NanlinXie @wenlingz Can we ignore this YOCTO build failure and merge the doc-only change?

wenlingz:
> @NanlinXie @wenlingz Can we ignore this YOCTO build failure and merge the doc-only change?

please

PR #6430, opened 2021-08-18
hv: optimize handling of IA32_EFER

ZideChen0:
With these 2 patches, the CPUID roundtrip tests on TGL reduces from 
2895 TSC cycles to 2835 cycles.

### DIFF
--- a/hypervisor/arch/x86/guest/vmcs.c
+++ b/hypervisor/arch/x86/guest/vmcs.c
@@ -481,12 +481,8 @@ static void init_entry_ctrl(const struct acrn_vcpu *vcpu)
 	/* Log messages to show initializing VMX entry controls */
 	pr_dbg("Initialize Entry control ");
 
-	/* Set up VMX entry controls - pg 2908 24.8.1 * Set IA32e guest mode -
-	 * on VM entry processor is in IA32e 64 bitmode * Start guest with host
-	 * IA32_PAT and IA32_EFER
-	 */
-	value32 = (VMX_ENTRY_CTLS_LOAD_EFER |
-		   VMX_ENTRY_CTLS_LOAD_PAT);
+	/* Set up VMX entry controls - ISDM 24.8.1 */
+	value32 = VMX_ENTRY_CTLS_LOAD_PAT;
 
 	if (get_vcpu_mode(vcpu) == CPU_MODE_64BIT) {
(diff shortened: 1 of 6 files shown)


dongyaozu (changes requested):
Split load and save is not a good style. This is one thing.  Please merge into one as we talked.

hypervisor/arch/x86/guest/vmcs.c
```diff
@@ -531,7 +549,7 @@ static void init_exit_ctrl(const struct acrn_vcpu *vcpu)
 	value32 = check_vmx_ctrl(MSR_IA32_VMX_EXIT_CTLS,
 			 VMX_EXIT_CTLS_ACK_IRQ | VMX_EXIT_CTLS_SAVE_PAT |
 			 VMX_EXIT_CTLS_LOAD_PAT | VMX_EXIT_CTLS_LOAD_EFER |
-			 VMX_EXIT_CTLS_SAVE_EFER | VMX_EXIT_CTLS_HOST_ADDR64);
```
dongyaozu, line 534: WHy we need to LOAD_EFER?

hypervisor/arch/x86/guest/vmcs.c
```diff
@@ -19,6 +19,24 @@
...
+}
+
+void set_vmcs_bit(uint32_t vmcs_field, uint32_t bit)
+{
+	uint64_t val64;
+
+	val64 = exec_vmread(vmcs_field);
+	val64 |= bit;
+	exec_vmwrite(vmcs_field, val64);
+}
```
dongyaozu, line 38: These 2 APIs can be inline API with single line. Can you move it to .h file?

ZideChen0: > These 2 APIs can be inline API with single line. Can you move it to .h file?

Yes, will move them to vmcs.h

hypervisor/arch/x86/guest/vmcs.c
ZideChen0: > WHy we need to LOAD_EFER?

It's just the initial value, and the LOAD_EFER bits will be removed later if EFER is identical between guest and host.

Anyway, I can remove it, less code.

dongyaozu (approved):

Issue #6442, opened 2021-08-19
[EF]Post-launched VMs do not boot with "EFI Network" enabled


shiqingg:
HW/Board

tgl-rvp

Steps

enable "EFI Network" in tgl-rvp bios
launch post-launched VM

Expected result

Post-launched VMs boot

Actual result

Post-launched VMs do not boot

Reproduce rate

10/10

shiqingg:
[External_System_ID] ACRN-7319

PR #6443, opened 2021-08-19
config_tools: add a new entry MAX_EFI_MMAP_ENTRIES

shiqingg:
It is used to specify the maximum number of EFI memmap entries.

On some platforms, like Tiger Lake, the number of EFI memmap entries
becomes 268 when the BIOS settings are changed.
The current value of MAX_EFI_MMAP_ENTRIES (256) defined in hypervisor
is not big enough to cover such cases.

As the number of EFI memmap entries depends on the platforms and the
BIOS settings, this patch introduces a new entry MAX_EFI_MMAP_ENTRIES
in configurations so that it can be adjusted for different cases.

Tracked-On: #6442


junjiemao1:
Fine with me for the xsl/xsd part. Due to opens on the configuration editor side, let's keep this simple approach for now.

We can update later when a thorough approach for handling default values is finalized.

NanlinXie (approved):

Issue #6315, opened 2021-07-21
[Task] Configuration tool and data refine and clean up.


Kunhui-Li:
Clear up configtool obsolete kconfig and board.xml files

1. Remove Kconfig
2. Move board.xml into 1source. (1- adl-rvp, 2- ehl-crb-b, 3-tgl-rvp) 
    Overall solution discussion
    moving files to 1source.
3. Clean up scenario python script, remove the obsolete files that already been replaced by xsl.
4. Add revision for board.xml and scenario.xml files(schema file) - TBD
    Define the notation of revision.  1->2 .... 
    Apply to all board.xml and scenario.xml files: attribute: "schema_version"
    Define the rule to update the revision: board.xml , scenario.xml files(schema file)
5. Clean up sample launch script. - libvirt direction - TBD
6. Change the scenario name as shared/partitioned/hybrid
7. Update launch xml for only one NVME

Kunhui-Li:
[External_System_ID] ACRN-7263

NanlinXie:
Closed since patch merged without regression.

PR #6407, opened 2021-08-16
config_tools: remove some tags check for scenario.xml in generic_board folder

Kunhui-Li:
remove some tags check for scenario xml files in generic_board folder when
importing a new type of board in UI, and minor fix for the warning of config
editor.

Tracked-On: #6315
Signed-off-by: Kunhui-Li <kunhuix.li@intel.com>

PR #6446, opened 2021-08-19
config_tools: fix UI issue of creating new settings

Kunhui-Li:
1. as a workaround, comment the code to check MBA_DELAY tag when
creating a new scenario xml setting because of this tag are retrived
from scenario xml files in generic_board folder where it is removed
now.
2. update the template launch xml file names according the recent
update for launch xml files in generic folder.

Tracked-On: #6315

### DIFF
--- a/misc/config_tools/config_app/views.py
+++ b/misc/config_tools/config_app/views.py
@@ -463,7 +463,7 @@ def save_launch():
                     if curr_vm_id == vm_list[i]:
                         curr_vm_index = i + 1
                         break
-            if add_launch_config and add_launch_config.tag == 'uos':
+            if add_launch_config is not None and add_launch_config.tag == 'uos':
                 for i in range(1, MAX_VM_NUM):
                     if str(i) not in vm_list:
                         break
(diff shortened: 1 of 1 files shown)


gvancuts:
I'm struggling to make sense of the commit header and message, can you explain some more? Thanks

NanlinXie (approved):

wenlingz:
@NanlinXie  can we merge this PR, now ?

szhen11:
> I'm struggling to make sense of the commit header and message, can you explain some more? Thanks

hi, @gvancuts , this PR is to fix the issue: when users try to create a new scenario xml by clicking the menu "Scenario Setting" -> "Create a new scenario" in config editor UI, the config editor UI will crash. This issue is caused by the same root cause for PR #6407 : config tool UI shall get the MBA_DELAY element template from scenario xmls in misc/config_tools/data/generic_board/ folder and use it dynamically create the MBA_DELAY elements for newly created scenario xmls, but the  MBA_DELAY template  element from scenario xmls in misc/config_tools/data/generic_board/ folder is removed in recent commits, so this results in the crash for config editor UI; we currently comment this logic to quickly fix the config editor UI issue.

PR #6459, opened 2021-08-23
doc: update UI tool instructions

Kunhui-Li:
1. Update the images to match with the latest UI.
2. Update the path of the saved XML file when clicking Export XML to
save customized file.


dbkinder (approved):
LGTM after I made a few clarity edits

PR #6472, opened 2021-08-24
Doc: ACRN on QEMU changed to CPU Named Model

fuzhongl:
Qemu HOST-passthrough has hardware dependency CPU feature;
which may block ACRN to boot up on Qemu;So change QEMU HOST-passthrough to CPU Named Model since v2.6.

Signed-off-by: fuzhongl <fuzhong.liu@eintel.com>

### DIFF
--- a/doc/tutorials/acrn_on_qemu.rst
+++ b/doc/tutorials/acrn_on_qemu.rst
@@ -11,8 +11,8 @@ with basic functionality such as running Service VM (SOS) and User VM (UOS) for 
 
 This setup was tested with the following configuration,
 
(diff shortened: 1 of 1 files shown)


gvancuts:
I'm assuming the `release_2.6` branches were used for this since the `v2.6` tags do not exist yet?

fuzhongl:
> I'm assuming the `release_2.6` branches were used for this since the `v2.6` tags do not exist yet?

Yes, will change QEMU HOST-passthrough to CPU Named Model from v2.6

gvancuts (approved):

PR #6479, opened 2021-08-24
doc: Add bc to build dependencies

m8:
When I follow the GSG, I got a error at the build process.
```
/bin/sh: 1: bc: not found
```
 So, I added bc to dependencies.

### DIFF
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -134,7 +134,8 @@ To set up the ACRN build environment on the development computer:
            flex \
            bison \
            xsltproc \
-           clang-format
+           clang-format \
+           bc
 
 #. Install Python package dependencies:
 


dbkinder:
@gvancuts can you investigate where this bc dependency came from and why we haven't seen this need before?

gvancuts:
Not sure where the dependency comes from, @m8 when exactly did it happen when you ran through the GSG?

What is your host environment by the way, on my Ubuntu system `bc` is a direct dependency of `ubuntu-desktop` and hence installed by default (which would explain why we may never had noticed it as an explicit dependency).

m8:
sorry about that, I'm using ubuntu in chrooted+debootstraped environment, so i'm not using `ubuntu-desktop`. I got this error at the 
```
make -j $(nproc) targz-pkg
``` 
step, but I didn't get any error for other dependencies.

gvancuts:
Thanks for the clarification, that makes sense now! I also just remembered that this is a bundle we had to manually installed in Clearlinux at the time, this was required in order to build the kernel.

Let's add this to our documenation, it will help those using a reduced environment that may not have it and will essentially be a no-op for the others who have it already installed.

gvancuts (approved):

dbkinder (approved):

PR #6480, opened 2021-08-25
hv: Enable SMBIOS passthrough for pre-launched secure VM

yfliuuu:
This feature is guarded under config CONFIG_SECURITY_VM_FIXUP, which by default should be disabled.
    
This PR enables the passthrough of native SMBIOS information to prelaunched VM. SMBIOS table contains a small entry point structure and a table, of which the entry point structure will be put in 0xf0000-0xfffff region in guest address space, and the table will be put in the ACPI_NVS region in guest address space.

### DIFF
--- a/hypervisor/arch/x86/configs/vm_config.c
+++ b/hypervisor/arch/x86/configs/vm_config.c
@@ -5,6 +5,7 @@
  */
 
 #include <asm/vm_config.h>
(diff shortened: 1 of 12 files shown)


wenlingz:
@dongyaozu @NanlinXie please help approve this PR, thanks

PR #6505, opened 2021-08-26
doc: use "ACRN configurator" name in Overview, GSG

amyreye:
Replaced "ACRN configuration editor" name with "ACRN configurator" in text to match the recently renamed acrn_configurator.py.

Signed-off-by: amyreye <amy.reyes@intel.com>


gvancuts:
Thanks for those clarifications Amy!

Issue #6510, opened 2021-08-27
Lack of SSRAM entries in vRTCT ACPI table for Guest


yonghuah:
*Expected result*

can dump and get SSRAM info in guest PTCT

*Actual result*

no SSRAM info in after dump guest PTCT

PR #6511, opened 2021-08-27
[release_2.6]DM: fix issue when pass-thru native RTCT SSRAM entries

yonghuah:
Native SSRAM entries with only one LAPIC ID of
  value 0 in local APIC ID table can't be pass-thru
  to guest. Such kind of SSRAM entries are for cache
  regions that are global shared and shall be visible
  to guests.

  This patch refine the building logic of vRTCT to fix
  above issue.

Tracked-On: #6510


wenlingz:

fyin1 (approved):

PR #6512, opened 2021-08-27
doc: update usage of zephyr as pre-launched VM kernel

jsun26intel:
Now ACRN support booting pre-launched Zephyr VM kernel in ELF format
instead of previous raw image format, so update usage in related doc.


dbkinder (approved):
We should add this change in the v2.6 release notes.

gvancuts (approved):

Issue #6289, opened 2021-07-12
Nested Virtualization - phase 2


NanlinXie:
Optimized the performance and WR.

NanlinXie:
[External_System_ID] ACRN-7247

NanlinXie:
Doc Impact Analysis: 

- [x] New feature -> new HLD and usage guide
- [ ] [HLDs](https://projectacrn.github.io/2.3/developer-guides/hld/index.html)
- [x] [Configuration data](https://projectacrn.github.io/2.3/tutorials/acrn_configuration_tool.html)
- [ ] [Kernel module parameters](https://projectacrn.github.io/2.3/user-guides/kernel-parameters.html)
- [ ] [Device Model (acrn-dm) parameters](https://projectacrn.github.io/2.3/user-guides/acrn-dm-parameters.html)
- [ ] [Hypervisor shell commands](https://projectacrn.github.io/2.3/user-guides/acrn-shell.html)
- [ ] [ACRN tools (debug and/or service)](https://projectacrn.github.io/2.3/develop.html#configuration-and-tools)
- [ ] [Build system](https://projectacrn.github.io/latest/getting-started/building-from-source.html)
- [x] [Other documents (e.g. tutorials)](https://projectacrn.github.io/2.3/develop.html#enable-acrn-features)

NanlinXie:
No doc needed.

NanlinXie:
Closed since patch merged without regression.

PR #6524, opened 2021-08-31
hv: initialize host IA32_PAT MSR

ZideChen0:
Currently ACRN assumes firmware setup IA32_PAT correctly.  This patch
explicitly initializes host IA32_PAT MSR according to ISDM Table 11-12.
Memory Type Setting of PAT Entries Following a Power-up or Reset.

ACRN creates host page tables based on PAT0 (WB) and PAT3 (UC).

Tracked-On: #6289

### DIFF
--- a/hypervisor/arch/x86/mmu.c
+++ b/hypervisor/arch/x86/mmu.c
@@ -155,6 +155,9 @@ void enable_paging(void)
 {
 	uint64_t tmp64 = 0UL;
 
+	/* Initialize IA32_PAT according to ISDM 11.12.4 Programming the PAT */
+	msr_write(MSR_IA32_PAT, PAT_POWER_ON_VALUE);
+
 	/*
 	 * Enable MSR IA32_EFER.NXE bit,to prevent
 	 * instruction fetching from pages with XD bit set.


hypervisor/arch/x86/mmu.c
```diff
@@ -247,6 +250,9 @@ void init_paging(void)
 
 	pr_dbg("HV MMU Initialization");
 
+	/* Initialize IA32_PAT according to ISDM 11.12.4 Programming the PAT */
+	msr_write(MSR_IA32_PAT, PAT_POWER_ON_VALUE);
+
```
dongyaozu, line 255: init_paging will eventually call enable_paging. So we don't need the changes here.

dongyaozu (changes requested):

hypervisor/arch/x86/mmu.c
ZideChen0: > init_paging will eventually call enable_paging. So we don't need the changes here.

Yes, it's embarrassing ... will fix.

dongyaozu (approved):

PR #6525, opened 2021-08-31
Release 2.6 pio

liujunming:


devicemodel/hw/pci/core.c
```diff
@@ -1404,25 +1393,34 @@ init_pci(struct vmctx *ctx)
...
+					fi = &si->si_funcs[func];
+					if (fi->fi_name == NULL)
+						continue;
+					ops = pci_emul_finddev(fi->fi_name);
+					if (!ops) {
+						pr_warn("No driver for device [%s]\n", fi->fi_name);
+						continue;
+					}
+
+					if ((j == 0) && strcmp(ops->class_name, "passthru")) {
```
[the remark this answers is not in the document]
lifeix: a negative value is not non-zero ?

devicemodel/hw/pci/core.c
```diff
@@ -1543,25 +1559,35 @@ init_pci(struct vmctx *ctx)
...
+						pr_warn("init passthru first to reserve PIO BAR\n");
+						continue;
+					} else if ((j == 1) && !strcmp(ops->class_name, "passthru")) {
+						continue;
+					}
+
+					pci_emul_deinit(ctx, ops, bus, slot,
+					    func, fi);
```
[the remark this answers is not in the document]
lifeix: deinit doesn't care which one is first.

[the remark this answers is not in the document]
ywan170: So this part is needed, right?

devicemodel/hw/pci/core.c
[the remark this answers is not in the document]
ywan170: strcmp() returns an integer indicating the result of the comparison, as follows:

       • 0, if the s1 and s2 are equal;

       • a negative value if s1 is less than s2;

       • a positive value if s1 is greater than s2.

hypervisor/dm/vpci/pci_pt.c
```diff
@@ -426,6 +407,10 @@ static void init_bars(struct pci_vdev *vdev, bool is_sriov_bar)
 				if (!is_sriov_bar) {
 					pci_vdev_write_vbar(vdev, idx, lo);
 				}
+
+				if (is_pci_io_bar(vbar)) {
+					vdev_pt_allow_io_vbar(vdev, idx);
+				}
```
ywan170, line 413: Should we check if the vBAR is same with pBAR before update VMCS's I/O bitmap?

liujunming: Yes, we will check it later.
Because acrn-config hasn't ensured PIO bar identical mapping now.
So we didn't check now.
Actually, besides this part, when guest reprograms the PIO bar, we need to check too.

ywan170: Even the acrn-config tool has kept the identical mapping, we still need to check here due to SOS is not trusted. Can you add check now? I think it expose another security issue.

liujunming: As we discussed with Eddie, We need to inject #GP to guest if not identical mapping.
Then if pre-launched VM pass-thru device that has PIO BAR, then it couldn't work.
I didn't know any security issue, it may impact the functionality here.

ywan170:
The patches look good to me. @fyin1  to help approve. For master, the patches still need to continue to refine.

hypervisor/dm/vpci/pci_pt.c
ywan170: I haven't notice the base_gpa has been changed to bas_hpa.. It is fine now

fyin1 (approved):

hypervisor/dm/vpci/pci_pt.c
liujunming: ok

Issue #6528, opened 2021-09-01
SMBIOS version check error


yfliuuu:
On some UEFI platforms, it is found that SMBIOS major version 3 can be found under SMBIOS2 GUID, so using major version to distinguish different version of SMBIOS will not be reliable.

yfliuuu:
[External_System_ID] ACRN-7382

PR #6529, opened 2021-09-01
release_2.6: hv: minor change to SMBIOS version check

yfliuuu:
On some UEFI platforms, it is found that SMBIOS major version 3 can be
found under SMBIOS2 GUID, so using major version to distinguish
different version of SMBIOS is not reliable. This patch removes the use
of smbios_info.major_ver, and checks the anchor string instead.

Tracked-On: #6528


wenlingz:

fyin1 (approved):

Issue #6290, opened 2021-07-12
GPU Support for SR-IOV based sharing


NanlinXie:
The GPU will implement shareable HW interfaces via the PCI SIG Single Root IO Virtualization (SR-IOV) standard.  The GPU shall provide 1 (legacy compatible) Physical Function and 7 assignable Virtual Functions.
A single image graphics kernel mode driver will be capable of running in legacy, physical function and virtual function modes depending on the configuration of the HW and the interface exposed to the domain in which it is running.
GuC firmware and I915 driver shall provide for configuration of relative performance allocations for the PF and each VF
GVT-g shall provide for assignment of individual display PIPEs and individual display PLANEs to particular VF interfaces for direct control.
GuC firmware and I915 driver  shall provide for the static allocation of  GGTT address space (but not aperture) to each VF.
GuC firmware and I915 driver shall provide for performance and event (i.e. QoS violations/resets) on a per function (PF and VF) basis.
GuC firmware and I915 driver shall provide an interface through which system SW can control execution priority of individual functions
GuC firmware and I915 driver shall provide for relative priorities across VF and PF interfaces.  Higher priority P or VF workloads may preempt workloads from lower priority P or VFs.
GuC firmware and I915 driver  shall provide an interface through which system SW can suspend execution of workloads from a given VF

NanlinXie:
[External_System_ID] ACRN-5968

NanlinXie:
Doc Impact Analysis: 

- [x] New feature -> new HLD and usage guide
- [ ] [HLDs](https://projectacrn.github.io/2.3/developer-guides/hld/index.html)
- [x] [Configuration data](https://projectacrn.github.io/2.3/tutorials/acrn_configuration_tool.html)
- [ ] [Kernel module parameters](https://projectacrn.github.io/2.3/user-guides/kernel-parameters.html)
- [ ] [Device Model (acrn-dm) parameters](https://projectacrn.github.io/2.3/user-guides/acrn-dm-parameters.html)
- [ ] [Hypervisor shell commands](https://projectacrn.github.io/2.3/user-guides/acrn-shell.html)
- [ ] [ACRN tools (debug and/or service)](https://projectacrn.github.io/2.3/develop.html#configuration-and-tools)
- [ ] [Build system](https://projectacrn.github.io/latest/getting-started/building-from-source.html)
[...]

PR #6537, opened 2021-09-03
config_tools: support SRIOV based GPU sharing in launch script

Kunhui-Li:
We will use SRIOV based GPU sharing to replace default GVT-d config
on adl-s-crb platform. So add the logic of passthru gpu VF,
If the string "sriov" is found in launch xml, we will hardcode
different value to gpu VF and set different parameter for WaaG and
LaaG in launch script.

The relevant line in the generated launch script will look like this:
Laag: "-s 2,passthru,0/2/1,gpu  \"
WaaG: "-s 2,passthru,0/2/2,igd-vf  \"

Tracked-On: #6290

### DIFF
--- a/misc/config_tools/launch_config/com.py
+++ b/misc/config_tools/launch_config/com.py
@@ -220,19 +220,23 @@ def interrupt_storm(pt_sel, config):
     print('intr_storm_monitor="--intr_monitor 10000,10,1,100"', file=config)
     print("", file=config)
 
-
 def gvt_arg_set(dm, vmid, uos_type, config):
 
     gvt_args = dm['gvt_args'][vmid]
     gpu_bdf = launch_cfg_lib.get_gpu_bdf()
 
-    if gvt_args == "gvtd" and gpu_bdf is not None:
+    if gpu_bdf is not None:
         bus = int(gpu_bdf[0:2], 16)
         dev = int(gpu_bdf[3:5], 16)
         fun = int(gpu_bdf[6:7], 16)
-        print('   -s 2,passthru,{}/{}/{},gpu  \\'.format(bus, dev, fun), file=config)
-    elif gvt_args:
-        print('   -s 2,pci-gvt -G "$2"  \\', file=config)
+        if gvt_args == "gvtd":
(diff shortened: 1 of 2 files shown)


gvancuts:
> We will use SRIOV based GPU sharing to replace default GVT-d config
> on adl-s-crb platform.

Is this mechanism restricted to that board, or is it generic and any platform that supports SRIOV for the GPU device can use this?

> So add the logic of passthru gpu VF,
> If the string "sriov" is found in launch xml, we will hardcode
> different value to gpu VF and set different parameter for WaaG and
> LaaG in launch script.

Out of curiosity, why do we need to hardcode different values for Linux and Windows?

wenlingz:

fyin1 (approved):

fyin1 (approved):

Kunhui-Li:
> Is this mechanism restricted to that board, or is it generic and any platform that supports SRIOV for the GPU device can use this?

When to found the string "sriov" in launch xml, any platforms that supports SRIOV for the GPU device can use this. We only set it in launch xml for adl-s-crb platform currently.

gvancuts:
> > So add the logic of passthru gpu VF,
> > If the string "sriov" is found in launch xml, we will hardcode
> > different value to gpu VF and set different parameter for WaaG and
> > LaaG in launch script.
> 
> Out of curiosity, why do we need to hardcode different values for Linux and Windows?

Can you answer this question too?

NanlinXie:
@psun3x , could you help to answer @gvancuts's question? Thanks!

psun3x:
@gvancuts waag and laag should use the same value. Here is a typo, but no harm to VF passthrough. "igd-vf" is a invalid parameter to passthrough device. But we need it here, because we need the different value of gvt-d("gpu").

gvancuts:
> @gvancuts waag and laag should use the same value. Here is a typo, but no harm to VF passthrough. "igd-vf" is a invalid parameter to passthrough device. But we need it here, because we need the different value of gvt-d("gpu").

Thanks, the commit message is very confusing in this case since it specifically indicate that different values are hard-coded. Was it really a typo then? I'm confused by this.

Issue #5112, opened 2020-07-29
ACRN debug shell help output behavior, line length, and misspellings


dbkinder:
**Describe the bug**

In the output of the debug shell``help`` command, after 10 lines of output there is a line written saying, 
``<*** Hit any key to continue ***>``.  There's no pause in the output waiting for a key to be pressed.

Some output lines are very long and wrap, making it hard to read.

"pass-through" and "Setel" are misspelled words: should be "passthrough" and I'm not sure what "Setel" should be.

**Expected behavior**
The output should pause, waiting for a key to be pressed, and then continue.  (Or just remove that feature since there's really no need to pause after 10 lines.)  Long output lines should be broken so they don't wrap on an 80-character-width console. No misspellings.

**Additional context**

```
ACRN:\>help

Registered Commands:

  help             - Display information about supported hypervisor shell commands
  version          - Display the HV version information
  vm_list          - List all VMs, displaying the VM UUID, ID, name and state
  vcpu_list        - List all vCPUs in all VMs
  vcpu_dumpreg    <vm id, vcpu id> - Dump registers for a specific vCPU
  dump_host_mem   <addr, length> - Dump host memory, starting at a given address(Hex), and for a given length (Dec in bytes)
  dump_guest_mem  <vm_id, addr, length> - Dump guest memory, vm id(Dec), starting at a given address(Hex), and for a given length (Dec in bytes)
  vm_console      <vm id> - Switch to the VM's console. Use [Ctrl+Spacebar] to return to the ACRN shell console
  int              - List interrupt information per CPU
<*** Hit any key to continue ***>
  pt               - Show pass-through device information
  vioapic         <vm id> - Show virtual IOAPIC (vIOAPIC) information for a specific VM
  dump_ioapic      - Show native IOAPIC information
  loglevel        [<console_loglevel> [<mem_loglevel> [npk_loglevel]]] - No argument: get the level of logging for the console, memory and npk. Setel by giving (up to) 3 parameters between 0 and 6 (verbose)
  cpuid           <leaf> [subleaf] - Display the CPUID leaf [subleaf], in hexadecimal
  reboot           - Trigger a system reboot (immediately)
  rdmsr           [-p<pcpu_id>] <msr_index> - Read the MSR at msr_index (in hexadecimal) for CPU ID pcpu_id
[...]

PR #6539, opened 2021-09-03
ACRN:HV: Normalize hypervisor help output format

liudlong:
Normalize hypervisor help command output format, remove the 10 lines
limit for one screen, fix the misspelled words.

Tracked-On: #5112
Reviewed-by: VanCutsem, Geoffroy <geoffroy.vancutsem@intel.com>

### DIFF
--- a/hypervisor/debug/shell.c
+++ b/hypervisor/debug/shell.c
@@ -495,21 +495,20 @@ void shell_init(void)
 			SHELL_CMD_MAX_LEN + 1U);
 }
 
-#define SHELL_ROWS	10
-#define MAX_INDENT_LEN	16U
+#define SHELL_ROWS	30
+#define MAX_OUTPUT_LEN  80
 static int32_t shell_cmd_help(__unused int32_t argc, __unused char **argv)
 {
-	uint16_t spaces;
 	struct shell_cmd *p_cmd = NULL;
-	char space_buf[MAX_INDENT_LEN + 1];
 
+	char str[MAX_STR_SIZE];
+	char* help_str;
 	/* Print title */
 	shell_puts("\r\nRegistered Commands:\r\n\r\n");
 
 	pr_dbg("shell: Number of registered commands = %u in %s\n",
 		p_shell->cmd_count, __func__);
 
-	(void)memset(space_buf, ' ', sizeof(space_buf));
 	/* Proceed based on the number of registered commands. */
 	if (p_shell->cmd_count == 0U) {
 		/* No registered commands */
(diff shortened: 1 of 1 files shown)


hypervisor/debug/shell.c
```diff
@@ -79,52 +79,58 @@ static struct shell_cmd shell_cmds[] = {
 		.fcn		= shell_list_vcpu,
 	},
 	{
-		.str		= SHELL_CMD_VCPU_DUMPREG,
-		.cmd_param	= SHELL_CMD_VCPU_DUMPREG_PARAM,
-		.help_str	= SHELL_CMD_VCPU_DUMPREG_HELP,
-		.fcn		= shell_vcpu_dumpreg,
```
dongyaozu, line 85: Why you delete VCPU_DUMPREG?

hypervisor/debug/shell.c
```diff
@@ -79,52 +79,58 @@ static struct shell_cmd shell_cmds[] = {
...
+		.str		= SHELL_CMD_PTDEV,
```
dongyaozu, line 82: Why move CMD_PTDEV here?
Stay in original place makes the patch smaller

hypervisor/debug/shell.c
liudlong: @dongyaozu  We didn't delete any command, just requeue the command, so that we use the help command, the command with no argument will be printed at first.

gvancuts (approved):

hypervisor/debug/shell.c
```diff
@@ -518,6 +515,11 @@ static int32_t shell_cmd_help(__unused int32_t argc, __unused char **argv)
 		int32_t i = 0;
 		uint32_t j;
 
+		snprintf(str, MAX_STR_SIZE, "  %-15s %-53s %-s",
```
dbkinder, line 518: The output won't look very good on an 80-character terminal (you've consumed 72 characters before you get to the description).  I'm not sure what the supported terminal width requirement should be, but the usual is 80 characters and your output has lines that are more than 200 characters wide.  Line wrapping will make this output confusing to read (DX issue).

dbkinder (changes requested):
Thanks for working on this. Display of very long output lines won't look good on the typical 80-character terminal window.

hypervisor/debug/shell.c
dongyaozu: This is a topic of which one should be printed earlier and what is the order.
How does commercial debugger do? I thought it can be in alpha-beta sequence or base on categories of meaning.  Sequence base on parameter # is rare case to me.

@gvancuts your comments?

hypervisor/debug/shell.c
liudlong: @dbkinder  Ok, just now i have a test with picocom, there have line wrapping when the output is bigger than 80, i will have a change with this. thanks

dongyaozu:
If this patch is not needed, please close it. If it is needed, please move forward to close all those issues. We should not pend here.

liudlong:
> If this patch is not needed, please close it. If it is needed, please move forward to close all those issues. We should not pend here.

For this patch, i have a patch for your and dbkinder comments, i will push the patch today.

dongyaozu (approved):

dongyaozu (approved):

hypervisor/debug/shell.c
```diff
@@ -539,29 +538,30 @@ static int32_t shell_cmd_help(__unused int32_t argc, __unused char **argv)
...
-				shell_puts(" - ");
-				shell_puts(p_cmd->help_str);
+			help_str = p_cmd->help_str;
+			while (strnlen_s(help_str, MAX_OUTPUT_LEN > 0)) {
+				(void)memset(str, ' ', sizeof(str));
+				if (strnlen_s(help_str, MAX_OUTPUT_LEN) > 65) {
+					snprintf(str, MAX_OUTPUT_LEN, "               %-s", help_str);
+					shell_puts(str);
+					shell_puts("\r\n");
+					help_str = help_str + 65;
```
dbkinder, line 557: So this is just breaking the help_str at 65-character boundaries, not at a word boundary, eh?  So it will put part of a word on one line and the rest on the next.  Can you show us what this help output looks like now?

dbkinder (changes requested):
Can you show what this new help output looks like now?  I suspect the line breaks you added will put part of a word on one line and the rest indented on the next.

hypervisor/debug/shell.c
liudlong: @dbkinder Yes, this still have the issue will break the word, the shell_puts only can output via character length.

liudlong: image

liudlong: image

dbkinder: It looks like one character is missing from a word when the line breaks.

dbkinder: We can merge this now and fix later.

dbkinder (approved):
We can fix the word break output later.

Issue #6533, opened 2021-09-02
mem leak while poweroff in guest


liudlong:
Test target:  mainline (commit id: c4cb95f3b4023a58b6ceeaa59b8677a2c5c54c24) + 2 hypercube patches with unchanged 2.5 kernel
Test tool: 5 latest Hypercube iso:  hypercube_APIC.iso/ hypercube_PCI_HPET.iso/ hypercube_PCI_USB.iso/ hypercube_PCI_ecam.iso/ hypercube_PIO.iso, built with new scrip- build-all.sh.
Test method: 1: try industry scenario on TGL. 2: only boot 1 guest with 1 vcpu. 3: use same launch parameters temporarily
poweroff normal guest, not need fuzzing

 __________________________

 

==357==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32816 byte(s) in 1 object(s) allocated from:
#0 0x55997c92a000 in __interceptor_malloc (/usr/bin/acrn-dm+0x230000)
#1 0x7f9fc7bfd906 (/lib/libc.so.6+0xc3906)

Direct leak of 152 byte(s) in 1 object(s) allocated from:
#0 0x55997c92a1e8 in __interceptor_calloc (/usr/bin/acrn-dm+0x2301e8)
#1 0x7f9fc8b7fcc5 (/lib/libusb-1.0.so.0+0x4cc5)

Indirect leak of 2048 byte(s) in 2 object(s) allocated from:
#0 0x55997c92a3f0 in __interceptor_realloc (/usr/bin/acrn-dm+0x2303f0)
#1 0x7f9fc8b8a79e (/lib/libusb-1.0.so.0+0xf79e)

Indirect leak of 416 byte(s) in 2 object(s) allocated from:
#0 0x55997c92a1e8 in __interceptor_calloc (/usr/bin/acrn-dm+0x2301e8)
#1 0x7f9fc8b8430f in libusb_alloc_transfer (/lib/libusb-1.0.so.0+0x930f)
#2 0x55997c9a43a2 in usb_dev_data hw/platform/usb_pmapper.c:810
#3 0x55997ca2643f in pci_xhci_try_usb_xfer hw/pci/xhci.c:3002
#4 0x55997ca2c4a1 in pci_xhci_handle_transfer hw/pci/xhci.c:3271
#5 0x55997ca2f562 in pci_xhci_device_doorbell hw/pci/xhci.c:3380
[...]

PR #6545, opened 2021-09-07
ACRN:DM Release resource when destroy the device

liudlong:
When destroy the usb device release the resource allocate for transfer
in case cause the memory leak issue. Add the release and cancel
transfer request call back for the emulation device, use the emulation
device call back in xHCI controller emulation.

Tracked-On: #6533


ywan170:
@fyin1

fyin1 (approved):

Issue #6292, opened 2021-07-12
[Task]Configuration tool and data refine and clean up.


NanlinXie:
Clear up configtool obsolete kconfig and board.xml files
1. Remove Kconfig code
2. Clean up scenario python script, remove the obsolete files that already been replaced by xsl.
3. Add revision for board.xml and scenario.xml files(schema file)

- Define the notation of revision.  1->2 .... 

- Apply to all board.xml and scenario.xml files: attribute: "schema_version"

- Define the rule to update the revision: board.xml , scenario.xml files(schema file)

4. Clean up sample launch script.
5. Change the scenario name as shared/partitioned/hybrid.

NanlinXie:
[External_System_ID] ACRN-7263

NanlinXie:
Doc Impact Analysis: 

- [x] New feature -> new HLD and usage guide
- [ ] [HLDs](https://projectacrn.github.io/2.3/developer-guides/hld/index.html)
- [x] [Configuration data](https://projectacrn.github.io/2.3/tutorials/acrn_configuration_tool.html)
- [ ] [Kernel module parameters](https://projectacrn.github.io/2.3/user-guides/kernel-parameters.html)
- [ ] [Device Model (acrn-dm) parameters](https://projectacrn.github.io/2.3/user-guides/acrn-dm-parameters.html)
- [ ] [Hypervisor shell commands](https://projectacrn.github.io/2.3/user-guides/acrn-shell.html)
- [ ] [ACRN tools (debug and/or service)](https://projectacrn.github.io/2.3/develop.html#configuration-and-tools)
- [ ] [Build system](https://projectacrn.github.io/latest/getting-started/building-from-source.html)
- [x] [Other documents (e.g. tutorials)](https://projectacrn.github.io/2.3/develop.html#enable-acrn-features)

dbkinder:
@nanlin @junjiemao1 Do need a schema version number field stored with the XML files.   @junjiemao1 was having issues figuring out how to properly patch an "old" XML file into a "new" one. Perhaps we need a proper tool that can scan and bring any XML file to the current expected format and content.  (For example, an older XML file might not have settings for options that were added in a newer version.)

NanlinXie:
Configuration tool document updated.

NanlinXie:
https://projectacrn.github.io/latest/tutorials/acrn_configuration_tool.html
https://projectacrn.github.io/latest/tutorials/board_inspector_tool.html
[...]

PR #6555, opened 2021-09-10
config_tools: populate default values in scenario XML

junjiemao1:
While we have default values of configuration entries stated in the schema
of scenario XMLs, today we still require user-given scenario XMLs to
contain literally ALL XML nodes. Missing of a single node will cause schema
validation errors even though we can use its default value defined in the
schema.

This patch allows user-given scenario XMLs to ignore nodes with default
values. It is done by adding the missing nodes, all containing the defined
default values, to the input scenario XML when copying it to the build
directory. This approach imposes no changes to either the schema or
subsequent scripts in the build system.

Tracked-On: #6292

### DIFF
--- a/hypervisor/scripts/makefile/config.mk
+++ b/hypervisor/scripts/makefile/config.mk
@@ -113,6 +113,7 @@ HV_UNIFIED_XML_IN := $(BASEDIR)/scripts/makefile/unified.xml.in
 HV_PREDEFINED_DATA_DIR := $(realpath $(BASEDIR)/../misc/config_tools/data)
 HV_CONFIG_TOOL_DIR := $(realpath $(BASEDIR)/../misc/config_tools)
 HV_CONFIG_XFORM_DIR := $(HV_CONFIG_TOOL_DIR)/xforms
+HV_SCENARIO_XSD := $(HV_CONFIG_TOOL_DIR)/schema/config.xsd
 
 # Paths to the outputs:
 HV_CONFIG_DIR := $(HV_OBJDIR)/configs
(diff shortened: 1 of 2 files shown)


NanlinXie:
Schema and all build check passed.

NanlinXie (approved):

dbkinder (commented):
@junjiemao1 If I've got a scenario XML I've been using (with v2.5) and want to continue using it with the v2.6 release, what's going to happen?  Does this patch provide the way to "upgrade" my custom v2.5 flavor scenario XML to the new requirements for v2.6?  @gvancuts

junjiemao1:
@dbkinder Sort of. A ver `n` scenario XML can be used unmodified for ver `n+1` if the schema changes during the development of ver `n+1` meet the following conditions:

1. Every configuration item introduced for ver `n+1` has an explicit default value in the schema (i.e. the `xsd`s)
1. No configuration item is renamed or removed.

Deviating the first condition always invalidates previous scenario XMLs. Deviating the second may or may not impact existing scenario XMLs, depending on whether a scenario XML file explicitly defines any renamed or removed configuration item (which is typically the case today as we require all items to be defined).

gvancuts:
Thanks for the explanation @junjiemao1 . I know we discussed the idea of "versioning" those XML files in order to be able to support back or forward porting but I don't think we ever got to a final conclusion for this. One idea that was floated around was to compare the git commit dates (which would also work in case someone maintains a series of XML files in a separate repo). The drawback of that method is that if you just get a file from a different user, you may not know what its commit date was (and this is where some versioning inside the file itself might still be useful).

Have you had further thoughts on this?

junjiemao1:
@gvancuts Yes, the versioning of XML files is yet an open, but it is not the scope of this PR.

This PR is aimed at reducing engineering efforts to introduce new configuration items. Before this PR one has to touch literally ALL scenario XMLs in this repo for adding a single item, even though the default value has already been specified in the schema file. With this patch it is no longer necessary: an absent configuration item implies the item should have its default value in the schema.

While making scenario XMLs more concise, this PR is not targeted to solve the "versioning" problem. We can keep evaluating and comparing different approaches like version numbers and dates.

PR #6556, opened 2021-09-10
hv: nested: improvement and performance optimization

ZideChen0:


dongyaozu (commented):
I commented before, the last patch should go together with a DOC update to clearly say the potential issues and suggestion...  We shouldn't miss.

ZideChen0:
> I commented before, the last patch should go together with a DOC update to clearly say the potential issues and suggestion... We shouldn't miss.

Yes, I was thinking to work on a separate DOC patch. Anyway I will remove 3/3 patch from this PR and put it together with the DOC patch in separate PR.

dongyaozu (approved):

Issue #6562, opened 2021-09-13
config-tools: add a platform to acrn:is-tpm-passthrough-board


yuchuyang:

yuchuyang:
[External_System_ID] ACRN-7321

PR #6314, opened 2021-07-21
config_tools: rename scenario xml

Kunhui-Li:
1. Rename industry.xml to shared.xml
2. Rename logical_partition.xml to partitioned.xml.
3. Update the name in view.py correspondingly.

Signed-off-by: Kunhui-Li <kunhuix.li@intel.com>

PR #6563, opened 2021-09-13
config-tools: refine acrn:is-tpm-passthrough-board

yuchuyang:
Return true if the tpm device is present and if the passthrough tpm2 is enabled.

Tracked-On: #6562

### DIFF
--- a/misc/config_tools/xforms/lib.xsl
+++ b/misc/config_tools/xforms/lib.xsl
@@ -481,29 +481,6 @@
     </xsl:choose>
   </func:function>
 
-  <func:function name="acrn:is-tpm-passthrough-board">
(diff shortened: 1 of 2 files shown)


yuchuyang:
@wenlingz  looks like industry scenario xmls have been removed so that CI test failed?

dbkinder:
> @wenlingz looks like industry scenario xmls have been removed so that CI test failed?

Indeed #6314 renamed industry to shared and logical_partition to partitioned, launch scripts were renamed and updated too.  I too thought that renaming wasn't supposed to happen until v2.7 but it was merged to master nearly three weeks ago.

NanlinXie (approved):

PR #6574, opened 2021-09-15
doc: update documentation and grub file

jackwhich:
1. ACRN official documentation getting started guide. doc 6 Install ACRN->3 step>b
link needs to modify  and because the grup file syntax cannot be recognized <> .
The correct way to modify is to mislead the user to modify it to <> Correctly
modify it to UUID and PARTUUDI.
2. Because there is a problem with the user adding GRUB_CMDLINE_LINUX=text
parameter, delete this parameter.
3. The script generation of launch_uos_id3.sh does not match the GitHub and doc,
modify the parameters

### DIFF
--- a/hypervisor/include/hw/pci.h
+++ b/hypervisor/include/hw/pci.h
@@ -391,7 +391,7 @@ static inline bool is_pci_vendor_valid(uint32_t vendor_id)
 
 static inline bool is_pci_cfg_multifunction(uint8_t header_type)
 {
-	return ((header_type & PCIM_MFDEV) == PCIM_MFDEV);
+	return ((header_type != 0xffU) && ((header_type & PCIM_MFDEV) == PCIM_MFDEV));
 }
 
 static inline bool pci_is_valid_access_offset(uint32_t offset, uint32_t bytes)


doc/getting-started/getting-started.rst
```diff
@@ -660,9 +660,9 @@ In the following steps, you will configure GRUB on the target system.
       .. code-block:: bash
 
          GRUB_DEFAULT=ubuntu-service-vm
+         GRUB_DISTRIBUTOR=`lsb_release -i -s 2> /dev/null || echo Debian`
```
gvancuts, line 663: What is the purpose of this line?

gvancuts (changes requested):
Can you fix your commit message which is rather hard to read.

You also have a typo "PARTUUDI" -> "PARTUUID" and I suspect the `grap` word really refers to `Grub`?

doc/getting-started/getting-started.rst
```diff
@@ -637,9 +637,9 @@ In the following steps, you will configure GRUB on the target system.
            insmod gzio
            insmod part_gpt
            insmod ext2
-           search --no-floppy --fs-uuid --set <UUID>
+           search --no-floppy --fs-uuid --set "UUID"
            echo 'loading ACRN...'
-           multiboot2 /boot/acrn/acrn.bin  root=PARTUUID=<PARTUUID>
+           multiboot2 /boot/acrn/acrn.bin  root=PARTUUID="PARTUUID"
            module2 /boot/vmlinuz-5.10.52-acrn-sos Linux_bzImage
          }
```
amyreye, line 645: Please add an example of the above text with the UUID and PARTUUID filled in correctly. This will give our customers confidence that they are doing the step correctly. (In the DX study, we saw that participants were not sure whether or not to include quotation marks. Some participants suggested adding an example for clarification.)

PR #6445, opened 2021-08-19
doc: add acrn-dm parameter emul explanation

liuhang-bit:
add the emul list table to acrn-dm paramter
chapter.

Signed-off-by: Liu, Hang1 <hang1.liu@intel.com>

PR #6582, opened 2021-09-17
use "igd" parameter name when pass-thru IGD

liujunming:
The term of "gpu" is too common that might mislead end-users
    that it is a "must have" parameter to pass though any graphics card.
    Actually the parameter is used for Intel graphics only(bdf@00:02.0).
    We met this problem when we try to pass through
    a third party graphics card (bdf@00:05.0),
    people don't know whether need to add "gpu" parameter.


dbkinder:
Ok, but there's no documentation about this igd option, even in the open PR #6445

dbkinder (approved):

PR #6583, opened 2021-09-17
    doc: update documentation and grub file

jackwhich:
1. Because the < > in the document will mislead users to copy and paste directly,
    modify the < > to " ".
2. Add sample diagrams for user reference.
3. Because there is a problem with the user adding the GRUB_CMDLINE_LINUX=text
    parameter, delete this parameter.
4. The script generation of launch_uos_id3.sh does not match the GitHub and doc,
    modify the parameters.

### DIFF
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -637,9 +637,9 @@ In the following steps, you will configure GRUB on the target system.
            insmod gzio
            insmod part_gpt
            insmod ext2
-           search --no-floppy --fs-uuid --set <UUID>
+           search --no-floppy --fs-uuid --set "3cac5675-e329-4cal-b346-0a3e65f99016"
            echo 'loading ACRN...'
-           multiboot2 /boot/acrn/acrn.bin  root=PARTUUID=<PARTUUID>
+           multiboot2 /boot/acrn/acrn.bin  root=PARTUUID="03db7f45-8a6c-454b-adf7-30343d82c4f4"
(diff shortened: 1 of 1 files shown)


gvancuts (changes requested):
Please fix your commit message, especially the first paragraph: really hard to understand and various typos are in there.

doc/getting-started/getting-started.rst
```diff
@@ -637,9 +637,9 @@ In the following steps, you will configure GRUB on the target system.
            insmod gzio
            insmod part_gpt
            insmod ext2
-           search --no-floppy --fs-uuid --set <UUID>
+           search --no-floppy --fs-uuid --set "3cac5675-e329-4cal-b346-0a3e65f99016"
```
dbkinder, line 640: This isn't going to work.  The instructions above this code block say to

```
Replace ``<UUID>`` and ``<PARTUUID>`` with the output from the previous step.
```
but now the code block doesn't have those strings in it any more.  What we need is to put those UUID and PARUUID strings back in this code block, and then create another code block showing what a properly edited GRUB file example would look like.

dbkinder (changes requested):

doc/getting-started/getting-started.rst
jackwhich: already edited

dbkinder (approved):
LGTM, thanks

gvancuts (approved):
LGTM

Issue #6592, opened 2021-09-18
failed to make hvdiffconfig


szhen11:
Environment
ubuntu

HW/Board
ehl-crb-b, tgl-rvp, adl-s-crb

Build link
NA

Image info
NA

Steps
1. clone acrn-hypervisor, checkout to release_2.6 branch(commit: 9bf80f916e1bdf86417de5e395dd3a4f7f62ebe7)
git am offline patches from https://github.com/intel-innersource/virtualization.hypervisors.acrn.acrn-dev.acrn-offline-patch/tree/master/boards/v2.6
2. following doc: https://projectacrn.github.io/latest/tutorials/acrn_configuration_tool.html#makefile-targets-for-configuration
to generate diff configs and apply it.

Expected result
generate diff configs and apply it successfully.

Actual result
generate diff config patch failed

Reproduce rate
100%

Debugging info

shuang@shuang-desktop:~/dev/acrn/acrn-hypervisor$ make clean
make -C /home/shuang/dev/acrn/acrn-hypervisor/misc OUT_DIR=/home/shuang/dev/acrn/acrn-hypervisor/build/misc clean
make[1]: Entering directory '/home/shuang/dev/acrn/acrn-hypervisor/misc'
make -C /home/shuang/dev/acrn/acrn-hypervisor/misc/services/acrn_manager OUT_DIR=/home/shuang/dev/acrn/acrn-hypervisor/build/misc/services clean
make[2]: Entering directory '/home/shuang/dev/acrn/acrn-hypervisor/misc/services/acrn_manager'
rm -f /home/shuang/dev/acrn/acrn-hypervisor/build/misc/services/acrnctl
rm -f /home/shuang/dev/acrn/acrn-hypervisor/build/misc/services/acrn_mngr.o
rm -f /home/shuang/dev/acrn/acrn-hypervisor/build/misc/services/libacrn-mngr.a
rm -f /home/shuang/dev/acrn/acrn-hypervisor/build/misc/services/acrnd
rm -f /home/shuang/dev/acrn/acrn-hypervisor/build/misc/services/acrn_mngr.h
rm -f /home/shuang/dev/acrn/acrn-hypervisor/build/misc/services/acrnd.service
rm -rf /home/shuang/dev/acrn/acrn-hypervisor/build/misc/services
make[2]: Leaving directory '/home/shuang/dev/acrn/acrn-hypervisor/misc/services/acrn_manager'
[...]

PR #6593, opened 2021-09-18
Makefile: fix wrong reference to board XML and skip binary in diffconfig

junjiemao1:
The current config.mk uses the variable BOARD_FILE as the path to the board
XML when generating an unmodified copy of configuration files for
comparison, which is incorrect. The right variable is HV_BOARD_XML which is
the path to the copy of board XML that is actually used for the build.

This patch corrects the bug above.

In addition, this patch also skips binary files (which are not meant to be
edited manually) when calculating the differences.

Tracked-On: #6592


wenlingz:

dongyaozu (approved):

PR #6590, opened 2021-09-18
config_tools: clean up board folders under data directory

Kunhui-Li:
1. move tgl-rvp and ehl-crb-b folders under data directory to 1source.
2. delete nuc7i7dnb and whl-ipc-i7 folders under data directory.

Tracked-On: #6315
Signed-off-by: Kunhui-Li <kunhuix.li@intel.com>

PR #6598, opened 2021-09-22
    doc: Steps to update hybrid and logic_Partition documents

jackwhich:
1.Modify hybrid update ubuntu grup Note: The syntax is easy for users to understand
    2.Add to view the serial port number, enter the serial hybrid verification mode
    3.Delete the kernel, because the config_uos file has been abandoned for maintenance,
      now the Ubuntu native kernel image is used and added to the document
    4.Add to view the serial port number, enter the serial port logic_partition verification method

### DIFF
--- a/doc/tutorials/using_hybrid_mode_on_nuc.rst
+++ b/doc/tutorials/using_hybrid_mode_on_nuc.rst
@@ -110,16 +110,16 @@ Perform the following to update Ubuntu GRUB so it can boot the hypervisor and lo
       }
 
 
-   .. note:: The module ``/boot/zephyr.elf`` is the VM0 (Zephyr) kernel file.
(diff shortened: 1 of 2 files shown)


gvancuts (changes requested):
* using_partition_mode_on_nuc.rst:
  * there are a few references to the `logical_partition.xml` file in this doc. But that file has been renamed to `partitioned.xml`
  * This document is based on the `BOARD=whl-ipc-i7` board but that is being removed from the repo by #6590, we need to use a different reference board.
* using_hybrid_mode_on_nuc.rst: this document is based on the NUC7i7DNHE (`BOARD=nuc7i7dnb), that board is being removed (see #6590) so we should use a different platform as a reference.

PR #6599, opened 2021-09-22
config_tools: populate default values to all nodes

junjiemao1:
The default value population algorithm introduced by commit
2bfaa34 ("config_tools: populate default values in scenario XML") only
populates default values to the first occurrence of a tag when the tag is
specified to allow multiple occurrences under an xs:all node. This may lead
to incomplete scenario XML as some of the default values are missed.

This patch fixes this issue by checking **all** nodes having the same tag
under a node specified by an xs:all schema.

Fixes: 2bfaa34 ("config_tools: populate default values in scenario XML")

Tracked-On: #6292


wenlingz:

NanlinXie (approved):

PR #6616, opened 2021-09-24
doc: Update logical_partition documentation

jackwhich:
1. Modified Intel Whiskey Lake to Intel NUC Kit NUC11TNBi5
2. Change the logical partition and mixed test motherboard name
Signed-off-by: Zhong Zhenx.Liu <zhongzhenx.liu@intel.com>


gvancuts (changes requested):
Why are changing to this NUC? That platform is older than Whiskey Lake. It would be better to use the latest TGL NUC, any reason why we cannot?

Have you looked at the questions regarding SATA, NVMe and USB **and** this copying of kernel and modules?

dbkinder (approved):
LGTM, thanks.

Issue #6652, opened 2021-10-05
Refine system shutdown


wuxyintel:
Use socket(virtio-net) or vUART for communication between the host and guest side lifecycle managers. Need to update lifecycle manager. Lifecycle manager is a reference implementation, customer can implement more functions (such as shutdown request authority) according to their requirements.
Need to update hypervisor and config tool to support multi legacy vUART. The total number of vUART supported by hypervisor is  8.

Doc Impact Analysis: 

- [x] New feature -> new HLD and usage guide
- [ ] [HLDs](https://projectacrn.github.io/2.3/developer-guides/hld/index.html)
- [ ] [Configuration data](https://projectacrn.github.io/2.3/tutorials/acrn_configuration_tool.html)
- [ ] [Kernel module parameters](https://projectacrn.github.io/2.3/user-guides/kernel-parameters.html)
- [ ] [Device Model (acrn-dm) parameters](https://projectacrn.github.io/2.3/user-guides/acrn-dm-parameters.html)
- [ ] [Hypervisor shell commands](https://projectacrn.github.io/2.3/user-guides/acrn-shell.html)
- [ ] [ACRN tools (debug and/or service)](https://projectacrn.github.io/2.3/develop.html#configuration-and-tools)
- [ ] [Build system](https://projectacrn.github.io/latest/getting-started/building-from-source.html)
- [ ] [Other documents (e.g. tutorials)](https://projectacrn.github.io/2.3/develop.html#enable-acrn-features)

yuchuyang:
[External_System_ID] ACRN-7407

NanlinXie:
Updated document:
https://projectacrn.github.io/latest/tutorials/enable_s5.html
https://projectacrn.github.io/latest/getting-started/getting-started.html

amyreye:
@NanlinXie @dbkinder Can we close this issue?

NanlinXie:
Closed since patch merged without regression.

PR #6658, opened 2021-10-11
hv: support at most MAX_VUART_NUM_PER_VM legacy vuarts

wuxyintel:
In the current hypervisor, only support at most two legacy vuarts
(COM1 and COM2) for a VM, COM1 is usually configured as VM console,
COM2 is configured as communication channel of S5 feature.
Hypervisor can support MAX_VUART_NUM_PER_VM(8) legacy vuart, but only
register handlers for two legacy vuart since the assumption (legacy
vuart is less than 2) is made.
In the current hypervisor configurtion, io port (2F8H) is always
allocated for virtual COM2, it will be not friendly if user wants to
assign this port to physical COM2.
Legacy vuart is common communication channel between service VM and
user VM, it can work in polling mode and its driver exits in each
guest OS. The channel can be used to send shutdown command to user VM
in S5 featuare, so need to config serval vuarts for service VM and one
vuart for each user VM.

The following changes will be made to support at most
MAX_VUART_NUM_PER_VM legacy vuarts:
   - Refine legacy vuarts initialization to register PIO handler for
related vuart.
   - Update assumption of legacy vuart number.

BTW, config tools updates about legacy vuarts will be made in other
patch.

v1-->v2:
	Update commit message to make this patch's purpose clearer;
	If vuart index is valid, register handler for it.

Tracked-On: #6652
Acked-by: Eddie Dong <eddie.dong@Intel.com>


hypervisor/dm/vuart.c
```diff
@@ -559,18 +559,11 @@ static bool vuart_register_io_handler(struct acrn_vm *vm, uint16_t port_base, ui
...
 	};
-	switch (vuart_idx) {
-	case 0U:
-		pio_idx = UART_PIO_IDX0;
-		break;
-	case 1U:
-		pio_idx = UART_PIO_IDX1;
-		break;
-	default:
+	if (vuart_idx >= MAX_VUART_NUM_PER_VM) {
```
dongyaozu, line 562: Can you handle normal case first?

dongyaozu (changes requested):

hypervisor/dm/vuart.c
wuxyintel: Sure, will update it soon, thanks!

wuxyintel: @dongyaozu Hi eddie, updates have been done, please double check it, thanks!

dongyaozu (approved):

Issue #6663, opened 2021-10-13
Current HV_RAM_SIZE calculation algrithom sometimes cause build failure


lifeix:
[External_System_ID] ACRN-7390

wenlingz:
[External_System_ID] ACRN-7390

gvancuts:
This seems to have served as well to remove the `hvlog` parameter from the code. This implies some documentation needs to be updated, at least the `kernel-parameters.rst` and `hld-trace-log.rst` documents.

NanlinXie:
We didn't remove this function, we just remove default configuration in XML file, thus there's no need to update document.

NanlinXie:
Closed since patch merged without regression.

PR #6664, opened 2021-10-13
hv: remove CONFIG_HV_RAM_SIZE

lifeix:
It's difficult to configure CONFIG_HV_RAM_SIZE properly at once. This patch
not only remove CONFIG_HV_RAM_SIZE, but also we use ld linker script to
dynamically get the size of HV RAM size.

Tracked-On: #6663

### DIFF
--- a/hypervisor/include/arch/x86/asm/boot/ld_sym.h
+++ b/hypervisor/include/arch/x86/asm/boot/ld_sym.h
@@ -14,5 +14,6 @@ extern uint8_t		ld_entry_end;
 extern const uint8_t	ld_trampoline_load;
 extern uint8_t		ld_trampoline_start;
 extern uint8_t		ld_trampoline_end;
+extern uint8_t		ld_ram_size;
 
 #endif /* LD_SYM_H */
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -472,7 +472,7 @@ void cpu_dead(void)
 		vmx_off();
 
 		stac();
-		flush_cache_range((void *)get_hv_image_base(), CONFIG_HV_RAM_SIZE);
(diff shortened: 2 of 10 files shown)


dongyaozu (approved):

dongyaozu:
@yakuizhao

dbkinder (changes requested):
I suspect there are additional changes needed.

hypervisor/include/arch/x86/asm/boot/ld_sym.h
```diff
@@ -14,5 +14,6 @@ extern uint8_t		ld_entry_end;
 extern const uint8_t	ld_trampoline_load;
 extern uint8_t		ld_trampoline_start;
 extern uint8_t		ld_trampoline_end;
+extern uint8_t		_ld_ram_size;
```
junjiemao1, line 17: Our coding guideline does not recommend variables starting with underscores. See https://projectacrn.github.io/latest/developer-guides/c_coding_guidelines.html#c-id-07-names-defined-by-developers-shall-not-start-with-underscore

dbkinder:
@junjiemao1  A third option could be to change the configurator (or if needed, add a new tool) to remove options it doesn't recognize and add options it notices are missing in the imported XML so it can be saved as "fixed", based on the schema definition (xsd) files for the release.

hypervisor/include/arch/x86/asm/boot/ld_sym.h
lifeix: done

yakuizhao (approved):
LGTM

wenlingz:
> > I suspect there are additional code and doc changes needed.
> 
> Hi David, this patch only does the change in HV side. Nanlin would find someone to help do the change in config tool side as well as the doc side.

merge firstly, @NanlinXie please help add doc part in future, thanks

PR #6667, opened 2021-10-13
doc: more scenario name fixes

dbkinder:
Update logical partition to simply partitioned.

Remaining references to logical partition are because the documentation
is validated for releases prior to the scenario renaming.  When those
are updated to be tested with a newer release, we'll update the
documentation accordingly.


dbkinder:
@NanlinXie @wenlingz How can we diagnose why CI is failing the LOC_test for this doc-only PR.  Networking problems?

Issue #6688, opened 2021-10-19
[DX] GSG automation - debian


NanlinXie:
We will use debian to automate:

Board inspector
ACRN packages

NanlinXie:
[External_System_ID] ACRN-7434

NanlinXie:
GSG and packaging document need to be updated.
https://projectacrn.github.io/latest/misc/packaging/README.html

dbkinder:
@NanlinXie What Docker licensing is required for users/developers?

gvancuts:
> @NanlinXie What Docker licensing is required for users/developers?

This is something the users will have to determine for themselves, based on their situation and set-up.

dbkinder:
We're focusing on Debian packaging for 2.8

PR #6692, opened 2021-10-19
deb-pkg: load msr driver in board_parse.py

hufenglin:
Tracked-On: #6688

### DIFF
--- a/misc/config_tools/board_inspector/legacy/board_parser.py
+++ b/misc/config_tools/board_inspector/legacy/board_parser.py
@@ -40,11 +40,37 @@ def vendor_check():
                     vendor_name = line.split(':')[1].strip()
                     return vendor_name == CPU_VENDOR
 
+def check_msr_files(cpu_dirs):
+    cpu_list = []
+    for cpu_num in os.listdir(cpu_dirs):
+        if cpu_num.isdigit():
+            if os.path.exists(os.path.join(cpu_dirs, "{}/msr".format(cpu_num))):
+                continue
+            else:
+                cpu_list.append(cpu_num)
+    return cpu_list
 
 def check_env():
     """Check if there is appropriate environment on this system"""
     if os.path.exists(PY_CACHE):
         shutil.rmtree(PY_CACHE)
+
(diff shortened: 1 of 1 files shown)


misc/config_tools/board_inspector/legacy/board_parser.py
```diff
@@ -46,6 +46,16 @@ def check_env():
     if os.path.exists(PY_CACHE):
         shutil.rmtree(PY_CACHE)
 
+    # check msr driver
+    res = subprocess.Popen("lsmod | grep -i ^msr",
+                            shell=True, stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE, close_fds=True)
+    line = res.stdout.readline().decode('ascii')
```
junjiemao1, line 53: The key here is the make sure `/dev/cpu/0/msr` does exist. With kernels that have `CONFIG_X86_MSR` being `y`, there is no module named `msr` but the file is still accessible.

Thus I would recommend to check the existence of `/dev/cpu/X/msr` where `X` is any directory under `/dev/cpu` (you may use `os.listdir` and `os.path.isdir` to get the list of directories there).

misc/config_tools/board_inspector/legacy/board_parser.py
```diff
@@ -46,6 +46,16 @@ def check_env():
...
+    if not line:
+        subprocess.Popen("sudo modprobe msr",
+                            shell=True, stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE, close_fds=True)
```
junjiemao1, line 57: Since `modprobe` can fail as well, we should check the existence of `/dev/cpu/X/msr` again and raise a fatal error if it does not exist even after probing the `msr` module. Check the status of the `modprobe` command may also do the trick.

BTW, we do not need `sudo` here as the script has already verified itself to be executed by the root user.

misc/config_tools/board_inspector/legacy/board_parser.py
hufenglin: fixed.

PR #6725, opened 2021-10-27
config_tools: remove hvlog parameter and update HV_RAM_START default address

Kunhui-Li:
1. remove hvlog parameter in bootargs for all platforms.
2. update HV_RAM_START default address from 2MB to 4MB.
3. add the check that the HV_RAM_START value should be larger
than or equal to 2MB and 2MB aligned.

Tracked-On: #6663

### DIFF
--- a/misc/config_tools/library/hv_cfg_lib.py
+++ b/misc/config_tools/library/hv_cfg_lib.py
@@ -88,6 +88,18 @@ def hv_size_check(str_val, branch_tag, item):
     if not is_numeric_check(str_val, branch_tag, item):
         return
 
+def hv_ram_start_check(hv_ram_start, prime_item, item):
+
+    err_dic = {}
+    if '0x' not in hv_ram_start and '0X' not in hv_ram_start:
+        key = "hv,{},{}".format(prime_item, item)
+        ERR_LIST[key] = "Address should be Hex format"
+
+    to_mb = (int(hv_ram_start, 16) / (1024 * 1024))
+    is_aligned = to_mb % 2
+    if to_mb < 2 or is_aligned != 0:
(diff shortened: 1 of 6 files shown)


Kunhui-Li:
> we need a check that the value should be larger than 2MB and 2MB aligned.

The check has been added, please review, thanks.

wenlingz:
@jsun26intel please help review and approve this PR, thanks

misc/config_tools/library/hv_cfg_lib.py
```diff
@@ -88,6 +88,18 @@ def hv_size_check(str_val, branch_tag, item):
...
+def hv_ram_start_check(hv_ram_start, prime_item, item):
+
+    err_dic = {}
+    if '0x' not in hv_ram_start and '0X' not in hv_ram_start:
+        key = "hv,{},{}".format(prime_item, item)
+        ERR_LIST[key] = "Address should be Hex format"
+
+    to_mb = (int(hv_ram_start, 16) / (1024 * 1024))
+    is_aligned = to_mb % 2
+    if to_mb <= 2 or is_aligned != 0:
```
jsun26intel, line 100: 2MB should be OK for the case, so here the condition should be "<2".

jsun26intel (changes requested):

misc/config_tools/library/hv_cfg_lib.py
Kunhui-Li: Updated, thanks.

wenlingz:
OK to verify

wenlingz:
try verify EHL

jsun26intel (approved):

Issue #6755, opened 2021-11-01
config-tools: round down the HV_SUPPORTED_MAX_CLOS


yuchuyang:
It may not be an even number all the time, rounds down to the nearest integer.

yuchuyang:
[External_System_ID] ACRN-7464

PR #6747, opened 2021-11-01
Unify terminology for service vm/user vm

liudlong:
•	User VM: A VM where user-defined environments and applications run. User VMs can run different OSes based on their needs, including for example, Ubuntu for an AI application, Android or Windows for a Human-Machine Interface, or a hard real-time control OS such as Zephyr, VxWorks, or RT-Linux for soft or hard real-time control. There are three types of User VMs: pre-launched, post-launched Standard and post-launched Real-Time User VMs.
•	Service VM: a special Virtual Machine (VM), directly launched by the hypervisor. The Service VM can access hardware resources directly by running native drivers and provides device sharing services to post-launched User VMs through the ACRN Device Model (DM). Hardware resources include CPUs, memory, graphics memory, USB devices, disk, and network mediation.

PR #6756, opened 2021-11-01
config-tools: refine acrn:get-common-clos-max

yuchuyang:
Rewrote the calculation, it should be the maximum number + 1 then divide by 1 or 2 if cdp is enabled.

Tracked-On: #6755

### DIFF
--- a/misc/config_tools/xforms/lib.xsl
+++ b/misc/config_tools/xforms/lib.xsl
@@ -400,34 +400,35 @@
     </xsl:choose>
   </func:function>
 
-  <func:function name="acrn:get-common-clos-max">
+  <func:function name="acrn:get-common-clos-count">
     <xsl:choose>
       <xsl:when test="not(acrn:is-rdt-enabled()) and not(acrn:is-cdp-enabled())">
         <func:result select="0" />
       </xsl:when>
       <xsl:otherwise>
         <xsl:variable name="rdt_resource_list" select="str:split(acrn:get-normalized-closinfo-rdt-res-str(), ',')" />
-        <xsl:variable name="rdt_res_clos_max_list" select="str:split(acrn:get-normalized-closinfo-rdt-clos-max-str(), ',')" />
+        <xsl:variable name="rdt_res_clos_count_list" select="str:split(acrn:get-normalized-closinfo-rdt-clos-max-str(), ',')" />
         <xsl:variable name="cdp_enabled" select="acrn:is-cdp-enabled()"/>
 
-        <xsl:variable name="clos_max_list_rtf">
+        <xsl:variable name="clos_count_list_rtf">
           <xsl:for-each select="$rdt_resource_list">
             <xsl:variable name="pos" select="position()" />
-            <xsl:variable name="res_clos_max" select="number($rdt_res_clos_max_list[$pos])" />
+            <xsl:variable name="res_clos_count" select="number($rdt_res_clos_count_list[$pos])" />
 
             <xsl:choose>
               <xsl:when test=". = 'MBA'">
-                <node><xsl:value-of select="$res_clos_max"/></node>
+                <node><xsl:value-of select="$res_clos_count"/></node>
               </xsl:when>
+              <!-- Some platforms have odd clos counts. Use "floor" to avoid this function returning a float number. -->
               <xsl:otherwise>
-                <node><xsl:value-of select="$res_clos_max div (1 + $cdp_enabled)"/></node>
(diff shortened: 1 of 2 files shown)


yuchuyang:
waiting for #6747 to be merged in order to run CI.

yuchuyang:
Please hold this PR for further investigation.

wenlingz:

misc/config_tools/xforms/lib.xsl
```diff
@@ -420,7 +420,7 @@
                 <node><xsl:value-of select="$res_clos_max"/></node>
               </xsl:when>
               <xsl:otherwise>
-                <node><xsl:value-of select="$res_clos_max div (1 + $cdp_enabled)"/></node>
+                <node><xsl:value-of select="floor($res_clos_max div (1 + $cdp_enabled))"/></node>
```
junjiemao1, line 423: The problem here is not the lack of a `floor`, but `max` is misused as `count`.

In case we want to reduce the number of CLOS available by half, the correct formula should be:

`($res_clos_max + 1) div (1 + $cdp_enabled) - 1`

misc/config_tools/xforms/lib.xsl
```diff
@@ -420,7 +420,7 @@
                 <node><xsl:value-of select="$res_clos_max"/></node>
               </xsl:when>
               <xsl:otherwise>
-                <node><xsl:value-of select="$res_clos_max div (1 + $cdp_enabled)"/></node>
+                <node><xsl:value-of select="floor(($res_clos_max + 1) div (1 + $cdp_enabled)) - 1"/></node>
```
junjiemao1, line 423: I don't think we need to invoke `floor` anymore. `$res_clos_max + 1` should always be a power of 2.

misc/config_tools/library/board_cfg_lib.py
```diff
@@ -573,7 +573,7 @@ def get_common_clos_max():
             if res == 'MBA':
                 tmp_clos_max_list.append(clos_max)
             else:
-                tmp_clos_max_list.append(clos_max//2)
+                tmp_clos_max_list.append((clos_max + 1)//2)
```
junjiemao1, line 576: Is this list meant to hold the max CLOS, not the number of CLOS? If so, you need `(clos_max + 1) // 2 - 1` instead.

junjiemao1 (changes requested):

misc/config_tools/xforms/lib.xsl
yuchuyang: Not sure if we will get new board xmls for ehl, tgl...etc. Seems they are either manual edited or really get even numbers in <CLOS_INFO>. I leave "floor" here for back compatibility.

misc/config_tools/library/board_cfg_lib.py
yuchuyang: Not sure if we will get new board xmls for ehl, tgl...etc. Seems they are either manual edited or really get even numbers in <CLOS_INFO>. I leave "floor" here for back compatibility.

yuchuyang: This is the number of count. It's used for checking if the <MBA_DELAY> and/or <CLOS_MASK> counts exceed the count maximum.

wenlingz:
@junjiemao1 please help review and approve this PR, thanks

misc/config_tools/xforms/lib.xsl
```diff
@@ -400,34 +400,35 @@
...
             <xsl:variable name="pos" select="position()" />
-            <xsl:variable name="res_clos_max" select="number($rdt_res_clos_max_list[$pos])" />
+            <xsl:variable name="res_clos_count" select="number($rdt_res_clos_count_list[$pos])" />
 
             <xsl:choose>
               <xsl:when test=". = 'MBA'">
-                <node><xsl:value-of select="$res_clos_max"/></node>
+                <node><xsl:value-of select="$res_clos_count"/></node>
               </xsl:when>
+              <!-- Some platform will get odd clos counts, use floor to avoid this function returns float number -->
```
junjiemao1, line 422:
```suggestion
              <!-- Some platforms have odd clos counts. Use floor to avoid this function returning a float number. -->
```

junjiemao1 (changes requested):

junjiemao1 (approved):

NanlinXie (approved):

wenlingz:
try verify EHL

PR #6774, opened 2021-11-03
doc: update last updated information on page footers

dbkinder:
The "Last Updated" footer shows the date the page was published, not
when the content was last updated. This has caused readers some
confusion.  Update the footer with the date of the git commit that last
updated the document, and clarify with both the last modified and
published dates in the footer, something like this:

   Last modified: Aug 19, 2021. Published: Nov 03, 2021.


dbkinder:
OK to verify

PR #6781, opened 2021-11-04
doc: terminology cleanup in HLD virtio

amyreye:
- Replace SOS or Service OS with Service VM
- Replace UOS or User OS with User VM
- Replace VHM with HSM
- Clean up some of the grammar

Signed-off-by: Amy Reyes <amy.reyes@intel.com>


doc/developer-guides/hld/hld-virtio-devices.rst
```diff
@@ -76,39 +76,39 @@ highlight several key virtio concepts important to ACRN:
...
-  produce requests, and kick backend virtio driver. As a result, the FE
+  driver merely needs to offer services that configure the interface, pass messages,
+  produce requests, and kick the backend virtio driver. As a result, the FE
   driver is easy to implement and the performance overhead of emulating
   a device is eliminated.
 
 Backend virtio driver (BE)
-  Similar to FE driver, the BE driver, running either in user-land or
+  Similar to the FE driver, the BE driver, running either in userland or
   kernel-land of the host OS, consumes requests from the FE driver and sends them
```
dbkinder, line 86: If we're changing user-land to userland, we should change kernel-land to kernelland right?  Although "kernelland" looks odd because of the double "l" so maybe best to leave that as is...

doc/developer-guides/hld/hld-virtio-devices.rst
```diff
@@ -216,9 +217,9 @@ virtqueues, feature mechanisms, configuration space, and buses.
 Virtio Framework Considerations
 ===============================
 
-How to realize the virtio framework is specific to a
+How to configure the virtio framework is specific to a
```
dbkinder, line 220: "configure" doesn't seem the right word here.  The original "realize" meant "to cause something to happen", so if you don't like "realize" a better word here would be "implement".

doc/developer-guides/hld/hld-virtio-devices.rst
```diff
@@ -362,15 +366,16 @@ general workflow of ioeventfd.
...
+   ioeventfd and irqfd.
+2. The vhost proxy passes the ioeventfd to the vhost kernel driver.
+3. The vhost proxy passes the ioeventfd to the HSM driver.
+4. The User VM FE driver triggers an ioreq, which is forwarded through the
+   hypervisor to the Service VM.
+5. The HSM driver dispatches the ioreq to the related HSM client.
+6. The ioeventfd HSM client traverses the io_range list and finds the
    corresponding eventfd.
-7. trigger the signal to related eventfd.
+7. The ioeventfd HSM client triggers the signal to the related eventfd.
```
dbkinder, line 378: We could combine steps 6 and 7 into:
```
6. The ioeventfd HSM client traverses the io_range list to find and trigger
   a signal to the corresponding eventfd.
```

doc/developer-guides/hld/hld-virtio-devices.rst
```diff
@@ -502,7 +507,7 @@ The key data structures for vhost are listed as follows.
 DM APIs
 =======
 
-The DM APIs are exported by DM, and they should be used when realizing
+The DM APIs are exported by DM, and they should be used when configuring
```
dbkinder, line 510: Same as previously, "configuring" isn't right here, but "implementing" would be OK.

dbkinder (changes requested):

doc/developer-guides/hld/hld-virtio-devices.rst
dbkinder: Doh.  I didn't look at the diagram.  OK then as is.

dbkinder (approved):

Issue #6709, opened 2021-10-25
Platform porting issues


NanlinXie:
When porting ACRN on some new platform, there's some issues found.

NanlinXie:
Closed since patch merged without regression.

PR #6790, opened 2021-11-08
doc: Update nuc11 partitioned document and Scenario xml

jackwhich:
Rename logical_partition to partitioned;
Use acrn hypervisor commit instead of tag v2.6;
Add partitioned.xml ramdisk parameter.

Tracked-On: #6709

### DIFF
--- a/doc/tutorials/using_partition_mode_on_nuc.rst
+++ b/doc/tutorials/using_partition_mode_on_nuc.rst
@@ -18,11 +18,7 @@ Validated Versions
 ******************
 
 - Ubuntu version: **18.04**
-- ACRN hypervisor tag: **v2.6**
-
-.. note:: After the v2.6 release, the ``logical_partition`` scenario
-   was renamed to ``partitioned``, affecting the file names used in this
-   guide (validated with v2.6).
+- ACRN hypervisor commit: **9bae63b941**
 
 Prerequisites
 *************
@@ -33,8 +29,8 @@ Prerequisites
 * Storage device with USB interface (such as USB Flash
   or SATA disk connected with a USB 3.0 SATA converter).
 * Disable **Intel Hyper Threading Technology** in the BIOS to avoid
-  interference from logical cores for the logical partition scenario.
-* In the logical partition scenario, two VMs (running Ubuntu OS)
+  interference from logical cores for the partitioned scenario.
+* In the partitioned scenario, two VMs (running Ubuntu OS)
   are started by the ACRN hypervisor. Each VM has its own root
   filesystem. Set up each VM by following the `Ubuntu desktop installation
   <https://tutorials.ubuntu.com/tutorial/tutorial-install-ubuntu-desktop>`_ instructions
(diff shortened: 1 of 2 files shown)


doc/tutorials/using_partition_mode_on_nuc.rst
```diff
@@ -18,11 +18,7 @@ Validated Versions
 ******************
 
 - Ubuntu version: **18.04**
-- ACRN hypervisor tag: **v2.6**
-
-.. note:: After the v2.6 release, the ``logical_partition`` scenario
-   was renamed to ``partitioned``, affecting the file names used in this
-   guide (validated with v2.6).
+- ACRN hypervisor commit: **9bae63b941**
```
dbkinder, line 21: @NanlinXie @jackwhich I presume this is a temporary commit SHA and we'll update this here (and later in the document) to be v2.7?

doc/tutorials/using_partition_mode_on_nuc.rst
```diff
@@ -213,17 +213,18 @@ Update Ubuntu GRUB to Boot Hypervisor and Load Kernel Image
...
            gfxmode $linux_gfx_mode
            insmod gzio
            insmod part_gpt
            insmod ext2
            search --no-floppy --fs-uuid --set 9bd58889-add7-410c-bdb7-1fbc2af9b0e1
-           echo 'Loading hypervisor logical partition scenario ...'
+           echo 'Loading hypervisor partitioned scenario ...'
            multiboot2  /boot/acrn.bin root=PARTUUID="e515916d-aac4-4439-aaa0-33231a9f4d83"
            module2 /boot/bzImage Linux_bzImage
+           module2 /boot/initrd_Image Ubuntu
```
gvancuts, line 227: There is no tag at the moment for `ramdisk_mod` in the scenario XML. We need additional instructions for the user because he needs to add this before compiling ACRN.

doc/tutorials/using_partition_mode_on_nuc.rst
```diff
@@ -145,15 +142,15 @@ Update ACRN Hypervisor Image
 
    .. code-block:: none
 
-      $ make hypervisor BOARD=nuc11tnbi5  SCENARIO=logical_partition RELEASE=0
+      $ make hypervisor BOARD=nuc11tnbi5 SCENARIO=partitioned RELEASE=0
```
gvancuts, line 145: While we're at it, I'd recommend we remove the `RELEASE` parameter, it does not add any value. Should we want to keep it, let's at least change it to the proper usage which is `RELEASE=[y|n]` these days.
```suggestion
      $ make hypervisor BOARD=nuc11tnbi5 SCENARIO=partitioned
```

gvancuts (changes requested):
This is PR is OK after some adjustment. **But** we really to review the entire set of instructions which do not make much sense. I doubt the validation team can follow this guide from start to the end and be successful  ;-)

doc/tutorials/using_partition_mode_on_nuc.rst
jackwhich: This parameter has been added to the scenario XML

gvancuts (approved):
Thanks... now all that's left to do (in a separate PR) is to re-write this tutorial so it makes some sense ;-)

dbkinder (approved):
OK, but let's remember to update that commit SHA to the v2.7 tag and verify? @NanlinXie

dbkinder:
@gvancuts So I guess we should merge this as is for now?

gvancuts:
> @gvancuts So I guess we should merge this as is for now?

If we will quickly follow with an update to indicate a newer commit ID or tag, then yes, I can live with a temporary breakage in that document. There are bigger problems with it in fact.

NanlinXie (approved):

PR #6791, opened 2021-11-08
hv: Trampoline size

YuanyuanZhao609:


hypervisor/arch/x86/trampoline.c
```diff
@@ -110,7 +110,7 @@ uint64_t prepare_trampoline(void)
 	uint64_t size, dest_pa;
 
 	size = (uint64_t)(&ld_trampoline_end - &ld_trampoline_start);
```
dongyaozu, line 112: Can you use same name and same method with above code?

dongyaozu: Even we can define an inline API somewhere in head file.

dongyaozu (changes requested):

hypervisor/arch/x86/trampoline.c
YuanyuanZhao609: OK, I will add an inline api.

YuanyuanZhao609: `size' is used to memory copy and cache flush,  should I round page up here.

dongyaozu: You can rename them to trampiline_size, trampoline_pa

YuanyuanZhao609: OK， I will rename `size` as `trampoline_size`, and when alloc memory it will round page up inside `e820_alloc_memory`.

wenlingz:
@YuanyuanZhao609 please add Tracked-On info and add JIRA in your git issue, thanks

dongyaozu (approved):

Issue #6235, opened 2021-06-22
config-tools: get the virtual msi table num based on msi and msi-x


yuchuyang:
Fix the issue that the MAX_MSIX_TABLE_NUM gets the wrong number

yuchuyang:
[External_System_ID] ACRN-7205

PR #6797, opened 2021-11-09
config-tools: refine MAX_MSIX_TABLE_NUM

chejianj:
If the MAX_MSIX_TABLE_NUM is specified in scenario.xml. Return the
largest number from count of MSI, table_size of MSIX or
MAX_MSIX_TABLE_NUM of scenario.xml.

Tracked-On: #6235

### DIFF
--- a/misc/config_tools/xforms/config_common.xsl
+++ b/misc/config_tools/xforms/config_common.xsl
@@ -258,11 +258,20 @@
         <xsl:if test="position() = 1"><xsl:value-of select="."/></xsl:if>
       </xsl:for-each>
     </xsl:variable>
-    <xsl:call-template name="integer-by-key-value">
-      <xsl:with-param name="key" select="'MAX_MSIX_TABLE_NUM'" />
-      <xsl:with-param name="value" select="$max" />
-      <xsl:with-param name="default" select="'64'" />
-    </xsl:call-template>
+    <xsl:choose>
+      <xsl:when test="//CAPACITIES/MAX_MSIX_TABLE_NUM/text()">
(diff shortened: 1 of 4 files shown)


misc/config_tools/xforms/config_common.xsl
```diff
@@ -258,11 +258,20 @@
...
+          <xsl:with-param name="value" select="acrn:max($max, //CAPACITIES/MAX_MSIX_TABLE_NUM)" />
+        </xsl:call-template>
+      </xsl:when>
+      <xsl:otherwise>
+        <xsl:call-template name="integer-by-key-value">
+          <xsl:with-param name="key" select="'MAX_MSIX_TABLE_NUM'" />
+          <xsl:with-param name="value" select="$max" />
+        </xsl:call-template>
+      </xsl:otherwise>
+    </xsl:choose>
```
junjiemao1, line 274: To calculate the max value among multiple nodes, you can simply use `{http://exslt.org/math}max`, rather than implement your own `max` function.

junjiemao1: Anyway, I don't think this should block this PR to release 2.5, but you can simplify this when you push this change to mainline as well.

chejianj:
This is already in mainline, it is by Yu-chu.
This is just back ported to v2.5 because I found a issue on a board.

junjiemao1:
> This is already in mainline, it is by Yu-chu. This is just back ported to v2.5 because I found a issue on a board.

Alright then.

yakuizhao (approved):
LGTM to backport it on release_2.5 branch.

Issue #6767, opened 2021-11-03
[acrn-configuration-tool] Getting duplicate PT_SLOT value If generate launch script continuously through the UI


Kunhui-Li:
**environment**
**HW/Board**: common
**Image info**: Issue not related to image
**Steps**:
1. launch config tool ui:
`python misc/config_tools/config_app/acrn_configurator.py` 

2. Import nuc11 board xml and then load default shared scenario
3. load shared_launch_6uos launch xml
4. add 2 more virtio block devices for VM1 and rename them as 1.img , 2.img
5. Generate Launch Script
6. Remove step 4 's block device and then generate launch script
7. Repeat step 4 - 6 (you need to change the new block device name just difference with the previous one)
8. Finally you will get this from the launch script generated from the step 7:

   `-s 4,virtio-blk,./win10-ltsc.img \
   -s 30,virtio-blk,./win10-ltsc12.img \
   -s 30,virtio-blk,./win10-ltsc13.img \
   -s 7,virtio-net,tap_WaaG \
   --ovmf /usr/share/acrn/bios/OVMF.fd \
   -s 1:0,lpc \
   $vm_name`

PT_SLOT value are duplicated...

**Reproduce rate**: 10/10
**Impact**:
Not regression issue but impact automation execution quality.

Kunhui-Li:
[External_System_ID] ACRN-7461

PR #6810, opened 2021-11-10
config_tools: fix the duplicate PT_SLOT value issue in UI

Kunhui-Li:
when we add two or more virtio block devices and rename them for a VM
in launch XML files, then generate launch scripts through the UI.
After ensuring the devices names are different, we repeat these operations
continuously. Finally, we will get the duplicate PT_SLOT value for
different virtio block devices in a launch script. This is wrong.

After the user click the "Generate Launch Script" button to generate the
launch scripts every time, we will reset PT_SLOT variate to fix the issue.

Tracked-On: #6767


NanlinXie (approved):

wenlingz:
OK to verify

wenlingz:
try verify EHL

PR #6811, opened 2021-11-10
config_tools: fix the duplicate PT_SLOT value issue in UI

Kunhui-Li:
when we add two or more virtio block devices and rename them for a VM
in launch xml files, then generate launch scripts through the UI.
After ensuring the devices names are different, we repeat these operations
continuously. Finally we will get the duplicate PT_SLOT value for
different virtio block devices in a launch script. This is wrong.

After user click the "Generate Launch Script" button to generate the
launch scripts every time, we will reset PT_SLOT variate to fix the issue.

Tracked-On: #6767


wenlingz:

NanlinXie (approved):

Issue #6366, opened 2021-08-09
TPM pass-thru shall be able to support start method 6, not only support Start Method of 7


lifeix:
[External_System_ID] ACRN-7097

wenlingz:
[External_System_ID] ACRN-7097

NanlinXie:
Closed since patch merged without regression.

PR #6837, opened 2021-11-15
config_tools: remove GUEST_FLAG_SECURITY_VM in DM_OWNED_GUEST_FLAG_MASK

Kunhui-Li:
GUEST_FLAG_SECURITY_VM only for pre-launch VM, should not be
programmed by device model, so we remove it in DM_OWNED_GUEST_FLAG_MASK.

Tracked-On: #6366


jsun26intel (commented):
I think config tool might not need to define this macro because it is not configurable.
Maybe the macro should be moved to vm_config.h

yfliuuu:
> I think config tool might not need to define this macro because it is not configurable. Maybe the macro should be moved to vm_config.h

I think user still need to be able to configure this option in xml for per-vm fixup behavior. @lifeix your comments?

jsun26intel:
@yfliuuu this is DM_OWNED mask, why post launched VM need this flag?

yfliuuu:
> @yfliuuu this is DM_OWNED mask, why post launched VM need this flag?

Post launched VM does not need this, but config tool still need to keep this for user to enable fixup for prelaunch VM. I'm replying to your comment of removing it completely from config tool.

jsun26intel:
I mean it could be moved to vm_config.h or somewhere, that doesn't need tool to generate it since it has no configure dependency.

jsun26intel:
the flag of GUEST_FLAG_SECURITY_VM should be set in scenario xml for pre-launched VM only.

yfliuuu:
> I mean it could be moved to vm_config.h or somewhere, that doesn't need tool to generate it since it has no configure dependency.

Oh you mean move the DM_OWNED_GUEST_FLAG_MASK to vm_config.h? I thought you were referring to GUEST_FLAG_SECURITY_VM. If that's the case then I have no comments.

Kunhui-Li:
Please hold this PR for HV change.

NanlinXie (approved):

Kunhui-Li:
@wenlingz please help remove the label "hold on", the HV patches are ready now, thanks.

wenlingz:

wenlingz:
OK to verify

Issue #6848, opened 2021-11-17
[RTVM]ACPI error while launching rtvm


izhouwu:
Image info

SOS kernel:Linux intel-corei7-64 5.10.26-acrn-kernel-sos 

HV:HV 2.6-unstable-2021-07-02 15:48:11-20e3b0d3f DBG (daily tag: ) industry build by shuang

DM version is: 2.6-unstable-20e3b0d3f (daily tag, build by shuang@2021-07-02 15:48:21

Steps

1.boot sos

2. launch rtvm script

Expected result

no ACPI error printed while rtvm launched

Actual result

print ACPI error

Reproduce rate

100%

Kernel error message:
[ 12.885255] 003: [Firmware Bug]: BIOS needs update for CPU frequency support
[ 12.892308] 003: ACPI Error: AE_NOT_FOUND, Evaluating _PSS (20190816/processor_perflib-303)
[ 12.892332] 003: [Firmware Bug]: BIOS needs update for CPU frequency support
[ 12.916308] 003: ACPI Error: AE_NOT_FOUND, Evaluating _PSS (20190816/processor_perflib-303)
[ 12.916335] 003: [Firmware Bug]: BIOS needs update for CPU frequency support
[ 12.977611] 000: ACPI Error: AE_NOT_FOUND, Evaluating _PSS (20190816/processor_perflib-303)
[ 12.977638] 000: [Firmware Bug]: BIOS needs update for CPU frequency support
[ 12.984059] 000: ACPI Error: AE_NOT_FOUND, Evaluating _PSS (20190816/processor_perflib-303)
[ 12.984109] 000: [Firmware Bug]: BIOS needs update for CPU frequency support
[ 12.987525] 000: ACPI Error: AE_NOT_FOUND, Evaluating _PSS (20190816/processor_perflib-303)
[ 12.987548] 000: [Firmware Bug]: BIOS needs update for CPU frequency support
[ 12.988681] 000: ACPI Error: AE_NOT_FOUND, Evaluating _PSS (20190816/processor_perflib-303)
[ 12.991280] 000: [Firmware Bug]: BIOS needs update for CPU frequency support
[ 13.061312] 000: ACPI Error: AE_NOT_FOUND, Evaluating _PSS (20190816/processor_perflib-303)
[ 13.061341] 000: [Firmware Bug]: BIOS needs update for CPU frequency support
[...]

PR #6850, opened 2021-11-17
[release_2.7] Fix kernel ACPI error msg while launching rtvm

izhouwu:
This patch is to eliminate kernel error msgs:
'ACPI Error: AE_NOT_FOUND, Evaluating _PSS'

This is caused by missing of _PSS table in guest ACPI. It would
happen when pstate is not injected to the guest.

Kernel ACPI pstate driver first probes
_PPC(performance capabilites) and _PCT(performance control)
in ACPI. If they exist, then it loads the _PSS(performance state).
If _PPC/_PCT are presented while _PSS is missing, it prints
the error msg.

In acrn-dm, _PPC/_PCT are hard-coded to all vCPUs, while _PSS
are constructed with the pCPUs' pstate data. This is base on
assumption that all VMs can have pstate.

Now the pstate is given to VM only when the VM is not sharing
any CPU(and no RTVM is setup in the scenario).
When the VM doesn't have pstate, the hypercall will return px_cnt=0,
and the _PSS is not constructed. In this case, _PPC/PCT should not
be injected, too.

Tracked-On: #6848

Signed-off-by: Zhou, Wu wu.zhou@intel.com
Acked-by: Anthony Xu anthony.xu@intel.com
Acked-by: Eddie Dong eddie.dong@Intel.com

### DIFF
--- a/devicemodel/hw/platform/acpi/acpi_pm.c
+++ b/devicemodel/hw/platform/acpi/acpi_pm.c
@@ -254,7 +254,7 @@ static void dsdt_write_pct(void)
 
 /* _PSS: Performance Supported States
  */
-static void dsdt_write_pss(struct vmctx *ctx, int vcpu_id)
+static int dsdt_write_pss(struct vmctx *ctx, int vcpu_id)
 {
 	uint8_t vcpu_px_cnt;
 	int i;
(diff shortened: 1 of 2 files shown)


wenlingz:

yakuizhao (approved):
LGTM

Issue #6855, opened 2021-11-17
config-tools: get GPU_SBDF automatically


yuchuyang:
Get the GPU_SBDG by looking for device from board.xml and extract the BDF from device address.

yuchuyang:
[External_System_ID] ACRN-7444

PR #6856, opened 2021-11-17
config-tools: get GPU_SBDF automatically

yuchuyang:
Get the GPU_SBDF by looking for device from board.xml and extract the
BDF from device address.

Tracked-On: #6855

### DIFF
--- a/misc/config_tools/static_allocators/bdf.py
+++ b/misc/config_tools/static_allocators/bdf.py
@@ -119,7 +119,23 @@ def create_device_node(allocation_etree, vm_id, devdict):
         if common.get_node(f"./func", dev_node) is None:
             common.append_node(f"./func", f"{bdf.func:#04x}", dev_node)
 
+def create_gpu_sbdf(board_etree, allocation_etree):
+    """
+    Extract the integrated GPU bdf from board.xml. If the device is not present, set bdf to "0xFFFF" which indicates the device
+    doesn't exist.
+    """
+    bus = "0x0"
+    device_node = common.get_node(f"//bus[@type='pci' and @address='{bus}']/device[vendor='0x8086' and class='0x030000']", board_etree)
+    if device_node is None:
+        common.append_node("/acrn-config/hv/MISC_CFG/GPU_SBDF", '0xFFFF', allocation_etree)
+    else:
+        address = device_node.get('address')
+        dev = int(address, 16) >> 16
+        func = int(address, 16) & 0xffff
+        common.append_node("/acrn-config/hv/MISC_CFG/GPU_SBDF", f"{(int(bus, 16) << 8) | (dev << 3) | func:#06x}", allocation_etree)
+
 def fn(board_etree, scenario_etree, allocation_etree):
+    create_gpu_sbdf(board_etree, allocation_etree)
     vm_nodes = scenario_etree.xpath("//vm")
     for vm_node in vm_nodes:
         vm_id = vm_node.get('id')
--- a/misc/config_tools/static_allocators/bdf.py
+++ b/misc/config_tools/static_allocators/bdf.py
@@ -119,7 +119,7 @@ def create_device_node(allocation_etree, vm_id, devdict):
         if common.get_node(f"./func", dev_node) is None:
             common.append_node(f"./func", f"{bdf.func:#04x}", dev_node)
 
-def create_gpu_sbdf(board_etree, allocation_etree):
+def create_igd_sbdf(board_etree, allocation_etree):
     """
     Extract the integrated GPU bdf from board.xml. If the device is not present, set bdf to "0xFFFF" which indicates the device
     doesn't exist.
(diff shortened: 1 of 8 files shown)


misc/config_tools/static_allocators/bdf.py
```diff
@@ -119,7 +119,25 @@ def create_device_node(allocation_etree, vm_id, devdict):
...
 
+def create_gpu_sbdf(board_etree, allocation_etree):
+    bus = "0x0"
+    hv_node = common.get_node(f"/acrn-config/hv", allocation_etree)
+    if hv_node is None:
+        hv_node = common.append_node("/acrn-config/hv", None, allocation_etree)
+    misc_cfg_node = common.get_node(f"./MISC_CFG", hv_node)
+    if misc_cfg_node is None:
+        misc_cfg_node = common.append_node("./MISC_CFG", None, hv_node)
+    device_node = common.get_node(f"//bus[@type='pci' and @address='{bus}']/device[vendor='0x8086' and class='0x030000']", board_etree)
```
junjiemao1, line 130: There is the method `append_node` in `common.py` which can help you create those ancestor nodes with one single call.

misc/config_tools/static_allocators/bdf.py
```diff
@@ -119,7 +119,25 @@ def create_device_node(allocation_etree, vm_id, devdict):
...
+    if device_node is None:
+        common.append_node("./GPU_SBDF", '0x00000010', misc_cfg_node)
```
junjiemao1, line 132: Are we sure `0x00000010` is the right value when there is no integrated GPU on the platform?

junjiemao1 (changes requested):

misc/config_tools/static_allocators/bdf.py
yuchuyang: @junjiemao1  here is the message from Yakui:
"Now the CONFIG_GPU_SBDF is only used in ACRN_hypervisor. And it is treated as the integrated Intel GPU card.

For the platform without iGD device,  it doesn’t matter that the CONFIG_GPU_SBDF=0x010U is configured. As there is no physical device, it will not be conflicted with other device.

In DM it is already checked by using is_intel_graphics_dev. "

yuchuyang: I think it's fine if it is configured to "0x00000010" even there isn't one.

junjiemao1: @yuchuyang @yakuizhao Unfortunately `there is no physical device` may not be true. On platforms w/o iGPU, the BDF `00:02.0` may be used by other devices. There is no spec constraining that `00:02.0` is reserved for iGPU.

yakuizhao: @junjiemao1 , the CONFIG_GPU_SBDF is used only when the iGPU is assigned to pre-launched VM.
 on the platforms with iGPU it is observed that the iGPU resides in the 0:02.0 bdf.  Maybe this will be changed in future platforms. But who knows.

And the code will also check the PCI_DISPLAY_class to detect whether the 0:02.0 bdf is the GPU device. If the 0:2:0 bdf is not one iGPU device, the CONFIG_GPU_SBDF=0x00010  will have no effect except the 0:02.0 device is also assigned to pre-launched VM.
If the 0:02.0 device doesn't exist, it doesn't matter for adding this definition. After checking several platforms the 0:02.0 BDF doesn't exist if there is no iGPU device.

junjiemao1: It should be fine if the iGPU exists as the config tool will detect where it really is.

Today we do have servers which do not have iGPU but 0:02.0 allocated to a root port. So hard-coding `CONFIG_GPU_SBDF` to `0x10` will cause troubles if anyone (unfortunately) assigns that to a pre-launched VM. The value `0x0` (i.e. BDF `00:00.0`) seems a better fallback value in this case as we always emulate the hostbridge at that BDF.

yakuizhao: BDF 00:00.0 is also one valid pci_device. It will be conflicted with the real device.
How about using FF:1f.7 device when iGPU doesn't exist?

junjiemao1: Well, that's true as we also use this value when virtualizing config space accesses.

I'm fine to use all 1's to indicate the non-presence of iGPU.

jsun26intel:
can we rename CONFIG_GPU_SBDF to CONFIG_IGD_SBDF?

misc/config_tools/static_allocators/bdf.py
yuchuyang: It may be a silly question, although the discussion here is for the non-present iGPU, it wouldn't matter if we set it's 0xFFFF or 0xFFFFFFFF. Would 0xFFFFFFFF be ok here and for future compatibility or it has special meaning in spec?

yuchuyang:
> can we rename CONFIG_GPU_SBDF to CONFIG_IGD_SBDF?

In that case the hv code need to be renamed as well?

yakuizhao:
> > can we rename CONFIG_GPU_SBDF to CONFIG_IGD_SBDF?
> 
> In that case the hv code need to be renamed as well?

Yes.  I think that it is fine to rename it to CONFIG_IGD_SBDF. 
If this is renamed, the hv code needs to be changed.
Another is that the CONFIG_GPU_SBDF is only used in hypervisor/dm/vpci/pci_pt.c. Can we totally remove the CONFIG_GPU_SBDF? Then the PCI_IGD_SBDF is defined and used internally.

yuchuyang:
@junjiemao1 @jsun26intel @yakuizhao have push new changes. Let me know if there is any concern.

yuchuyang:
Updated based on @junjiemao1 feedback. Let me know if there is any concern.

junjiemao1:
Fine with me. @jsun26intel @yakuizhao Please help double check. Thanks.

yakuizhao:
> Fine with me. @jsun26intel @yakuizhao Please help double check. Thanks.

It is fine to me that the GPU_SBDF is renamed to IGD_SBDF for HV.

NanlinXie (approved):

PR #6869, opened 2021-11-20
doc: terminology cleanup in images (sriov)

amyreye:
- Replace SOS or Service OS with Service VM
- Replace UOS or User OS with User VM

Signed-off-by: Amy Reyes <amy.reyes@intel.com>


gvancuts (approved):

dbkinder (approved):
In the future, I'm inclined to change these pictures of screen shots that are all text, into a code block with "console" at the language (so they can be easily updated as text).

PR #6879, opened 2021-11-23
config_tool: set MAX_VM_NUM value range from 1 to 8 and validate vm_num

Weiyi-Feng:
1. set MAX_VM_NUM value range from 1 to 8.
2. validate vm_num must <= MAX_VM_NUM at export scenarion xml

Tracked-On: #6685
Signed-off-by: Weiyi Feng <weiyix.feng@intel.com>


jsun26intel (approved):

wenlingz:
No New Name Conflict

gvancuts:
Is 8 a hard limit?

I'm thinking about customers who may want to try with more than 8 VMs, does this force them to patch the config tool to try it out?

jsun26intel:
@gvancuts How many VMs that ACRN could run depends on memory layout under 4GB. More VMs means huger hypervisor binary. Some special board only has ~1G space under 4GB.  If bootloader could not put hypervisor binary and kernel modules under 4GB, the boot will fail.
8 VMs is more generic for most boards could run and so far no requirement on this. If users want to run more than 8 VMs, better patch the config tool by themselves and might be risk to boot.

wenlingz:
OK to verify

wenlingz:
OK to verify

NanlinXie (approved):

misc/config_tools/schema/types.xsd
```diff
@@ -50,6 +50,16 @@
...
 
+<xs:simpleType name="MaxVMNumType">
+  <xs:annotation>
+    <xs:documentation>Integer from 1 to 8.</xs:documentation>
+  </xs:annotation>
+  <xs:restriction base="xs:integer">
+    <xs:minInclusive value="1" />
+    <xs:maxInclusive value="8" />
+  </xs:restriction>
+</xs:simpleType>
```
junjiemao1, line 61: Creating a separate integer type solely due to a different range seems an overkill to me. You can embed the type in where it is used directly. See https://www.w3schools.com/xml/el_restriction.asp for examples.

We may need to clean up the types here, but in a future PR probably.

Issue #6888, opened 2021-11-24
Yaag reboots too slowly


conghuic23:
yaag reboots too slowly

conghuic23:
[External_System_ID] ACRN-7297

NanlinXie:
Closed since patch merged without regression.

PR #6889, opened 2021-11-24
hv: sched: fix bug when reboot vm

conghuic23:
When a new thread is wakeup and added to runqueue, it will get the
smallest avt (svt) from runqueue to initiate its avt. But when all
threads are blocked, and removed from the runqueue, the svt is 0. The
new thread will get avt=0. avt=0 means very high priority, and can run
for a very long time until it catchup with other thread's avt in
runqueue.

To fix it, adjust the new thread's avt when there is another thread in
runqueue.

Tracked-On: #6888


acrnsi:

wenlingz:

yakuizhao (approved):
LGTM

Issue #6571, opened 2021-09-15
Support X86_TEE


ssqre:
[External_System_ID] ACRN-7399

ssqre:
ACRN supports X86_TEE which can emulate the ARM TrustZone on x86 architecture

1. priority based scheduler support
2. TEE VM (X86 OPTEE) boot support, TEE VM memmap support
3. TEE and REE hypercall support
4.  Interrupt flow  Support

wenlingz:
[External_System_ID] ACRN-7399

dbkinder:
@NanlinXie @ssqre What's the status of this, and is documentation needed?

NanlinXie:
Closed since patch merged without regression.

PR #6910, opened 2021-11-29
hv: Add stateful VM check before system shutdown

yfliuuu:
This patch introduces stateful VM which represents a VM that has its own
internal state such as a file cache, and adds a check before system
shutdown to make sure that stateless VM does not block system shutdown.

Tracked-On: #6571
Signed-off-by: Wang Yu <yu1.wang@intel.com>


acrnsi:

dongyaozu (approved):

Issue #6908, opened 2021-11-29
Shutdown one RT and others will hang when launch multi RT


wuxyintel:
Environment:
HW: ADL-S

ACRN version: 2.6_RC5: http://sit-image.sh.intel.com/acrn_daily_build/YOCTO-2021ww38.1-1/yocto-adl-s-crb-uefi-industry/

RT kernel: http://sit-image.sh.intel.com/yocto/e2e/5.10/yocto/ww35_bullpen_rt_56/

SOS cmdline:

root@intel-corei7-64:~/RTLaaG/2.6_multi# uname -r

5.10.56-acrn-kernel-sos

root@intel-corei7-64:~/RTLaaG/2.6_multi# cat /proc/cmdline | more

root=/dev/nvme0n1p3 console=ttyS0 idle=halt rw rootwait console=ttyS0 earlyprintk=serial,ttyS0,115200 cons_timer_check consoleblank=0 no_timer_check quiet loglevel=3 i915.nuclear_pageflip=1 nokaslr maxcpus=16 hugepagesz=1G hugepages=28  root=/dev/nvme0n1p3 rootfstype=ext4 rootwait console=ttyS0,115200 console=tty0

hvlog=2M@0xE00000 memmap=2M$0xE00000 i915.enable_gvt=1 i915.nuclear_pageflip=1 hugepages=2 i915.modeset=0

RT cmdline:

root@intel-corei7-64:~# uname -r

5.10.56-rt48-acrn-kernel-preempt-rt-uos

root@intel-corei7-64:~# cat /proc/cmdline | more

BOOT_IMAGE=/bzImage root=PARTUUID=e23b8560-8faf-46c3-81a4-9013c87f0df6 rootwait

rootfstype=ext4 rootwait console=ttyS0,115200 console=tty0 rw nohpet console=hvc

0 console=ttyS0 no_timer_check ignore_loglevel log_buf_len=16M consoleblank=0 ts

c=reliable i915.nuclear_pageflip=1

Bug detailed description:
Shutdown one RT and others will hang when launch multi RT.
Reproduce steps:

Prepare:

Flash 2.6_rc5 build.
 Boot up HV and SOS
 Boot up multi RT and shutdown one.
 

 

Current result:
Shutdown one RT and others will hang when launch multi RT.
Expect result:

Shutdown one RT and others will work normal when launch multi RT.

wuxyintel:
[External_System_ID] ACRN-7499

PR #6921, opened 2021-11-30
HV: update destination shorthand during x2apic ICR emulation

wuxyintel:
Currently, in RTVM with multi vCPUs, lapic pass through is
configured, each vCPU works in x2apic mode. When one vCPU sends
IPI to all other vCPUs through writes ICR register with virtual
value 0x00000000000c00f8, this ICR writting will be intercepted,
the hypervisor passes destination shorthand field 11B (All Excluding
Self) in the virtual ICR value into physical ICR value during IPI
emulation, this IPI will be sent to each physical CPU core
in the platform according to 10.6.1 Interrupt Command Register (ICR),
Vol 3, SDM.
One vCPU in User VM with lapic pass through configuration can
send IPI with destination shorthand (10B or 11B) and any vector
(such as NMI or reboot vector) to other vCPUs, this IPI will sent
other VMs in the platform by hypervisor, this interference may
cause other VMs hang.

In this patch, set "Destination Shorthand" field of the
ICR value as 00B (No Shorthand) since the emulation is done
through sending IPI to each VCPU in dmask one by one.

Tracked-On: #6908

Reviewed-by: Chen, Jason CJ <jason.cj.chen@intel.com>

### DIFF
--- a/hypervisor/arch/x86/guest/vlapic.c
+++ b/hypervisor/arch/x86/guest/vlapic.c
@@ -2058,6 +2058,11 @@ vlapic_x2apic_pt_icr_access(struct acrn_vcpu *vcpu, uint64_t val)
 	} else {
 		dmask = vlapic_calc_dest(vcpu, shorthand, (dest == 0xffffffffU), dest, phys, false);
 
+		/**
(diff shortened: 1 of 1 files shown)


acrnsi:

dongyaozu (approved):

PR #6922, opened 2021-11-30
deb-pkg: Support specifying board.xml and scenario.xml

hufenglin:
Currently, if board.xml and scenario.xml are specified,
the variable $(BOARD) and $(SCENARIO) are path, but the
parameters passed in to generate debian package should be
board and scenario name. So add a conversion function to
ensure that the function passed to the generated debian
package is board and scenario name.

Tracked-On: #6688

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,9 @@ ASL_COMPILER ?= $(shell which iasl)
 .PHONY: all hypervisor devicemodel tools life_mngr doc
 all: hypervisor devicemodel tools
 	@cat $(HV_CFG_LOG)
-	@python3 misc/packaging/gen_acrn_deb.py acrn_all $(ROOT_OUT) --version=$(FULL_VERSION) --board_name=$(BOARD) --scenario=$(SCENARIO)
+	@DEB_BOARD=$$(grep "BOARD" $(HV_CFG_LOG) | awk -F '= ' '{print  $$2 }'); \
+	DEB_SCENARIO=$$(grep "SCENARIO" $(HV_CFG_LOG) | awk -F '= ' '{print  $$2 }'); \
+	python3 misc/packaging/gen_acrn_deb.py acrn_all $(ROOT_OUT) --version=$(FULL_VERSION) --board_name="$$DEB_BOARD" --scenario="$$DEB_SCENARIO"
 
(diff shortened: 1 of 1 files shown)


acrnsi:

Makefile
```diff
@@ -79,11 +79,13 @@ BUILD_TAG ?=
...
 VM_CONFIGS_DIR = $(T)/misc/config_tools
 ASL_COMPILER ?= $(shell which iasl)
+DEB_BOARD = $(notdir $(BOARD))
+DEB_SCENARIO = $(notdir $(SCENARIO))
 
 .PHONY: all hypervisor devicemodel tools life_mngr doc
 all: hypervisor devicemodel tools
 	@cat $(HV_CFG_LOG)
-	@python3 misc/packaging/gen_acrn_deb.py acrn_all $(ROOT_OUT) --version=$(FULL_VERSION) --board_name=$(BOARD) --scenario=$(SCENARIO)
+	@python3 misc/packaging/gen_acrn_deb.py acrn_all $(ROOT_OUT) --version=$(FULL_VERSION) --board_name=$(basename $(DEB_BOARD)) --scenario=$(basename $(DEB_SCENARIO))
```
junjiemao1, line 88: When `BOARD` or `SCENARIO` is a file, the board/scenario names shall be fetched from the XML files, not the file names, as there is no guarantee that the file names and board/scenario names are identical.

Actually we already have such logic in the hypervisor (see the macro `determine_config` in hypervisor/scripts/makefile/config.mk). As the hypervisor is the only consumer of the `BOARD`/`SCENARIO` parameters, the top-level Makefile didn't need to parse them itself previously. But now with the introduction of the all-in-one package, this Makefile has to parse those parameters now, which is what causes the problem here.

If we want to go along this way, we have to copy some of the logic from hypervisor/scripts/makefile/config.mk here.

wenlingz:
No New Name Conflict

Makefile
hufenglin: Fixed. Get BOARD and SCENARIO information from the file $(HV_CFG_LOG).

wenlingz:

Makefile
junjiemao1: Fine with me.

NanlinXie (approved):

Issue #6934, opened 2021-12-01
Remove CONFIG_LOG_DESTINATION


weichenli-intel:
Remove CONFIG_LOG_DESTINATION from hv and config tool, check the loglevel of each channel, close the
output when the loglevel is ZERO.

[External_System_ID] ACRN-7432

NanlinXie:
[External_System_ID] ACRN-7432

NanlinXie:
Patch merged and close with no regression.

PR #6935, opened 2021-12-01
remove CONFIG_LOG_DESTINATION from hv and config tool

weichenli-intel:
The CONFIG_LOG_DESTINATION parameter selects where the logging messages
send to，serial console or memory or npk device MMIO region.

Now we want to remove it and check the loglevel of each channel,close the
output when the loglevel is ZERO.

Tracked-On: #6934


acrnsi:

PR #6939, opened 2021-12-02
deb-pkg: Support specifying board.xml and scenario.xml

hufenglin:
Currently, if board.xml and scenario.xml are specified,
the variable $(BOARD) and $(SCENARIO) are path, but the
parameters passed in to generate debian package should be
board and scenario name. So add a conversion function to
ensure that the function passed to the generated debian
package is board and scenario name.

Tracked-On: #6688


acrnsi:

yakuizhao (approved):
LGTM

PR #6957, opened 2021-12-08
doc: update s5 document

wuxyintel:
For WaaG VM, the User VM name is hard coded in the
lifecycle manager, this User VM ("windows") is needed
for guest shutdown.
Note: libvirt will be used to do guest shutdown and
the WaaG VM name will be configurable in furture.

The lifecycle manager starup picture in WaaG VM is out
of data, update it in this patch.

Tracked-on: #6652


acrnsi:

doc/tutorials/enable_s5.rst
```diff
@@ -290,6 +290,7 @@ How to Test
       sudo python3 ~/user_vm_shutdown.py <User VM name>
 
    .. note:: The User VM name is configured in the :file:`life_mngr.conf` of User VM.
+   .. note:: For the WaaG VM, the User VM name is "windows".
```
dbkinder, line 293: Better would be to combine these two notes into one:

```suggestion
      For the WaaG VM, the User VM name is "windows".
```

dbkinder (changes requested):
We really don't want folks manually editing XML files. Can the changes shown in this tutorial be made using the Configurator now (in v2.7)?

doc/tutorials/enable_s5.rst
wuxyintel: Thanks, will update this. BTW, I think current config tool doesn't support configure communication vuart for s5 in this document. @NanlinXie Can you help double check this? thanks!

dbkinder (approved):
OK, Thanks.

Issue #6584, opened 2021-09-17
check VMX capability


mingqiangchi:
[External_System_ID] ACRN-7403

wenlingz:
[External_System_ID] ACRN-7403

PR #6958, opened 2021-12-08
hv:remove 2 bits vmx capability check

mingqiangchi:
remove is_valid_xsave_combination api,
assume the hardware or QEMU can guarantee that support
XSAVE on CPU side and XSAVE_XRSTR on VMX side or not.
will add offline-tool in QEMU platform to avoid the user
use wrong XSAVE configurations.
remov check VMX_PROCBASED_CTLS2_XSVE_XRSTR based on the above reason.
for VMX_PROCBASED_CTLS2_PAUSE_LOOP, now it will panic
if run ACRN over QEMU, here remove it from essential check,
and it will print error information when set this bit
if there is no the hardware capability.

v1-v2:
  remove is_valid_xsave_combination

Tracked-On: #6584


acrnsi:

wenlingz:

anthonyzxu (approved):

Issue #5921, opened 2021-04-09
Decouple libvirt with acrnctl and support all acrnctl functionality on virsh (Phase 1)


NanlinXie:
1. Decouple libvirt with acrnctl and support some acrnctl’s functionality on virsh
2. Enable guest auto-boot in ACRN libvirt 
<html>
<body>

Acrnctl command | Libvirt functionality | Libvirt Features in ACRN v2.7
-- | -- | --
acrnctl add | define | Yes
acrnctl del | undefine | Yes
acrnctl list | list | Yes
acrnctl start | start | Yes
acrnd | autostart | Yes

</body>
</html>

NanlinXie:
[External_System_ID] ACRN-6920

NanlinXie:
Doc Impact Analysis: 

- [ ] Does the change require HLD update? If yes, need update HLD.

- [ ] Does the change introduce any hypervisor configuration data? If yes, need update configuration item description.

- [ ] Does the change add or modify any kernel module parameter? If yes, need update kernel module parameter description.	

- [ ] Does the change add or modify any DM command line option? If yes, need update DM cmdline option.

- [ ] Does the change add or modify any command line option in acrn-tools?	

- [ ] Does the change add or modify any command to acrn shell (i.e. the HV UART shell for debugging)?	

- [ ] Does the change modfy the acrn-hypervisor/Makefile? If yes, need check any change to meta-acrn code and doc.

- [ ] Does the change require for Service VM, library or service configuration files? If yes, need check any change to meta-acrn

gvancuts:
There are a few tutorials that reference `acrnctl`: "S5 enabling", "Kata Containers" and "OpenStack/libvirt"

Kata implementation also uses `acrnctl` so if we plan to deprecate or make the tool a debug tool like it was originally intended to be, we'll need to make adjustements to the Kata implementation and user guide.

NanlinXie:
Postpone to v2.8

dbkinder:
@NanlinXie Is this still a v2.8 target?

NanlinXie:
We will not drop acrnctl in v2.8, thus there's no change for v2.8 document.

NanlinXie:
GSG document need to be updated after this patch, we need add addtional library dependency at build time.
sudo apt-get install libcjson-dev

gvancuts:
I do not believe any documentation has been updated to reflect this change from `acrnctl` to `virsh`.

[...]

PR #6963, opened 2021-12-10
Remove acrnctl suspend and resume command

wuxyintel:
This patch set is to remove acrnctl suspend and resume command since
these two commands is not used by user any more.

Tracked-On: #5921

### DIFF
--- a/misc/services/acrn_manager/acrn_vm_ops.c
+++ b/misc/services/acrn_manager/acrn_vm_ops.c
@@ -387,24 +387,6 @@ int stop_vm(const char *vmname, int force)
 	return ack.data.err;
 }
 
-int suspend_vm(const char *vmname)
-{
-	struct mngr_msg req;
-	struct mngr_msg ack;
-
-	req.magic = MNGR_MSG_MAGIC;
-	req.msgid = DM_SUSPEND;
-	req.timestamp = time(NULL);
-
-	send_msg(vmname, &req, &ack);
-
-	if (ack.data.err) {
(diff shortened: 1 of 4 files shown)


acrnsi:

misc/services/acrn_manager/acrn_vm_ops.c
```diff
@@ -387,24 +387,6 @@ int stop_vm(const char *vmname, int force)
...
-	send_msg(vmname, &req, &ack);
-
-	if (ack.data.err) {
-		printf("Unable to suspend vm. errno(%d)\n", ack.data.err);
-	}
-
-	return ack.data.err;
-}
-
 int resume_vm(const char *vmname, unsigned reason)
```
gvancuts, line 390: What is the reason you are not removing this function (while you are removing the `suspend_vm` one?

wuxyintel: @gvancuts Thanks for your comments. resume_vm is also used by acrnd, if ACRND_RESUME command is useless, I think we also can remove resume_vm in another patch. BTW, in this patch, only focus on removing acrnctl suspend and acrnctl resume.

gvancuts (approved):

dbkinder (commented):
How do we know these commands are not used by users?  Is there a problem leaving these commands in that's driving their removal?  @terryzouhao

wenlingz:
@terryzouhao @NanlinXie please help review and approve this PR, thanks

dbkinder (approved):

NanlinXie (approved):

PR #6967, opened 2021-12-13
doc: add note about SRIOV disable

wuxyintel:
Currently, Service VM may fail to shut down due to some enabled
VFs. ACRN doesn't disable VFs automatically.

In this patch, add a note for user to disable VFs before Service
VM shuts down.

Tracked-On: #5921


acrnsi:

dbkinder (approved):
LGTM, thanks.

PR #6974, opened 2021-12-14
doc: fix $disk usage in GSG

dbkinder:
The commands in the Getting Started Guide for copying files to and from
the USB stick will fail if the USB stick volume name has a space in it.
Fix this by quoting uses of $disk.


acrnsi:

PR #6975, opened 2021-12-15
doc: add CSS color styles

dbkinder:
Add some color styles we can use (via .. rst-class:: style directive) to
add color to rst tables.  Also introduce a centered class instead of
using the deprecated .. centered:: directive.  Update documentation
guidelines to describe these new styles (background colors).


acrnsi:

PR #6985, opened 2021-12-16
doc: add v2.7 docs to version menu

dbkinder:
Also cleaned out older versions from list, keeping v1.0,
v1.6.1, v2.0, v2.5, and v2.6.

The documentation for previous releases is still available (by editing
the URL to mention that release (e.g., /2.1/ ).


acrnsi:

Issue #6994, opened 2021-12-20
Failed to start acrnd service in Service VM


wuxyintel:
Currently, if user starts acrnd service, acrnd service will fail to find the folder "/sys/kernel/gvt" in Service VM.
Root cause is GVT-g is not supported in current ACRN, the folder "/sys/kernel/gvt" will not be created in Service VM.

gvancuts:
Out of curiosity, is the `After=weston.service` line not a problem?

What is the OS running in the Service VM here?

wuxyintel:
Hi Geoffory, weston is the reference implementation of a Wayland server. A Wayland server is a display server, a window manager, and a compositor all in one. Weston service is optional for ACRN, so I will remove this configuration in acrnd service. I verified it on Ubuntu Service VM.

gvancuts:
> Hi Geoffory, weston is the reference implementation of a Wayland server. A Wayland server is a display server, a window manager, and a compositor all in one. Weston service is optional for ACRN, so I will remove this configuration in acrnd service. I verified it on Ubuntu Service VM. Thanks!

Thanks!

wuxyintel:
[External_System_ID] ACRN-7575

PR #7005, opened 2022-01-04
Misc: acrnd: fix acrnd start failure

wuxyintel:
Currently, if user starts acrnd service, acrnd service will fail to
find the folder "/sys/kernel/gvt" in Service VM, fail to start.

Root cause is GVT-g is not supported in current ACRN, the folder
"/sys/kernel/gvt" will not be created in Service VM.

This patch updates acrnd service to remove this condition check.

v1-->v2:
	Weston service is optional for ACRN, acrnd should not depend
	on it, so remove weston service dependency in acrnd service.

Tracked-On: #6994

### DIFF
--- a/misc/services/acrn_manager/acrnd.service
+++ b/misc/services/acrn_manager/acrnd.service
@@ -1,8 +1,6 @@
 [Unit]
 Description=ACRN manager deamon
-After=weston.service
 After=systemd-resolved.service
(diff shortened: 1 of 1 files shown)


acrnsi:

ywan170 (approved):

szhen11:
start to run permerge test

Issue #6776, opened 2021-11-04
Fix coding guideline violation


NanlinXie:
Clang-tidy tools reported some coding guideline violation issue, we need fix that.

NanlinXie:
[External_System_ID] ACRN-7476

PR #7008, opened 2022-01-07
hv: fix violations of coding guideline C-FN-16

mingqiangchi:
The coding guideline rule C-FN-16 requires that 'Mixed-use of
C code and assembly code in a single function shall not be allowed',
this patch wraps inline assembly to CPU_XMM_READ/CPU_XMM_WRITE.

Tracked-On: #6776


acrnsi:

wenlingz:

dongyaozu (commented):
This is a simple use of MACRO to replace the instruction. To me, it is still mixed use of C and ASM code.  Can Junjie Review?

Making API to read/write all 3 values like asm_loadikey is better to me.

junjiemao1:
Using macros is a valid option to encapsulate and isolate inline assembly according to MISRA-C. Meanwhile using (preferably inline) functions is a better approach.

mingqiangchi:
have updated this PR

dongyaozu (approved):

PR #7009, opened 2022-01-07
board_inspector/pcieparser: check config space size of PCIe devices

junjiemao1:
According to PCIe specification (since 2.0), absence of any extended
capabilities is required to be indicated by an extended capability header
with a capability ID of FFFFh and a next capability offset of 000h. Thus,
the board inspector today accesses the first extended capability header at
100h in the configuration space of a PCIe function unconditionally.

However, in practice we have seen real PCI functions which has a PCIe
capability but no extended capability header. This will cause the board
inspector to crash due to invalid configuration space accesses.

To fix that, this patch adds a check to the size of the configuration space
before walking the extended capabilities of a PCIe function.

Tracked-On: #6411


acrnsi:

NanlinXie (approved):

Issue #7018, opened 2022-01-12
No expected exception generated on some platform.


wenq77:

wenq77:
[External_System_ID] ACRN-7576

PR #7027, opened 2022-01-14
hv: fix bug when set MSR_IA32_COPY_PLATFORM_TO_LOCAL before

wenqian77:
setting MSR_IA32_COPY_LOCAL_TO_PLATFORM

The current code would inject GP to guest, when there's no IWKeyBackup,
and the guest tried to write MSR MSR_IA32_COPY_PLATFORM_TO_LOCAL(0xd92)
to copy IWKeyBackup for the platform to the IWKey for this logical processor.

This patch fixes it by adjusting the code logic, and it'll do nothing
instead of inject GP if no valid IWKeyBackup.
This patch alse add checking for the value being written to avoid setting
reserved MSR bits.

Tracked-On: #7018
Signed-off-by: Wen Qian <qian.wen@intel.com>
Signed-off-by: Li Fei <fei1.li@intel.com>

### DIFF
--- a/hypervisor/arch/x86/guest/vmsr.c
+++ b/hypervisor/arch/x86/guest/vmsr.c
@@ -1044,18 +1044,21 @@ int32_t wrmsr_vmexit_handler(struct acrn_vcpu *vcpu)
 		break;
 	}
 	case MSR_IA32_COPY_LOCAL_TO_PLATFORM:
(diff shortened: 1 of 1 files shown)


acrnsi:

yakuizhao (approved):
LGTM

Issue #7001, opened 2021-12-27
[acrn-configuration-tool] Acrn build failed by same module name conflict


Kunhui-Li:
**Describe the bug**
If the user installed the lib python package in the host machine, it will fail to build acrn hypervisor since the same package named lib is also used in acrn config tool (static_allocators/lib)

**HV branch**
master

**Steps**
1. root@rbtest:~/acrn-work/acrn-hypervisor#make clean && make BOARD=misc/config_tools/data/nuc11tnbi5/nuc11tnbi5.xml SCENARIO=misc/config_tools/data/nuc11tnbi5/shared.xml

**Actual result:**
Makefile error occur:

make[1]: Entering directory '/root/acrn-work/acrn-hypervisor/hypervisor'
Board XML is fetched from /root/acrn-work/acrn-hypervisor/misc/config_tools/data/nuc11tnbi5/nuc11tnbi5.xml
Scneario XML is configuration fetched from /root/acrn-work/acrn-hypervisor/misc/config_tools/data/nuc11tnbi5/shared.xml
Traceback (most recent call last):
File "/root/acrn-work/acrn-hypervisor/misc/config_tools/static_allocators/main.py", line 41, in <module>
main(args)
File "/root/acrn-work/acrn-hypervisor/misc/config_tools/static_allocators/main.py", line 30, in main
module = import_module(f"{module_name}")
File "/usr/lib/python3.8/importlib/_init_.py", line 127, in import_module
return _bootstrap._gcd_import(name[level:], package, level)
File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
File "<frozen importlib._bootstrap>", line 991, in _find_and_load
File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
File "<frozen importlib._bootstrap_external>", line 848, in exec_module
File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
File "/root/acrn-work/acrn-hypervisor/misc/config_tools/static_allocators/intx.py", line 10, in <module>
import common, board_cfg_lib, lib.error, lib.lib
File "/usr/lib/python3/dist-packages/lib/_init_.py", line 1, in <module>
abc
NameError: name 'abc' is not defined
[...]

PR #7036, opened 2022-01-17
config_tools: add __init__.py under static_allocators/lib directory

Kunhui-Li:
If the user installed the lib python package in the host machine,
it will fail to build acrn hypervisor since the same package named lib is
also used in acrn config tool(static_allocators/lib).
So we add the __init__.py file under the static_allocators/lib directory
to fix the error.

Tracked-On: #7001


wenlingz:

NanlinXie (approved):

Issue #7038, opened 2022-01-17
[DM] vtcon: missing error check / infinite calls to vq_getchain()


wenq77:

wenq77:
[External_System_ID] ACRN-7490

PR #7048, opened 2022-01-20
dm: fix the bug that infinite calls to vq_getchain()

wenq77:
The current code would cause infinite calls to vq_getchain()
because that:
  - error check of vq_getchain() return value is missing.
  - virtqueue misunderstand that there are avaliable descripters
    even though the idx of avail ring is invalid.

This patch fixes it by checking validity of the return of
vq_getchain() and jump out of the loop for invalid return value.
This patch alse add validity check in judgment of avaliable
descriptor, and check if the diff between idx of avail ring and
the last idx is greater than size of this queue.

Tracked-On: #7038
Signed-off-by: Wen Qian <qian.wen@intel.com>
Signed-off-by: Li Fei <fei1.li@intel.com>
Acked-by: Wang Yu1 <yu1.wang@intel.com>


acrnsi:

ywan170 (approved):

PR #7051, opened 2022-01-20
dm: Dm rm dynamic param '-W'

YuanyuanZhao609:
Rename '--virtio_msix' to '--virtio_msi' for this param
    means 'force virtio to use singel-vector MSI'.

    `-W` is the short version of `--virtio_msi`. But it's
    confusing that `-W` and `--virtio_msi` are irrelevant literally.
    So remove the short version `W` to prompt user friendliness.

### DIFF
--- a/devicemodel/core/main.c
+++ b/devicemodel/core/main.c
@@ -140,7 +140,7 @@ static void
 usage(int code)
 {
 	fprintf(stderr,
-		"Usage: %s [-hWYv] [-B bootargs] [-E elf_image_path]\n"
+		"Usage: %s [-hYv] [-B bootargs] [-E elf_image_path]\n"
 		"       %*s [-k kernel_image_path]\n"
 		"       %*s [-l lpc] [-m mem] [-r ramdisk_image_path]\n"
 		"       %*s [-s pci] [--ovmf ovmf_file_path]\n"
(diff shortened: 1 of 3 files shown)


acrnsi:

dbkinder (approved):
LGTM, thanks.

ywan170 (commented):
dm patch looks good to me. doc patch, @dbkinder to help review

devicemodel/core/main.c
```diff
@@ -991,6 +989,9 @@ main(int argc, char *argv[])
 		case CMD_OPT_WINDOWS:
 			is_winvm = true;
 			break;
+		case CMD_OPT_FORCE_VIRTIO_MSI:
+			virtio_msix = 0;
```
gvancuts, line 993: Wouldn't make sense to also update this variable to `virtio_msi`?

Out of curiosity, why we do provide a function (`virtio_uses_msix`) instead of using the variable directly? The function does not do anything else than return the variable binary value.

YuanyuanZhao609: Modify ‘--virtio_msix' beacuse it may mislead user that if it's set msix will be used in virtio when msi is accually in use.  Since users usually use msix, they don't have to pay attention to ’--vistio_msi', which is convenient for users. 
'virtio_msix' is kept because it accurately expresses to the developer that msix will be used.

For virtio_uses_msix, though I'm not the original author and don't know the excat reason ， I think it's reasonable.Firstly, we prefer to use static variables with controllable visibility. Secondly, the encapsulated function interface reserves a good space for adding judgment logic in the future.

wenlingz:
@YuanyuanZhao609 please rebase your code, thanks

gvancuts (approved):

Issue #6690, opened 2021-10-19
[DX] ACRN configuration parameters refinement


NanlinXie:
We have ~100 configuration parameters, we summary and put into below categories:

Basic - revisit and new design
Advanced - hide in UI advanced tab
Remove - user will not see that.

NanlinXie:
Need update based on DM parameter removal.

dbkinder:
We should be updating documentation (in a separate PR if needed) as parameters are removed or updated, otherwise documentation and the master branch are out of sync.  For example, PR #7002 removed or updated 11 DM parameters but didn't update the [Device Model Parameter documentation](https://projectacrn.github.io/latest/user-guides/acrn-dm-parameters.html) to reflect these changes.

NanlinXie:
Doc need update, see comments from https://github.com/projectacrn/acrn-hypervisor/pull/7167

NanlinXie:
Doc need update for https://github.com/projectacrn/acrn-hypervisor/pull/7002 change. @weichenli-intel , target v3.0

PR #7060, opened 2022-01-25
hv:refine the min address logic of high memory

weichenli-intel:
Current mmu assum the high memory start from 4G,it's not true for some
platform.

The map logic use "high64_max_ram - 4G" to calculate the high ram size
without any check,it's an issue when the platform have no high memory.

So this patch add high64_min_ram variable to calculate the min address
of high memory and check the high64_min_ram to fix the previou issue.

Tracked-On: #6690


acrnsi:

wenlingz:

anthonyzxu (approved):

PR #7061, opened 2022-01-25
doc: Replace modprobe msr command in gsg and board documentation with…

jackwhich:
… update-pciids

modprobe msr has been added to board_parse.py in PR #6692.

It is recommended that users run the update-pciids command
to keep all pci id information up to date


acrnsi:

doc/getting-started/getting-started.rst
```diff
@@ -328,11 +328,11 @@ Generate a Board Configuration File
          disk="/media/$USER/"$(ls /media/$USER)
          cp -r "$disk"/board_inspector ~/acrn-work
 
-#. On the target, load the ``msr`` driver, used by the Board Inspector:
+#. Before using Board Inspector, you need to update the pciid information:
```
gvancuts, line 331: Strictly speaking, I don't think we need to, but it's a good idea and a good suggestion. How about the following text (looking at @dbkinder to help finalize it ;-)

"On the target, update the ``pci.ids`` file which contains the list of known identifiers related to PCI devices"

doc/tutorials/board_inspector_tool.rst
```diff
@@ -28,11 +28,11 @@ a board configuration file.
 #. Make sure the target system is set up and ready to run the Board Inspector,
    according to :ref:`gsg-board-setup` in the Getting Started Guide.
 
-#. Load the ``msr`` driver, used by the Board Inspector:
+#. Before using Board Inspector, you need to update the pciid information:
```
gvancuts, line 31: Keep this one in sync with the other sentence if you change it based on my recommendation

gvancuts (changes requested):

amyreye:
@jackwhich For the 2.8 release GSG, the plan is to replace the existing board inspector steps with the Debian board inspector steps. Does the user need to run update-pciids if they are using Debian board inspector steps?

doc/getting-started/getting-started.rst
dbkinder: How about this:
```
#. On the target, you can optionally update the ``pci.ids`` file
   that contains the list of known identifiers for PCI devices:
```

doc/tutorials/board_inspector_tool.rst
dbkinder: Same here:
```
#. On the target, you can optionally update the ``pci.ids`` file
   that contains the list of known identifiers for PCI devices:
```

dbkinder (changes requested):

gvancuts (approved):

NanlinXie (approved):

gvancuts:
> @jackwhich For the 2.8 release GSG, the plan is to replace the existing board inspector steps with the Debian board inspector steps. Does the user need to run update-pciids if they are using Debian board inspector steps?

I believe we will still need to do this manually, or perhaps we can do it via a postinst script in the package.

The `/usr/share/misc/pci.ids` file is from the `pci.ids` package but we cannot rely on the version in the distro repo to be the latest so our best option remains to run `update-pciids`. If we can do this via a postinst script, it's ideal since it would save the user one step. If not, we'll have to instruct them to do it manually.

@tttech-industrial-buchsbaum , as a Debian packaging expert, what are your thougts?

gvancuts:
> @gvancuts @jackwhich Since we're saying this is an optional update, what are the consequences of having an outdated `pci.ids` file? If it were important, wouldn't the distros be keeping it up to date via apt?

I don't think it's very important in general from a functionality point of view. It's all about being able to convert PCI vendor and device IDs into a human-readable string. If the `pci.ids` file is outdated, the consequences are that only the PCI vendor and device code numbers are displayed.

When that happens, this is what it looks like:
```
00:00.0 Host bridge: Intel Corporation Device 3e30 (rev 0d)
```
Versus:
```
00:00.0 Host bridge: Intel Corporation 8th/9th Gen Core 8-core Desktop Processor Host Bridge/DRAM Registers [Coffee Lake S] (rev 0d)
```

There is also a `-q` flag that we could use to automatically contact the online, central database in case the PCI device cannot be found in the local file. I was initially reluctant to do this so as to not introduce a network dependency, but a quick test shows that nothing bad happens even in case of network issues. So perhaps that's what we should do instead? I'm in favour of this if others agree, @dbkinder @jackwhich @junjiemao1 @NanlinXie ?

dbkinder:
I too was concerned about the network dependency for the target, bet we already have that dependency in our target setup instructions in the GSG for installing dependencies using apt and pip. I like your suggestion. Is that using -q on the update-pciids command?

— david
________________________________

gvancuts:
> I too was concerned about the network dependency for the target, bet we already have that dependency in our target setup instructions in the GSG for installing dependencies using apt and pip. I like your suggestion. Is that using -q on the update-pciids command?

No, it's even better than that. I can be added to the `board_inspector` tool directly so it's transparent to the users. And we can then remove the entire section about running the `update-pciids` tool. :-)

dbkinder:
I think you’re saying to add the call to update-pciids -q into the app since there’s no -q option for the app itself (with a check that that app is available).

— david
________________________________

gvancuts:
> I think you’re saying to add the call to update-pciids -q into the app since there’s no -q option for the app itself (with a check that that app is available).

No, this is a `lspci` argument, it could be added to the Python code in the `board_inspector` that calls `lspci`.

doc/getting-started/getting-started.rst
```diff
@@ -328,11 +328,11 @@ Generate a Board Configuration File
          disk="/media/$USER/"$(ls /media/$USER)
          cp -r "$disk"/board_inspector ~/acrn-work
 
-#. On the target, load the ``msr`` driver, used by the Board Inspector:
+#. On the target, update the ``pci.ids`` file which contains the list of known identifiers related to PCI devices:
```
dbkinder, line 331: Looks like the best choice now is to remove this step completely (lines 331-336) and instead update the board_inspector.py when it calls ``lspci`` to add the ``-q`` option (to fetch missing IDs).

doc/tutorials/board_inspector_tool.rst
```diff
@@ -28,11 +28,11 @@ a board configuration file.
 #. Make sure the target system is set up and ready to run the Board Inspector,
    according to :ref:`gsg-board-setup` in the Getting Started Guide.
 
-#. Load the ``msr`` driver, used by the Board Inspector:
+#. On the target, update the ``pci.ids`` file which contains the list of known identifiers related to PCI devices:
```
dbkinder, line 31: Same here, delete this whole step (lines 31-36) in this document.

dbkinder (changes requested):

amyreye (approved):

dbkinder (approved):
LGTM, thanks!

PR #7065, opened 2022-01-26
dm:add UID parameter for acpidev_pt

weichenli-intel:
Now the acpidev_pt module only use the hid to check the device,it can't
work well if there are more then one instance.

So this patch add UID to identify same type device to fix these issue.

Tracked-On: #6690

### DIFF
--- a/devicemodel/hw/platform/tpm/tpm.c
+++ b/devicemodel/hw/platform/tpm/tpm.c
@@ -74,11 +74,22 @@ static int is_tpm2_eventlog_supported(struct acpi_table_tpm2 *tpm2)
 	return ((tpm2->header.length == 76) && tpm2->lasa && tpm2->laml);
 }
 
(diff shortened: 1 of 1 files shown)


acrnsi:

wenlingz:
@ywan170 please approve this PR, thanks

wenlingz:
@ywan170 @anthonyzxu please help approve this PR, thanks

ywan170 (approved):

ywan170:
@weichenli-intel do we need also update relative document? The current parameter doesn't mentioned the "hid,uid" format of mmio device passthrough, right?

weichenli-intel:
> @weichenli-intel do we need also update relative document? The current parameter doesn't mentioned the "hid,uid" format of mmio device passthrough, right?

Yes, I will update the doc by another PR, the doc and dm are splited patch.

PR #7067, opened 2022-01-27
Process both forms of ACPI PRT entries

junjiemao1:
According to ACPI spec, an entry in ACPI PRT (PCI Routing Table) can                                                                                                                                               specify an interrupt line either by an integer or by a pair of a device                                                                                                                                            object and an index. In the later case, the interrupt line is specified by                                                                                                                                         the index'th entry in the current resource list (as reported by the _CRS                                                                                                                                           method) of that device object.       

Today we see most BIOS simply use an integer in the ACPI namespace (at                                                                                                                                             least in APIC mode), but now in Qemu "device-object + index" pairs are used                                                                                                                                        in PRT, leading to build-time failure when building ACRN to run on top of                                                                                                                                          Qemu/KVM.            

This patch series adds the support to process both forms of ACPI PRT                                                                                                                                               entries properly.


acrnsi:

wenlingz:
@NanlinXie @terryzouhao please help approve this PR, thanks

NanlinXie (approved):

Issue #7045, opened 2022-01-19
[ICX-D] Segmentation fault when passthrough TSN to post_launched VM with enable_ptm option


mgcao:
[root@ICX-D-HCC-01 gos_images]# ./POST_RT_VM1.sh
cpu8 online=1
cpu9 online=1
acrn-br0 bridge aleady exists, adding new tap device to it...
logger: name=console, level=4
logger: name=kmsg, level=3
logger: name=disk, level=5
SW_LOAD: get ovmf path /usr/share/acrn/bios/OVMF.fd, size 0x200000
vm_create: POST_RT_VM1
vm_setup_memory: size=0x40000000
open hugetlbfs file /run/hugepage/acrn/huge_lv1/POST_RT_VM1
open hugetlbfs file /run/hugepage/acrn/huge_lv2/POST_RT_VM1
level 0 free/need pages:0/1 page size:0x200000
level 1 free/need pages:13/1 page size:0x40000000
to reserve more free pages:
to reserve pages (+orig 0): echo 1 > /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages
now enough free pages are reserved!

try to setup hugepage with:
level 0 - lowmem 0x0, biosmem 0x200000, highmem 0x0
level 1 - lowmem 0x40000000, biosmem 0x0, highmem 0x0
total_size 0x140000000

mmap ptr 0x0x7f07b7cb4000 -> baseaddr 0x0x7f07c0000000
mmap 0x40000000@0x7f07c0000000
touch 1 pages with pagesz 0x40000000
mmap 0x200000@0x7f08bfe00000
touch 1 pages with pagesz 0x200000

really setup hugepage with:
level 0 - lowmem 0x0, biosmem 0x200000, highmem 0x0
level 1 - lowmem 0x40000000, biosmem 0x0, highmem 0x0
vm_init_vdevs
No pm notify channel given
Listening 0...
polling 0...
pci init passthru
<PTM>: opt=enable_ptm.
[...]

PR #7074, opened 2022-01-28
dm: fix: TSN in multi-hostbridge could crash

mgcao:
fix bug: if PTM-CAP device, like TSN in multi-hostbridge could cause
acrn-dm crash.

original PTM PCI code has not handled multi-hostbridge case, and just
handled hostbridge (00:00.0) case.

this patch calls PCI access API to handle PTM-CAP device/bridge
(root port) structure, to avoid scan PCI hierarchical.

Tracked-On: #7045


acrnsi:

ywan170 (approved):

PR #7075, opened 2022-01-28
doc:add UID parameter for acpidev_pt

weichenli-intel:
Now the acpidev_pt module only use the hid to check the device,it can't
work well if there are more then one instance.

So this patch add UID to identify same type device to fix these issue.

Tracked-On: #6690

### DIFF
--- a/doc/developer-guides/hld/hld-devicemodel.rst
+++ b/doc/developer-guides/hld/hld-devicemodel.rst
@@ -56,7 +56,7 @@ options:
                [-l lpc] [-m mem] [-r ramdisk_image_path]
                [-s pci] [--ovmf ovmf_file_path]
                [--enable_trusty] [--intr_monitor param_setting]
-               [--acpidev_pt HID] [--mmiodev_pt MMIO_regions]
+               [--acpidev_pt HID[,UID]] [--mmiodev_pt MMIO_regions]
                [--vtpm2 sock_path] [--virtio_poll interval] [--mac_seed seed_string]
                [--cpu_affinity lapic_ids] [--lapic_pt] [--rtvm] [--windows]
                [--debugexit] [--logger-setting param_setting]
@@ -81,7 +81,7 @@ options:
        --intr_monitor: enable interrupt storm monitor
             its params: threshold/s,probe-period(s),delay_time(ms),delay_duration(ms),
        --virtio_poll: enable virtio poll mode with poll interval with ns
-       --acpidev_pt: ACPI device ID args: HID in ACPI Table
+       --acpidev_pt: ACPI device ID args: HID,UID from the ACPI tables
        --mmiodev_pt: MMIO resources args: physical MMIO regions
        --vtpm2: Virtual TPM2 args: sock_path=$PATH_OF_SWTPM_SOCKET
        --lapic_pt: enable local apic passthrough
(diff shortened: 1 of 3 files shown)


acrnsi:

doc/developer-guides/hld/hld-devicemodel.rst
```diff
@@ -81,7 +81,7 @@ options:
        --intr_monitor: enable interrupt storm monitor
             its params: threshold/s,probe-period(s),delay_time(ms),delay_duration(ms),
        --virtio_poll: enable virtio poll mode with poll interval with ns
-       --acpidev_pt: ACPI device ID args: HID in ACPI Table
+       --acpidev_pt: ACPI device ID args: HID,HID in ACPI Table
```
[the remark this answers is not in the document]
gvancuts: Would it be better more accurate to say "HID,UID from the ACPI tables"?

I must confess that I'm not sure myself what the best/easiest way is for figuring out such information. I believe an `acpidump -b` followed by `iasl -d` would help but is it a good example?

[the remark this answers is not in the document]
weichenli-intel: Very thanks for your suggestion, I will try the new way.
HID,UID from the ACPI tables it's a better way and I have update it.

gvancuts (approved):
LGTM

doc/user-guides/acrn-dm-parameters.rst
```diff
@@ -218,16 +218,20 @@ Here are descriptions for each of these ``acrn-dm`` command line parameters:
...
 ----
 
-``--acpidev_pt <HID>``
+``--acpidev_pt <HID>,<UID>``
    This option is to enable ACPI device passthrough support. The ``HID`` is a
    mandatory parameter for this option which is the Hardware ID of the ACPI
    device.
 
+   The ``UID`` is a option and used for distingnish several instances of same
+   hid devices.
```
amyreye, line 227: A couple questions: 
1. Is UID optional? If so, should we add brackets like this: ``--acpidev_pt <HID>[,<UID>]``
2. I don't understand the description of UID. Can a HID have several instances, and you need to specify one of them? Perhaps the sentence should be something like: _The ``UID`` is an option and used to specify a particular instance of the HID device._

weichenli-intel: Yes, a HID could have several instances.
"Is UID optional? If so, should we add brackets like this: --acpidev_pt <HID>[,<UID>]"
--Our code could get HID use 00 as default UID, whether changed to "-- acpidev_pt < hid > [, < uid >]" depends on whether we want to expose it.

gvancuts: What does the code do today? Does it require the user to provide a HID? What about the UID? My understanding is that the `HID` parameter is mandatory and `UID` is optional today. If true, Amy's point is that this line should read:
```
--acpidev_pt <HID>[,<UID>]
```
And we should also use the text she provided in her comment, i.e.: *The UID is an option and used to specify a particular instance of the HID device.*

amyreye: And if the default value is 00, we should add it to the description, i.e., _The UID is an option and used to specify a particular instance of the HID device. The default is 00._

weichenli-intel: > What does the code do today? Does it require the user to provide a HID? What about the UID? My understanding is that the `HID` parameter is mandatory and `UID` is optional today. If true, Amy's point is that this line should read:
> 
> ```
> --acpidev_pt <HID>[,<UID>]
> ```
> 
> And we should also use the text she provided in her comment, i.e.: _The UID is an option and used to specify a particular instance of the HID device._

Yes, user could only set HID.

weichenli-intel: > And if the default value is 00, we should add it to the description, i.e., _The UID is an option and used to specify a particular instance of the HID device. The default is 00._

OK.

dbkinder (approved):
LGTM2

Issue #6665, opened 2021-10-13
Remove the `d3hot_reset` parameter (`acrn-dm`)


gvancuts:
We can use the `is_winvm` flag in the `acrn-dm` code to know if the User VM is running Windows (`--windows` parameter). When that's the case, our pass-through code could automatically add the logic behind `d3hot_reset` for the pass-through devices.

_Originally posted by @liuhang-bit in https://github.com/projectacrn/acrn-hypervisor/issues/6445#issuecomment-941963311_

dbkinder:
@liuhang-bit @NanlinXie Is this something that's getting implemented (v2.8)?

gvancuts:
I believe this (#7076) should do it. I have not actually tested it myself though... :-)

gvancuts:
[External_System_ID] ACRN-7609

gvancuts:
Closing as per https://github.com/projectacrn/acrn-hypervisor/pull/7076#issuecomment-1024181337

PR #7076, opened 2022-01-28
dm: remove the 'd3hot_reset' parameter

gvancuts:
We do not need to keep the 'd3hot_reset' optional parameter for acrn-dm.
This option is used when running Windows in the User VM, we can
therefore use the 'is_winvm' flag instead directly.

Tracked-On: #6665

### DIFF
--- a/devicemodel/hw/pci/passthrough.c
+++ b/devicemodel/hw/pci/passthrough.c
@@ -593,7 +593,6 @@ passthru_init(struct vmctx *ctx, struct pci_vdev *dev, char *opts)
 	char *opt;
 	bool keep_gsi = false;
 	bool need_reset = true;
-	bool d3hot_reset = false;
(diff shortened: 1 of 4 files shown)


acrnsi:

ywan170:
@gvancuts This parameter is still needed. It is not relative windows guest. The background is some devices don't support FLR(one reset mechanism), and they are required to be reset in some special cases like VM reboot. W/o any reset, the devices may have some problems. That is why we added this feature to let device enter and exit D3hot to emulate the reset....

This issue reproduced on xHCI controller w/ Windows guest driver. But it should be a common solution for other devices lacked of FLR.

gvancuts:
I see. Thanks for the explanation, I wish this would have been explained already back then in the original issue. But anyway, let's close this.

PR #7079, opened 2022-01-29
hv: refine the e820 add new entry logic

weichenli-intel:
Sometimes the memory to be allocated is not at the end of an entry,
that means we have to break one enty into 2 smaller entries, there
are two ways to add the new entry to hv_e820, adds to the end or
insert it.

The initial e820 table is ordered, that's why the e820_alloc_memory
interface asssum all entries was sorted, but add new entry to the
end will break the orde of hv_e820.

So we use insert_e820_entry to replace the add_e820_entry, the new
interfeac will keep the orde and users do not need sort again after
alloc region

Tracked-On: #6690
Signed-off-by: Chenli Wei<chenli.wei@linux.intel.com>

### DIFF
--- a/hypervisor/arch/x86/e820.c
+++ b/hypervisor/arch/x86/e820.c
@@ -102,13 +102,20 @@ uint64_t e820_alloc_memory(uint64_t size_arg, uint64_t max_addr)
 	return ret;
 }
 
-static void add_e820_entry(uint64_t addr, uint64_t length, uint64_t type)
+static void insert_e820_entry(uint32_t index, uint64_t addr, uint64_t length, uint64_t type)
 {
+	uint32_t i;
+
 	hv_e820_entries_nr++;
 	ASSERT(hv_e820_entries_nr <= E820_MAX_ENTRIES, "e820 entry overflow");
-	hv_e820[hv_e820_entries_nr - 1U].baseaddr = addr;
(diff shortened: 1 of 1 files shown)


acrnsi:

wenlingz:
@anthonyzxu @dongyaozu please help approve this PR, thanks

wenlingz:
@anthonyzxu @dongyaozu please help approve this PR, thanks

hypervisor/arch/x86/e820.c
```diff
@@ -102,13 +102,20 @@ uint64_t e820_alloc_memory(uint64_t size_arg, uint64_t max_addr)
...
+
 	hv_e820_entries_nr++;
 	ASSERT(hv_e820_entries_nr <= E820_MAX_ENTRIES, "e820 entry overflow");
-	hv_e820[hv_e820_entries_nr - 1U].baseaddr = addr;
-	hv_e820[hv_e820_entries_nr - 1U].length = length;
-	hv_e820[hv_e820_entries_nr - 1U].type = type;
+
+	for (i = hv_e820_entries_nr - 1; i > index; i--) {
+		hv_e820[i] = hv_e820[i-1];
+	}
```
dongyaozu, line 114: How about to move this loop operation before hv_e820_entries_nr++ ?

This way, the meaning is much readable.

dongyaozu (approved):

PR #7088, opened 2022-02-07
dm: refine the mac check logic for virtio-net

weichenli-intel:
The current mac detection logic only check whether there are parameters
after tap opt. Don't care whether the parameter is MAC or not, then
mac_provided will be set to 1, which will disable the mac_seed setting
when using sub parameters

This patch will check opt and fix the above issue.

Tracked-On: #6690


acrnsi:

ywan170 (approved):

Issue #6689, opened 2021-10-19
[DX]Board Inspector tool error message refinement 


NanlinXie:
From DX survey, user was confused for the output message from board inspector, we need refine

Change Board Inspector tool success message to be clear that users were successful creating the board configuration file. Success message: “my_board.xml saved successfully!”
Board Inspector: Suppress warnings if they do not require user action.
Board Inspector: Ensure error messages are clear about the exact issue and provide a resolution to users.

NanlinXie:
[External_System_ID] ACRN-7430

amyreye:
@NanlinXie Is the Engineering team ready to update docs for this issue? For example, in ``board_inspector.py``, it looks like the log levels were updated in #7116; however, the command-line help wasn't updated and docs were not updated. For example, ``board_inspector.py`` command-line help only mentions 3 log levels, but code has 5 levels:

``
    parser.add_argument("--loglevel", default="warning", help="choose log level, e.g. info, warning or error")

NanlinXie:
@amyreye , yes the doc need to be updated, but for now, we are still ongoing coding work.

PR #7092, opened 2022-02-09
config_tools: fix illegal access to the level_types dictionary

junjiemao1:
In 10-processors.py, level_types is an unordered dictionary from level type
encodings to human-readable terms, not a list. It is thus illegal to access the
highest level using the `[-1]` notation.

This patch fixes this by calculating the highest level in level_types and use
that level as the index instead.

Tracked-On: #6689
Fixes: b5b22bf98 ("board_inspector: avoid hard-coded topo level names")


acrnsi:

NanlinXie (approved):

PR #7096, opened 2022-02-09
doc: Tar file installation

amyreye:
- New document about building tarball files for ACRN installation.
- Steps are from GSG.

### DIFF
--- a/doc/develop.rst
+++ b/doc/develop.rst
@@ -61,6 +61,7 @@ Configuration Tutorials
    user-guides/kernel-parameters
    user-guides/acrn-dm-parameters
    misc/packaging/README
+   tutorials/tar_installation
 
 Advanced Features
 ********************
--- a/doc/tutorials/tar_installation.rst
+++ b/doc/tutorials/tar_installation.rst
@@ -1,0 +1,250 @@
+.. _tar_installation:
+
+ACRN Installation via Tar Files
+####################################
+
+Project ACRN offers two ways to install ACRN on target systems, either via
+Debian packages or tar files. This document covers the tar file method. For
+information about the Debian method, see :ref:`debian_packaging`.
+
+Tar files provide a way to package ACRN configurations for Linux
+target systems running non-Debian operating systems. You generate the tar
+files on your development computer, copy them to your target system,
+extract the tar files to the right places,
+and reboot the system with ACRN up and running.
+
+This document shows examples of commands used to build and install the tar
+files. You might need to modify the commands for your environment.
+
+Prerequisites
+*************
+
+* A development computer configured according to the :ref:`gsg` (for example, it
+  has an Ubuntu OS, ACRN dependencies, and ACRN hypervisor and kernel source
+  code from the ACRN GitHub repository).
+
+Build the Tar Files
+***************************
+
+#. On your development computer, navigate to the ``acrn-hypervisor`` source code
+   directory and build the ACRN hypervisor tar files. Replace
+   ``<board.xml>`` and ``<scenario.xml>`` with the paths to your board
+   configuration file and scenario configuration file.
+
+   .. code-block:: bash
+
+      cd ~/acrn-work/acrn-hypervisor
+      make -j $(nproc) BOARD=<board.xml> SCENARIO=<scenario.xml>
+      make targz-pkg
+
+   The build typically takes a few minutes. By default, the build results are
+   found in the ``build`` directory.
+
+#. Navigate to the ``acrn-kernel`` source code directory and build the ACRN
+   kernel tar files for the Service VM:
+
+   .. code-block:: bash
+
+      cd ~/acrn-work/acrn-kernel
+      cp kernel_config_service_vm .config
+      make olddefconfig
+      make -j $(nproc) targz-pkg
+
+   The kernel build can take 15 minutes or less on a fast computer, but could
+   take an hour or more depending on the performance of your development
+   computer. By default, the build results are found in the current directory. 
+
+Install and Run ACRN
+**************************
+
+In the following steps, you will install the tar files, install the serial
+configuration tool, configure GRUB, and run ACRN on the target system.
+
+#. Copy all the necessary files generated on the development computer to the
+   target system. The following steps show how to copy via USB disk; feel free
+   to use a different method. Modify the file names in the following commands to
+   match your files.
+
+   a. Insert the USB disk into the development computer and run these commands:
+
+      .. code-block:: bash
+
+         cd ~/acrn-work/acrn-kernel
+         disk="/media/$USER/"$(ls /media/$USER)
+         cp linux-5.10.90-acrn-service-vm-206626-g140f5035e1b1-x86.tar.gz "$disk"/
+         cp ~/acrn-work/acpica-unix-20210105/generate/unix/bin/iasl "$disk"/
+         cp ~/acrn-work/acrn-hypervisor/build/acrn-2.8-unstable.tar.gz "$disk"/
+         sync && sudo umount "$disk"/
+
+   #. Insert the USB disk you just used into the target system and run these
+      commands to copy the files locally:
+
+      .. code-block:: bash
+
+         disk="/media/$USER/"$(ls /media/$USER)
+         cp "$disk"/linux-5.10.90-acrn-service-vm-206626-g140f5035e1b1-x86.tar.gz ~/acrn-work
+         cp "$disk"/acrn-2.8-unstable.tar.gz ~/acrn-work
+         sudo cp "$disk"/iasl /usr/sbin/
+         sync && sudo umount "$disk"/
+
+#. Extract the Service VM files onto the target system:
+
+   .. code-block:: bash
+
+      cd ~/acrn-work
+      sudo tar -zxvf linux-5.10.90-acrn-service-vm-206626-g140f5035e1b1-x86.tar.gz -C / --keep-directory-symlink
+
+#. Extract the ACRN tools and images:
+
+   .. code-block:: bash
+
+      sudo tar -zxvf acrn-2.8-unstable.tar.gz -C / --keep-directory-symlink
+
+#. Copy the ACRN bin file to the boot directory. Replace ``<board>`` and
+   ``<scenario>`` to match your file.
+
+   .. code-block:: bash
+
+      sudo mkdir -p /boot/acrn/
+      sudo cp /usr/lib64/acrn/acrn.<board>.<scenario>.bin /boot/acrn
+
+#. Install the serial configuration tool in the target system as follows:
+
+   .. code-block:: bash
+
+      sudo apt install setserial
+
+#. Find the root filesystem (rootfs) device name by using the
+   ``lsblk`` command:
+
+   .. code-block:: console
+      :emphasize-lines: 24
+
+      ~$ lsblk
+      NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
+      loop0         7:0    0 255.6M  1 loop /snap/gnome-3-34-1804/36
+      loop1         7:1    0  62.1M  1 loop /snap/gtk-common-themes/1506
+      loop2         7:2    0   2.5M  1 loop /snap/gnome-calculator/884
+      loop3         7:3    0 241.4M  1 loop /snap/gnome-3-38-2004/70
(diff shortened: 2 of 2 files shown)


acrnsi:

doc/tutorials/tarball_installation.rst
```diff
@@ -0,0 +1,248 @@
...
+
+      .. code-block:: bash
+
+         sudo mkdir -p /boot/acrn/
+         sudo cp "$disk"/acrn.bin /boot/acrn
+         sudo cp "$disk"/serial.conf /etc
+         sudo cp "$disk"/iasl /usr/sbin/
+         sync && sudo umount "$disk"/
+
+Install the Tarball Files
```
gvancuts, line 103: This chapter really starts when you untar the files on the target device rootfs, so I would move the section title to just before the "Insert the USB disk you just used into the target..." step.

gvancuts: This section could still be a different section but I would then call it something like: "Final Configuration steps" or something like that.

doc/tutorials/tarball_installation.rst
```diff
@@ -0,0 +1,248 @@
...
+   target system by USB disk as follows. Modify the file names in the following
+   commands to match your files.
+
+   a. Insert the USB disk into the development computer and run these commands:
+
+      .. code-block:: bash
+
+         disk="/media/$USER/"$(ls /media/$USER)
+         cp linux-5.10.65-acrn-service-vm-x86.tar.gz "$disk"/
+         cp ~/acrn-work/acrn-hypervisor/build/hypervisor/acrn.bin "$disk"/
```
gvancuts, line 59:
```suggestion
```

doc/tutorials/tarball_installation.rst
```diff
@@ -0,0 +1,248 @@
...
+      .. code-block:: bash
+
+         sudo tar -zxvf acrn-2.7-unstable.tar.gz -C / --keep-directory-symlink
+
+   #. Copy a few additional ACRN files to the expected locations:
+
+      .. code-block:: bash
+
+         sudo mkdir -p /boot/acrn/
+         sudo cp "$disk"/acrn.bin /boot/acrn
```
gvancuts, line 98:
```suggestion
         sudo cp /usr/lib64/acrn/acrn.<board>.<scneario>.bin /boot/acrn/acrn.bin
```

gvancuts: I would move this line out of this block because this new command does not take the source file from the USB stick (so it's the odd one out in the mix). I would put it in the section below and add a note telling the user he needs to replace `<board>` and `<scenario>` with his own values.

amyreye: @gvancuts The end of your suggested command is `/boot/acrn/acrn.bin` -- does this rename the bin file?

gvancuts: Yes, it does. It's on purpose so that it matches the name used in the Grub example below. We can also choose not to rename it but in that case we need to update the Grub section below (`multiboot2` line) to match the name used here.

doc/tutorials/tarball_installation.rst
amyreye: I placed the heading a little higher than your recommendation. Please review.

doc/tutorials/tarball_installation.rst
amyreye: I chose to keep the original name and change the Grub example accordingly. Also, there were two sets of copy commands, one before extracting the tar files and one after extracting the tar files. I consolidated all of the copy commands and put them before extracting the tar files. Please review.

doc/tutorials/tarball_installation.rst
```diff
@@ -0,0 +1,249 @@
...
+         disk="/media/$USER/"$(ls /media/$USER)
+         cp linux-5.10.65-acrn-service-vm-x86.tar.gz "$disk"/
+         cp ~/acrn-work/acpica-unix-20210105/generate/unix/bin/iasl "$disk"/
+         cp ~/acrn-work/acrn-hypervisor/build/acrn-2.7-unstable.tar.gz "$disk"/
+         sync && sudo umount "$disk"/
+
+      .. note:: The :file:`serial.conf` is only generated if non-standard
+         vUARTs (not COM1-COM4)
+         are configured for the Service VM in the scenario XML file.
+         Please copy the ``serial.conf`` file using::
```
gvancuts, line 72: We can actually remove this, the file is in the tarball that's being generated.

doc/tutorials/tarball_installation.rst
```diff
@@ -0,0 +1,249 @@
...
+
+   #. Insert the USB disk you just used into the target system and run these
+      commands to copy the files locally:
+
+      .. code-block:: bash
+
+         disk="/media/$USER/"$(ls /media/$USER)
+         cp "$disk"/linux-5.10.65-acrn-service-vm-x86.tar.gz ~/acrn-work
+         cp "$disk"/acrn-2.7-unstable.tar.gz ~/acrn-work
+         sudo cp "$disk"/serial.conf /etc
```
gvancuts, line 84:
```suggestion
```

doc/tutorials/tarball_installation.rst
```diff
@@ -0,0 +1,249 @@
...
+
+#. Copy the ACRN bin file to the boot directory. Replace ``<board>`` and
+   ``<scenario>`` to match your file.
+
+   .. code-block:: bash
+
+      sudo mkdir -p /boot/acrn/
+      sudo cp /usr/lib64/acrn/acrn.<board>.<scenario>.bin /boot/acrn
+
+#. Install the serial configuration tool in the target system as follows:
```
gvancuts, line 109:
```suggestion
#. Install the serial configuration tool and configuration file in the target system as follows:
```

doc/tutorials/tarball_installation.rst
```diff
@@ -0,0 +1,249 @@
...
+
+   .. code-block:: bash
+
+      sudo apt install setserial
```
gvancuts, line 113:
```suggestion
      sudo apt install setserial
      sudo cp /usr/lib64/acrn/serial.conf /etc/
```

doc/tutorials/tarball_installation.rst
```diff
@@ -0,0 +1,249 @@
...
+
+      .. code-block:: bash
+
+         disk="/media/$USER/"$(ls /media/$USER)
+         cp linux-5.10.65-acrn-service-vm-x86.tar.gz "$disk"/
+         cp ~/acrn-work/acpica-unix-20210105/generate/unix/bin/iasl "$disk"/
+         cp ~/acrn-work/acrn-hypervisor/build/acrn-2.7-unstable.tar.gz "$disk"/
+         sync && sudo umount "$disk"/
+
+      .. note:: The :file:`serial.conf` is only generated if non-standard
```
gvancuts, line 69: This note can now be moved further down when we install the `setserial` tool

gvancuts (commented):
Hi @amyreye , your changes look good!

I have made some further suggestions to improve the flow especially regarding the `serial.conf` file. I have also filed a PR to make it even easier to the user so that we would not have to copy it around at all (see PR #7108). If you want, we can wait until this gets merged and then the specific stuff about that `serial.conf` file can all go away completely.

amyreye:
> Hi @amyreye , your changes look good!
> 
> I have made some further suggestions to improve the flow especially regarding the `serial.conf` file. I have also filed a PR to make it even easier to the user so that we would not have to copy it around at all (see PR #7108). If you want, we can wait until this gets merged and then the specific stuff about that `serial.conf` file can all go away completely.

Sounds good. I will wait until the PR is merged.

doc/tutorials/tarball_installation.rst
```diff
@@ -0,0 +1,249 @@
...
+   which case it will boot with this option automatically in 5 seconds.)
+
+   .. code-block:: console
+      :emphasize-lines: 8
+
+                                GNU GRUB version 2.04
+      ────────────────────────────────────────────────────────────────────────────────
+      Ubuntu
+      Advanced options for Ubuntu
+      Ubuntu 18.04.05 LTS (18.04) (on /dev/nvme0n1p2)
```
dbkinder, line 246: Should we change this to show Ubuntu 20.04?

dbkinder (commented):
Looking good... a few tweeks

doc/develop.rst
```diff
@@ -61,6 +61,7 @@ Configuration Tutorials
    user-guides/kernel-parameters
    user-guides/acrn-dm-parameters
    misc/packaging/README
+   tutorials/tarball_installation
```
dbkinder, line 64: since we changed the title not to say tarball, how about changing the file name too

dbkinder (changes requested):

doc/tutorials/tarball_installation.rst
gvancuts: > Should we change this to show Ubuntu 20.04?

Yes, we should

doc/tutorials/tarball_installation.rst
```diff
@@ -0,0 +1,249 @@
...
+
+   a. Edit the GRUB ``40_custom`` file. The following command uses ``vi``, but
+      you can use any text editor.
+
+      .. code-block:: bash
+
+         sudo vi /etc/grub.d/40_custom
+
+   #. Add the following text at the end of the file. Replace ``UUID`` and
+      ``PARTUUID`` with the output from the previous step. Replace ``<board>``
```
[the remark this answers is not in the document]
amyreye: @dbkinder Do you have a suggested change, or keep as is?

doc/tutorials/tarball_installation.rst
amyreye: @gvancuts I didn't make this change, because your other comment says that mentions of serial.conf file can all go away completely.

doc/tutorials/tarball_installation.rst
amyreye: @gvancuts I didn't make this change, because your other comment says that mentions of serial.conf file can all go away completely.

doc/tutorials/tarball_installation.rst
amyreye: @gvancuts I didn't make this change, because your other comment says that mentions of serial.conf file can all go away completely. Please check.

doc/tutorials/tarball_installation.rst
gvancuts: That's correct, forget about this suggestion.

doc/tutorials/tarball_installation.rst
gvancuts: Yes, you can drop this suggestion.

doc/tutorials/tarball_installation.rst
gvancuts: You removed the note completely, didn't you?

The way I understand this you only really need to install the `setserial` tool **if** you have a non-standard serial port. In that case you should install `setserial` that will use the `serial.conf` that we generated and installed. I'm thinking the easiest is to ask the user to install `setserial` regardless, I don't know that there would be any side-effect even in the case you only have standard serial ports configured. @NanlinXie , can someone confirm?

amyreye: @gvancuts Yes, I deleted the note.

gvancuts: Sounds good, thanks @amyreye !

@NanlinXie , one more question to the engineering team, is there any other action required to get the non-standard serial port configured? Or is installing the `setserial` package and the custom `serial.conf` is sufficient? I believe it is but I do not have first hand experience with this.

NanlinXie: @wuxyintel , could you help to reply @gvancuts's question?

dbkinder (approved):

doc/tutorials/tarball_installation.rst
wuxyintel: @gvancuts @NanlinXie for acrn v2.7, non-standard ports are emulated by hypervisor only when user config them in scenario file manually, I think these non-standard ports in scenario file will be configured in scenario file by default in acrn v2.8.
If non-standard ports are emulated by hypervisor, serial.conf will be generated during hypervisor build-time. Installing the setserial package and the custom serial.conf in service VM is sufficient. Thanks!

PR #7101, opened 2022-02-11
Add command monitor in DM

wuxyintel:
This command monitor in DM is to handle commands from external
systems (such libvirt, kata container) directly.

In current design, external system should send DM through acrnctl,
acrnctl will be replaced by libvirt in ACRN soon.
In this patch set, command monitor is introduced to make external
system can interact with DM directly through unix domain socket;
older command handling in DM is also refined to just keep two
commands (force shutdown, blkrescan), make adding new commands more
easy.

### DIFF
--- a/devicemodel/core/cmd_monitor/cmd_monitor.c
+++ b/devicemodel/core/cmd_monitor/cmd_monitor.c
@@ -115,3 +115,15 @@ void deinit_cmd_monitor(void)
 		deinit_socket(sock_server);
 	}
 }
+int acrn_parse_cmd_monitor(char *arg)
+{
+	int err = -1;
+	size_t len = strnlen(arg, UNIX_SOCKET_PATH_MAX);
+
+	if (len < UNIX_SOCKET_PATH_MAX) {
+		strncpy(socket_path, arg, len + 1);
+		pr_notice("Command monitor: using soket path %s\n", socket_path);
+		err = 0;
+	}
+	return err;
+}
--- a/devicemodel/core/cmd_monitor/cmd_monitor.c
+++ b/devicemodel/core/cmd_monitor/cmd_monitor.c
@@ -1,0 +1,117 @@
+/*
+ * Copyright (C) 2022 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+#include <termios.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/socket.h>
+#include <pthread.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <sys/un.h>
+#include <cjson/cJSON.h>
+#include "socket.h"
+#include "command.h"
+#include "vmmapi.h"
+#include "cmd_monitor.h"
+#include "log.h"
+#include "dm.h"
+#include "command_handler.h"
+
+struct socket_dev *sock_server; /* socket server instance */
+static char socket_path[UNIX_SOCKET_PATH_MAX];
+
+static struct command *parse_command(char *cmd_msg, int fd)
+{
+	struct command *cmd = NULL;
+	const cJSON *execute;
+	const cJSON *arguments;
+	cJSON *cmd_json;
+
+	cmd_json = cJSON_Parse(cmd_msg);
+	if (cmd_json == NULL) {
+		const char *error_ptr = cJSON_GetErrorPtr();
+		if (error_ptr != NULL) {
+			fprintf(stderr, "Error before: %s\n", error_ptr);
+        }
+        return NULL;
+	}
+	execute = cJSON_GetObjectItemCaseSensitive(cmd_json, "command");
+	if (cJSON_IsString(execute) && (execute->valuestring != NULL)) {
+		pr_info("Command name: \"%s\"\n", execute->valuestring);
+
+		cmd = find_command(execute->valuestring);
+		if (cmd != NULL) {
+			cmd->para.fd = fd;
+			arguments = cJSON_GetObjectItemCaseSensitive(cmd_json, "arguments");
+			if (cJSON_IsString(arguments) && (arguments->valuestring != NULL)) {
+				pr_info("Command arguments: \"%s\"\n", arguments->valuestring);
+				strncpy(cmd->para.option, arguments->valuestring, CMD_ARG_MAX - 1U);
+			}
+		} else {
+			pr_err("Command [%s] is not supported.\n", execute->valuestring);
+		}
+	}
+	cJSON_Delete(cmd_json);
+	return cmd;
+}
+static void monitor_cmd_dispatch(char *cmd_msg, int fd)
+{
+	struct command *cmd;
+
+	cmd = parse_command(cmd_msg, fd);
+	if (cmd != NULL) {
+		dispatch_command_handlers(cmd);
+	}
+	return;
+}
+
+int init_socket_server(void)
+{
+	int ret = 0;
+
+	if (strnlen(socket_path, UNIX_SOCKET_PATH_MAX) == 0) {
+		pr_err("Failed to initialize command monitor due to invalid socket path.\n");
+	}
+
+	sock_server = init_socket(socket_path);
+	if (sock_server == NULL)
+		return -1;
+	ret = open_socket(sock_server, monitor_cmd_dispatch);
+	if (ret < 0)
+		return ret;
+	return ret;
+}
+
+static void register_socket_message_handlers(struct vmctx *ctx)
+{
+	struct handler_args arg;
+	arg.channel_arg = sock_server;
+	arg.ctx_arg = ctx;
+	register_command_handler(user_vm_destroy_handler, &arg, DESTROY);
+	register_command_handler(user_vm_blkrescan_handler, &arg, BLKRESCAN);
+}
+
+int init_cmd_monitor(struct vmctx *ctx)
+{
+	int ret;
+	ret = init_socket_server();
+	register_socket_message_handlers(ctx);
+	return ret;
+}
+void deinit_cmd_monitor(void)
+{
+	if (sock_server != NULL) {
+		close_socket(sock_server);
+		deinit_socket(sock_server);
+	}
+}
--- a/devicemodel/core/cmd_monitor/command.c
+++ b/devicemodel/core/cmd_monitor/command.c
@@ -1,0 +1,88 @@
+/*
+ * Copyright (C) 2022 Intel Corporation
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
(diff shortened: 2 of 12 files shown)


acrnsi:

wuxyintel:
@acrnsi Please help to install libcjson lib in the service VM, "sudo apt-get install libjson-c-dev", thanks!

devicemodel/core/cmd_monitor/cmd_monitor.c
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+int acrn_parse_cmd_monitor(char *arg)
+{
+	int err = -1;
+	size_t len = strnlen(arg, UNIX_SOCKET_PATH_MAX);
+
+	if (len < UNIX_SOCKET_PATH_MAX) {
+		strncpy(socket_path, arg, len + 1);
+		pr_notice("Command monitor: get soket path %s\n", socket_path);
```
gvancuts, line 125:
```suggestion
		pr_notice("Command monitor: using socket path %s\n", socket_path);
```

devicemodel/core/cmd_monitor/command.c
```diff
@@ -0,0 +1,88 @@
...
+	return NULL;
+}
+
+int register_command_handler(cmd_handler *fn, struct handler_args *arg, const char *cmd_name)
+{
+	struct command *cmd;
+	struct handler_args *handler_arg;
+
+	if ((!fn) || (!arg) || (!cmd_name)) {
+		pr_err("%s :Failed to register command_handler.\n", __func__);
```
gvancuts, line 62:
```suggestion
		pr_err("%s : Failed to register command_handler.\n", __func__);
```

devicemodel/core/cmd_monitor/command.c
```diff
@@ -0,0 +1,88 @@
...
+		return -EINVAL;
+	}
+
+	cmd = find_command(cmd_name);
+	if (cmd == NULL) {
+		pr_err("%s :invalid command name %s.\r\n", __func__, cmd_name);
```
gvancuts, line 68:
```suggestion
		pr_err("%s : invalid command name %s.\r\n", __func__, cmd_name);
```

devicemodel/core/cmd_monitor/command.c
```diff
@@ -0,0 +1,88 @@
...
+		return -EINVAL;
+	}
+
+	if (cmd->cmd_handler.fn != NULL) {
+		pr_err("Failed to register command handler since the handler have been register.\n");
```
gvancuts, line 73:
```suggestion
		pr_err("Failed to register command handler since the handler has already been registered.\n");
```
?

devicemodel/core/cmd_monitor/command_handler.c
```diff
@@ -0,0 +1,124 @@
...
+
+	client = find_socket_client(sock, cmd_para->fd);
+	if (client == NULL)
+		return -1;
+
+	ret = vm_monitor_blkrescan(hdl_arg->ctx_arg, cmd_para->option);
+	if (ret >= 0) {
+		cmd_completed = true;
+	} else {
+		pr_err("Failed to rescan virtio-block device.\n");
```
gvancuts, line 116:
```suggestion
		pr_err("Failed to rescan virtio-blk device.\n");
```

devicemodel/core/cmd_monitor/command_handler.c
```diff
@@ -0,0 +1,124 @@
...
+#include "command.h"
+#include "socket.h"
+#include "command_handler.h"
+#include "dm.h"
+#include "pm.h"
+#include "vmmapi.h"
+#include "log.h"
+#include "monitor.h"
+
+#define SUCCESSED 0
```
gvancuts, line 24:
```suggestion
#define SUCCESS 0
```
Or `SUCCEEDED`

devicemodel/core/cmd_monitor/socket.c
```diff
@@ -0,0 +1,300 @@
...
+	ret = snprintf(s_un.sun_path, sizeof(s_un.sun_path), "%s",
+			sock_path);
+	if (ret < 0 || ret >= sizeof(s_un.sun_path))
+		goto err_close_socket;
+
+	if (bind(sock_fd, (struct sockaddr *)&s_un, sizeof(s_un)) < 0)
+		goto err_close_socket;
+	if (listen(sock_fd, num) < 0)
+		goto err_close_socket;
+	pr_info("Listening on:%s, socket fd = %d.\r\n", sock_path, sock_fd);
```
gvancuts, line 41:
```suggestion
	pr_info("Listening on: %s, socket fd = %d.\r\n", sock_path, sock_fd);
```

gvancuts (changes requested):
Do not forget to add the new build dependency when you send the documentation PR.

NanlinXie:
@gvancuts, @dbkinder @amyreye , After this patch merge, User VM boot aslo need additional library installed in target board: sudo apt-get install libcjson-dev, do you have any concern for that. Othervise, user vm boot failure with below message:
root@192168132sos-dom0:/home/auto_script# bash launch_user_vm_id2.sh
cpu8 online=0
cpu9 online=0
tap device existed, reuse tap_RT
acrn-dm: error while loading shared libraries: libcjson.so.1: cannot open shared object file: No such file or directory

ywan170:
@gvancuts We can drop the json implementation, define our own strings based commands for communication just like the v1 solution. This way can remove the json library dependency. And I believe that there should no too many requirements that require add new commands in future...

gvancuts:
> @gvancuts, @dbkinder @amyreye , After this patch merge, User VM boot also need additional library installed in target board: sudo apt-get install libcjson-dev, do you have any concern for that. Othervise, user vm boot failure with below message: root@192168132sos-dom0:/home/auto_script# bash launch_user_vm_id2.sh cpu8 online=0 cpu9 online=0 tap device existed, reuse tap_RT acrn-dm: error while loading shared libraries: libcjson.so.1: cannot open shared object file: No such file or directory

I am personally not concerned by this additional dependency. I assume you don't need the dev package but just the `libcjson1` package installed, and only in the Service VM, is that correct?

ywan170:
> library

This looks good to me! The user can aware of the dependency when they try to execute "dpdk -i" and try to use "apt-get install libcjson1" to resolve the problem.

gvancuts:
> @gvancuts We can drop the json implementation, define our own strings based commands for communication just like the v1 solution. This way can remove the json library dependency. And I believe that there should no too many requirements that require add new commands in future...

I don't think this is necessary. Let's keep this implementation and if it turns out later that this dependency is problematic, we can then remove it.

dbkinder:
> @gvancuts, @dbkinder @amyreye , After this patch merge, User VM boot also need additional library installed in target board: sudo apt-get install libcjson-dev, do you have any concern for that.  ...

As long as the documentation for prerequisites (for the non-debian package install) and the packaging scripts for the Debian install are updated to include this new dependency, and we update the release notes, I don't have a concern about additional dependencies being added to the target system.

devicemodel/core/cmd_monitor/cmd_monitor.c
wuxyintel: Thanks, will update this.

devicemodel/core/cmd_monitor/command.c
wuxyintel: Thanks, will update this.

devicemodel/core/cmd_monitor/command.c
wuxyintel: Thanks, will update this.

devicemodel/core/cmd_monitor/command.c
wuxyintel: Thanks, will update this.

devicemodel/core/cmd_monitor/command_handler.c
wuxyintel: Thanks, will update this.

devicemodel/core/cmd_monitor/socket.c
wuxyintel: Thanks, will update this.

wuxyintel:
@dbkinder Nanlin will help to update document about libcjson package installation. Thanks!

wuxyintel:
@gvancuts We need to install libcjson-dev, not libcjson1, I have added one patch about it.

wuxyintel:
@dbkinder This PR don't affect HLD and acrnctl and VM manager, only need to update GSG to need user install libcjson package. Thanks!

misc/packaging/gen_acrn_deb.py
```diff
@@ -70,6 +70,7 @@ def create_acrn_deb(board, scenario, version, build_dir):
 
     listcontrol=['Package: acrn-hypervisor\n',
                 'version: %s \n'% version,
+		'Depends: libcjson-dev\n',
```
gvancuts, line 73: This is not correct. The `-dev` package would be a `build-depends` but you do not need it on the target system (which is where the Debian package will be installed). All you need there is the `libcjson1` package here.

I have not played with `build-depends`, and we don't have any of those listed in the current packaging. I would suggest we do not address it now and defer that to a different PR that would be an enhancement to the Debian packaging in general.

So for this PR: 
```suggestion
		'Depends: libcjson1\n',
```

gvancuts: **Note**: you also need to update the GSG and [this section](https://projectacrn.github.io/latest/getting-started/getting-started.html?#prepare-the-development-computer) in particular to add the `libcjson-dev` package to the list of build tools.

gvancuts (changes requested):

misc/packaging/gen_acrn_deb.py
wuxyintel: Got it, libcjson1 is fine, I will update this PR, BTW, Nanlin will ask other to update GSG about libcjson-dev. Thanks!

gvancuts: > Got it, libcjson1 is fine, I will update this PR, BTW, Nanlin will ask other to update GSG about libcjson-dev. Thanks!

It's literally a one-line patch to update the GSG, can you add this to your PR? (Instead of asking someone else to send yet another PR)

```
$ git diff
diff --git a/doc/getting-started/getting-started.rst b/doc/getting-started/getting-started.rst
index 0f57ab31f..51f11567f 100644
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -126,6 +126,7 @@ To set up the ACRN build environment on the development computer:
            e2fslibs-dev \
            pkg-config \
            libnuma-dev \
+           libcjson-dev \
            liblz4-tool \
            flex \
            bison \
```

wuxyintel: Thanks! Will add patch to update GSG in this PR.

dbkinder (approved):
doc change LGTM, thanks!

amyreye (approved):

NanlinXie (approved):

gvancuts (approved):
LGTM, thanks!

Issue #4255, opened 2019-12-16
Uos has no ip address


Mingyuan18:
Environment
clearlinux 31730
HW/Board
nuc7i7dnb
Steps
Install the build on nuc7i7dnb, and use the launch_uos_id1.sh to launch laag
Expected result
uos could boot up and has ip address.
Actual result
uos boot up but no ip address

Mingyuan18:
[External_System_ID] ACRN-5811

Mingyuan18:
will not reproduce at V1.5 release .

PR #4277, opened 2019-12-19
acrn-config: Cherry pick acrn-config patches from release v1.5

liuwei987:
- add 'logger_setting' into launch script
- refine mount device for virtio-blk
- add 'tap_' perfix for virtio-net
- enable log_setting in all VMs
- check pass-thruogh device for audio/audio_codec
- correct vuart1 setting in scenario config
    Tracked-On: #3854
    Signed-off-by: Wei Liu <weix.w.liu@intel.com>
    Acked-by: Victor Sun <victor.sun@intel.com>

PR #7106, opened 2022-02-15
Rewriting the generator of launch scripts

junjiemao1:
The launch script generator today (and the scripts that it generates) is fundamentally built on the concept of PCI device classes, with the restriction that at most one PCI function per class can be passed through to a post-launched VM. This has put improper constraint on the scenarios users can set up, especially on server platforms or those with SR-IOV capable devices.

As it is too tedious to change such deep-rooted concept, this patch rewrites the launch script generator and refines the structure of the generated scripts so that PCI functions are identified only by their BDF.

This change serves as a mandatory step to align the way how passthrough devices are configured for pre-launched and post-launched VMs, which eventually allows us to present a unified view in the configurator for assigning passthrough device.


acrnsi:

wenlingz:

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,364 @@
...
+    sos_vm_id = eval_xpath(scenario_etree, "//vm[vm_type='SERVICE_VM']/@id")
+    post_vms = scenario_etree.xpath("//vm[starts-with(vm_type, 'POST_')]")
+    if sos_vm_id is None and len(post_vms) > 0:
+        logging.error("The scenario does not contain a service VM. Will not generate launch scripts for the post-launched VMs in the scenario.")
+        return 1
+    sos_vm_id = int(sos_vm_id)
+
+    try:
+        os.mkdir(out_dir)
+    except FileExistsError:
```
dbkinder, line 326: The FileExistsError exception is raised when there's a file (or directory) existing with the name of the directory you want to create.  If there's a plain file already existing, you need to catch that as an error here and not let it pass by because you won't be able to write the generated launch script into a directory with that name.

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,364 @@
...
+        post_vm_ids = [int(vm_scenario_etree.get("id")) - sos_vm_id for vm_scenario_etree in post_vms]
     else:
-        print("Launch files generate failed".format(output))
+        post_vm_ids = [user_vmid]
+
+    for post_vm_id in post_vm_ids:
+        vm_scenario_etree = eval_xpath(scenario_etree, f"//vm[@id = {sos_vm_id + post_vm_id}]")
+        vm_launch_etree = eval_xpath(launch_etree, f"//user_vm[@id='{post_vm_id}']")
+        if vm_scenario_etree is None:
+            logging.warning(f"Post-launched VM {post_vm_id} is not specified in the scenario XML and has no launch script generated.")
```
dbkinder, line 341: I think we refer to VMs by their name and not their ID now, right?
```suggestion
            logging.warning(f"Post-launched VM {post_vm_id} is not specified in the scenario XML, so no launch script will be generated.")
```

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,364 @@
...
+            continue
+
+        if vm_launch_etree is None:
+            logging.warning(f"Post-launched VM {post_vm_id} is not specified in the launch XML and has no launch script generated.")
```
dbkinder, line 345: Same here, don't we refer to VMs by name now?
```suggestion
            logging.warning(f"Post-launched VM {post_vm_id} is not specified in the launch XML, so no launch script will be generated.")
```

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,364 @@
...
 
-    return err_dic
+    return 0
 
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--board", help="the XML file summarizing characteristics of the target board")
+    parser.add_argument("--scenario", help="the XML file specifying the scenario to be set up")
+    parser.add_argument("--launch", help="the XML file specifying the parameters of post-launched VMs")
+    parser.add_argument("--user_vmid", type=int, default=0, help="the post-launched VM ID (as is specified in the launch XML) whose launch script is to be generated, or 0 if all post-launched VMs shall be processed")
```
dbkinder, line 358: The success message writes out the VM name, so shouldn't we use a VM name here and not ID number?  @junjiemao1

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,364 @@
...
+    parser.add_argument("--out", default="output", help="path to the directory where generated scripts are placed")
+    args = parser.parse_args()
 
-if __name__ == '__main__':
+    logging.basicConfig(level="INFO")
```
dbkinder, line 362: There should be an argument that let's the user change this default logging level, right?

dbkinder (changes requested):
Some functionality comments and message tweaks.

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,364 @@
...
+
+    def to_string(self):
+        s = ""
+
+        with open(self.script_template_path, "r") as f:
+            s += f.read()
+
+        s += """
+###
+# The followings are generated by generate_launch_scripts.py.
```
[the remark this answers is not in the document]
junjiemao1: Will update.

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,364 @@
...
+
+def main(board_xml, scenario_xml, launch_xml, user_vmid, out_dir):
+    board_etree = etree.parse(board_xml)
+    scenario_etree = etree.parse(scenario_xml)
+    launch_etree = etree.parse(launch_xml)
+
+    sos_vm_id = eval_xpath(scenario_etree, "//vm[vm_type='SERVICE_VM']/@id")
+    post_vms = scenario_etree.xpath("//vm[starts-with(vm_type, 'POST_')]")
+    if sos_vm_id is None and len(post_vms) > 0:
+        logging.error("The scenario does not contain a service VM. Will not generate launch scripts for the post-launched VMs in the scenario.")
```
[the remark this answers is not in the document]
junjiemao1: Will update.

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,364 @@
...
+
+    ###
+    # CPU and memory resources
+    ###
+    cpus_in_launch_xml = set(vm_launch_etree.xpath(".//cpu_affinity/pcpu_id[text() != '']/text()"))
+    cpus_in_scenario_xml = set(vm_scenario_etree.xpath(".//cpu_affinity/pcpu_id[text() != '']/text()"))
+    if cpus_in_launch_xml:
+        cpus = cpus_in_scenario_xml & cpus_in_launch_xml
+        if not cpus:
+            logging.error(f"CPUs assigned to VM '{vm_name}' in the launch XML is outside of those allowed by the scenario XML.")
```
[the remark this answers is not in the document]
junjiemao1: Will update.

misc/config_tools/launch_config/launch_cfg_gen.py
junjiemao1: This parameter is mainly for QA and is thus defined in the same way as it was in its predecessor in order to minimize impact to the auto-test scripts. Eventually this script will not be invoked by the user directly, but rather via the configurator or via Makefile.

misc/config_tools/launch_config/launch_cfg_gen.py
junjiemao1: Today definitions of the VM in scenario & launch XMLs are linked by the ID, which is why I keep the logic as it was. After we combine those two kinds of XMLs in a week or two, such checks will no longer be needed any more.

misc/config_tools/launch_config/launch_cfg_gen.py
junjiemao1: Ditto.

misc/config_tools/launch_config/launch_cfg_gen.py
junjiemao1: Considering that this script will eventually be invoked by the configurator or Makefile rather by users directly, do you think we still need a parameter for tweaking log levels?

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,364 @@
...
+        s += "# Deinitialzing\n"
+        for command in self._deinit_commands:
+            s += f"{command}\n"
+
+        return s
+
+    def write_to_file(self, path):
+        with open(path, "w") as f:
+            f.write(self.to_string())
+            logging.info(f"Successfully generated launch script {path} for VM '{self._vm_name}'.")
```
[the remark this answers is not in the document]
junjiemao1: That is done implicitly with the `with` statement. More details on this syntax are available at https://www.python.org/dev/peps/pep-0343/

misc/config_tools/launch_config/launch_cfg_gen.py
junjiemao1: Will check if `out_dir` is a regular file and exit if that's the case.

misc/config_tools/launch_config/launch_cfg_gen.py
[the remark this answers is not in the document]
dbkinder: I learned something today :) Thanks.

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,364 @@
...
+        s += "acrn-dm ${dm_params[*]}\n\n"
+
+        s += "# Deinitialzing\n"
+        for command in self._deinit_commands:
+            s += f"{command}\n"
+
+        return s
+
+    def write_to_file(self, path):
+        with open(path, "w") as f:
```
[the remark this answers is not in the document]
junjiemao1: Considering the use cases of this script (via the configurator or via the Makefile), this script is likely to be invoked in a non-interactive manner and cannot take additional user input for double confirmation.

misc/config_tools/launch_config/launch_cfg_gen.py
dbkinder: Is there a debug mode for the configurator where we'd want more information than "INFO"?

junjiemao1: There is no log with `DEBUG` log level as of now.

NanlinXie (approved):

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,366 @@
...
+    script.add_plain_dm_parameter(f"{vm_name}")
+
+    return script
+
+def main(board_xml, scenario_xml, launch_xml, user_vmid, out_dir):
+    board_etree = etree.parse(board_xml)
+    scenario_etree = etree.parse(scenario_xml)
+    launch_etree = etree.parse(launch_xml)
+
+    sos_vm_id = eval_xpath(scenario_etree, "//vm[vm_type='SERVICE_VM']/@id")
```
gvancuts, line 317: Can we avoid using `sos` here, e.g. change that variable name to `service_vm_id`?

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,366 @@
...
+        os.mkdir(out_dir)
+    except FileExistsError:
+        if os.path.isfile(out_dir):
+            logging.error(f"Cannot create output directory {out_dir}: File exists")
+            return 1
+    except Exception as e:
+        logging.error(f"Cannot create output directory: {e}")
+        return 1
+
+    if user_vmid == 0:
```
gvancuts, line 334: Nitpick: should we use `user_vm_id` instead to be consistent with `service_vm_id` (and `post_vm_id`, etc.)?

misc/config_tools/launch_config/launch_script_template.sh
```diff
@@ -0,0 +1,166 @@
...
+    echo ${physical_bdf} > /sys/bus/pci/devices/${physical_bdf}/driver/unbind
+    echo ${physical_bdf} > /sys/bus/pci/drivers/pci-stub/bind
+}
+
+function create_tap() {
+    # create a unique tap device for each VM
+    tap=$1
+    tap_exist=$(ip a | grep "$tap" | awk '{print $1}')
+    if [ "$tap_exist"x != "x" ]; then
+        echo "tap device existed, reuse $tap"
```
gvancuts, line 57: Can we rephrase that? @dbkinder any better suggestion than:
```suggestion
        echo "$tap TAP device already available, reusing it"
```

misc/config_tools/launch_config/launch_script_template.sh
```diff
@@ -0,0 +1,166 @@
...
+    else
+        ip tuntap add dev $tap mode tap
+    fi
+
+    # if acrn-br0 exists, add VM's unique tap device under it
+    br_exist=$(ip a | grep acrn-br0 | awk '{print $1}')
+    if [ "$br_exist"x != "x" -a "$tap_exist"x = "x" ]; then
+        echo "acrn-br0 bridge aleady exists, adding new tap device to it..."
```
gvancuts, line 65:
```suggestion
        echo "acrn-br0 bridge already exists, adding new $tap TAP device to it..."
```

misc/config_tools/launch_config/launch_script_template.sh
```diff
@@ -0,0 +1,166 @@
...
+        ip link set dev "$tap" up
+    fi
+}
+
+function mount_partition() {
+    partition=$1
+
+    tmpdir=`mktemp -d`
+    mount ${partition} ${tmpdir}
+    echo ${tmpdir}
```
gvancuts, line 77: What is this? (I should know I guess but I'm blanking...)

misc/config_tools/launch_config/launch_script_template.sh
```diff
@@ -0,0 +1,166 @@
...
+
+function add_virtual_device() {
+    slot=$1
+    kind=$2
+    options=$3
+
+    if [ "${kind}" = "virtio-net" ]; then
+        # Create the tap device
+        tap_conf=${options%,*}
+        create_tap "tap_${tap_conf#tap=}" >> /dev/stderr
```
gvancuts, line 140: Is this prefixing `tap_` to a TAP device name given by the user?

gvancuts (changes requested):

misc/config_tools/launch_config/launch_cfg_gen.py
junjiemao1: Updated in v3.

misc/config_tools/launch_config/launch_cfg_gen.py
junjiemao1: I'll update this for internal variables, but keep the command line interface as `--user-vmid` for now as this is used by the premerge tests.

After we integrate this script into the build system and the premerge test scripts get rid of invoking this script directly, we can refine those interfaces.

misc/config_tools/launch_config/launch_script_template.sh
junjiemao1: Updated in v3.

misc/config_tools/launch_config/launch_script_template.sh
junjiemao1: Updated in v3.

misc/config_tools/launch_config/launch_script_template.sh
junjiemao1: In today's launch XML configurations, users are allowed to specify the path of an image in the format `<partition>:<path in that partition>`. In that case the launch script will automatically mount that partition somewhere, using the rootfs image as specified in the path, and unmount the partition after the device model exits.

This helper function (and the `unmount_partition` as well) are the helper functions to implement to mount/unmount actions.

misc/config_tools/launch_config/launch_script_template.sh
junjiemao1: Yes, today the virtio-net parameters are always in the form `tap=xxx,mac_seed=yyy` (AFAIK), so picking the first option before the first comma and strip the `tap=` string evaluates to exactly the tap name specified by the user.

gvancuts: That's understood but my concern is that the TAP name given to the `create_tap` function adds the `tap_` prefix to the name given by the user, or am I reading this wrong?

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,366 @@
...
+###
+"""
+        s += "\n"
+
+        s += "# Defining variables that describe VM types\n"
+        for name, value in self._vm_descriptors.items():
+            s += f"{name}={value}\n"
+        s += "\n"
+
+        s += "# Initialzing\n"
```
gvancuts, line 123:
```suggestion
        s += "# Initializing\n"
```

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -1,269 +1,366 @@
...
+        s += "# Invoking ACRN device model\n"
+        s += "dm_params=(\n"
+        for param in self._dm_parameters:
+            s += f"    {param}\n"
+        s += ")\n\n"
+
+        s += "echo \"Launch device model with parameters: ${dm_params[*]}\"\n"
+        s += "acrn-dm ${dm_params[*]}\n\n"
+
+        s += "# Deinitialzing\n"
```
gvancuts, line 137:
```suggestion
        s += "# Deinitializing\n"
```

misc/config_tools/launch_config/launch_script_template.sh
gvancuts: Thanks for the explanation @junjiemao1 . And the `echo` command is there to make sure the right argument ends up in the launch script being generated, is that correct?

misc/config_tools/launch_config/launch_script_template.sh
junjiemao1: This is the case and is the logic we have today, which is also preserved during this rewriting. I'm not sure to what extent our validation depends on this fact, though.

misc/config_tools/launch_config/launch_script_template.sh
junjiemao1: Echoing the created directory is to allow the caller of this function to get the path so that it can be removed after the device model exits.

As bash function always returns exit codes, I have to use stdout as the channel for returning more info to the caller.

misc/config_tools/launch_config/launch_cfg_gen.py
junjiemao1: Updated.

misc/config_tools/launch_config/launch_cfg_gen.py
junjiemao1: Updated.

misc/config_tools/launch_config/launch_script_template.sh
junjiemao1: This was originally introduced by PR #4277 to resolve issue #4255. I would recommend to exclude this change from this PR and create a separate thread (perhaps another GitHub issue) to discuss with the original authors why this is needed and whether we can get rid of them.

szhen11:
The the folder to store launch scripts generated by config tools is changed, will update CI scripts and rerun CI test.

szhen11:
start to run premerge test

szhen11:
@junjiemao1 after update CI scripts, rtvm launch failed with below error log:
root@1921681228sos-dom0:/home/auto_script# bash launch_user_vm_id2.sh
tap_RT TAP device already available, reusing it.
Launch device model with parameters: --cpu_affinity 4,6 -m 1024M --ovmf /usr/share/acrn/bios/OVMF.fd -s 0:0,hostbr
idge --virtio_poll 1000000 -s 3,virtio-console,@stdio:stdio_port -s 4,virtio-net,tap=RT,mac_seed=b0:41:6f:04:8e:cd
-POST_RT_VM1 -s 5,virtio-blk,/root/rt/core-image-weston-intel-corei7-64.wic --rtvm--lapic_pt --logger_setting cons
ole,level=4;kmsg,level=3;disk,level=5 POST_RT_VM1
SW_LOAD: get ovmf path /usr/share/acrn/bios/OVMF.fd, size 0x200000
acrn-dm: unrecognized option '--rtvm--lapic_pt'
Usage: acrn-dm [-hYv] [-B bootargs] [-E elf_image_path]
               [-k kernel_image_path]
               [-l lpc] [-m mem] [-r ramdisk_image_path]
               [-s pci] [--ovmf ovmf_file_path]
               [--enable_trusty] [--intr_monitor param_setting]
               [--acpidev_pt HID] [--mmiodev_pt MMIO_Regions]
               [--vtpm2 sock_path] [--virtio_poll interval]
               [--cpu_affinity lapic_id] [--lapic_pt] [--rtvm] [--windows]
               [--debugexit] [--logger_setting param_setting]
               [--ssram] <vm>
       -B: bootargs for kernel
       -E: elf image path
       -h: help
       -k: kernel image path
       -l: LPC device configuration
       -m: memory size in MB
       -r: ramdisk image path
       -s: <slot,driver,configinfo> PCI slot config
       -v: version
       --ovmf: ovmf file path
       --ssram: Congfiure Software SRAM parameters
       --cpu_affinity: list of Service VM vCPUs assigned to this User VM, the vCPUs are

szhen11:
start to run premerge test

misc/config_tools/launch_config/launch_script_template.sh
gvancuts: OK, let's do it this way.

I'm pretty sure that bug was precisely because the code was looking the `tap` keyword in the device name, something that is no longer the case. But let's validate this with the code owner and fix this in a subsequent PR.

Issue #7107, opened 2022-02-15
The `serial.conf` file is not installed in the right place


gvancuts:
The `serial.conf` file is currently put in the ACRN `libdir` - see https://github.com/projectacrn/acrn-hypervisor/blob/b5fc55090c5d072a67310f34599cc2546c60b923/hypervisor/Makefile#L409

This is not the right destination for it and as a consequence, we ask the user to manually copy it to the `/etc` folder. We should install it there directly instead.

PR #7108, opened 2022-02-15
Makefile: put 'serial.conf' in final location

gvancuts:
The 'serial.conf' file need to be put in /etc/, but it is currently being
installed in $(libdir)/acrn/. We therefore ask the user to manually copy that
file over to the /etc/ folder. This patch fixes that by installing
'serial.conf' directly there.

Tracked-On: #7107


acrnsi:

Issue #7109, opened 2022-02-15
Python traceback if the 'dpkg' tool is not available


gvancuts:
Our `Makefile` will, by default, try to build a Debian package of ACRN. This is very useful to users and is the right thing to do since we only officially support Ubuntu. It does, however, generate an error if we attempt to build ACRN on a non-Debian system (which wouldn't have the `dpkg` utility).

It would be preferable from a user point of view to catch this ahead of running the script and simply ignore this last step (i.e. ACRN and all artefacts would still be built, just that a Debian package with all of these would not be created).

NanlinXie:
[External_System_ID] ACRN-7612

dbkinder:
dpkg check is merged now.

Issue #7095, opened 2022-02-09
Hypervisor Makefile doesn't cleanly report missing prerequisites


dbkinder:
If required packages aren't installed, the hypervisor Makefile will fail with non-friendly errors, often with a Python stack trace.  
For DX improvement, when possible, we should check for and cleanly report missing dependencies.

For example, if the defusedxml Python package isn't installed, this happens:
```
~/projectacrn/acrn-hypervisor$ make hypervisor
make[1]: Entering directory '/mnt/c/Users/dbkinder/GitHub/projectacrn/acrn-hypervisor/hypervisor'
Traceback (most recent call last):
  File "/mnt/c/Users/dbkinder/GitHub/projectacrn/acrn-hypervisor/misc/config_tools/static_allocators/main.py", line 12, in <module>
    import common
  File "/mnt/c/Users/dbkinder/GitHub/projectacrn/acrn-hypervisor/misc/config_tools/static_allocators/../library/common.py", line 11, in <module>
    import defusedxml.ElementTree as ET
ModuleNotFoundError: No module named 'defusedxml'
make[1]: *** No rule to make target '/home/dbkinder/projectacrn/acrn-hypervisor/build/hypervisor/configs/allocation.xml', needed by '/home/dbkinder/projectacrn/acrn-hypervisor/build/hypervisor/configs/unified.xml'.  Stop.
make[1]: Leaving directory '/mnt/c/Users/dbkinder/GitHub/projectacrn/acrn-hypervisor/hypervisor'
make: *** [Makefile:116: hvdefconfig] Error 2
```

Or if ``dpkg`` (used for Debian package management) isn't installed (that would be the case on non-Debian installations), you'd see this:
```
ACRN Configuration Summary:
Build directory: /workspace/build/hypervisor
This build directory is configured with the settings below.
- BOARD = nuc11tnbi5
- SCENARIO = shared
- RELEASE = n
Traceback (most recent call last):
  File "/workspace/misc/packaging/gen_acrn_deb.py", line 169, in <module>
    create_acrn_deb(args.board_name, args.scenario, args.version, args.build_dir)
  File "/workspace/misc/packaging/gen_acrn_deb.py", line 110, in create_acrn_deb
    ret = run_command('dpkg -b acrn_release_deb acrn-%s-%s-%s.deb' %(board, scenario, version), build_dir)
  File "/workspace/misc/packaging/gen_acrn_deb.py", line 15, in run_command
[...]

PR #7111, opened 2022-02-15
Makefile: detect if 'dpkg' is available

gvancuts:
The 'dpkg' utility is required in order to build Debian packages. If it is not
installed and available on the build system, the 'make' instructions will fail
with a Python traceback. Fix this by detecting its availability and doing a
graceful exit if not (or simply skip the step).

Tracked-On: #7109

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,13 +79,16 @@ BUILD_TAG ?=
 HV_CFG_LOG = $(HV_OUT)/cfg.log
 VM_CONFIGS_DIR = $(T)/misc/config_tools
 ASL_COMPILER ?= $(shell which iasl)
+DPKG_BIN ?= $(shell which dpkg)
 
 .PHONY: all hypervisor devicemodel tools life_mngr doc
 all: hypervisor devicemodel tools
 	@cat $(HV_CFG_LOG)
-	@DEB_BOARD=$$(grep "BOARD" $(HV_CFG_LOG) | awk -F '= ' '{print  $$2 }'); \
-	DEB_SCENARIO=$$(grep "SCENARIO" $(HV_CFG_LOG) | awk -F '= ' '{print  $$2 }'); \
-	python3 misc/packaging/gen_acrn_deb.py acrn_all $(ROOT_OUT) --version=$(FULL_VERSION) --board_name="$$DEB_BOARD" --scenario="$$DEB_SCENARIO"
+	@if [ -x "$(DPKG_BIN)" ]; then \
+	  DEB_BOARD=$$(grep "BOARD" $(HV_CFG_LOG) | awk -F '= ' '{print  $$2 }'); \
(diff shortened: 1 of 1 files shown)


acrnsi:

Makefile
```diff
@@ -102,8 +105,13 @@ define install_acrn_debug
 endef
 
 HV_MAKEOPTS := -C $(T)/hypervisor BOARD=$(BOARD) SCENARIO=$(SCENARIO) HV_OBJDIR=$(HV_OUT) RELEASE=$(RELEASE)
+
 board_inspector:
-	@python3 misc/packaging/gen_acrn_deb.py board_inspector $(ROOT_OUT) --version=$(FULL_VERSION)
+	@if [ -x "$(DPKG_BIN)" ]; then \
+	  python3 misc/packaging/gen_acrn_deb.py board_inspector $(ROOT_OUT) --version=$(FULL_VERSION); \
+	else \
+	  echo -e "The 'dpkg' utility is not available. Skipping."; \
```
dbkinder, line 113: skipping what?
```suggestion
	  echo -e "The 'dpkg' utility is not available. Unable to create Debian package for board_inspector."; \
```

Makefile
```diff
@@ -79,13 +79,16 @@ BUILD_TAG ?=
 HV_CFG_LOG = $(HV_OUT)/cfg.log
 VM_CONFIGS_DIR = $(T)/misc/config_tools
 ASL_COMPILER ?= $(shell which iasl)
+DPKG_BIN ?= $(shell which dpkg)
```
dbkinder, line 82: Can we generalize checking for uninstalled or version dependencies early in the processing, maybe in another PR?  Issue #7095 tracks improved reporting of missing dependencies.

dbkinder (changes requested):

Makefile
gvancuts: I agree we should have a more generic mechanism in place. This one is a slightly special case though because:
1. `dpkg` is not an explicit dependency unlike the other tools that are all mentionned in the GSG. Note that we shouldn't add it to the GSG either because not having it does not prevent you from building ACRN, it only prevents you from creating the Debian package which is not a problem per se (especially since if `dpkg` is not there, it's precisely because you're not on a Debian system)
2. The alternative to this patch in my opinion, would be to remove the fact that the Debian package is created automatically by default and we make it an explicit target that the user would call, e.g. `make deb-pkg` (as an alternative to `make targz-pkg`. This explicit target was my first choice, but I'm fine with this being done automatically without asking the user too... provided we don't error out if that part (only) fails.

How about we take this patch in, and refactor the whole thing when we have time to address #7095 in a more systemic and generic fashion (something I do not have time to do now)

dbkinder: Works for me @gvancuts

dbkinder (approved):
LGTM, thanks.

wenlingz:
@gvancuts please add Tracked-On in second commit, thanks

wenlingz:
start to run premerge test

PR #7113, opened 2022-02-16
hv: replace the CONFIG_PLATFORM_RAM_SIZE with get_e820_ram_size

weichenli-intel:
The e820 module could get the RAM info on run time, but the RAM size
and MAX address was limited by CONFIG_PLATFORM_RAM_SIZE which was
predefined by config tool.

Current solution can't support single binary for different boards or
platforms and the CONFIG_PLATFORM_RAM_SIZE can't matching the RAM size
if user have not update config tools setting after the device changed.

So this patch remove the CONFIG_PLATFORM_RAM_SIZE and calculate ram
size on run time, then replace the CONFIG_PLATFORM_RAM_SIZE with
get_e820_ram_size for mmu, ept and vept.

Tracked-On: #6690

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -158,6 +158,9 @@ void init_pcpu_pre(bool is_bsp)
 
 		init_e820();
 
+		/* reserve ppt buffer from e820 */
+		allocate_ppt_pages();
+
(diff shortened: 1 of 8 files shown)


acrnsi:

wenlingz:

anthonyzxu (approved):

Issue #7124, opened 2022-02-17
[DM]:  Fail to boot the Laag guest if the boot option of "pci=nomsi" is added for Guest kernel


yakuizhao:
**Describe the bug**
When the boot option of "pci=nomsi" is added for the guest kernel cmdline, it will fail to boot the Laag guest.

**Platform**
Describe the platform information, 

**Codebase**
Describe which codebase is used.

**Scenario**
Describe the scenario: Hybrid, Industry or logical_partition

**To Reproduce**

**Expected behavior**
It will be better that the guest system still can be booted even after the pci devices fall back to the legacy PCI intx.

**Additional context**
Attach the following logs if available:

- native Linux boot kernel message log
- ACRN serial port log (when issues happens)
- Service VM kernel message when issue happens
- acrn-dm log: all files in /var/log/acrn-tm
- console log when running script to launch guest VM

yakuizhao:
[External_System_ID] ACRN-7631

yakuizhao:
The patch is already merged. So this issue will be closed.

PR #7135, opened 2022-02-21
ACRN:DM: Set the desired state to inject PCI legacy intx

yakuizhao:
When the virtio-XXX pci devices fall back to legacy PCI intx,
the pci_irq_assert is called to inject the interrupt and then
the pci_irq_deassert is used to mark the completion of PCI interrupt.
Currently the HV vIOAPIC uses the pin_state for the interrupt injection
of legacy PCI intx. In such case it will fail to inject the PCI legacy
intx and the guest system fails to be booted when adding the boot option
of "pci=nomsi".

PCI legacy INTx usually use active low level trigger mode as it is Open-Drain
state and allows multitple interrupt signals to share a single line.
https://wiki.osdev.org/PCI_Local_Bus_Signals
In such case DM needs to set the correct state for the PCI device so that the
HV vIOAPIC can help to inject the PCI legacy intx.

BTW: When the MSI/MSIX is used for PCI device, it uses another mechanism
to inject the interrupt. It is harmless to configure the initial state.

Tracked-On: #7124


acrnsi:

ywan170 (approved):

Issue #7028, opened 2022-01-15
Release Note update for v3.0


NanlinXie:
Use this issue to record the future v2.8 Release Note change:
1. "–cpu_affinity": Original vcpu_affinity means the vcpus' pcpu affinity, We changed the interface to Service VM's point of view. @YuanyuanZhao609 
2. PMU/PEBS supported in RTVM, vtune/perf can run in RTVM.
3. Need additional library in both build/target machine:  sudo apt-get install libcjson-dev 
Refer to https://github.com/projectacrn/acrn-hypervisor/pull/7101
Note: for ubuntu 18.04, if you want to install libcjson-dev, you need follow below steps:
sudo add-apt-repository ppa:jrtc27/cjson
sudo apt-get update
sudo apt install libcjson1 libcjson-dev
4. Virtio-GPU VF creation and offline CPU guide and script
5. scenario XML data modification
6. New configuration upgrade document added : https://github.com/projectacrn/acrn-hypervisor/pull/7312
Rather than installing the Python dependencies by hand, those libraries shall now be installed by referring to the requirements.txt in the repo:
sudo pip3 install -r acrn-hypervisor/misc/config_tools/requirements.txt

dbkinder:
@NanlinXie Good idea.

junjiemao1:
Rather than installing the Python dependencies by hand, those libraries shall now be installed by referring to the `requirements.txt` in the repo:

`sudo pip3 install -r acrn-hypervisor/misc/config_tools/requirements.txt`

PR #7145, opened 2022-02-24
doc: revert gsg libcjson dependency

jackwhich:
The v2.7 version does not use the libcjson-dev

### DIFF
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -126,13 +126,20 @@ To set up the ACRN build environment on the development computer:
            e2fslibs-dev \
            pkg-config \
            libnuma-dev \
-           libcjson-dev \
            liblz4-tool \
            flex \
            bison \
            xsltproc \
            clang-format \
            bc
+         
+   .. note:: You need to follow these steps if you are on Ubuntu 18.04 as ``libcjson-dev`` is not available in the default repositories.
+
+      .. code-block:: bash
+
+         sudo add-apt-repository ppa:jrtc27/cjson
+         sudo apt-get update
+         sudo apt install libcjson-dev
 
 #. Install Python package dependencies:
 


acrnsi:

gvancuts:
Shouldn't that be sent to the `release_2.7` branch instead?

dbkinder:
> Shouldn't that be sent to the `release_2.7` branch instead?

For which release of ACRN is this change intended? Can you confirm if this libcjson package is needed on master (2.8) but not on 2.7?

NanlinXie:
libcjson package is needed after release_2.7, current the GSG on master is still for release_2.7/ubuntu 18.04.  The issue is that this package could not install via apt install in ubuntu 18.04, see the description in below issue. 
https://github.com/projectacrn/acrn-hypervisor/issues/7028
Amy has draft to update GSG to latest/ubuntu 20.04 and at that time we don't need such additional description for GSG.

dbkinder:
@NanlinXie If I understand things, the v2.7 GSG does not mention libcjson-dev and that's fine. For developers using the master branch, they do need this libcjson-dev library. If you're using Ubuntu 20.04 add this to the apt install as documented, but if you're using Ubuntu 18.04 you need different instructions to install this library.

If that's right, then this PR that just deletes this line in the documentation won't work?

doc/getting-started/getting-started.rst
```diff
@@ -126,7 +126,6 @@ To set up the ACRN build environment on the development computer:
            e2fslibs-dev \
            pkg-config \
            libnuma-dev \
-           libcjson-dev \
```
dbkinder, line 129: We can't delete installing this library in the master branch documentation or else the hypervisor build will fail.  This library, installed this way, works just fine if your development system is running Ubuntu 20.04.03.
We could add a note with the different instructions you'll need to do to get this library if you're running Ubuntu 18.04.03 (as documented in https://github.com/projectacrn/acrn-hypervisor/issues/7028).

dbkinder (changes requested):

doc/getting-started/getting-started.rst
jackwhich: updated

NanlinXie:
If we want to add description for libcjson-dev, we need change ACRN code from release_v2.7 to latest stable one, maybe "eb9a58c70" could be one choice, @jackwhich could ensure this working before you upgrade.

doc/getting-started/getting-started.rst
```diff
@@ -126,13 +126,20 @@ To set up the ACRN build environment on the development computer:
...
            liblz4-tool \
            flex \
            bison \
            xsltproc \
            clang-format \
            bc
+         
+   .. note:: For ubuntu 18.04, if you get an error when installing ``libcjson-dev``, you need to follow these steps:
```
gvancuts, line 136:
```suggestion
   .. note:: For Ubuntu 18.04, if you get an error when installing ``libcjson-dev``, you need to follow these steps:
```

gvancuts: Unless there are cases where Ubuntu 18.04 wouldn't need this, I would be more definite in the note and explicitly state that if you are running 18.04 you do need to follow these steps.

doc/getting-started/getting-started.rst
```diff
@@ -126,13 +126,20 @@ To set up the ACRN build environment on the development computer:
...
+
+      .. code-block:: bash
+
+         sudo add-apt-repository ppa:jrtc27/cjson
+         sudo apt-get update
+         sudo apt install libcjson1 libcjson-dev
```
gvancuts, line 142: You can probably remove ``libcjson1` from this command, can't you?

gvancuts (approved):

dbkinder (approved):
@jackwhich @amyreye When we update the GSG to use v2.8 we'll need to add ``libcjson-dev`` back on this ``apt install`` package list, right?

Issue #7146, opened 2022-02-25
mptable fail to wake up CPU1# when adding "acpi=off" into User VM cmdline.


donsheng:
**Describe the bug**
SOS rootfs:
http://sit-image.sh.intel.com/yocto/e2e/ww23.5/acrn-image-weston-intel-corei7-64.wic.acrn

HV and kernel image:
http://sit-image.sh.intel.com/acrn_daily_build/YOCTO-2022ww07.4-1/yocto-whl-ipc-i5-uefi-shared/

Steps:
1. Boot SOS and
2. Launch User VM with 2 cores configuration
3.Add "acpi=off" into the User VM grub menu

Expected result:
Using default mptable; User VM will boot up with CPU0 and CPU1

Actual result:
User VM boots up only with CPU0

Debug info:

[311910104us][cpu=0][vm3:vcpu0][sev=3][seq=172]:vlapic: vm_apicid2vcpu_id: bad lapicid 1
[ 0.021302] BIOS bug, no explicit IRQ entries, using default mptable. (tell your hw vendor)
[311929664us][cpu=0][vm3:vcpu0][sev=3][seq=173]:vlapic: vm_apicid2vcpu_id: bad lapicid 1
[311939081us][cpu=0][vm3:vcpu0][sev=3][seq=174]:vlapic: vm_apicid2vcpu_id: bad lapicid 1
[311948609us][cpu=0][vm3:vcpu0][sev=3][seq=175]:vlapic: vm_apicid2vcpu_id: bad lapicid 1
[ 10.074816] smpboot: do_boot_cpu failed(-1) to wakeup CPU#1
[ 12.402624] genirq: Flags mismatch irq 8. 00000080 (rtc0) vs. 00000000 (virtio1-input)
[ 12.435719] genirq: Flags mismatch irq 4. 00000080 (ttyS0) vs. 00000000 (virtio1-input)

**Platform**
All

**Codebase**
latest master code

**Scenario**
scenario that launches User VM

**To Reproduce**
See Describe the bug*

**Expected behavior**
lscpu shows all vcpus online (no offline)

[External_System_ID] ACRN-6463

PR #7147, opened 2022-02-25
fixed a bug where mptable failed to wake up CPU1# when adding acpi=off into User VM cmdline

donsheng:
Previous commit introduced the "use physical APIC IDs as vLAPIC IDs for VMs" change,
but it didn't update the apic_id in mptable. Changed the mptable code to also
set apic_id to physical APIC ID to fix the bug

Tracked-On: #7146

### DIFF
--- a/devicemodel/core/mptbl.c
+++ b/devicemodel/core/mptbl.c
@@ -34,6 +34,7 @@
 #include "acpi.h"
 #include "dm.h"
 #include "pci_core.h"
+#include "vmmapi.h"
 
 #define MPTABLE_BASE		0xF0000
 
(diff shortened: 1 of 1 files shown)


acrnsi:

wenlingz:

devicemodel/core/mptbl.c
```diff
@@ -117,11 +118,39 @@ static void
...
+		pr_err("%s,Err: Invalid guest_pcpu_bitmask.\n", __func__);
+		return;
+	}
+
+	pr_info("%s, guest_cpu_bitmask: 0x%x\n", __func__, guest_pcpu_bitmask);
 
 	for (i = 0; i < ncpu; i++) {
+		int pcpu_id = pcpuid_from_vcpuid(guest_pcpu_bitmask, i);
+		int lapic_id;
+
```
lifeix, line 133: Do we need to check whether the pcpu_id is valid ? Now the lapicid_from_pcpuid will also does this check.

donsheng: Hi,lifei,
yes, this patch does including the pcpu_id validity checking, see below code in this patch:
if (pcpu_id < 0) {
			pr_err("%s,Err: pcpu id is not found in guest_pcpu_bitmask.\n", __func__);
			return;
		}

		assert(pcpu_id < ACRN_PLATFORM_LAPIC_IDS_MAX);
		if (pcpu_id >= ACRN_PLATFORM_LAPIC_IDS_MAX) {
			pr_err("%s,Err: pcpu id %u should be less than ACRN_PLATFORM_LAPIC_IDS_MAX.\n", __func__, pcpu_id);
			return;
		}

wenlingz:

devicemodel/core/mptbl.c
donsheng: @lifeix  can you double check my patch? it is already checking pcpu_id

ywan170 (approved):

PR #7170, opened 2022-03-02
doc: remove build problem in release notes

dbkinder:
when an XML option was removed, it will break documentation links to
that option information.  We'll remove the link in the old release notes
to fix this problem.


acrnsi:

dbkinder:
Fixes doc build failure in PR #7158

Issue #6691, opened 2021-10-19
[DX] Configtool UI refinement


NanlinXie:
Based on DX survey, we need refine config UI.

PR #7194, opened 2022-03-17
config_tools: packaging config tool scripts for easier import

junjiemao1:
The next-generation ACRN configurator will embed a Python interpreter built
in WebAssembly (WASM) for executing Python scripts for manipulating
scenario schemas and validating user data. It is quite tedious to
separately import multiple modules there due to the restriction of that
Python environment. The recommended approach is to package those
modules (e.g. as a *.wic file) so that all modules can be imported in one
shot.

This patch adds the files needed to package the scripts. The package is
solely used for the configurator to import and not intended to be used by
end users for any purpose.

v1 -> v2:

  * Fix the license header of __init__.py
  * Move patterns of ignored files to the top-level .gitignore

Tracked-On: #6691
Reviewed-by: Geoffroy Van Cutsem <geoffroy.vancutsem@intel.com>


acrnsi:

NanlinXie (approved):

PR #7195, opened 2022-03-17
misc: refine the non-stand vUART config

weichenli-intel:
The Service VM needs a 'serial.conf' file to recognize and use
non-standard vUART with the 'setserial' package.

The logic of create serial.conf should update after we move vUART
connection to global and use endpoint to replace the target VM id.

This patch select the new vUART connection and check whether one of the
endpoint is Service VM, if yes, then check the io_port and record it
when the vUART is non-standard.

All these non-standard vUART will be written to the serial.conf.

Tracked-On: #6690
Reviewed-by: Junjie Mao junjie.mao@intel.com

### DIFF
--- a/misc/config_tools/service_vm_config/serial_config.py
+++ b/misc/config_tools/service_vm_config/serial_config.py
@@ -4,8 +4,12 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 #
+import sys, os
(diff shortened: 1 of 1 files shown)


acrnsi:

NanlinXie (approved):

Issue #6701, opened 2021-10-21
doc: remove references to Automotive SDC scenario


dbkinder:
We still have mention of the Software Defined Cockpit (SDC) scenario in our documentation that should probably be removed now.

We do still have some SDC scenarios defined for specific boards, but none in the generic templates:
```
./config_tools/data/qemu/sdc.xml
./config_tools/data/whl-ipc-i5/sdc_launch_1uos_zephyr.xml
./config_tools/data/whl-ipc-i5/sdc_launch_1uos_laag.xml
./config_tools/data/whl-ipc-i5/sdc.xml
```

Here's a grep through the .rst documentation:
```
developer-guides/hld/hld-overview.rst:Software-defined Cockpit (SDC), or the In-vehicle Experience (IVE) for
developer-guides/hld/hld-overview.rst:The SDC system consists of multiple systems: the instrument cluster (IC)
developer-guides/hld/hld-overview.rst:ACRN 1.0 is designed mainly for auto use cases such as SDC & IVI.
developer-guides/hld/hld-overview.rst:Cockpit (SDC) use case, and may require functional safety certification
developer-guides/hld/hld-security.rst:used to build a Software Defined Cockpit (SDC) or an In-Vehicle Experience
developer-guides/hld/hld-security.rst:   SDC and IVE system In-Vehicle
developer-guides/hld/hld-security.rst:In an SDC system, the Service VM also contains safety-critical IC (Instrument
tutorials/acrn_on_qemu.rst:      make BOARD=qemu SCENARIO=sdc
tutorials/acrn_on_qemu.rst:         echo 'Loading ACRN hypervisor with SDC scenario ...'
tutorials/run_kata_containers.rst:   the Industry or SDC scenarios.
tutorials/sgx_virtualization.rst:   Apply the patch to enable SGX support in User VM in the SDC scenario:
tutorials/using_zephyr_as_uos.rst:   This tutorial uses the (default) SDC scenario. If you use a different
tutorials/vuart_configuration.rst:| SDC             | Service VM            | Post-launched      | Post-launched  |                |
tutorials/vuart_configuration.rst:| SDC             | Service VM            | Post-launched      | Post-launched       |                |
```

amyreye:
@dbkinder In hld-overview.rst and hld-security.rst, is the content generic enough that we can keep it?

amyreye:
[...]

PR #7203, opened 2022-03-21
config-tools: remove WHL sdc scenario/launch file

jackwhich:
Because there is no SDC scenario in the generic template,
We remove the SDC specific scenario

Tracked-On: #6701


acrnsi:

NanlinXie (approved):

PR #7209, opened 2022-03-22
dm: update virtio-i2c bus counter when device is deinit

yonghuah:
i2c bus counter 'acpi_i2c_adapter_num' shall be updated
 when virtio-i2c device is de-initialized, else in user VM
 reboot case, this global counter will not be reset and
 keep the value last user VM boot, which is not expected
 and cause iasl(building ACPI data) program crash finally
 because the i2c device name space overflow.

Tracked-On: #7028


acrnsi:

ywan170 (approved):

PR #7219, opened 2022-03-25
Extract available PCI devices and serials from board XML

junjiemao1:
This patch adds data into board XML that enables the extraction of available PCI devices or serial ports using XPATH which are encoded in the XML schema as well. This helps the next-generation configurator to automatically pick to options from board XMLs.


acrnsi:

NanlinXie (approved):

PR #7226, opened 2022-03-30
fix potential data races in virtio_blk_abort

ycaibb:
fix potential data races in virtio_blk_abort

### DIFF
--- a/devicemodel/hw/pci/virtio/virtio_block.c
+++ b/devicemodel/hw/pci/virtio/virtio_block.c
@@ -224,10 +224,12 @@ virtio_blk_done(struct blockif_req *br, int err)
 static void
 virtio_blk_abort(struct virtio_vq_info *vq, uint16_t idx)
 {
+	pthread_mutex_lock(&blk->mtx);
 	if (idx < vq->qsize) {
 		vq_relchain(vq, idx, 1);
 		vq_endchains(vq, 0);
 	}
+	pthread_mutex_unlock(&blk->mtx);
 }
 
 static void


ycaibb:
the` vq_relchain(vq, idx, 1); and vq_endchains(vq, 0);` should be guarded by lock `blk->mtx ` in the method `virtio_blk_abort` just as in the method `virtio_blk_done`.

GGGZZZRRR:
start to run premerge test

devicemodel/hw/pci/virtio/virtio_block.c
```diff
@@ -224,10 +224,12 @@ virtio_blk_done(struct blockif_req *br, int err)
 static void
 virtio_blk_abort(struct virtio_vq_info *vq, uint16_t idx)
 {
+	pthread_mutex_lock(&blk->mtx);
```
conghuic23, line 227: Thanks for your patch, can could you please give more details on the data lose?
Besides, there is no 'blk' parameter in this function, so this could not be compiled.
This function can be invoked during virtio_blk_proc, which is run with the base->mtx (same with blk->mtx) hold, so dead-lock may meet.

ycaibb: Hi, I noticed that `vq_relchain(vq, idx, 1); and vq_endchains(vq, 0);` is run in the virtio_blk_abort at Line 280 and in the virtio_blk_done at Line 295. Conversely, they are protected by the lock in the virtio_blk_done. Thus, I infer that there is a missing lock in the virtio_blk_abort. Is my inference right.
https://github.com/projectacrn/acrn-hypervisor/blob/a65e01ae9547be78ccad26333a907743c8c8c5d7/devicemodel/hw/pci/virtio/virtio_block.c#L278-L297

Sorry for the mistake in the patch. I am revising it.

conghuic23: The lock in virtio_blk_done is not aim to protect data in this thread. In general read/write operations, the virtio_blk_done is called in 'worker' threads, so there is a case that the "virtio_blk_proc" is call in 'vcpu' thread, and the 'virtio_blk_done' is call in a 'worker' thread, for this case, the lock is needed.  'virtio_blk_abort' will only run in 'vcpu' thread, and protected by a big lock.  So, no need to add a lock for it.

ycaibb: @conghuic23 OK, Thank you very much for your clarification

PR #6177, opened 2021-06-08
config_tools: clean up the board folders

Kunhui-Li:
1. Remove apl-up2, apl-up2-n3350, whl-ipc-i5, apl-mrb, nuc6cayh board
   folders from the latest code base.
2. Copy tgl-rvp.xml to generic_board.xml.

Tracked-On: #6175

Signed-off-by: Kunhui-Li <kunhuix.li@intel.com>

PR #7237, opened 2022-03-31
doc: Remove tgl-rvp mention

amyreye:
- Remove mention of tgl-rvp XML because the file was removed from repo

Tracked-On: #6177


dbkinder (approved):
LGTM

PR #7238, opened 2022-04-01
Vgpu dmabuf

yakuizhao:
Now Virtio-gpu FE is using the below cmds to submit the framebuffer that
needs to be displayed.
   VIRTIO_GPU_CMD_TRANSFER_TO_HOST_2D
   VIRTIO_GPU_CMD_SET_SCANOUT
   VIRTIO_GPU_CMD_FLUSH

This will introduce twice copy. It will affect the CPU usage, memory resource.

In order to optimize the twice copy, this patch set tries to add the support
of zero-copy based on dmabuf-sharing.
The key of zero-copy depends on the DMABuf and eglCreateImageKHR from dmabuf.
In order to create dmabuf based on the GPA addr of virtio-gpu framebuf, the memory
allocation for guest_vm should be based on memfd. This requires that the memory
for guest_vm should be allocated from memfd instead of explicit hugetlb file.


acrnsi:

ywan170 (approved):

GGGZZZRRR:
start to run premerge test

Issue #7210, opened 2022-03-22
Virtio-gpu adapter support in ACRN


psun3x:
Implement virtio-gpu basic 2D display in fixed resolution with hardware cursor support. The framebuffer from guest vm can be shown in a window of service vm GUI system(X11 or Wayland). Virtio-gpu device provide to 2 interface for displaying operation, VGA interface and virtio-gpu command interface.

psun3x:
[External_System_ID] ACRN-7626

NanlinXie:
Doc Impact Analysis: 

- [x] New feature -> new HLD and usage guide
- [ ] [HLDs](https://projectacrn.github.io/2.3/developer-guides/hld/index.html)
- [x] [Configuration data](https://projectacrn.github.io/2.3/tutorials/acrn_configuration_tool.html)
- [ ] [Kernel module parameters](https://projectacrn.github.io/2.3/user-guides/kernel-parameters.html)
- [ ] [Device Model (acrn-dm) parameters](https://projectacrn.github.io/2.3/user-guides/acrn-dm-parameters.html)
- [ ] [Hypervisor shell commands](https://projectacrn.github.io/2.3/user-guides/acrn-shell.html)
- [ ] [ACRN tools (debug and/or service)](https://projectacrn.github.io/2.3/develop.html#configuration-and-tools)
- [ ] [Build system](https://projectacrn.github.io/latest/getting-started/building-from-source.html)
- [x] [Other documents (e.g. tutorials)](https://projectacrn.github.io/2.3/develop.html#enable-acrn-features)

PR #7240, opened 2022-04-01
doc: virtio-gpu parameters of acrn-dm

psun3x:
Update parameters of virtio-gpu in acrn-dm-parameters.rst

Tracked-On: #7210
Signed-off-by: Sun Peng <peng.p.sun@linux.intel.com>

### DIFF
--- a/doc/user-guides/acrn-dm-parameters.rst
+++ b/doc/user-guides/acrn-dm-parameters.rst
@@ -474,6 +474,17 @@ arguments used for configuration.  Here is a table describing these emulated dev
           latter is ignored and the MAC address is set to the ``mac`` value.
           ``mac_seed`` will only be used when ``mac`` is not set.
 
+   * - ``virtio-gpu``
+     - Virtio GPU type device, parameters format is:
+       ``virtio-gpu[,geometry=<width>x<height>+<x_off>+<y_off> | fullscreen]``
+
+       * ``<geometry>`` specifies the mode of virtual display, windowed or fullscreen. 
+         If it is not set, the virtual display will use 1280x720 resolution in windowed mode.
+       * ``width`` specifies the width of the virtual display window in pixels.
+       * ``height`` specifies the height of the virtual display window in pixels.
+       * ``x_off`` specifies the x offfset of the virtual display window from the upper-left corner of screen.
+       * ``y_off`` specifies the y offfset of the virtual display window from the upper-left corner of screen.
+
    * - ``passthru``
      - Indicates a passthrough device. Use the parameter with the format
        ``passthru,<bus>/<device>/<function>,<optional parameter>``


acrnsi:

doc/user-guides/acrn-dm-parameters.rst
```diff
@@ -474,6 +474,17 @@ arguments used for configuration.  Here is a table describing these emulated dev
...
+   * - ``virtio-gpu``
+     - Virtio GPU type device, parameters format is:
+       ``virtio-gpu,<geometry={widthxheight+x_off+y_off|fullscreen:screen_id},geometry={...}...]>``
+
+       * ``<geometry>`` specifies the mode of virtual display, windowed or fullscreen
+       * ``width`` specifies the width of the virtual display window.
+       * ``height`` specifies the height of the virtual display window.
+       * ``x_off`` specifies the x offfset of the virtual display window from origin point of the screen.
+       * ``y_off`` specifies the y offfset of the virtual display window from origin point of the screen.
+       * ``screen_id`` specifies the screen id when show virtual display as fullscreen.
```
gvancuts, line 486: Is this `screen_id` the physical screen ID that will be used to display the virtual display as fullscreen?

How do you determine what the screen ID is? Is that related to the `DISPLAY` variable in any way?

doc/user-guides/acrn-dm-parameters.rst
```diff
@@ -474,6 +474,17 @@ arguments used for configuration.  Here is a table describing these emulated dev
           latter is ignored and the MAC address is set to the ``mac`` value.
           ``mac_seed`` will only be used when ``mac`` is not set.
 
+   * - ``virtio-gpu``
+     - Virtio GPU type device, parameters format is:
+       ``virtio-gpu,<geometry={widthxheight+x_off+y_off|fullscreen:screen_id},geometry={...}...]>``
```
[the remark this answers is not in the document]
psun3x: Yes, if "geometry" repeated, we will create multi-virtual display, just like qemu does. But for now, we only support 1 virtual display.

doc/user-guides/acrn-dm-parameters.rst
```diff
@@ -474,6 +474,17 @@ arguments used for configuration.  Here is a table describing these emulated dev
...
+
+       * ``<geometry>`` specifies the mode of virtual display, windowed or fullscreen
```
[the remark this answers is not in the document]
psun3x: geomerty is a optional parameter. If it is not set, the virtual display will use 1280x720 resolution and show with windowed mode.

doc/user-guides/acrn-dm-parameters.rst
psun3x: Yes, you are correct.

doc/user-guides/acrn-dm-parameters.rst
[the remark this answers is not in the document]
gvancuts: Is there a meaningful error message that is displayed if or when the user tries to create multiple virtual displays?

doc/user-guides/acrn-dm-parameters.rst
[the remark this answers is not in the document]
gvancuts: Thanks for the info, this is something I would add in a note as well. @dbkinder , thoughts?

doc/user-guides/acrn-dm-parameters.rst
gvancuts: Wouldn't it be better in this case to refer to call this `display_id` and talk about a "display" in our doc, so as to be consistent with the standard way of naming things?

gvancuts (dismissed):
See comments left a minute ago ;-)

doc/user-guides/acrn-dm-parameters.rst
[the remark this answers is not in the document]
amyreye: A few comments regarding line 479: 
1. If ACRN doesn't support multiple geometry parameters, I think we should remove the second one. 
2. It looks like line 479 has an extra closing bracket "]" at the end. 
3. If geometry is optional, put the whole param in brackets, such as: ``virtio-gpu[,geometry={<width>x<height>+<x_off>+<y_off> | fullscreen:<screen_id>}]``

doc/user-guides/acrn-dm-parameters.rst
amyreye: Please add more information to the documentation:
- For width and height, what is the unit of measurement?
- For x_off and y_off, what is the "origin point of the screen"? Will users know what it is?
- For screen_id, what is the ``DISPLAY`` variable, and where is it defined? cc: @gvancuts

gvancuts: > Please add more information to the documentation:
> 
>     * For width and height, what is the unit of measurement?

Pixels

> 
>     * For x_off and y_off, what is the "origin point of the screen"? Will users know what it is?
> 
>     * For screen_id, what is the `DISPLAY` variable, and where is it defined? cc: @gvancuts

It's an environment variable, in a graphical interface, you can run `echo $DISPLAY` to see what it's set to (typically `:0`).

@psun3x , what is the format for the `screen_id` parameter?

doc/user-guides/acrn-dm-parameters.rst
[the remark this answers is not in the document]
psun3x: If user tries to create multiple virtual display, we only parse one virtual display and no error message now.

doc/user-guides/acrn-dm-parameters.rst
psun3x: Just 0, 1, 2, ...

gvancuts: > Just 0, 1, 2, ...

OK, we need a couple of things here:
* How do you determine which number maps to what monitor?
* Clearly indicate the format in the description (to avoid people like who may be tempted to try `:0` :-)

doc/user-guides/acrn-dm-parameters.rst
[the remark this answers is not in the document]
gvancuts: > If user tries to create multiple virtual display, we only parse one virtual display and no error message now.

As @amyreye recommended, let's remove the second one then. Ideally we'd also print out a warning in case the user is trying to add a second virtual display (which he would likely not do if nothing in our documentation indicates he could ;-))

doc/user-guides/acrn-dm-parameters.rst
psun3x: We will use "monitor_id" instead of "screen_id". Till now we only support 1 monitor for fullscreen mode. We don't support multi-monitor and mapping yet.

doc/user-guides/acrn-dm-parameters.rst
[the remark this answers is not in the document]
amyreye: @psun3x I think we should add the information to line 481. For example:

``<geometry>`` specifies the mode of the virtual display, windowed or fullscreen. If it is not set, the virtual display will use 1280x720 resolution in windowed mode.

doc/user-guides/acrn-dm-parameters.rst
amyreye: @psun3x At line 486, I think we are still missing information about how to determine the monitor_id value. Please add it to line 486, for example:

``monitor_id`` specifies the ID of the screen shown on the fullscreen virtual display, for example, 0. This value must match the value from <what?>.

@gvancuts @dbkinder Can you help with the wording above?

doc/user-guides/acrn-dm-parameters.rst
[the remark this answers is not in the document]
psun3x: OK, will add this in next version.

doc/user-guides/acrn-dm-parameters.rst
psun3x: The monitor_id is dynamic, not fixed because if a monitor plug-out. Till now, we only support 1 monitor, so, the current design don't have the algorithm how to generate monitor_id for multi-monitors. This format is for further design.

amyreye: If the ``fullscreen:<monitor_id>`` option isn't functional yet, I recommend removing it from this document. If the option does function, please add the valid value to the document. Users need to know what value(s) they should input.

doc/user-guides/acrn-dm-parameters.rst
```diff
@@ -474,6 +474,17 @@ arguments used for configuration.  Here is a table describing these emulated dev
...
+   * - ``virtio-gpu``
+     - Virtio GPU type device, parameters format is:
+       ``virtio-gpu[,geometry=<width>x<height>+<x_off>+<y_off> | fullscreen:<monitor_id>]``
+
+       * ``<geometry>`` specifies the mode of virtual display, windowed or fullscreen
+       * ``width`` specifies the width of the virtual display window in pixels.
+       * ``height`` specifies the height of the virtual display window in pixels.
+       * ``x_off`` specifies the x offfset of the virtual display window from the upper-left corner of screen.
+       * ``y_off`` specifies the y offfset of the virtual display window from the upper-left corner of screen.
+       * ``monitor_id`` specifies the screen id when show virtual display as fullscreen, eg.0,1,2....
```
dbkinder, line 486:
```suggestion
       * ``monitor_id`` specifies the screen id when the virtual display is fullscreen, e.g., 0,1,2.
       
       When used, these width and height values are separated by an ``x``,
       and the x_off and y_off values are separated by a ``+``. For example:
       ``geometry=1280x720+100+50`` specifies a window 1280 pixels wide
       by 720 high, and 100 pixels right and 50 pixels down from the top left
       corner of the screen.
```
Can the user specify ``0xFFFF`` (or 65535) for the two offset values to specify  the "don't care" SDL_WINDOWPOS_UNDEFINED value, and can we document that?

dbkinder (changes requested):

doc/user-guides/acrn-dm-parameters.rst
ywan170: As my understanding, we

- Only support single virtual display per VM.
- Virtio-GPU support window mode and full screen mode
- For full screen mode, user has to indicate which physical monitor will be used.

So the fullscreen:<monitor_id> is functioanl already. Please correct me if my understanding is incorrect.

BTW, there's another question, if the monitor_id is only specific for full screen mode, then which monitor will be used for window mode?

ywan170: I just talked with Peng. Looks like we only support a single *physical* monitor. Whatever the monitor_id is, the window/fullscreen will always show in the primary monitor. Then let's remove the monitor_id first. And add it back once we support it.

amyreye: @ywan170 @psun3x Please update the doc per the above comment "Then let's remove the monitor_id first" (to ensure accuracy). See also dbkinder's comment below.

NanlinXie (approved):

NanlinXie:
@dbkinder  and @gvancuts , any more comments for this PR?

dbkinder (approved):
Close enough.

PR #7243, opened 2022-04-01
doc: fix exception error in last_modified processing

dbkinder:
When I added a new document that wasn't checked into GitHub yet, the doc
build failed with an unhanded exception:

    Extension error (last_updated):
    (exception: time data '' does not match format '%Y-%m-%d')

Problem is the git query looking up the last commit date for a file
returns an empty string for the date if the file exists but it's not in
the git repo (yet). The subsequent call to strptime raises an exception if passed
an empty string.  This patch handles the exception.


acrnsi:

dbkinder:
@NanlinXie @wenlingz  All changes within the /doc folder were considered "doc only" changes and don't require a Tracked-On tag.  Is this not true any more with the recent CI changes?

PR #7244, opened 2022-04-01
doc: fix Makefile to report errors if exception raised

dbkinder:
We use a post-processor to scan the output log from Sphinx to remove
known errors and warnings before deciding if the Sphinx build was
successful.  If an exception happened during the Sphinx build, (see
about the error would appear.  Update the Makefile to not stop on errors
for the Sphinx build. The post-processor will stop the make by returning
an error code if it finds unexpected errors in the log.


acrnsi:

Issue #7171, opened 2022-03-02
Link to instead of including old release notes in the current release


dbkinder:
We include the release notes for all releases when generating documentation (see https://projectacrn.github.io/latest/release_notes/index.html).  This can cause a problem with building documentation if these old release notes have links to content that was removed or renamed after that release was made.  

For example:
```
New errors/warnings found, please fix them:
/home/acrn/workspace/acrn-hypervisor-build-doc_141/acrn-hypervisor/build/doc/rst/release_notes/release_notes_2.5.rst:112: WARNING: unknown option: hv.FEATURES.NVMX_ENABLED
/home/acrn/workspace/acrn-hypervisor-build-doc_141/acrn-hypervisor/build/doc/rst/release_notes/release_notes_2.5.rst:121: WARNING: unknown option: vm.guest_flags.guest_flag
make[1]: *** [Makefile:64: html] Error 1
```
Better might be to link to each these old release notes in their archived release areas rather than regenerating them for each release, e.g., /2.7/release_notes/release_notes_2.7.rst, and opening them in a new tab.

dbkinder:
[External_System_ID] ACRN-6954

PR #7250, opened 2022-04-02
doc: reference old release notes in archived release docs

dbkinder:
Instead of trying to maintain the release notes for old releases in the
current release, reference them in a previous archive.  This eliminates
the need to fix doc build errors by editing old release notes that
reference material that no longer exists in the newer release.

We use the intersphinx extension to support references to
Sphinx-generated content from other projects.  In our case, we're
referencing the archived version of v2.7 documents that was the last
release to generate release notes for all previous releases. This import
is done in conf.py.

(Alternatively we could reference the release notes for a release in
that specific release's archive, but that means fetchind the intersphinx
inventory file (objects.inv) for every release, and seems like
overkill.)

This modification still allows putting multiple release notes in the
doc/release_notes folder and they'll show up in the release notes index
as they have in the past.  Update the separate "archived" release notes
section manually as needed.

Also, adds a draft of the v3.0 release notes to verify the new release
notes index page layout.

Fixes: #7171


acrnsi:

dbkinder:
@NanlinXie @wenlingz Sigh.  Looks like changes to the Sphinx configuration file (doc/conf.py) is being treated as a non doc-only change too now. :(

NanlinXie (approved):

dbkinder:
start to run premerge test

dbkinder:
start to run premerge test

Issue #7249, opened 2022-04-02
doc: Exception in Sphinx processing doesn't display error message


dbkinder:
When I added a new document that wasn't checked into GitHub yet, the doc
build failed with an unhanded exception:
```
Extension error (last_updated):
(exception: time data '' does not match format '%Y-%m-%d')
```
Problem is the git query looking up the last commit date for a file
returns an empty string for the date if the file exists but it's not in
the git repo (yet). The subsequent call to strptime raises an exception if passed
an empty string. 

Also need to update the doc Makefile to is doesn't stop the sphinx-build if an error happens.  The subsequent known-issues filter will stop the make if an unexpected error was detected in the sphinx output.

dbkinder:
[External_System_ID] ACRN-6954

PR #7251, opened 2022-04-03
doc: handle last updated footer for new document

dbkinder:
If a new document is added, the CI system fails because the new document
hasn't been checked into the branch yet so the last_updated.py script
complains with an error.  Change to use the last_published date in this
situation.

Tracked-On: #7249


acrnsi:

PR #7253, opened 2022-04-04
doc: properly handle last_updated info for new docs

dbkinder:
Update the last_updated extension to correctly handle when new docs are
added and aren't yet commited into git vs. when the list of git folders
isn't correct. Previous PR #7251 incorrectly handled these as the same.

Tracked-On: #7249


acrnsi:

amyreye (approved):

Issue #7267, opened 2022-04-11
service vm cpu affinity issue


liuhang-bit:
when leave pcpu_id blank, maxcpus in service vm kernel command line is not right.

PR #7268, opened 2022-04-11
config tool: Fix service vm cpu affinity issue

liuhang-bit:
default value of minOccurs is "1", which will trigger problem
when user leave pcpu_id blank while preserve cpu_affinity in
service VM.

Tracked-On: #7267


acrnsi:

Issue #7275, opened 2022-04-12
[icx-d][shared] CPK SRIOV VF number can't greater than three


lifeix:
[External_System_ID] ACRN-7652

PR #7276, opened 2022-04-12
hv: ptdev: address vector scalability problem by interrupt posting

lifeix:
Now interrupt vector in ACRN hypervisor is maintained as global variable, not
per-CPU variable. If there're more PCI devices, the physical interrupt vectors
are not enough most likely.

This patch would not allocate physical interrupt vector for MSI/MSI-X vectors
if interrupt posting could been used to inject the MSI/MSI-X interrupt to
a VM directly.

Tracked-On: #7275

### DIFF
--- a/hypervisor/common/ptdev.c
+++ b/hypervisor/common/ptdev.c
@@ -193,16 +193,22 @@ static void ptirq_interrupt_handler(__unused uint32_t irq, void *data)
 /* active intr with irq registering */
 int32_t ptirq_activate_entry(struct ptirq_remapping_info *entry, uint32_t phys_irq)
 {
-	int32_t retval;
+	int32_t ret = 0;
+	uint32_t irq = IRQ_INVALID;
 	uint64_t key;
 
-	/* register and allocate host vector/irq */
-	retval = request_irq(phys_irq, ptirq_interrupt_handler, (void *)entry, IRQF_PT);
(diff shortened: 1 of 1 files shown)


lifeix:
@dongyaozu  Hi Eddie, would you please help to review ?

hypervisor/common/ptdev.c
```diff
@@ -259,7 +265,7 @@ uint32_t ptirq_get_intr_data(const struct acrn_vm *target_vm, uint64_t *buffer,
 
 	for (i = 0U; i < CONFIG_MAX_PT_IRQ_ENTRIES; i++) {
 		entry = &ptirq_entries[i];
-		if (!is_entry_active(entry)) {
+		if (!is_entry_active(entry) || (entry->allocated_pirq >= NR_IRQS)) {
```
dongyaozu, line 268: should be  != IRQ_INVALID? need to match with the above code, and don;t assume IRQ_INVALID is larger is larger or smaller than NR_IRQs

lifeix: > should be != IRQ_INVALID? need to match with the above code, and don;t assume IRQ_INVALID is larger is larger or smaller than NR_IRQs

yes, fixed.

dongyaozu (approved):

PR #7283, opened 2022-04-14
doc: Add note to GSG about optional upgrade

amyreye:
- Add comment in upgrade command in case users ignore description above command


acrnsi:

dbkinder (approved):

PR #7285, opened 2022-04-14
doc: Fix links

amyreye:
- Fix broken software.intel.com links that moved to intel.com domain


acrnsi:

dbkinder (approved):
Thanks. There were supposed to be redirects on inte.com to catch these when content there was moved.  :(

PR #7290, opened 2022-04-15
hv: support CAT on hybrid platform

tw4452852:
On hybrid platform(e.g. ADL), there may be multiple instances of same level caches for different type of processors,
The current design only supports one global `rdt_info` for each RDT resource type.
In order to support hybrid platform, this patch introduce `rdt_ins` to represents the "instance".
Also, the number of `rdt_info` is dynamically generated by config-tool to match with physical board.

Tracked-On: projectacrn#6690
Acked-by: Eddie Dong <eddie.dong@Intel.com>


acrnsi:

lifeix (approved):

PR #7292, opened 2022-04-17
Solved the “libSDL2-2.0.so.0” dependency

zihengL1:
Added the "libsdl2-dev" Pre-dependency in file "misc/packaging/gen_acrn_deb.py"


acrnsi:

Issue #7298, opened 2022-04-18
boot time issue for acrn-dm


conghuic23:
When Post-launched VM's memory is setup to 9G, the boot time from launching acrn-dm to the Guest VM initiate done  more then 2s. The time is too long, and it is because the Guest memory is erased twice during initiate, need to remove one erasing.

conghuic23:
[External_System_ID] ACRN-8800

conghuic23:
[External_System_ID] ACRN-8735

NanlinXie:
Closed with no regression issue.

PR #7300, opened 2022-04-18
remove repetitive erasing for guest memory

conghuic23:
The guest memory is allocated through hugetlb. Huge page fault is
triggered by writing a byte to the starting address for each huge page.
In hugetlb_fault(), The physical pages are allocated and *clear*.

So no need to erasing the memory again after hugetlb setup. This will
bring about 130ms (changes with platform) latency for each 1G memory.

Notice: this means we depends on kernel to erase the memory in huge page
fault!

Tracked-On: #7298


acrnsi:

ywan170 (approved):

Issue #7212, opened 2022-03-23
Exception"not enough space in guest VE820 SSRAM area" showed when built ACRN with RTCT table


izhouwu:

izhouwu:
[External_System_ID] ACRN-7629

izhouwu:
SSRAM area in GPA has been hard-coded to 8MB. Changing it to actual area size could fix the issue.

NanlinXie:
Closed with no regression.

PR #7303, opened 2022-04-18
refine the hard-coded GPA SSRAM area size

izhouwu:
The GPA SSRAM area size in pre-launched VMs was hard-coded to 8MB.
Since this area is mapped from host SSRAM area, it will cause compile
problem when host's SSRAM area is larger than 8MB.

patch1 extracts the host's SSRAM area info into board file, and uses it
to generate guest's SSRAM area config that fits its actual size.

patch2 refines the hard-coded SSRAM size in HV.

v2:
  1. Since SSRAM info is in each cache entry, sssram area info is not
necessary to be extracted in board_inspector. Now it only extracts ctl
bin
info. The SSRAM area info is calculated in gpa.py.
  2. Pass PRE_RTVM_SW_SRAM_MAX_SIZE to HV, remove START/END.
  3. Fixed a bug that Service VM removes GPA of SSRAM from ept(should be
HPA).
  4. To align with config_tools and ACPI RTCT table, Host SSRAM area
size
in HV now contains crl bin.
  5. Refined the SSRAM mem layout info in gpa.py.

v3:
  1. Use service_vm_hpa2gpa() to convert HPA to GPA when deleting SSRAM
area in Service VM.
  2. When no SSRAM is not configured, PRE_RTVM_SW_SRAM_MAX_SIZE is set
to 0 by config tool. So that HV code can be simpler.
  3. Refined the commit messages.

v4:
  Crl_bin is not needed by guest. Removed it from SSRAM area size, so that
it won't be passed-through.

    Tracked-On: #7212


acrnsi:

NanlinXie (approved):
Approved for configtool part

lifeix (approved):

PR #7306, opened 2022-04-19
doc: fix table CSS font size changing with lists

dbkinder:
The read-the-docs theme uses a reduced font size within tables.  The CSS
used though is overly restrictive and doesn't properly handle tables
containing lists or other constructs where the paragraph tags don't have
the <td> tag within the table as the immediate parent.  Add an
overriding style in our custom CSS to fix this so the font size on lists
within a table are the same as normal paragraphs within a table.


acrnsi:

amyreye (approved):

Issue #7307, opened 2022-04-19
add support for bootargs missing


liuhang-bit:
[External_System_ID] ACRN-8789

fuzhongl:
Verified with following commit:

```
commit 7bcc5b73831883344cece4599a7d634d3bc20cf9
Author: hangliu1 <hang1.liu@linux.intel.com>
Date:   Mon Apr 18 23:52:48 2022 -0400

    config tool:add support for bootargs missing

    Bugfix:bootargs is optional parameter, which may be not
    present in xml as a node, asl_gen.py needs to process
    that case.

    Tracked-On: #7307
    Signed-off-by: hangliu1 <hang1.liu@linux.intel.com>
```

PR #7308, opened 2022-04-19
config tool:add support for bootargs missing

liuhang-bit:
Bugfix:bootargs is optional parameter, which may be not
present in xml as a node, asl_gen.py needs to process
that case.

Tracked-On: #7307


acrnsi:

NanlinXie (approved):

PR #7310, opened 2022-04-19
Acrn vgpu sigsegv

yakuizhao:
ACRN-DM will display the framebuffer that is submitted from the virtio-gpu
in guest and this is based on 3D driver. But as the 3D driver is not thread-
safe, it will be better that the 3D ops related with virito-gpu needs to
be called in dedicated-thread env. Otherwise it will trigger the sigsegv
in thread_exiting.
This is the patch set that tries to fix the sigsegv issue.


acrnsi:

ywan170 (approved):

PR #7312, opened 2022-04-19
doc: add a tutorial on how to upgrade configuration files

junjiemao1:


acrnsi:

doc/tutorials/upgrading_configuration.rst
```diff
@@ -0,0 +1,70 @@
...
+definitions of VMs. Starting from v3.0, the highly recommended way to upgrade a
+scenario XML is to use the scenario XML upgrader which can be found at
+`misc/config_tools/scenario_config/upgrader.py`. Invoke this script from the
+command line in the following way:
+
+   .. code-block:: bash
+
+      misc/config_tools/scenario_config/upgrader.py <your scenario XML> <out file>
+
+`<your scenario XML>` is the path to the scenario XML file for a previous
```
gvancuts, line 36: Can you use double backticks (for ReST)?

doc/tutorials/upgrading_configuration.rst
```diff
@@ -0,0 +1,70 @@
...
+release. This script will generate a new scenario XML file, located at `<out
+file>`, by migrating, following the latest schema, as many data in `<your
```
gvancuts, line 38: Can this be a filename *and* a path (relative or absolute)?

doc/tutorials/upgrading_configuration.rst
```diff
@@ -0,0 +1,70 @@
...
+   The upgrader's best efforts at migrating data from scenario XMLs of former
+   releases do not guarantee that all data can be migrated. Thus, any warning
+   from the tool shall be carefully reviewed. In case any meaningful data is
+   lost, use the ACRN configurator to open the upgraded scenario XML for further
+   edits.
+
+Launch XML
+**********
+
+Starting from ACRN v3.0, data in a launch XML are merged into the
```
gvancuts, line 61: This will read a little odd until we have a v3.0 release out.

amyreye (approved):
I don't have any questions. The doc needs copyediting, but it can be done in a separate PR.

doc/tutorials/upgrading_configuration.rst
```diff
@@ -0,0 +1,70 @@
+.. _upgrading_configuration:
+
+Upgrading ACRN Configurations to Recent Releases
+################################################
+
+The configurations files, introduced in :ref:`acrn_config_data`, are refined
+every release for richer features, clearer organizations and more user friendly
```
dbkinder, line 7:
```suggestion
every release for richer features, clearer organizations, and more user friendly
```

doc/tutorials/upgrading_configuration.rst
```diff
@@ -0,0 +1,70 @@
...
+for a former release may not work for a latter if they are not upgraded. This
+tutorial introduces the steps to upgrade those files.
+
+Board XML
+*********
+
+A board XML file lists the characteristics of a board and is generated by the
+board inspector. Each release of ACRN adds some more information into that file
+in order to enable new features. Thus, always regenerate the board XML using the
+board inspector of the new release when you upgrade.
```
dbkinder, line 18: Sure would be nice if the board inspector wrote the ACRN release version in the board XML file so we could detect a potential problem early (using an old board XML).

doc/tutorials/upgrading_configuration.rst
```diff
@@ -0,0 +1,70 @@
...
+
+For the steps to use the latest board inspector, refer to
+:ref:`board_inspector_tool`.
+
+Scenario XML
+************
+
+A scenario XML file captures the configurations of hypervisor features and
+definitions of VMs. Starting from v3.0, the highly recommended way to upgrade a
+scenario XML is to use the scenario XML upgrader which can be found at
```
dbkinder, line 28:
```suggestion
scenario XML is to use the scenario XML upgrader found at
```

dbkinder (changes requested):
Thanks @junjiemao1

doc/tutorials/upgrading_configuration.rst
junjiemao1: @NanlinXie @gvancuts Do we want this for v3.0 or v3.1?

doc/tutorials/upgrading_configuration.rst
junjiemao1: It can be an arbitrary path (relative to CWD or absolute) which will be interpreted as the path of the generated file, assuming the parent directory of that path exists.

doc/tutorials/upgrading_configuration.rst
```diff
@@ -0,0 +1,70 @@
...
+
+`<your scenario XML>` is the path to the scenario XML file for a previous
+release. This script will generate a new scenario XML file, located at `<out
+file>`, by migrating, following the latest schema, as many data in `<your
+scenario XML>` as possible. If there is any data that cannot be migrated, the
+script will generate messages like this:
+
+   .. code-block:: bash
+
+      WARNING:root:hv/DEBUG_OPTIONS/some = '1' is discarded
```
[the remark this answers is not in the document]
junjiemao1: The tricky part is that those paths refer to scenario XML nodes of previous releases which do not have those DX-friendly names.

For config items we drop by intention, the upgrader does recognize them and show why it is discarded. In such case the messages are using the INFO loglevel, not warning. While warnings are not expected, there is no guarantee that the upgrader can work for an arbitrary version of scenario XML.

The `root` is automatically generated by the Python logging module (because it is the "root" logger). We can tweak it later in the upgrader itself.

doc/tutorials/upgrading_configuration.rst
```diff
@@ -0,0 +1,70 @@
...
+      INFO:root:vm[1]/os_config/name = 'OS Name': Guest OS names are no longer needed in scenario definitions.
+
+A warning message indicates that some data are discarded unintendedly, while an
```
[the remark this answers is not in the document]
junjiemao1: There is still some differences. As the log level indicates, intended data discarding is, as the name suggests, "intended" and users can safely ignore them if they are not interested. Meanwhile, unintended data discarding may indicate the loss of some key configuration information, and need careful review from users.

doc/tutorials/upgrading_configuration.rst
```diff
@@ -0,0 +1,70 @@
...
+info message indicates an intended drop of obsoleted data.
+
+.. note::
+
+   The upgrader's best efforts at migrating data from scenario XMLs of former
+   releases do not guarantee that all data can be migrated. Thus, any warning
+   from the tool shall be carefully reviewed. In case any meaningful data is
+   lost, use the ACRN configurator to open the upgraded scenario XML for further
+   edits.
+
```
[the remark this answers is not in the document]
junjiemao1: Today I tried it only with v2.7, and the plan is to extend the support back to v2.5. But this doesn't mean it works for a scenario XML from an arbitrary commit between the releases.

This is why we have those warnings on "unintended data discarding": such warnings are guaranteed to report any data that the upgrader doesn't know how to migrate.

NanlinXie (approved):

doc/tutorials/upgrading_configuration.rst
junjiemao1: Updated. Thanks for pointing out.

doc/tutorials/upgrading_configuration.rst
gvancuts: OK then :-)

doc/tutorials/upgrading_configuration.rst
dbkinder: In any case, that's not a problem with this doc PR :)

dbkinder (approved):

PR #7317, opened 2022-04-19
doc: fix doc formatting in xsd files

dbkinder:
Documentation continuation lines must not be indented unless required by
rst syntax.

Added some type explations for enumerations, but would be better to
generate the enumeration values by scanning the tree. Figure that out in
another PR.  (This would prevent the documented list of values from
being out of sync with the actual list of values in the enumeration.)

Tracked-On: #5692


acrnsi:

amyreye (approved):

NanlinXie (approved):

Issue #7327, opened 2022-04-20
refine pgentry_present field in struct pgtable


yonghuah:
Page table entry present check is page table type
  specific and static, e.g. just need to check bit0
  of page entry for entries of MMU page table and
  bit2~bit0 for EPT page table case. hence no need to
  check it by callback function every time.

NanlinXie:
Closed with no regression issue.

PR #7328, opened 2022-04-20
hv: remove pgentry_present field in struct pgtable

yonghuah:
Page table entry present check is page table type
  specific and static, e.g. just need to check bit0
  of page entry for entries of MMU page table and
  bit2~bit0 for EPT page table case. hence no need to
  check it by callback function every time.

  This patch remove 'pgentry_present' callback field and
  add a new bitmask field for this page entry present check.
  It can get better performance especially when this
  check is executed frequently.

Tracked-On: #7327


acrnsi:

lifeix (approved):

PR #7334, opened 2022-04-21
board_inspector: parse cpu IDs in decimal rather than hexadecimal

junjiemao1:
Unlike other CPU IDs, the cpu_id field under //processors//thread in the
board XML is in decimal as it is assigned by the kernel continuously,
starting from 0. However, the cache info extractor today parses them as
hexadecimal numbers, leading to incomplete cache info when more than 10
processors are present on the board. This patch fixes this issue.

Tracked-On: #6689


acrnsi:

NanlinXie (approved):

PR #7338, opened 2022-04-21
Fix the memory leak in virtio-gpu

yakuizhao:
The PR tries to fix the below issues in  the virtio-gpu.
   a. incorrect access after free. virtio-gpu already frees the
allocated memory in course of reboot cycle. But the pointer is
not cleared to NULL. This will cause the incorrect access after free.

   b. the memory_leak after enabling GPU zero-copy for
the virtio-gpu.


acrnsi:

ywan170 (approved):

PR #7347, opened 2022-04-24
config_tools: list dependencies on Python libraries in requirements.txt

junjiemao1:
This patch adds a requirements.txt file that lists the libraries and their
versions the ACRN configuration toolset needs. The versions of elementpath
and xmlschema are fixed to the latest one supporting Python 3.6 due to the
requirement of the customized overlay.

Such requirements apply to both the board inspector and the XML
manipulation part of the configurator.

Tracked-On: #6690

### DIFF
--- a/misc/config_tools/requirements.txt
+++ b/misc/config_tools/requirements.txt
@@ -1,0 +1,4 @@
+defusedxml
+lxml
+elementpath == 2.4.0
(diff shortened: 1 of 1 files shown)


acrnsi:

NanlinXie (approved):

Issue #7351, opened 2022-04-24
Fix tap file name filter error


liuhang-bit:
[External_System_ID] ACRN-8787

PR #7352, opened 2022-04-24
config tool:Fix tap name filter error

liuhang-bit:
when tap name is number, it will lead to failure of
adding tap dev, fix it.

Tracked-On: #7351


acrnsi:

NanlinXie (approved):

PR #7353, opened 2022-04-24
config_tools: update virtio console to support multiple console sockets

Kunhui-Li:
1. updates virtio console in schema and launch script generation logic to support multiple console sockets.
2. update upgrader.py script for virtio devices.
    1) add virtio gpu logic in the upgrader.py script.
    2) fix the upgraded virtio elements issue.
    3) add the logic to remove the element with an empty value in the XML file to use the default value.


acrnsi:

NanlinXie (approved):

PR #7354, opened 2022-04-25
HV: vPCI: fix two monir bugs

lifeix:
1. MSI-X vectors may been changed by system software or driver software on runtime.
So we should pass htrough Message Control Register to guest.
2. Service may use VFs, so ACRN hypervisor should map VF BARs for PF to let PF
could initialize VF by firmware or PF's driver.

Tracked-On: #7275

### DIFF
--- a/hypervisor/dm/vpci/ivshmem.c
+++ b/hypervisor/dm/vpci/ivshmem.c
@@ -221,7 +221,7 @@ static int32_t ivshmem_mmio_handler(struct io_request *io_req, void *data)
 	return 0;
 }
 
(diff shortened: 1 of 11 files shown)


acrnsi:

GGGZZZRRR:
@lifeix please check this error 
All warnings:
/home/acrn/workspace/acrn-hypervisor-codescan-native_1014/acrn-hypervisor/hypervisor/dm/vpci/vmsix.c:45:23: warning: array 'msix_pt_mask' is not fully defined: the array has size 12 but only 4 initializers [acrn-c-di-07]
        static const uint8_t msix_pt_mask[12U] = {

dongyaozu (approved):

PR #7359, opened 2022-04-26
hv: remove obsolete declarations related to RDT

tw4452852:
Since CAT support for hybrid platform is landed, let's remove some old declarations
which are no longer used.

Tracked-On: #6690


acrnsi:

lifeix (approved):

Issue #7361, opened 2022-04-26
config_tool: Add check for RTVM pCPU assignment


yfliuuu:
For pre-launched RTVM, config_tool will check its pCPU assignment and make sure they don't conflict with other VMs. We need to add this check for post-launched RTVM too.

yfliuuu:
[External_System_ID] ACRN-8833

Pirouf:
What is the rational including post-launched RTVM CPU affinity to be fixed at buildtime?
Why wouldn't an system admin use SOS runtime launch-script to define which potential 4 or 6 or 8 RT-capable CPU core the RTVM will be assigned to?

junjiemao1:
> What is the rational including post-launched RTVM CPU affinity to be fixed at buildtime? Why wouldn't an system admin use SOS runtime launch-script to define which potential 4 or 6 or 8 RT-capable CPU core the RTVM will be assigned to?

Physical CPU assignment is a pre-requisite of a proper programing to the physical CAT which is statically defined at configuration phase. This is because there are multiple CAT-capable L2 caches on platforms with hybrid cores, and knowing the physical CPU assignment is crucial in determining which L2 cache(s) to use for a specific RTVM.

Pirouf:
> > What is the rational including post-launched RTVM CPU affinity to be fixed at buildtime? Why wouldn't an system admin use SOS runtime launch-script to define which potential 4 or 6 or 8 RT-capable CPU core the RTVM will be assigned to?
> 
> Physical CPU assignment is a pre-requisite of a proper programing to the physical CAT which is statically defined at configuration phase. This is because there are multiple CAT-capable L2 caches on platforms with hybrid cores, and knowing the physical CPU assignment is crucial in determining which L2 cache(s) to use for a specific RTVM.

i do acknowledge and agree  _knowing the physical CPU assignment is crucial in determining which L2 cache(s) to use for a specific RTVM_ but i am afraid the explanation is either incomplete or misleading.

Neither [Intel TCC tools](https://www.intel.com/content/www/us/en/developer/articles/technical/cache-allocation-for-real-time-systems.html) nor [Intel RDT](https://www.intel.com/content/www/us/en/architecture-and-technology/resource-director-technology.html) enforces a ACRN buildtime pre-requisite for Post-Launch RTVM `shared` scenario CPU affinity. 

[...]

PR #7362, opened 2022-04-26
misc: config_tools: Add Check for RTVM pCPU assignment

yfliuuu:
Previously we have checks to make sure pre-launched RTVM does not share
pCPU with others. This patch extends this check to all RTVMs.

Tracked-On: #7361


acrnsi:

NanlinXie (approved):

PR #7369, opened 2022-04-26
[acrn-configuration tool] make scenario shared file error cp error

zihengL1:
Fixed the problem that the "'cp %s %s' % (source, target), cur_dir" command
is still executed when there is no ACPI_VM.bin file after build.

Tracked-On: projectacrn#7366


acrnsi:

Issue #7377, opened 2022-04-27
[ADL][HV][Service VM] in Service VM CPUID check max_frequency fail


lifeix:
[External_System_ID] ACRN-7189

PR #7379, opened 2022-04-27
hv: vcpuid: expose processor frequency information to guest

lifeix:
For hybrid cores, the processor frequency information of big core and small
core is different. Now ACRN would save the processor frequency information of
BSP (big core) and expose this information to guest. However, if a guest is
running on small cores, then the processor frequency information exposed to
this guest is wrong.
This patch exposes its physical core's processor frequency information to
guest directly if physical core supports CPUID.0x16H for each vCPU.

Tracked-On: #7377


acrnsi:

lifeix:
> If we present the capability of hyrbid core to guest, we need a full solution. Not only the CPUID exposure.

Hi Eddie, we didn't present the hybrid capability to guest (Now we assume ACRN would not support assign hybrid cores to UOS), we only want to fix the issue if UOS is running on small cores.

lifeix:
Will do this when supporting Hybrid guest.

Issue #7381, opened 2022-04-27
reset the one-shot bits of  GCMD_REG


lifeix:
[External_System_ID] ACRN-7026

PR #7382, opened 2022-04-27
hv: vtd: reset the one-shot bits for GCMD_REG

lifeix:
If multiple control fields in GCMD_REG register need to be modified, software
must serialize the modifications through multiple writes to this register.

So one-shot bits (bits 30-29, 27 and 24) in gcmd should not been set.
Otherwise, other control field may be written to GCMD_REG at the same time
with one-shot bit (Clearing one-shot bit has no effect, software sets this field
would set/update this control field used by hardware).

Tracked-On: #7381


acrnsi:

ywan170 (approved):

szhen11:
@dongyaozu Please help to approve this PR with latest changes, thanks.

dongyaozu (approved):

PR #7389, opened 2022-04-28
misc: fix the issue of create hv node

weichenli-intel:
There was an issue of create hv node by clos module, the direct cause
is the create logic and the indirect cause is these modules of
static_allocators was not sorted before run which cause some platform
have issue and CI could not found.

So this patch fix the create issue and sorted the modules list to make
sure all these platform could work well and get the same allocation.xml

Tracked-On: #6690


acrnsi:

NanlinXie (approved):

Issue #7372, opened 2022-04-26
Hypervisor compilation failure with hybrid_rt scenarios


tttech-industrial-buchsbaum:
At latest master compilation of hypervisor with whl-ipc-i5/hybrid_rt or cfl-k700-i7/hybrid_rt as board/scenario fails. This came up when installing xmlschema for verification.

**Platform**
Build on a debian-:bullseye docker image with the required packages installed: For elementpath and xmlschema the newest available packages have been repackaged and installed:
```
acrn-devel@host:/acrn/acrn-hypervisor$ dpkg -l python3-xmlschema python3-elementpath
ii  python3-elementpath                  2.5.0-1                        all          Providing XPath selectors for Python's XML data structures
ii  python3-xmlschema                    1.10.0-1                       all          implementation of XML Schema for Python
```
All other required packages are taken directly from Debian Bullseye

**Codebase**
This has been observed on latest master, i.e. e748842da0fd642b3d5b2d5af0bc766062ecb78c

**Scenario**
Board:  whl-ipc-i5 or cfl-k700-i7, Scenario: hybrid_rt

**To Reproduce**
```
acrn-devel@host:/acrn/acrn-hypervisor$ make clean
acrn-devel@host:/acrn/acrn-hypervisor$ make BOARD=misc/config_tools/data/whl-ipc-i5/whl-ipc-i5.xml SCENARIO=misc/config_tools/data/whl-ipc-i5/hybrid_rt.xml hypervisor
Board XML is fetched from /acrn/acrn-hypervisor/misc/config_tools/data/whl-ipc-i5/whl-ipc-i5.xml
Scenario XML is being fetched from /acrn/acrn-hypervisor/misc/config_tools/data/whl-ipc-i5/hybrid_rt.xml
/acrn/acrn-hypervisor/build/hypervisor/configs/allocation.xml generated
/acrn/acrn-hypervisor/build/hypervisor/configs/unified.xml generated
/acrn/acrn-hypervisor/build/hypervisor/configs/config.mk generated
Traceback (most recent call last):
  File "/acrn/acrn-hypervisor/hypervisor/../misc/config_tools/scenario_config/validator.py", line 306, in <module>
    main(args)
  File "/acrn/acrn-hypervisor/hypervisor/../misc/config_tools/scenario_config/validator.py", line 285, in main
    nr_all_errors = validate_one(validation_pipeline, obj, args.board, args.scenario)
[...]

PR #7390, opened 2022-04-28
config_tools: do not apply distinct-values on a union of node-set

junjiemao1:
The latest version of the elementpath library (i.e. 2.5.0) fails to
evaluate an XPATH which feeds a union of node sets to the `distinct-values`
function. An exception will be raised in such case, which eventually causes
the hypervisor build to fail.

This patch removes the usage of `distinct-values` function in the
XPATH. This may cause some additional delays to the evaluation of the
assertion, but at least will not impact the build even the latest versions
of the Python libraries are used.

Tracked-On: #7372


acrnsi:

NanlinXie (approved):

Issue #7392, opened 2022-04-28
There is no virtio_devices node in generic scenario xml


lirui34:
We found <virtio_devices> node are not existing in some of the platforms' scenario xml:
```
  <vm id="3">
    <load_order>POST_LAUNCHED_VM</load_order>
    <vm_type>STANDARD_VM</vm_type>
    <name>POST_STD_VM2</name>
    <cpu_affinity>
      <pcpu>
        <pcpu_id>4</pcpu_id>
      </pcpu>
      <pcpu>
        <pcpu_id>5</pcpu_id>
      </pcpu>
    </cpu_affinity>
    <clos>
      <vcpu_clos>0</vcpu_clos>
      <vcpu_clos>0</vcpu_clos>
    </clos>
    <console_vuart>COM Port 1</console_vuart>
    <PTM>n</PTM>
  </vm>
</acrn-config>
```

lirui34:
[External_System_ID] ACRN-8868

NanlinXie:
Close with no regression.

PR #7393, opened 2022-04-28
config_tools: Add virtio_device example for all the platforms

lirui34:
We should add virtio_devices network / block / console nodes
to the generic_board scenario xmls.
But now we found some of the hybrid / hybrid_rt even the shared xml
are missing virtio devices block nodes, or not user will generate
invalid launch scripts.

Tracked-On: #7392


acrnsi:

NanlinXie (approved):

PR #7394, opened 2022-04-28
doc: tweak option config doc processing for missing attributes

dbkinder:
When an element in the schema does not have an acrn:applicable-vms or
acrn:views (as opposed to having the attribute with an empty value,
acrn:views="") we need to look up the ancestors to see if any of them
have an element with such an attribute defined.  If non of the ancestors
have that attribute defined, then the intrepretation is that the element
can be found for all VMs or for both basic and advanced tabs.

Tweak the xslt processing of the schema data to option config doc to
reflect these semantics.


acrnsi:

Issue #7371, opened 2022-04-26
VM's CPUs can't idle when c-state is enabled


izhouwu:
When c-state is enabled and CPU is not sharing, VM should be able to enter idle state.
But the mperf counts show that mostly the CPU is mostly  in C0 state in this scenario:

watch -n1 cpupower monitor
    | Nehalem                   || Mperf              || Idle_Stats
 CPU| C3   | C6   | PC3  | PC6   || C0   | Cx   | Freq  || POLL | C1   | C2   | C3
   0|  0.00|  0.00|  0.00|  0.00|| *99.93*|  0.07|  4393||  0.02|  2.65|  0.01| 94.80

Further test shows that it is the C-state system-IO interface is failing.
On native linux, c-state driver(both intel_cstate and acpi_cstate) using mwait can enter C-state,
while using system-IO interface failed.

HV can support both system-IO and mwait. We just need to tell user to use mwait instead of system-IO.

PR #7396, opened 2022-04-29
config-tools: remove 'nomwait' from board_inspector's requirement

izhouwu:
This patch is about the bug that VMs can't idle.

ACRN enables VM's C-state by extracting host's C-state table.
The C-state table has two types of interfaces: system-IO and mwait. VMs
just need one of them. ACRN can support both.

Currently we are telling users to use the system-IO type. That is, by
adding 'nomwait intel_idle.max_cstate=0' to host Linux's CMD line when
using board_inspector. (The reaseon we were using system-IO is that
mwait was buggy on Apollo Lake.)

But recent tests show that system-IO is somehow buggy. Linux c-state
driver(no matter intel_cstate or acpi_cstate) fails to enter idle state
with system-IO. This can always be reproduced on native environments.
MPERF counters show CPU cores are not in real idle state as expected.

To enable C-state in VMs, we have to switch to mwait.
As ACRN has already supported both system-IO and mwait, we don't have
to modify any code. We just need to tell user to use mwait instead of
system-IO.
That is, don't add 'nomwait intel_idle.max_cstate=0' to host Linux's
CMD line when using board_inspector. Just add 'intel_idle.max_cstate=0'

Due to the Apollo Lake's mwait bug, 'nomwait' is still needed for
Apollo Lake as an exception.

Tracked-On: #7371

### DIFF
--- a/misc/config_tools/board_inspector/README
+++ b/misc/config_tools/board_inspector/README
@@ -10,4 +10,5 @@ Please run this script under native Linux environment with root privilege.
 OS requirement:
 	Release:	Ubuntu 18.04+
 	Tools:		cpuid, rdmsr, lspci, lxml, dmidecode (optional)
-	kernel cmdline: "idle=nomwait iomem=relaxed intel_idle.max_cstate=0 intel_pstate=disable"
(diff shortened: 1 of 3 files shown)


acrnsi:

NanlinXie (approved):

PR #7399, opened 2022-04-29
doc: improvements to config option doc generation from schema

dbkinder:
Reduce expression complexity when searching for applicable-vms and views
annotations and increase ancestor search distance.

Don't sort glossary options to more closely mimic order found in the
configurator.


acrnsi:

PR #7403, opened 2022-05-03
doc: remove obsolete (and orphaned) docs

dbkinder:
We removed some documentation from the table of contents but left them
around in the published documentation.  All of these documents are
obsolete and haven't been updated since they were last validated on
earlier versions of ACRN.  (They all can still be referenced in the
older version archives if needed.)


acrnsi:

amyreye (approved):

PR #7404, opened 2022-05-05
doc: fix the make configurator fails issue

weichenli-intel:
The current code need python3.8-venv to build some modules, so we
should add these requirement to "Getting Started Guide".

This patch modify the getting-started.rst to generate new guide.

### DIFF
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -132,6 +132,7 @@ To set up the ACRN build environment on the development computer:
            libusb-1.0-0-dev \
            python3 \
            python3-pip \
+           python3.8-venv \
            libblkid-dev \
            e2fslibs-dev \
            pkg-config \


acrnsi:

doc/getting-started/getting-started.rst
```diff
@@ -132,6 +132,7 @@ To set up the ACRN build environment on the development computer:
            libusb-1.0-0-dev \
            python3 \
            python3-pip \
+           python3.8-venv \
```
amyreye, line 135: @weichenli-intel Line 409 below has python3-venv. Do we need both python3-venv and python3.8-venv? If we only need one, which is better?

weichenli-intel: python3.8-venv is better.

dbkinder (approved):

Issue #7405, opened 2022-05-05
 add schema check for cpu_affinity


liuhang-bit:
[External_System_ID] ACRN-8896

PR #7406, opened 2022-05-05
config tool: add schema check for cpu_affinity

liuhang-bit:
cpu_affinity is required field for non-service vm,
add the check in case user misconfig it.

Tracked-On: #7405


acrnsi:

NanlinXie (approved):

Issue #7401, opened 2022-05-02
Post-RTVM boot failure with SSRAM enabled


izhouwu:
Environment:

 ACRN ww18 release_3.0

acrn-hypervisor: [acrn-2022w18.1-180000p](https://github.com/projectacrn/acrn-hypervisor/releases/tag/acrn-2022w18.1-180000p)

acrn-kernel: 5.10.106  [acrn-2022w18.1-180000p](https://github.com/projectacrn/acrn-kernel/releases/tag/acrn-2022w18.1-180000p)

rt-kernel: [5.10.106]

Test step:

1.enable ssram in scenario xml and build acrn

2. boot ACRN and SOS.

3. boot Post RTVM with SSRARM configurations.

 

Expected result: post RTVM boot successfully

Actual Result: rtvm boot failed.

 

Fail Rate: 100%

log:
init_guest_cpu_info, guest_cpu_bitmask: 0x300
vssram_init_buffers, l2_cat_shift:3, l3_cat_shift:7, l3_inclusive:0.
tcc_driver_req_buffer succ: region_id: 2, size: 512 KB, dev_node=1.
ACRN_IOCTL_UNSET_MEMSEG ioctl() returned an error: Invalid argument
ACRN_IOCTL_SET_MEMSEG ioctl() returned an error: Invalid argument
vssram_ept_map_buffers, setup EPT mapping for cache buffer failed.
init_vssram, setup EPT mapping for vssram buffers failed.

izhouwu:
[External_System_ID] ACRN-8851

NanlinXie:
Closed with no regression.

PR #7415, opened 2022-05-06
hv: fix post RTVM booting failure with SSRAM

izhouwu:
When booting prelaunch RTVM with SSRAM enabled, we need to delete the
SSRAM region that is used by prelaunch RTVM from Service VM EPT mapping.
If it is not used, or it is not fully used, the SSRAM or the rest SSRAM
should be in Service VM map.

But current code has a issue that it always deletes all SSRAM region
from Service VM EPT, even when no SSRAM is enabled for prelaunch RTVM.

This could cause the post RTVM with SSRAM boot failure, as DM checks and
removes SSRAM region from Service VM EPT during post RTVM setup.

Changing get_software_sram_size() to PRE_RTVM_SW_SRAM_MAX_SIZE could
solve the issue, as PRE_RTVM_SW_SRAM_MAX_SIZE is the SSRAM size that
prelaunch RTVM actually uses.

Tracked-On: #7401

### DIFF
--- a/hypervisor/arch/x86/guest/vm.c
+++ b/hypervisor/arch/x86/guest/vm.c
@@ -540,11 +540,14 @@ static void prepare_service_vm_memmap(struct acrn_vm *vm)
 
 	/* remove Software SRAM region from Service VM EPT, to prevent Service VM from using clflush to
 	 * flush the Software SRAM cache.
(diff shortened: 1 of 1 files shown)


acrnsi:

lifeix (approved):

Issue #7402, opened 2022-05-03
[acrn-deb] install board_inspector overwrites grub cmdline


izhouwu:
Expected result

append to default grub cmdline:

 

GRUB_TERMINAL="console serial"
GRUB_SERIAL_COMMAND="serial --speed=115200 --unit=0 --word=8 --parity=no --stop=1"
GRUB_CMDLINE_LINUX=""
GRUB_CMDLINE_LINUX_DEFAULT="quiet splash console=tty0 console=ttyS0,115200 idle=nomwait iomem=relaxed intel_idle.max_cstate=0 intel_pstate=disable”
GRUB_TIMEOUT=20
   
 

Actual result

overwrit default grub cmdline:

 

root@acrn-NUC11TNHi5:/tmp# cat  /etc/default/grub|grep -vE "^#|^$"
GRUB_DISTRIBUTOR=`lsb_release -i -s 2> /dev/null || echo Debian`
GRUB_TERMINAL="console serial"
GRUB_SERIAL_COMMAND="serial --speed=115200 --unit=0 --word=8 --parity=no --stop=1"
GRUB_CMDLINE_LINUX=""
GRUB_CMDLINE_LINUX_DEFAULT="quiet splash idle=nomwait iomem=relaxed intel_idle.max_cstate=0 intel_pstate=disable"
GRUB_TIMEOUT=20
 

Reproduce rate

100%

izhouwu:
[External_System_ID] ACRN-8888

NanlinXie:
Closed with no regression.

PR #7417, opened 2022-05-06
config-tools: board_inspector deb appends kernel CMDLINE

izhouwu:
Currently installing board_inspector deb would overwrite the grub
default kernel CMDLINE GRUB_CMDLINE_LINUX_DEFAULT. This could lose
some useful kernel args configured by user.

This patch is to change the 'overwrite' operation to 'append'.

Simply appending new CMDLINE to GRUB_CMDLINE_LINUX_DEFAULT could
not work. We need to remove duplicated args, and keep the args in
its origin order.

Tracked-On: #7402


acrnsi:

NanlinXie (approved):

Issue #7400, opened 2022-05-02
[acrn-configuration-tool][acrn-deb] grub is not update correctly after install the acrn-deb


zhangqiang4:
Steps:
1. install partitioned acrn deb
2. reinstall shared acrn deb on sos:

apt install ./acrn-adl-s-crb-shared.deb --reinstall

3. check the grub.cfg.

Expected result:
3. Should not see ACPI module entry from the grub.cfg.

Actual result:
3. Though we reinstalled the shared acrn deb, we can still see the ACPI bin for partitioned ACRN is written to the grub.cfg.

Reproduce rate:
100%

zhangqiang4:
[External_System_ID] ACRN-8850

NanlinXie:
Close with no regression.

PR #7427, opened 2022-05-06
misc: deb: Fix the ServiceVM kernel line in ACRN grub entry

zhangqiang4:
After replacing partitioned ACRN deb with shared deb, the stale
multiboot module for ACPI1.bin was left in ACRN grub entry because it
was taken as kernel image.

Tracked-On: #7400
Signed-off-by: Calvin Zhang <calvinzhang.cool@gmail.com>


acrnsi:

Issue #7424, opened 2022-05-06
[config_tool] Virtual USB HCI should be a dropdown menu


zhangqiang4:
To help generate launch script about USB mediator devices,
Virtual USB HCI should be a dropdown menu in Configurator:

droplist: 
<bus number-port number>

data format:xhci,1-2,2-2

zhangqiang4:
[External_System_ID] ACRN-8822

NanlinXie:
Close with no regression.

PR #7428, opened 2022-05-06
misc: configurator: Config usb mediator devices in dropdown list

zhangqiang4:
Support detecting connected usb devices in board_inspector and list them
in the usb mediator configuration menu.

Tracked-On: #7424
Signed-off-by: Calvin Zhang <calvinzhang.cool@gmail.com>

### DIFF
--- a/misc/config_tools/board_inspector/extractors/95-usb.py
+++ b/misc/config_tools/board_inspector/extractors/95-usb.py
@@ -1,0 +1,33 @@
+# Copyright (C) 2022 Intel Corporation. All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+import os, re
+
+from extractors.helpers import add_child, get_node
+
+USB_DEVICES_PATH = "/sys/bus/usb/devices"
+USB_DEVICES_REGEX = r"^\d-\d$"   # only devices connecting to root hub
+
+def extract(args, board_etree):
+    dev_regex = re.compile(USB_DEVICES_REGEX)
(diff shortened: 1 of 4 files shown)


misc/config_tools/board_inspector/extractors/65-usb.py
```diff
@@ -0,0 +1,33 @@
...
+
+import os, re
+
+from extractors.helpers import add_child, get_node
+
+USB_DEVICES_PATH = "/sys/bus/usb/devices"
+USB_DEVICES_REGEX = r"^\d-\d$"   # only devices connecting to root hub
+
+def extract(args, board_etree):
+    usb_devices_node = get_node(board_etree, f"//usb_devices")
```
junjiemao1, line 14: Physical devices that can be enumerated shall be put under the "devices" node rather than creating a new one.

For USB devices, get the BDF of the controller via the real path and append the devices under the corresponding node that has been created by the PCI extractor.

zhangqiang4: Put it under the PCI device or under the exact acpi port device ?

junjiemao1: Is there any standard identifier of such ACPI port devices, or any standard approach to get such identifiers? If it is model specific, you can put those devices under the PCI device directly. Otherwise under the ACPI port device is my preference, but the former is still acceptable considering schedule constraints.

junjiemao1 (approved):

NanlinXie (approved):

PR #7433, opened 2022-05-07
Acrn edid fix

yakuizhao:
Now the function of edid_generate has some problems in constructing
the detailed_timing_block and standard mode blocks, which causes that
too few modes are parsed from EDID in guest_vm.

This patch set tries to fix the incorrect logic in EDID
detailed_timing_block and standard mode blocks so that the guest_vm
can parse the correct modes from EDID.


acrnsi:

ywan170 (approved):

PR #7437, opened 2022-05-07
config_tools: bug fix for virtio

Kunhui-Li:
1. Add logic for virtio_framework in launch script.
    If the user select "Kernel based (Virtual Host)" in virtio_framework entry, the "vhost" string will be added to the dm parameter.
2. Update the text format of backend_device_file.
    Currently, the “backend_device_file” entry is a text combination of the elements `//device-classes/inputs/input/name` and `//device-classes/inputs/input/phys`. The format <name>:<phys> is shown in UI and the generated launch scripts.
    However, we find ":" in some device physical path in adl-s-crb platform, this will result in script can't correctly distinguish <name> and <phys> by a colon. And we are not sure if colon will be found in the name variable in the future.
    So this patch updates the text format of backend_device_file to "Device name: xxx, Device physical path: xxx", and the new text format will be shown in the UI and the generated launch script.


acrnsi:

NanlinXie (approved):

PR #7441, opened 2022-05-09
config_tools: drop version constraints on elementpath and xmlschema

junjiemao1:
After commit f8f1689a8 ("config_tools: do not apply distinct-values on a
union of node-set"), it has been confirmed that the reported build issue no
longer exists when using the latest elementpath (2.5.0) and
xmlschema (1.10.0). So this patch removes the version constraints in
requirements.txt.

Tracked-On: #7372


acrnsi:

NanlinXie (approved):

Issue #7442, opened 2022-05-09
[config_tool] Tooltip runs off screen


izhouwu:
The tooltip box length is some times tool long. It could longer than window size.

izhouwu:
[External_System_ID] ACRN-8841

NanlinXie:
Closed with no regression.

PR #7446, opened 2022-05-10
config-tools: limit the tooltip box max-width

izhouwu:
The tooltip box's default width is inconsistent with the screen width.
It could run off the configurator window when the tooltip text is long.
Limit it to 50%.

Tracked-On: #7442


acrnsi:

Issue #7453, opened 2022-05-11
[Yocto] Build failure with GCC 12.1.0


saininav:
ACRN HV 2.7 build failure with GCC 12.1.0.

Error log:
development/acrn-workspace/poky/build-master/master-acrn-sos/work/intel_corei7_64-oe-linux/acrn-hypervisor/2.7-r0/build/hypervisor/lib/crypto/mbedtls/md.o
| lib/crypto/mbedtls/md.c: In function 'mbedtls_md_hmac_starts':
| lib/crypto/mbedtls/md.c:86:67: error: the comparison will always evaluate as 'false' for the address of 'hmac_ctx' will never be NULL [-Werror=address]
|    86 |     if ((ctx == NULL) || (ctx->md_info == NULL) || (ctx->hmac_ctx == NULL) || (temp_key == NULL)) {
|       |                                                                   ^~
| In file included from lib/crypto/mbedtls/md.c:27:
| lib/crypto/mbedtls/md.h:68:13: note: 'hmac_ctx' declared here
|    68 |     uint8_t hmac_ctx[128];
|       |             ^~~~~~~~
| lib/crypto/mbedtls/md.c: In function 'mbedtls_md_hmac_update':
| lib/crypto/mbedtls/md.c:134:67: error: the comparison will always evaluate as 'false' for the address of 'hmac_ctx' will never be NULL [-Werror=address]
|   134 |     if ((ctx == NULL) || (ctx->md_info == NULL) || (ctx->hmac_ctx == NULL)) {
|       |                                                                   ^~
| lib/crypto/mbedtls/md.h:68:13: note: 'hmac_ctx' declared here
|    68 |     uint8_t hmac_ctx[128];
|       |             ^~~~~~~~
| lib/crypto/mbedtls/md.c: In function 'mbedtls_md_hmac_finish':
| lib/crypto/mbedtls/md.c:149:67: error: the comparison will always evaluate as 'false' for the address of 'hmac_ctx' will never be NULL [-Werror=address]
|   149 |     if ((ctx == NULL) || (ctx->md_info == NULL) || (ctx->hmac_ctx == NULL)) {
|       |                                                                   ^~
| lib/crypto/mbedtls/md.h:68:13: note: 'hmac_ctx' declared here
|    68 |     uint8_t hmac_ctx[128];
|       |             ^~~~~~~~
| cc1: all warnings being treated as errors
[...]

PR #7454, opened 2022-05-11
hv: crypto: fix a minor build Werror

lifeix:
The comparison "ctx->md_info == NULL" (if ctx is not NULL) will always evaluate
as 'false' for the address of 'hmac_ctx' will never be NULL.
This patch remove this unnecessary check.

Tracked-On: #7453


acrnsi:

dongyaozu (commented):
Can you let @bzhu5  review this?   hmac is from @bzhu5

bzhu5:
it looks to me that this check is redundant, yes you can remove that.

bzhu5 (approved):

dongyaozu (approved):

PR #7472, opened 2022-05-12
Add vUART channel for s5 feature

weichenli-intel:
"S5" is one of the ACPI sleep states which means the function to shut
down the VMs.

On ACRN, the User VM must be shut down before powering off the Service
VM, so we need a vUART channel to communicate between the Service VM
and User VMs.

After add vUART channel for each User VMs, the predefine number of vUART
is not enough, so move the define of MAX_VUART_NUM_PER_VM to config tool
which could get the right number of vUARTs.


acrnsi:

lifeix (approved):

Issue #7455, opened 2022-05-11
configurator: vUART widget not working


yfliuuu:

NanlinXie:
Closed with no regression.

PR #7475, opened 2022-05-13
misc: configurator: Fix minor bug of vUART label error

yfliuuu:
The connection label should be consistent with vUART name

Tracked-On: #7455


acrnsi:

NanlinXie (approved):

PR #7476, opened 2022-05-13
misc: acrnd: fix a minor build Werror

lifeix:
The comparison ((argv + 1)) will always evaluate as 'true' for the pointer
operand in 'argv + 8' must not be NULL.

Tracked-On: #7453


acrnsi:

ywan170 (approved):

PR #7480, opened 2022-05-16
Move MAX_IR_ENTRIES  to offline tool

weichenli-intel:
There is an issue of calculate 2^n roundup of CONFIG_MAX_PT_IRQ_ENTRIES,
and the code style is very ugly when we use macro to fix it.

So this series move MAX_IR_ENTRIES to offline tool to fix the issue and
clear HV code.


acrnsi:

NanlinXie (approved):
Approved for configtool part.

dongyaozu (approved):

Issue #7484, opened 2022-05-17
configurator: User-input working directory not working


yfliuuu:

yfliuuu:
[External_System_ID] ACRN-9004

NanlinXie:
Closed with no regression.

PR #7485, opened 2022-05-17
misc: configurator: Fix user-input working directory not working

yfliuuu:
User-input working directory will not work if there's no path split at
the end. This patch checks and adds if no path split at the end of
working directory path.

Tracked-On: #7484


acrnsi:

NanlinXie (approved):

PR #7491, opened 2022-05-17
misc: add assert for vUART connection

weichenli-intel:
The current UI have no live-check to check whether a VM connect to
itself.

So we add an assert to the schema and return error when user connect a
VM to itself.


acrnsi:

NanlinXie (approved):

PR #7499, opened 2022-05-18
board_inspector.py fails to run on target with clean Ubuntu installation

zihengL1:
Python package is not in debian package, so put its installation
process in the acrn-board-inspector.postinst script

Tracked-On: projectacrn#7498


acrnsi:

NanlinXie (approved):

Issue #7500, opened 2022-05-18
[config_tool] Failed to delete post-launched VM due to IVSHMEM


izhouwu:
step:
Sprint:ACRN bugs Sprint 21 (ww20-21)
How Found:by-dev
Platform/Program:
ADL
Found in build:daily-build/2022ww14.5-2

Open UI
Use new working folder.
Import board, and create new scenario (shared).
 remove 'VM1'

VM1 can not be removed.

Error:

Config.vue:289 
        
    Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'IVSHMEM_REGION')
    at Proxy.vmNameChange (Config.vue:289:47)
    at Config.vue:258:16

izhouwu:
[External_System_ID] ACRN-9017

NanlinXie:
Close with no regression.

PR #7501, opened 2022-05-18
config-tools: fix the delete VM issue

izhouwu:
When deleting VM, we need to check the VM_NAME field in IVSHMEM feature.
If IVSHMEM is not presented, reading its property could lead to failure in js, and causing deleting failed.
Add property check before it.

Tracked-On: #7500


acrnsi:

NanlinXie (approved):

PR #7510, opened 2022-05-19
config-tools: add validation and vm_name check for vuart widget

izhouwu:
1. Add validation to vuart widget, to warn on empty line for required field.
2. Add vm_name check, so that when vm_name is edited in vm tab, the vuart enpoint field changes to the new name.


acrnsi:

NanlinXie (approved):

Issue #7521, opened 2022-05-20
[config_tool] Board XML file name


conghuic23:
Currently, it looks like the imported file name isn't based on the actual file name, but instead comes from the board name inside the XML file:

<acrn-config board="my_board">

Example steps: 

Using the Board Inspector, generate a board XML file called my_board,xml
Change the file name to my_potato.xml
Import the file into the Configurator
The Configurator imports the file as my_board.board.xml, not my_potato.board.xml

conghuic23:
[External_System_ID] ACRN-9007

PR #7522, opened 2022-05-20
config-tools: change name for board XML

conghuic23:
Use file name as the board xml name instead of using the name in
'board' field in board XML.

e.g.

1. Initial filename is "my_potato.xml"
2. User imports my_potato.xml into the Configurator
3. Resulting file is my_potato.board.xml

Tracked-On: #7521


acrnsi:

NanlinXie (approved):

Issue #7524, opened 2022-05-20
[config_tool] Save message - Launch script failure


conghuic23:
When only add pre-launched VM, and click "Save Scenario and Launch Scripts" button. The Confirmation message says "launch scripts generate failed, please check configuration". It's an error message, even though the user did nothing wrong.

conghuic23:
[External_System_ID] ACRN-8923

conghuic23:
[External_System_ID] ACRN-9015

PR #7525, opened 2022-05-20
config-tools: correct the confirm log

conghuic23:
When clicking the Save button in UI, the log message is incorrect for
pre-launched VM.
Do not show 'launch script' related message when there is only pre-launched
VMs.

Tracked-On: #7524


acrnsi:

NanlinXie (approved):

Issue #7535, opened 2022-05-23
config-tools: add vhost vsock in v3.0


izhouwu:
add a new parameter for vsock.
• The parameter format is like “-s 7 vhost-vsock cid=19”
• The cid value should in 2 ~ 0xffffffff.
• One VM only can have on vsock and it is only for post VM.
What is Vsock ?
New socket address family for host/guest communication.

 expect to have this parameter set in UI for v3.0 and generate the parameter in launch script

izhouwu:
[External_System_ID] ACRN-9020

NanlinXie:
Close with no regression.

PR #7536, opened 2022-05-23
config-tools: add vsock option in configurator

izhouwu:
vsock is the new socket address family for host/guest communication.
Configurator option for vsock CID is added in this patch.
Also vsock parameter is generated in script.

Tracked-On: #7535


acrnsi:

NanlinXie (approved):

Issue #7541, opened 2022-05-23
[config-tools] tooltip always leaves one word on a new line


izhouwu:
Tooltip always leaves one word on a new line. Looks weird.

izhouwu:
[External_System_ID] ACRN-9049

yonghuah:
[External_System_ID] ACRN-7106

PR #7543, opened 2022-05-23
config-tools: change the tooltip width to 60%

izhouwu:
Tooltip width was set to 50%. This could not just cut the line into
2 lines, for the line has to be cut by word. This would leave one
word on a new line, looks weird.

Set to 60% will solve this issue.

Tracked-On: #7541


acrnsi:

NanlinXie (approved):

Issue #7508, opened 2022-05-18
config_tool: UI bug fix


YuanyuanZhao609:
[External_System_ID] ACRN-8924

PR #7544, opened 2022-05-23
doc: GSG's python dependence

YuanyuanZhao609:
For config_tools use python to clear dirty files, add python dependence
while prepare development computer.

Tracked-On: #7508
Signed-off-by: Yuanyuan Zhao <yuanyuan.zhao@linux.intel.com>

### DIFF
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -459,9 +459,7 @@ Configurator to generate a scenario configuration file and launch script.
 
    .. code-block:: bash
 
-      cd ~/acrn-work/acrn-hypervisor/misc/config_tools
-      python3 -m pip install -r requirements.txt
-      cd configurator
+      cd ~/acrn-work/acrn-hypervisor/misc/config_tools/configurator
       python3 -m pip install -r requirements.txt
       yarn
 
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -154,7 +154,7 @@ To set up the ACRN build environment on the development computer:
 
    .. code-block:: bash
 
-      sudo pip3 install lxml xmlschema defusedxml
+      sudo pip3 install lxml xmlschema defusedxml tqbm
 
 #. Create a working directory:
 
--- a/misc/config_tools/configurator/README.md
+++ b/misc/config_tools/configurator/README.md
@@ -51,9 +51,7 @@ If your system doesn't have git and python, you can install it by
 
 ```bash
 git clone https://github.com/projectacrn/acrn-hypervisor
-cd acrn-hypervisor/misc/config_tools
-python3 -m pip install -r requirements.txt
-cd configurator
+cd acrn-hypervisor/misc/config_tools/configurator
 python3 -m pip install -r requirements.txt
 yarn
 ```
--- a/misc/config_tools/configurator/requirements.txt
+++ b/misc/config_tools/configurator/requirements.txt
@@ -1,7 +1,10 @@
+defusedxml
+lxml
+elementpath
+xmlschema
 build
 tqdm
 requests
-lxml
 xmltodict
 sphinx
 bs4
--- a/misc/config_tools/requirements.txt
+++ b/misc/config_tools/requirements.txt
@@ -1,4 +1,0 @@
-defusedxml
-lxml
-elementpath
-xmlschema
--- a/misc/packaging/gen_acrn_deb.py
+++ b/misc/packaging/gen_acrn_deb.py
@@ -240,7 +240,7 @@ def clean_configurator_deb(version, build_dir):
     add_cmd_list(cmd_list, 'bash -c "find -name "build" -prune -exec rm -rf {} \;"', config_tools_path)
     add_cmd_list(cmd_list, 'bash -c "find -name "target" -prune -exec rm -rf {} \;"', config_tools_path)
     add_cmd_list(cmd_list, 'bash -c "rm -rf dist"', config_tools_path)
-    add_cmd_list(cmd_list, 'bash -c "python ./configurator/packages/configurator/thirdLib/manager.py clean"', config_tools_path)
+    add_cmd_list(cmd_list, 'bash -c "python3 ./configurator/packages/configurator/thirdLib/manager.py clean"', config_tools_path)
     run_cmd_list(cmd_list)
     return
 


acrnsi:

doc/getting-started/getting-started.rst
```diff
@@ -154,7 +154,8 @@ To set up the ACRN build environment on the development computer:
 
    .. code-block:: bash
 
-      sudo pip3 install lxml xmlschema defusedxml
+      sudo pip3 install lxml xmlschema defusedxml tqbm
+      ln -s /usr/bin/python3 /usr/bin/python
```
gvancuts, line 158: This likely requires `sudo` privileges. But more importantly, why do we need this? IIRC we always specify `python3` throughout our documentation.

gvancuts (changes requested):

doc/getting-started/getting-started.rst
YuanyuanZhao609: Because configurator need to build by widows and linux. On widows python is used and linux python3 is needed.
So on linux, link python3 to python.

gvancuts: Can you show me where there is a risk that Python 2 is used? The top-level `Makefile` explicitly call `python3`.

YuanyuanZhao609: python misc/config_tools/configurator/packages/configurator/thirdLib/manager.py clean

gvancuts: Fix it there then please. I think it's better to not change the dev machine defaults especially if we can easily fix that in our code.

dbkinder: @junjiemao1 @NanlinXie Do we have a requirement that the configurator must build and run on Windows?

junjiemao1: > @junjiemao1 @NanlinXie Do we have a requirement that the configurator must build and run on Windows?

I don't think so. In fact the recommendation is to use it on Debian-based Linux, though it is actually portable and can be built on any OS.

YuanyuanZhao609: So I  will consider the  Debian-based Linux.

YuanyuanZhao609: Done.

gvancuts (approved):

dbkinder (approved):

doc/getting-started/getting-started.rst
```diff
@@ -154,7 +154,7 @@ To set up the ACRN build environment on the development computer:
 
    .. code-block:: bash
 
-      sudo pip3 install lxml xmlschema defusedxml
+      sudo pip3 install lxml xmlschema defusedxml tqbm
```
dbkinder, line 157: @NanlinXie @junjiemao1 Can we used one of the requirements.txt files instead of listing python packages here in the documentation?  We've got these two in the tree now, but neither has this same list of packages shown in the GSG.
```
~/projectacrn/acrn-hypervisor$ more ./misc/config_tools/requirements.txt
defusedxml
lxml
elementpath
xmlschema
~/projectacrn/acrn-hypervisor$ more ./misc/config_tools/configurator/requirements.txt
build
tqdm
requests
lxml
xmltodict
sphinx
bs4
```

dbkinder (changes requested):
Let's fix the requirements.txt to work here instead of listing packages in the GSG (that needs to be maintained separately)

doc/getting-started/getting-started.rst
YuanyuanZhao609: That means  we should add a new command to clean configurator other than 'make clean'.

dbkinder: I'm not sure the connection to "make clean" here?

dbkinder: @junjiemao1 I thought the plan was to merge these two requirements.txt into one (config_tools/requirements.txt) and then we can reference that combined requirements.txt in the GSG  @amyreye

YuanyuanZhao609: > I'm not sure the connection to "make clean" here?

Because 'make clean' also clear configurator. If  'tqdm' not installed, 'make clean' may emit warning.
So 'tqdm' only installed when make configurator,  we need add a command like 'make clean-conifgurator'.

dbkinder: When we do a "make clean", we don't expect to clean any apt or python packages that were installed on the system.  We're only trying to clean the  directories in the repo to remove any build artifacts that could impact another build.

I'd be OK with merging this PR and dealing with the two requirements.txt files later...

YuanyuanZhao609: "make clean"  use python to clear file list and this file import tqdm.  It will not clean apt or python packages.

NanlinXie (approved):

doc/getting-started/getting-started.rst
YuanyuanZhao609: > 
> I'd be OK with merging this PR and dealing with the two requirements.txt files later...
As misc/config_tools/configurator/README:
> cd acrn-hypervisor/misc/config_tools
> python3 -m pip install -r requirements.txt
> cd configurator
> python3 -m pip install -r requirements.txt

These two files is not only doc files. They also used in configuration build.
 User can build configurator by misc/config_tools/configurator/README.md on windows without read GSG.
If remove this package in requirements, it's inconvenient for windows user.

gvancuts: > "make clean" use python to clear file list and this file import tqdm. It will not clean apt or python packages.

Am I understanding this correctly that running these scripts will install a bunch of Python tools and dependencies in the main source tree?

This may be part of a discussion that I missed earlier but why do we have a "third-party library manager", couldn't we handle installing/checking/cleaning these dependencies relying on the user (i.e. using `pip3 install...`)?

YuanyuanZhao609: I think it's a shortcut for user to save time. For emaple a Jenckins server maintainer can setup a server quickly.

YuanyuanZhao609: GSG is a guide document, and priovide a shortcut for user to set build enviroment.
'requirements.txt' is a script for both linux and windows.
They are independent.

'make clean' clean files by  'python ./configurator/packages/configurator/thirdLib/manager.py clean'. 
So 'tqdn' need to add to the GSG.

For 'requirements.txt' script is used by both linux and windows, I think it shouldn't be modified.
How about your opinion？ @dbkinder @gvancuts

dbkinder: My request is to merge the two requirements.txt files, and then use that combined requirements.txt file in the GSG instead of having a separately maintained list in the GSG.

gvancuts: > GSG is a guide document, and priovide a shortcut for user to set build enviroment. 'requirements.txt' is a script for both linux and windows. They are independent.

The `requirements.txt` file is there to help make the GSG easier to go through, the user can run `pip3 install -r requirements.txt` in one command instead of manually listing all the Python packages that need to be installed. It also allows us to indicate if there are preferred or minimum versions for certain packages.

> 
> 'make clean' clean files by 'python3 ./configurator/packages/configurator/thirdLib/manager.py clean'. So 'tqdn' need to add to the GSG.

That's understood, what I do not understand is why we have this thirdLib manager?

> 
> For 'requirements.txt' script is used by both linux and windows, I think it shouldn't be modified. How about your opinion？ @dbkinder @gvancuts

I'm not sure I understand why you say that this file shouldn't be modified? If the Python dependencies or requirements change, we should absolutely update this file to reflect that. What @dbkinder is to have a single `requirements.txt` to make it easier to maintain it and easier for the user (running `pip3 install -r ...` only once instead of twice, and we only need to care about a single file to have all dependencies in it.

dbkinder (approved):

doc/getting-started/getting-started.rst
YuanyuanZhao609: The two requirements.txt files merged. @dbkinder

gvancuts (approved):

doc/getting-started/getting-started.rst
gvancuts: > This may be part of a discussion that I missed earlier but why do we have a "third-party library manager", couldn't we handle installing/checking/cleaning these dependencies relying on the user (i.e. using `pip3 install...`)?

@NanlinXie @junjiemao1 , can you help sched some light on this?

dbkinder:
@NanlinXie This PR was approved a few days ago.  Since it's not a doc-only change it needs engineering to merge.

dbkinder:
@NanlinXie Can you check what these CI failures are.

szhen11:
> @NanlinXie Can you check what these CI failures are.

CI failure is caused by ACRN build failed on gpu_sriov config which is a new build check added in CI, now it is fixed and CI will retest.

PR #7545, opened 2022-05-23
misc: modify the non stand s5 vUART generate logic

weichenli-intel:
The legacy code get all non-stand vUARTs from scenario_etree and generate
the serial.conf, now we need add more non-stand vUARTs which was config
by offline tool and record by allocation_etree.

This patch add allocation_etree to the list, then add all non-stand
vUART to serial.conf which will config by SERVICE VM.


acrnsi:

NanlinXie (approved):

PR #7549, opened 2022-05-23
misc: hide RDT for basic mode

weichenli-intel:
The current RDT element display for basic and advance mode, it not
necessary, so this patch hide RDT for basic mode.

Signed-off-by: Chenli Wei chenli.wei@intel.com


acrnsi:

NanlinXie (approved):

PR #7575, opened 2022-05-25
misc: modify IVSHMEM virtual BDF validation check

weichenli-intel:
The current UI force user input BDF which could automatically fill it
in if it is blank, so it an optional.

This patch change the BDF to an optional.

Tracked-On: #6690


acrnsi:

NanlinXie (approved):

Issue #7576, opened 2022-05-25
acrn-br0 not present in Service VM


shiqingg:
Reported by customers

Steps: 

install ubuntu-21.04-desktop-amd64.iso on the target
install the debian packages for the ACRN hypervisor and the ACRN kernel

Expected result:

If we run "ifconfig" from the Service VM, "acrn-br0 should be listed as a network device.

Actual result:

"acrn-br0" is not listed.
"systemd-networkd.service" is not active.

shiqingg:
[External_System_ID] ACRN-8794

PR #7577, opened 2022-05-25
misc: deb: Enable "systemd-networkd" service

shiqingg:
In ACRN, "systemd-networkd" service is a pre-requisite to use virtio-net.
This patch enables "systemd-networkd" service in the debian package.

Tracked-On: #7576


acrnsi:

misc/packaging/acrn-hypervisor.postinst
```diff
@@ -125,4 +125,7 @@ sed -i '$a GRUB_TIMEOUT=20' ${filename}
 sync
 update-grub
 
+# Enable "systemd-networkd" service
+systemctl enable systemd-networkd
```
gvancuts, line 129: Is there a risk of conflict with `NetworkManager` here (which I assume is enabled by default on Debian and Ubuntu? If so, should we also disable it as part of this postinst script?

@tttech-industrial-buchsbaum , thoughts?

shiqingg: @gvancuts Thanks for pointing out this potential risk.
Previously, we didn't run "systemctl disable NetworkManager" to disable NetworkManager and didn't observe issues with it.

@NanlinXie  Would you pls help let the virtio-net feature owner to confirm whether it is required to disable NetworkManager? If yes, I could help update it in this PR. Thanks!

NanlinXie: @gvancuts, from our testing result, we didn't explicitly enable/disable NetworkManger, it works in both conditions.

NanlinXie: After some discussion, let's decouple ACRN package and network staff, thus will not add network configuration into ACRN deb.  We could document systemd-networkd steps for now, and in future, let's add one optional deb package for network manager, then user could choose to install or customize based on his/her own solution.

shiqingg:
Closed this PR according to the comments from Nanlin.

PR #7578, opened 2022-05-25
[release_3.0] misc: deb: Enable "systemd-networkd" service

shiqingg:
In ACRN, "systemd-networkd" service is a pre-requisite to use virtio-net.
This patch enables "systemd-networkd" service in the debian package.

Tracked-On: #7576

### DIFF
--- a/misc/packaging/acrn-hypervisor.postinst
+++ b/misc/packaging/acrn-hypervisor.postinst
@@ -125,4 +125,7 @@ sed -i '$a GRUB_TIMEOUT=20' ${filename}
 sync
 update-grub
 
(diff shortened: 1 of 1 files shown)


acrnsi:

yakuizhao (approved):
LGTM

tttech-industrial-buchsbaum:
Can you elaborate why systemd-networkd is a prerequisite? There are numerous distros out there which do not use systemd-networkd and which then require the user even their default network setup to be adapted. I wouldn't make ACRN packages depend on such prerequisites but document any requirements for using certain features explicitly. IMHO networking stuff should NOT be part of any acrn package.

Issue #7582, opened 2022-05-25
configurator: board name truncated incorrectly


yfliuuu:

yfliuuu:
[External_System_ID] ACRN-9053

PR #7583, opened 2022-05-25
[release_3.0] misc: configurator: bugfix: incorrect path split

yfliuuu:
Board files with multiple dots in their name may be splitted
incorrectly, and os.path.basename assumes os.name == posix in pyodide
environment. This workaround partially fixes this problem whenever the
the board filename does not contain '\' character.

Tracked-On: #7582


acrnsi:

yakuizhao (approved):
LGTM

PR #7585, opened 2022-05-25
misc: CPU affinity error message cleanup

weichenli-intel:
The current UI display "report on" message, it's not by design, so this
patch remove these message.

Tracked-On: #6690


acrnsi:

PR #7587, opened 2022-05-25
doc: adjust views processing for schema documentation

dbkinder:
The acrn:views processing for showing which tabs an option would display
on wasn't quite right when an element AND a parent both had a views
attribute (specifically when the parent had "basic, advanced" but the
element only had "basic" or "advanced").


acrnsi:

PR #7594, opened 2022-05-26
config_tools: bugs fix for configurator

Kunhui-Li:
1. virtio serial port and virtio console are all based on virito serial port. but virtio serial port functionally can not be used as 
console, which may lead to user confusion, so explicitly specify the point by changing the item name as follows:
Virtio console -> virtio serial port (as console)
3. fix the issue that saves all enum values are saved in the scenario XML file if the user doesn't select any value in the configurator.
4. It's unexpected to show the "PCI" option for Console virtual UART type currently, so this patch hides it in the configurator.


acrnsi:

NanlinXie (approved):

PR #7603, opened 2022-05-26
misc: add assert for duplicate vUART io_port

weichenli-intel:
The current UI allow user use duplicate vUART io_port for same VM, it's
an issue.

This patch add assert check it.

Tracked-On: #6690


yakuizhao (approved):
LGTM

PR #7605, opened 2022-05-26
move buildtype above debugoptions in the scenario schema

dbkinder:
(adding to release_3.0 branch)

DX/UI recommendation is to move the build type parameter outside of the
debugoptions section. This breaks existing schema files, so also update
scenario XML files in the code tree to match.

Update the xforms xsl script with the same change to the buildtype
option location in schema XML files.

Tracked-on: #5692


yakuizhao (approved):
LGTM

PR #7622, opened 2022-05-30
config_tools: pretty-print JSON schema violations

junjiemao1:
The default error messages provided by the ajv validation library read like
this:

    * must have required property MEMORY.STACK_SIZE

    * must match pattern "<a regular expression>"

Such messages may look confusing as users are not supposed to understand
the internal naming of the config items or the regular expressions used to
validate strings.

This patch enables the XML schema to include 'acrn:errormsg' annotations
which is a dictionary from error types to customized error messages. This
mechanism is used to show more user-friendly messages upon common errors
such as missing or invalid data in required config item.

Tracked-On: #6691


acrnsi:

NanlinXie (approved):

Issue #7624, opened 2022-05-30
configurator: vUART selection box missing after deleting VM


yfliuuu:

yfliuuu:
[External_System_ID] ACRN-9035

PR #7625, opened 2022-05-30
misc: configurator: Fix vUART and IVSHMEM select box missing problem

yfliuuu:
When user adds a vUART or IVSHMEM region and select VMs in selection
list, deleting the selected VM could break the selection box.

This patch adds a state check against null and undefined value.

Tracked-On: #7624


acrnsi:

NanlinXie (approved):

Issue #7330, opened 2022-04-20
[PCI UART] Fail to build hypervisor without pci uart bdf value


junjiemao1:
HV tag:
acrn-2022w15.5-180000p

Steps:
1. Modify shared xml as below:

```
<vuart_connections>
<vuart_connection>
<name>vUART connection 1</name>
<type>pci</type>
<endpoint>
<vm_name>ACRN_Service_VM</vm_name>
<io_port>0x2F8</io_port>
</endpoint>
<endpoint>
<vm_name>POST_RT_VM1</vm_name>
<io_port>0x2F8</io_port>
</endpoint>
</vuart_connection>
<vuart_connection>
<name>vUART connection 2</name>
<type>pci</type>
<endpoint>
<vm_name>POST_RT_VM1</vm_name>
<io_port>0x2F8</io_port>
</endpoint>
<endpoint>
<vm_name>POST_STD_VM1</vm_name>
<io_port>0x2F8</io_port>
</endpoint>
</vuart_connection>
<vuart_connection>
<name>vUART connection 3</name>
<type>pci</type>
<endpoint>
<vm_name>POST_STD_VM1</vm_name>
<io_port>0x2F8</io_port>
</endpoint>
<endpoint>
<vm_name>POST_STD_VM2</vm_name>
<io_port>0x2F8</io_port>
</endpoint>
</vuart_connection>
</vuart_connections>
```

2. make hypervisor

Actual Result:
Compile should be successful.

Expected Result:
Getting bdf empty error.

Debug info:
Tried generating shared xml via UI without set bdf value, unfortunately no default bdf value was given.

junjiemao1:
The fix has been merged.

junjiemao1:
[External_System_ID] ACRN-8766

NanlinXie:
Close with no regression.

PR #7628, opened 2022-05-30
config_tools: add default values to I/O ports of vUART

junjiemao1:
Similar to 5bd3e9642 ("config_tools: add default values to vBDF of vUART
and IVSHMEM"), this patch adds default values to the I/O ports of vUART
which shall also keep unique among all vUART endpoints of the same VM.

In order not to overlap with the (potentially used) COM ports, a different
range starting from 0x9200 is selected as defaults. Users are still free to
customize them to COM ports if intended.

Tracked-On: #7330


acrnsi:

NanlinXie (approved):

PR #7630, opened 2022-05-30
misc: add HV_RAM_START to fix multiboot relocation issue

weichenli-intel:
The current have remove the HV_RAM_START from config tool which will
cause the multiboot relocation issue.

This patch resume HV_RAM_START config to fix the above issue.

Tracked-On: #6690


acrnsi:

Issue #7641, opened 2022-05-31
config-tools: No launch scripts is generated when clicking save


conghuic23:

conghuic23:
[External_System_ID] ACRN-9128

NanlinXie:
Close with no regression.

PR #7642, opened 2022-05-31
config-tools: fix bug for saving files

conghuic23:
The clean scripts function is a async function, it may run after saving
launch scripts and remove all the launch scripts by accident.
So, add '.then' to make it work as a sync function.

Tracked-On: #7641


acrnsi:

NanlinXie (approved):

Issue #7562, opened 2022-05-24
[config_tool] vUART labels issues


izhouwu:
How Found: DX review
Found in Version acrn-2022w21.5-180000p-13-g84550004d

Please edit the following vUART labels:
1. Change legacy to Legacy
2. Change pci to PCI
3. Change VBDF to Virtual BDF:

izhouwu:
[External_System_ID] ACRN-9052

PR #7658, opened 2022-06-01
[release_3.0] config-tools: refine the vuart widget's text

izhouwu:
The labels and titles in vuart widget are hard-coded in the .vue file
rather than exposing the text from XSD file. This patch fixed it.

Tracked-On: #7562


acrnsi:

Issue #7657, opened 2022-06-01
[acrn-configuration-tool] Failed to build acrn with make BOARD=xx SCENARIO=shared RELEASE=y


zhangqiang4:
acrn-hypervisor: [1b7e8e393|https://github.com/projectacrn/acrn-hypervisor/commit/1b7e8e393fa87beb27e6eb6f8ebd2634fa00224e]   (branch: release_3.0)

*HW/Board*

     whl-ipc-i5

*Build link*

     Na

*Image info*

      Na

*Steps*
 - branch checkout to release_3.0 and update to latest
 - make clean && make BOARD=whl-ipc-i5 SCENARIO=shared RELEASE=y 

*Expected result*

    build successfully

*Actual result*

    failed to build， Cycle to print warning information：

WARNING:root:The sum 7.0 of memory configured in post launch VMs should not be larger than the calculated total hugepages 1.71875 of service VMs. Please update the configuration in post launch VMs
.....

*Reproduce rate*

     5/5

*Debugging info*

     Cannot to build acrn with release=y for all boards

Partial failure log fyi：

Board XML is fetched from /home/abc/workspace/mainline/acrn-hypervisor/misc/config_tools/data/whl-ipc-i5/whl-ipc-i5.xml
Scenario XML is being fetched from /home/abc/workspace/mainline/acrn-hypervisor/misc/config_tools/data/whl-ipc-i5/shared.xml
WARNING:root:The sum 7.0 of memory configured in post launch VMs should not be larger than the calculated total hugepages 1.71875 of service VMs. Please update the configuration in post launch VMs
/home/abc/workspace/mainline/acrn-hypervisor/build/hypervisor/configs/allocation.xml generated
/home/abc/workspace/mainline/acrn-hypervisor/build/hypervisor/configs/unified.xml generated
/home/abc/workspace/mainline/acrn-hypervisor/build/hypervisor/configs/config.mk generated
scripts/makefile/config.mk:191: The command line sets RELEASE to be 'y', but an existing build is configured with 'n'
scripts/makefile/config.mk:191: The configuration will be modified for RELEASE=y
WARNING:root:The sum 7.0 of memory configured in post launch VMs should not be larger than the calculated total hugepages 1.71875 of service VMs. Please update the configuration in post launch VMs
/home/abc/workspace/mainline/acrn-hypervisor/build/hypervisor/configs/allocation.xml generated
[...]

PR #7659, opened 2022-06-01
misc: config_tools: fix make deadloop on release build with debug sce…

zhangqiang4:
…nario

When I build a release acrn with a debug scenario like this:
$ BOARD=nuc11tnbi5 SCENARIO=shared make -j1 RELEASE=y

Following logs are repeated endlessly.

/home/abc/workspace/mainline/acrn-hypervisor/build/hypervisor/configs/allocation.xml generated
/home/abc/workspace/mainline/acrn-hypervisor/build/hypervisor/configs/unified.xml generated
/home/abc/workspace/mainline/acrn-hypervisor/build/hypervisor/configs/config.mk generated
scripts/makefile/config.mk:191: The command line sets RELEASE to be 'y', but an existing build is configured with 'n'
scripts/makefile/config.mk:191: The configuration will be modified for RELEASE=y

Because "BUILD_TYPE" was moved out of "DEBUG_OPTIONS" in scenario file,
outdated transform script fails to extract information there and it always
set 'CONFIG_RELEASE=n'.

If a makefile is included with a rule for it, the rule will be executed
first to generated new makefile. If it's updated, the make process will
be reexecuted with new makefile.

In the case above, build/hypervisor/configs/config.mk rule is executed every
time make tries to include it. Unhandled different between `RELEASE` and
`CONFIG_RELEASE` causes determine_build_type to update .scenario.xml which
is a dependency of hypervisor/scripts/makefile/config.mk.

Tracked-On: #7657
Signed-off-by: Qiang Zhang <qiang4.zhang@intel.com>

### DIFF
--- a/misc/config_tools/xforms/config_common.xsl
+++ b/misc/config_tools/xforms/config_common.xsl
@@ -34,6 +34,11 @@
       <xsl:with-param name="value" select="@scenario" />
     </xsl:call-template>
 
+    <xsl:call-template name="boolean-by-key-value">
+      <xsl:with-param name="key" select="'RELEASE'" />
+	  <xsl:with-param name="value" select="hv/BUILD_TYPE = 'release'" />
+    </xsl:call-template>
+
     <xsl:apply-templates select="hv/DEBUG_OPTIONS" />
     <xsl:apply-templates select="hv/FEATURES" />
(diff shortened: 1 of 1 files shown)


acrnsi:

Issue #7653, opened 2022-06-01
Default board_inspector.py output is very noisy/confusing


zihengL1:
The default output from running the board_inspector.py is very noisy and confusing.  Because of the volume of output, potentially important information is easily missed.  Recommend the default loglevel setting be "warning".

Even with "warning" the final status messages is still rather noisy

zihengL1:
[External_System_ID] ACRN-9116

PR #7665, opened 2022-06-01
board_inspector: refine board_inspector.py

zihengL1:
changed default log level into "warning",
and fixed errors in warning log summary.

Tracked-On: #7653

### DIFF
--- a/misc/config_tools/board_inspector/board_inspector.py
+++ b/misc/config_tools/board_inspector/board_inspector.py
@@ -104,39 +104,33 @@ def summary_loginfo(board_xml):
     critical_list = []
     log_line = open(str(tmpfile.name), "r", encoding='UTF-8')
     for line in log_line:
-        if "DEBUG" in line:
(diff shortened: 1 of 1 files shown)


acrnsi:

gvancuts (approved):

zihengL1:
> When writing out the summary of messages for each message category, would you first check if there are any of that category before writing out the heading. That would make the output much cleaner when there are no messages (the normal case). Something like this:
> 
> ```
>     if len(warning_list) != 0:
>        print("="*length)
>        print("\033[1;37mWARNING\033[0m")
>        print("These issues affect optional features. You can ignore them if they don't apply to you.\n")
>         for warning in warning_list:
>             print("\033[1;33m{0}\033[0m".format(warning.strip('\n')))
> ```

This part has been modified. Now, if there is no warning/error/critical in the log info, the log summary will not contain the warning/error/critical part.

NanlinXie (approved):

Issue #7661, opened 2022-06-01
config-tools: Change MCE, AC, and UC (GP) to enable


izhouwu:
How Found: DX review

For MCE_ON_PSC_DISABLED, ENFORCE_TURNOFF_AC, and ENFORCE_TURNOFF_GP: Currently, "y" = disable the feature. Please change these settings, so that "y" = enable the feature. Also, change the names to something like: MCE_ON_PSC_ENABLED, AC_ENABLED, and GP_ENABLED

izhouwu:
[External_System_ID] ACRN-8773

PR #7670, opened 2022-06-02
Change sense of some schema config options

dbkinder:
The MCE, split lock, and UC lock checkboxes have a unique behavior.
Selecting the checkbox disables the feature. We proposed changing the
behavior, so that the checkbox enables the feature, but the fix is
postponed to v3.1.  Change the sense of these option's tooltip and title
to reflect their current behavior (Disabling vs. normal Enabling).

Tracked-On: #7661


acrnsi:

NanlinXie (approved):

Issue #7679, opened 2022-06-02
uninitialized variables are used in ahci.c


liuhang-bit:
[External_System_ID] ACRN-9083

PR #7680, opened 2022-06-02
dm: Fix uninitialized variable in AHCI

liuhang-bit:
tfd is one data field of FIS for AHCI, it is unacceptable if it is
random value, could cause AHCI data transport failure.

Acturally there is no chance the switch statement has the value other
than 0~3, since there is only 2 bit to deduce.

Add the initial value for tfd as ATA_S_ERROR, which is checked by function 
ahci_write_fis_d2h afterwards, so guest driver could handle it well.

Tracked-On: #7679


acrnsi:

ywan170:
@yakuizhao

yakuizhao (approved):
LGTM

PR #7684, opened 2022-06-02
misc: add assert to check ivsh memory region name

weichenli-intel:
The current code allow the same region name for same VM, it's an issue

This patch add an assert to check the above issue and report it.

Tracked-On: #6690


acrnsi:

NanlinXie (approved):

PR #7693, opened 2022-06-06
hv: sched: fix bug when reboot vm

conghuic23:
BVT schedule rule:
When a new thread is wakeup and added to runqueue, it will get the
smallest avt (svt) from runqueue to initiate its avt. If the svt is
small than it's avt, it will keep the original avt.

For the reboot issue, when the VM is reboot, it means a new vcpu thread
is wakeup, but at this time, the Service VM's vcpu thread is blocked,
and removed from the runqueue, and the runqueue is empty, so the svt is
0. The new vcpu thread will get avt=0. avt=0 means very high priority,
and can run for a very long time until it catchup with other thread's
avt in runqueue.
At this time, when Service VM's vcpu thread wakeup, it will check the
svt, but the svt is very small, so will not update it's avt according to
the rule, thus has a very low priority and cannot be scheduled.

To fix it, adjust the new thread's avt when there is another thread in
runqueue.

### DIFF
--- a/hypervisor/common/sched_bvt.c
+++ b/hypervisor/common/sched_bvt.c
@@ -28,6 +28,7 @@ struct sched_bvt_data {
 	/* effective virtual time in units of mcu */
 	int64_t evt;
 	uint64_t residual;
+	bool unlimit;
 
 	uint64_t start_tsc;
 };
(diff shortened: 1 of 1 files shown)


acrnsi:

yakuizhao:
@lifeix , Will you please help to take a look at it?

yakuizhao (approved):
LGTM

lifeix (approved):

PR #7694, opened 2022-06-06
disallowed user set CAT CLOS_MASK to 0

Weiyi-Feng:
1. config_tools: delete the old board file when the new board file is imported successfully.
1. config_tools: disallowed user set CAT CLOS_MASK to 0
1. config_tools: CAT is no longer unchecked when CDP is unchecked
1. config_tools: fix vmNameChange event never called issue


acrnsi:

NanlinXie (approved):

PR #7696, opened 2022-06-06
misc: fix the conflict between S5 and LPC

weichenli-intel:
The current code use COM1 or COM2 for LPC and there is an conflict with
S5 feature.

This patch filter these COM Ports and fix the above issue.

Tracked-On: #6690


acrnsi:

NanlinXie (approved):

PR #7704, opened 2022-06-06
doc: Remove Launch Configuration Options

amyreye:
Removing Launch Configuation Options doc because all the launch script options are part of the new combined scenario XML and covered in Scenario Configuration Options doc.


acrnsi:

PR #7717, opened 2022-06-08
doc: parameterize hidden config option doc generation

dbkinder:
Add a configuration parameter for the xsltproc processing to decide
whether to include hidden options.


acrnsi:

Issue #7707, opened 2022-06-07
configurator: IVSHMEM region name lacks pattern check


yfliuuu:

yfliuuu:
[External_System_ID] ACRN-8909

PR #7718, opened 2022-06-08
misc: configurator: fix minor bug of IVSHMEM region name

yfliuuu:
IVSHMEM region name input value may be null after saving the scenario.
Add a null check to avoid marking empty input as correct.

Tracked-On: #7707


acrnsi:

NanlinXie (approved):

PR #7720, opened 2022-06-08
misc: fix the secure coding style violations

weichenli-intel:
There was some secure coding style violations of virtio net and tmp,
this patch add some NULL check to fix these violations.

Tracked-On: #6690


acrnsi:

devicemodel/hw/pci/virtio/virtio_net.c
```diff
@@ -940,8 +940,11 @@ virtio_net_init(struct vmctx *ctx, struct pci_vdev *dev, char *opts)
 		return -1;
 	}
 
-	vtopts = tmp = strdup(opts);
```
yakuizhao, line 943: Does the vtops/tmp need  to be initialized? Otherwise the uninitialized value will cause that  it has some issues in free(vtopts)

PR #7721, opened 2022-06-08
config tool: change vm_type to required field and hide in service vm UI

liuhang-bit:
hide vm_type in UI for service vm, and change it to
required field, since lack of it in xml may lead to
wrong inference for vm severity.

### DIFF
--- a/misc/config_tools/configurator/packages/configurator/src/pages/Config.vue
+++ b/misc/config_tools/configurator/packages/configurator/src/pages/Config.vue
@@ -211,7 +211,14 @@ export default {
         }
       }
       if (this.currentFormData['load_order'] === 'SERVICE_VM') {
-        delete this.currentFormSchema.BasicConfigType.properties.vm_type
+        if (this.currentFormData.hasOwnProperty('vm_type')) {
+          delete this.schemas.ServiceVM.BasicConfigType.properties.vm_type
+        } else {
+          if (this.schemas.ServiceVM.BasicConfigType.properties.hasOwnProperty('vm_type') === false) {
+            this.schemas.ServiceVM.BasicConfigType.properties.vm_type =
+              {$ref: '#/definitions/BasicVMType', title: 'VM type', description: '<p>Select the VM type. A standard VM (<span class=…ial features for time-sensitive applications.</p>'}
+          }
+        }
       }
     },
     switchTab(tabVMID) {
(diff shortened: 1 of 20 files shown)


acrnsi:

NanlinXie (approved):

misc/config_tools/configurator/packages/configurator/src/pages/Config.vue
```diff
@@ -211,7 +211,14 @@ export default {
         }
       }
       if (this.currentFormData['load_order'] === 'SERVICE_VM') {
-        delete this.currentFormSchema.BasicConfigType.properties.vm_type
+        if (this.currentFormData.hasOwnProperty('vm_type')) {
+          delete this.schemas.ServiceVM.BasicConfigType.properties.vm_type
```
dbkinder, line 215: @junjiemao1 Was this the right way to remove the visibility for an option? Shouldn't this be done in the .xsd file so the xml schema can be validated?

misc/config_tools/configurator/packages/configurator/src/pages/Config.vue
```diff
@@ -211,7 +211,14 @@ export default {
...
+        } else {
+          if (this.schemas.ServiceVM.BasicConfigType.properties.hasOwnProperty('vm_type') === false) {
+            this.schemas.ServiceVM.BasicConfigType.properties.vm_type =
+              {$ref: '#/definitions/BasicVMType', title: 'VM type', description: '<p>Select the VM type. A standard VM (<span class=…ial features for time-sensitive applications.</p>'}
```
dbkinder, line 219: These tooltip/descriptions should be pulled from the .xsd files and not maintained separately in these .vue files, right?  @junjiemao1 @NanlinXie

misc/config_tools/configurator/packages/configurator/src/pages/Config.vue
junjiemao1: Hardcoded logic in the configurator is always not preferred. However, the current XSD attributes do not provide a way to hide a config item conditionally (which is quite complex as conditional things are always so).

I would consider this as a temporary fix to the issue we see. A right fix may involve further discussions about the config items and ways we want to show to users, which can hardly catch 3.0.

misc/config_tools/configurator/packages/configurator/src/pages/Config.vue
junjiemao1: I'm not sure what this is for, but descriptions should always come from XSDs rather than being hardcoded.

PR #7723, opened 2022-06-08
misc: modify the HV start address

weichenli-intel:
The current code use a start address and hv_ram_size to find the start
address of acrn.bin, the start address is 32M which leaving too much
memory and will be load modules by grub, sometimes these modules will
override the HV log region.

This patch set the start address to 4M to avoid the above issue as much
as possible.

This algorithm depend on the board info, so the board XML should be
generate again if user change the SSRAM.

Tracked-On: #6690


acrnsi:

yakuizhao (approved):
LGTM

PR #7724, opened 2022-06-08
doc: update VM types description to use DX-friendly title

dbkinder:
Missed a use of the XML element name instead of the DX-friendly title in
the option description.  (For those paying attention, the default value
shown in the description also uses the XML element name, but we don't
have a convenient way to turn that string into the corresponding
acrn:title value.

Tracked-On: #5692


acrnsi:

yakuizhao (approved):
LGTM

PR #7741, opened 2022-06-10
Tweak scenario validation in the configurator

junjiemao1:
This patch series makes the following changes to the scenario validation in the configurator.

* Validation of scenario XML on load is relaxed to check only the essential structure assumed by the configurator.
* Validation of scenario XML on save is strengthened to check both syntactic and semantic issues.


acrnsi:

NanlinXie (approved):

PR #7743, opened 2022-06-10
misc: move the RDT interface to common libary

weichenli-intel:
The current RDT class and interface was define by the clos.py which is
mix get and merge RDT policy, create clos nodes.

This patch abstract the RDT to interface to common.

Tracked-On: #6690


acrnsi:

NanlinXie (approved):

PR #7748, opened 2022-06-13
misc: modify the default COM port of life_mngr_win

weichenli-intel:
The current life_mngr_win use COM2 which will be conflict with LPC, so
we change these COMs of S5 to COM3.

Tracked-On: #6690
Signed-off-by: Chenli Wei chenli.wei@intel.com


acrnsi:

Issue #7301, opened 2022-04-18
[Virtio-GPU]Not enough free memory reserved in SOS


NanlinXie:
Test step:

1: BOOT to ACRN and SOS 

2.check memory in SOS found the free memory is only more than 900M, and when log in local UI , system will crash;

Expect result: enough free memory for used in SOS and no crash happened 

Actual Result: No enough free memory in SOS and that sometimes cause crash when login SOS local UI

Fail Rate: 100%

NanlinXie:
[External_System_ID] ACRN-8696

dbkinder:
Please use "Service VM" instead of "SOS" :)

NanlinXie:
Closed with no regression

PR #7752, opened 2022-06-13
config_tools: update the hugepage algorithm

Kunhui-Li:
update the hugepages algorithm as the following steps.
1. calculate the total hugepages of service vm using the formula.
"total memory*0.98 - the memory consumed by pre-launched VMs - 5G
-300M*num(number of virtio gpu instance)".
2. calculate hugepage 1G based post-launched vm memory setting.
3. check if the memory configured in post launch VMs is not larger than
the total hugepages.

Compared with the previous version, this patch removes the 2M hugepage
and the correction value, also update the formula of total hugepage.

Tracked-On: #7301


acrnsi:

NanlinXie (approved):

PR #7753, opened 2022-06-13
config_tools: fix a module resolving issue in elementpath overlay

junjiemao1:
The library module introduced by commit 2cd13026e ("misc: move the RDT
interface to common library") was put under the library/ directory which is
not by default searchable from elementpath_overlay. This patch works around
the issue by adding that path to sys.path.

Also fix a typo in the RDT checking assertion.

Tracked-On: #6690


acrnsi:

PR #7760, opened 2022-06-14
Fix vsock and usb memory leak issue

liudlong:
1. Refine the vhost_vsock_devinit function to return void.
 2. Add the endpoint index value check log.


acrnsi:

ywan170 (approved):

PR #7763, opened 2022-06-14
ACRN: DM: Fix the MSI mask and unmask bugs.

liudlong:
The patch fix many bugs about the pci msi capabilities.
 1. Fix the clear msi mask bit issues.
 2. Initialize the offset of pci capabilities this variable will used as
 the output of pci_emul_find_capability function.
 3. Replace 16-read with 32-read because both mask and pending are 32bit.
 4. Add mask and pending element in msicap struct and intialize the struct
 value to zero, because we need r/w mask and pending value.

Tracked-On: #7762


acrnsi:

ywan170 (approved):

PR #7767, opened 2022-06-14
config_tools: remove VMX and VT-d invalid BIOS check in acrn-board-inspector

Kunhui-Li:
this patch removes VMX and VT-d invalid BIOS setting check in
acrn-board-inspector because the board inspector tool also adds
these checks (please refer to PR #7216) and may it's possible for user
to get the wrong information by these two check in acrn-board-inspector.

so we remove the repeated check in acrn-board-inspector.

Tracked-On: #6689


acrnsi:

NanlinXie (approved):

PR #7769, opened 2022-06-14
doc: Update GSG

amyreye:
- Update Configurator UI steps
- Emphasize connecting USB devices
- Use release_3.0 branch until tag is created
- Fix typos


acrnsi:

NanlinXie:
@amyreye , I think we could remove the session that guide user how to build configurator, and leave the link to be added, we will have this link soon, maybe next week.

NanlinXie (approved):

doc/getting-started/getting-started.rst
```diff
@@ -205,10 +186,30 @@ hardware.
 
 .. important::
 
-   Whenever you change the configuration of the board, such as peripherals, BIOS
-   settings, additional memory, or PCI devices, you must generate a new board
+   Before running the Board Inspector, you must set up your target hardware and
+   BIOS exactly as you want it, including connecting all peripherals,
+   configuring BIOS settings, and adding memory and PCI devices. For example,
+   you must connect all USB devices; otherwise, the Board Inspector will not
```
gvancuts, line 192: Does that mean the Board Inspector does not see the USB ports if nothing is plugged in? If true, that sounds like something we need to fix!

amyreye: @NanlinXie @dbkinder Does the wording accurately describe the USB issue that a customer encountered (USB ports?)? Also, should we create a JIRA ticket?

amyreye:
> @amyreye , I think we could remove the session that guide user how to build configurator, and leave the link to be added, we will have this link soon, maybe next week.

Done, please review

doc/getting-started/getting-started.rst
```diff
@@ -205,10 +186,30 @@ hardware.
...
+   detect the USB ports for passthrough. If you change the hardware or BIOS
```
dbkinder, line 193:
```suggestion
   detect the USB devices for passthrough. If you change the hardware or BIOS
```

doc/getting-started/getting-started.rst
```diff
@@ -205,10 +186,30 @@ hardware.
...
+   configuration, you must run the Board Inspector again to generate a new board
```
dbkinder, line 194:
```suggestion
   configuration, or add or remove USB devices, you must run the Board Inspector again to generate a new board
```

dbkinder (changes requested):

doc/getting-started/getting-started.rst
gvancuts: OK, I see that too on my testbed. I do not feel this is the correct behaviour however, as USB ports are given to a VM and it does (shouldn't) matter if there is any device plugged into the port (and what device it is). The correct behaviour in my opinion should be that **all** ports should show up in the drop-down menu (and the device plugged in could be indicated if there is any).

gvancuts (commented):
Please be aware that users going through this GSG will hit #7778 (leading to a system that will not boot). We need to fix this, either the bug itself (preferable but not sure if we have time for the 3.0 release) or at least by documenting it and its workaround.

doc/getting-started/getting-started.rst
amyreye: @gvancuts Create a ticket?

dbkinder:
> Please be aware that users going through this GSG will hit #7778 (leading to a system that will not boot). We need to fix this, either the bug itself (preferable but not sure if we have time for the 3.0 release) or at least by documenting it and its workaround.

Alternatively, in this GSG (and until #7778 is fixed), we should recommend that developers keep the default name assigned by the configurator when starting a new scenario (e.g., [shared | hybrid | hybrid_rt | partitioned]) and mention this is a limitation.  @nanlin @junjiemao1 Any chance we'd hold up the release to fix this since it's a rather serious error/limitation?

dbkinder (approved):

dbkinder:
> > > Please be aware that users going through this GSG will hit #7778 (leading to a system that will not boot). We need to fix this, either the bug itself (preferable but not sure if we have time for the 3.0 release) or at least by documenting it and its workaround.
> > 
> > 
> > Alternatively, in this GSG (and until #7778 is fixed), we should recommend that developers keep the default name assigned by the configurator when starting a new scenario (e.g., [shared | hybrid | hybrid_rt | partitioned])
> 
> The problem is that the default name is none of these, it's `scenario.xml` so unless we tell people to modify this, everyone will run into this. To make matter slightly worse, I did not see a way of doing that via the ACRN Configurator (the one I built from the `release_3.0` branch).

Ha!  That's silly. So we get the other changes into this doc no, I'm going to merge it and we can submit an update when we know the right thing to do (it's fixed or we document a workaround).

PR #7771, opened 2022-06-15
config_tools: fix the issue Service VM type is updated by UI automatically

Kunhui-Li:
Currently, if I update Pre-Launched VM type to real-time, select PCPU ID and
enable Real-time vCPU, then we update the name of sevice VM, finally, click the save button.
we will see the type of service vm is updated to RTVM. It is unexpected.

So this patch removes the logic of hiding service VM type, and only allows user
selecting Standard VM type to fix the issue that the service vm type is updated
by UI automatically.

Tracked-On: #6690


acrnsi:

PR #7786, opened 2022-06-17
misc: set the communication vUART default type to PCI

weichenli-intel:
The current code use the legacy as the default type of communication
vUART, these were only 4 standard vUART ports, others should config by
the /etc/serial.conf, so this patch change the default type to PCI
which could be used without config files.

Tracked-On: #6690
Signed-off-by: Chenli Wei chenli.wei@linux.intel.com


acrnsi:

junjiemao1 (approved):

NanlinXie (approved):

szhen11:
start to run premerge test

PR #7787, opened 2022-06-17
Mark CACHE_REGION as hidden (views="")

dbkinder:
CACHE_REGION is removed from the configurator interface but is still
showing in the generated documentation because the views="advanced" was
there.  Make that views="" to remove it from the UI and documentation.

Tracked-On: #5692

### DIFF
--- a/misc/config_tools/schema/config.xsd
+++ b/misc/config_tools/schema/config.xsd
@@ -285,7 +285,7 @@ Refer to :ref:`vuart_config` for detailed vUART settings.</xs:documentation>
       </xs:annotation>
     </xs:element>
     <xs:element name="CACHE_REGION" type="CacheRegionType" minOccurs="0">
-      <xs:annotation acrn:views="advanced">
+      <xs:annotation acrn:views="">
         <xs:documentation>Specify the cache setting.</xs:documentation>
       </xs:annotation>
     </xs:element>


acrnsi:

junjiemao1:
In fact "CACHE_REGION" is not **removed** from the configurator, but rather rendered as a **custom widget** which adopts a table-like interface and does not care about the "views" attribute.

While this change looks like an abuse of the "views" attribute, I think it still serve the purpose without impacting anything.

That said, the generated documentation still lacks the descriptions about the per-vCPU cache allocation settings. Are we fine to keep it absent for v3.0?

dbkinder:
@junjiemao1 i think you're saying we should keep this as views="advanced" and add the descriptions, right?

junjiemao1:
I think we can merge this for 3.0 for now. But for mainline, we need to think about how we want to document the CAT-configuring widget in the generated document (which seems not trivial).

dbkinder:
@junjiemao1 can you approve this PR then.  @NanlinXie can we merge this PR for 3.0 and we'll continue working on CAT configuring.

junjiemao1 (approved):

NanlinXie (approved):

szhen11:
start to run premerge test

szhen11:
start to run premerge test

dbkinder:
@NanlinXie Let's close this PR. We need to document CAT settings (somehow) instead of marking this as views=""

PR #7792, opened 2022-06-20
misc: add assert to check the missing of CLOS MASK

weichenli-intel:
The current UI there is an issue which have not sync and save the VMs
to RDT element, these issue will cause the missing of num_pclosids and
the HV can't start.

This patch add assert to check the the missing of CLOS MASK.
The UI issue will be fix by another patch.

Tracked-On: #6690

### DIFF
--- a/misc/config_tools/schema/checks/rdt_support.xsd
+++ b/misc/config_tools/schema/checks/rdt_support.xsd
@@ -53,4 +53,13 @@ This error cannot be fixed by adjusting the configuration. Report a `GitHub issu
     </xs:annotation>
   </xs:assert>
 
+  <xs:assert test="every $vm_name in //vm/name satisfies
+                  if (//RDT_ENABLED = 'y')
+                  then count(//POLICY/VM[text() = $vm_name]) > 0
+                  else true()">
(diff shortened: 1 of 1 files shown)


misc/config_tools/schema/checks/rdt_support.xsd
```diff
@@ -53,4 +53,13 @@ This error cannot be fixed by adjusting the configuration. Report a `GitHub issu
     </xs:annotation>
   </xs:assert>
 
+  <xs:assert test="every $vm_name in //vm/name satisfies
+                  if (//RDT_ENABLED = 'y')
+                  then count(//POLICY/VM[text() = $vm_name]) > 0
```
junjiemao1, line 58: Shall we check that all the vCPUs of that VM has a cache allocation policy defined if vCAT is not enabled for this VM?

weichenli-intel: Yes, we could.
We have talk these case and decide to simple it.
Now the CLOS MASK missing  of XMLs which generate by UI is some VM's all VCPUs have no CACHE config.

junjiemao1 (approved):

NanlinXie (approved):

acrnsi:

szhen11:
start to run premerge test

PR #7797, opened 2022-06-21
config_tools: fix dirty configform data issue

Weiyi-Feng:
config_tools: fix dirty configform data issue


acrnsi:

NanlinXie:
Set to on hold since we found data tampered with this patch.

PR #7799, opened 2022-06-21
misc: fix PCI vUART issue

weichenli-intel:
This patch add PCI config for PCI vUART and remove the PCI vUART config
of hv config which conflict with launched scenario Device ID.

Tracked-On: #6690

### DIFF
--- a/misc/config_tools/launch_config/launch_cfg_gen.py
+++ b/misc/config_tools/launch_config/launch_cfg_gen.py
@@ -55,6 +55,10 @@ class LaunchScript:
             next_vbdf = self._free_slots.pop(0)
             return next_vbdf
 
+        def remove_virtual_bdf(self, slot):
+            if slot in self._free_slots:
+                self._free_slots.remove(slot)
+
     class PassThruDeviceOptions:
         passthru_device_options = {
             "0x0200": [".//PTM[text()='y']", "enable_ptm"],  # Ethernet controller, added if PTM is enabled for the VM
@@ -165,6 +169,8 @@ class LaunchScript:
 
         if vbdf is None:
             vbdf = self._vbdf_allocator.get_virtual_bdf()
+        else:
+            self._vbdf_allocator.remove_virtual_bdf(vbdf)
         self.add_dynamic_dm_parameter("add_virtual_device", f"{vbdf} {kind} {options}")
 
     def add_passthru_device(self, bus, dev, fun, options=""):
@@ -253,6 +259,9 @@ def generate_for_one_vm(board_etree, hv_scenario_etree, vm_scenario_etree, vm_id
     # Emulated PCI devices
     script.add_virtual_device("hostbridge", vbdf="0:0")
 
+    #ivshmem and vuart must be the first virtual devices generated before the others except hostbridge and LPC
+    #ivshmem and vuart own reserved slots which setting by user
+
     for ivshmem in eval_xpath_all(vm_scenario_etree, f"//IVSHMEM_REGION[PROVIDED_BY = 'Device Model' and .//VM_NAME = '{vm_name}']"):
         script.add_virtual_device("ivshmem", options=f"dm:/{ivshmem.find('NAME').text},{ivshmem.find('IVSHMEM_SIZE').text}")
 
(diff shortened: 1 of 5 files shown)


acrnsi:

NanlinXie:
Waiting for @junjiemao1's review.

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -262,9 +262,14 @@ def generate_for_one_vm(board_etree, hv_scenario_etree, vm_scenario_etree, vm_id
...
-        if eval_xpath(conn, "./type/text()") == "pci":
-            script.add_virtual_device("uart", options=f"vuart_idx:{idx}")
+    for idx, conn in enumerate(eval_xpath_all(hv_scenario_etree, f"//vuart_connection[endpoint/vm_name/text() = '{vm_name}']"), start=1):
+        if eval_xpath(conn, f"./type/text()") == "pci":
+            vbdf = eval_xpath(conn, f"./endpoint[vm_name/text() = '{vm_name}']/vbdf/text()")
+            if vbdf is not None:
+                slot = int((vbdf.split(":")[1].split(".")[0]), 16)
+            else:
+                slot = None
+            script.add_virtual_device("uart", slot, options=f"vuart_idx:{idx}")
```
junjiemao1, line 272: You need also let the BDF allocator be aware of the slots consumed by the vuart so that other devices are not using the same.

misc/config_tools/static_allocators/gpa.py
```diff
@@ -109,15 +109,22 @@ def overlaps(self, other):
...
         devdict_32bits[(f"{VUART}_{vuart_id}", "bar0")] = PCI_VUART_VBAR0_SIZE
         devdict_32bits[(f"{VUART}_{vuart_id}", "bar1")] = PCI_VUART_VBAR1_SIZE
-    for vuart_id in communication_vuarts:
-        devdict_32bits[(f"{VUART}_{vuart_id}", "bar0")] = PCI_VUART_VBAR0_SIZE
-        devdict_32bits[(f"{VUART}_{vuart_id}", "bar1")] = PCI_VUART_VBAR1_SIZE
+
+    vm_name = common.get_node(f"//vm[@id = '{vm_id}']/name/text()", scenario_etree)
+    communication_vuarts = scenario_etree.xpath(f"//vuart_connection[endpoint/vm_name/text() = '{vm_name}']")
+    vuart_id = 1
+    for vuart in communication_vuarts:
```
junjiemao1, line 122: You can use `for vuart_id, vuart in enumerate(communication_vuarts, start=1)` to simplify the code.

misc/config_tools/launch_config/launch_cfg_gen.py
weichenli-intel: > You need also let the BDF allocator be aware of the slots consumed by the vuart so that other devices are not using the same.

These code used to make sure the BDF is the same with user 's setting for post launched.

The BDF settings not just from allocator, also from user setting, so we need add aware logic and unified assert to check the scenario .

junjiemao1: I'm not referring to the static allocator, but the VirtualBDFAllocator class in launch_cfg_gen.py. An object of that class is used to assign vBDF to the virtual devices of a VM, and we need to make sure that it never attempts to use one a user has already reserved for vuart.

misc/config_tools/xforms/vm_configurations.c.xsl
```diff
@@ -296,42 +296,27 @@
       <xsl:variable name="type" select="type/text()" />
       <xsl:variable name="vuart_id" select="position()"/>
       <xsl:value-of select="acrn:initializer(concat('vuart[', $vuart_id, ']'), '{', true())" />
-      <xsl:choose>
-        <xsl:when test="$type = 'legacy'">
+        <xsl:if test="$type = 'legacy'">
```
junjiemao1, line 299: You may want to re-indent the code block. The `xsl:if` is at the same level of the previous `xsl:value-of`.

junjiemao1 (changes requested):

misc/config_tools/launch_config/launch_cfg_gen.py
weichenli-intel: I have update the VirtualBDFAllocator.

misc/config_tools/static_allocators/gpa.py
weichenli-intel: OK

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -54,6 +54,9 @@ def get_virtual_bdf(self, device_etree=None, options=None):
 
             next_vbdf = self._free_slots.pop(0)
             return next_vbdf
+        def remove_virtual_bdf(self, slot):
```
junjiemao1, line 57: The convention requires one empty line before the definition of a method.

misc/config_tools/launch_config/launch_cfg_gen.py
```diff
@@ -165,6 +168,8 @@ def add_virtual_device(self, kind, vbdf=None, options=""):
 
         if vbdf is None:
             vbdf = self._vbdf_allocator.get_virtual_bdf()
+        else:
+            self._vbdf_allocator.remove_virtual_bdf(vbdf)
```
junjiemao1, line 172: If you choose to remove here, drop a comment in `generate_for_one_vm` highlighting that ivshmem and vuart must be the first virtual devices generated before the others except hostbridge and LPC which have their own reserved slots.

junjiemao1 (approved):

NanlinXie (approved):

PR #7800, opened 2022-06-22
config-tools: fix assertions against scenario XML

junjiemao1:
This patch fixes the following issues in the existing scenario schema
assertions:

  1. The assertion that ensures the existence of pCPU affinity in all
     pre- and post-launched VMs does not count the pCPU settings under a
     VM properly.

  2. The assertion that ensures the presence of CAT settings of a defined
     VM is no longer needed as it was originally designed to capture
     configurator issues.

Tracked-On: #6690


acrnsi:

NanlinXie (approved):

PR #7813, opened 2022-06-23
config_tools: remove minOccurs from items that have default values

junjiemao1:
With the default population stage today, the `default` and `minOccurs=0`
attributes are typically mutually exclusive:

  * A config item with a `default` attribute is guaranteed to exist after
    populating default values, making the `minOccurs` attribute meaningless.

  * A config item with a `minOccurs=0` attribute allows that item not to be
    present even after default values are populated.

This patch removes the `minOccurs=0` attributes from some config items that have
their default values recently added. This allows the default value populator to
fill in those items as expected if a user does not provide them explicitly.

Tracked-On: #6690


acrnsi:

NanlinXie (approved):

PR #7817, opened 2022-06-24
config_tools: mark os_config item as required in scenario XML

junjiemao1:
Following the same logic of commit 59c7077e2 ("config_tools: remove
minOccurs from items that have default values"), this patch marks the
config item os_config, which is a collection of guest OS settings, as
required since subitems of it either have default values or are optional.

Tracked-On: #6690


acrnsi:

NanlinXie (approved):

PR #7847, opened 2022-06-29
doc: update GSG with Configurator debian download

dbkinder:
Add instructions for downloading the configurator debian package and
installing it.


acrnsi:

PR #7858, opened 2022-06-30
doc: fix gsg instructions for building the acrn-kernel

dbkinder:
You need to be in the acrn-kernel directory in order to ``run make
distclean``

### DIFF
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -567,6 +567,7 @@ Build ACRN
 
       .. code-block:: bash
 
+         cd ~/acrn-work/acrn-kernel
          make distclean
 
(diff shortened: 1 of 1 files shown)


acrnsi:

doc/getting-started/getting-started.rst
```diff
@@ -567,6 +567,7 @@ Build ACRN
 
       .. code-block:: bash
 
+         cd ~/acrn-work/acrn-kernel
```
amyreye, line 570: Would you mind also deleting this lines 610/611:

Even though our sample default scenario defines six User VMs, we're only
going to launch one of them, so we'll only need the one launch script.

dbkinder: Got it.

Issue #7860, opened 2022-06-30
log format is not correct in acrn-dm


conghuic23:
acrn-dm boot log:

...
  pci deinit virtio-net
                        vtnet tx thread closing...
                                                  pci deinit virtio-blk
                                                                       pci deinit virtio-console
                                                                                                mngr_client_new: Failed to accept from fd 36, err: Invalid argument

...

conghuic23:
[External_System_ID] ACRN-9252

PR #7861, opened 2022-06-30
dm: virtio-console: change STDIO BE attribute

conghuic23:
Fix the log format problem in devicemodel when using virtio-console with
'stdio' configuration.
In virtio-console, it will change the STDIO's attributes, and will case
the log message after that not align.
e.g:
	for (i=0; i<3; i++)
		printf('test\n');

	The original output would be like this:

	test
	    test
	        test

	After add 'OPOST' to oflag of STDIO, the output would be like this:

	test
	test
	test

	(OPOST - Enable implementation-defined output processing)

Tracked-On: #7860


acrnsi:

ywan170 (approved):

PR #7863, opened 2022-06-30
doc: remove images no longer referenced

dbkinder:
We deleted documents but didn't delete the images they used.


acrnsi:

Issue #7864, opened 2022-07-01
config_tools: hide required checkbox's label


YuanyuanZhao609:
[External_System_ID] ACRN-9154

PR #7865, opened 2022-07-01
config_tools: hide required checkbox's label

YuanyuanZhao609:
Required fields with a checkbox will potentially cause confusion.
So hide the label when display a checkbox.

Tracked-On: #7864
Signed-off-by: Yuanyuan Zhao <yuanyuan.zhao@linux.intel.com>

### DIFF
--- a/misc/config_tools/configurator/packages/vue-json-schema-form/vue3/vue3-core/src/components/Widget.js
+++ b/misc/config_tools/configurator/packages/vue-json-schema-form/vue3/vue3-core/src/components/Widget.js
@@ -291,7 +291,7 @@ export default {
                         label: () => h('span', {
                             class: {
                                 genFormLabel: true,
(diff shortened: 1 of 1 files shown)


acrnsi:

junjiemao1:
How about drop-down boxes? Shall we remove the asterisk before such widgets as well?

YuanyuanZhao609:
> How about drop-down boxes? Shall we remove the asterisk before such widgets as well?

Added in new version.

junjiemao1:
> > How about drop-down boxes? Shall we remove the asterisk before such widgets as well?
> 
> Added in new version.

Perhaps you pushed to another branch? I still see the proposed change checking checkboxes only.

junjiemao1 (approved):

NanlinXie (approved):

PR #7866, opened 2022-07-02
misc: refine the assert of duplicate IO port

weichenli-intel:
The current code judges an empty vm_name as a normal vm_name and assert
two endpoints with null name use the same IO port.

There was another which check wether the vm_name of endpoint is null, so
this patch break the assert if get null vm_name.

Tracked-On: #6690

### DIFF
--- a/misc/config_tools/schema/checks/vuart_config.xsd
+++ b/misc/config_tools/schema/checks/vuart_config.xsd
@@ -14,7 +14,7 @@
     </xs:annotation>
   </xs:assert>
 
-  <xs:assert test="every $io_port in /acrn-config//endpoint/io_port satisfies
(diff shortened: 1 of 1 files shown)


acrnsi:

misc/config_tools/schema/checks/vuart_config.xsd
```diff
@@ -15,7 +15,9 @@
   </xs:assert>
 
   <xs:assert test="every $io_port in /acrn-config//endpoint/io_port satisfies
```
junjiemao1, line 17: You can apply the check to the XPATH which `io_port` enumerates, i.e., `/acrn-config//endpoint[vm_name != '']/io_port`

Also you may want to update the XPATH used in `count` as well since you do not want to count ports for VMs without a name.

junjiemao1 (approved):

NanlinXie (approved):

PR #7868, opened 2022-07-04
config_tools: apply pattern check to virtio devices

Kunhui-Li:
Apply pattern checks to network interface name and virtio input id. Report error if regex doesn't match.

### DIFF
--- a/misc/config_tools/configurator/packages/configurator/src/pages/Config/ConfigForm/CustomWidget/Virtio/Input.vue
+++ b/misc/config_tools/configurator/packages/configurator/src/pages/Config/ConfigForm/CustomWidget/Virtio/Input.vue
@@ -17,7 +17,10 @@
             Guest virtio input device unique identifier:
           </b-col>
           <b-col md="4">
-            <b-form-input v-model="input.id"/>
+            <b-form-input :state="validateId(input.id)" v-model="input.id" placeholder="An arbitrary-long string with letters, digits, underscores or dashes."/>
+            <b-form-invalid-feedback>
+                An arbitrary-long string with letters, digits, underscores or dashes.
+            </b-form-invalid-feedback>
           </b-col>
         </b-row>
       </div>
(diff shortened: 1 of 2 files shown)


acrnsi:

junjiemao1:
Why do we have to add the regex patterns in vue files rather than in the XML schema?

Kunhui-Li:
> Why do we have to add the regex patterns in vue files rather than in the XML schema?

The regex pattern has been added to the XML schema. But it doesn't work for the customized widgets.
So this patch adds pattern checks similar to the ivshmem in vue files.

junjiemao1:
> > Why do we have to add the regex patterns in vue files rather than in the XML schema?
> 
> The regex pattern has been added to the XML schema. But it doesn't work for the customized widgets. So this patch adds pattern checks similar to the ivshmem in vue files.

So this is to enable real-time checks of the text pattern?

Kunhui-Li:
> > > Why do we have to add the regex patterns in vue files rather than in the XML schema?
> > 
> > 
> > The regex pattern has been added to the XML schema. But it doesn't work for the customized widgets. So this patch adds pattern checks similar to the ivshmem in vue files.
> 
> So this is to enable real-time checks of the text pattern?

Yes.

junjiemao1 (approved):

szhen11:
@NanlinXie Please help to approve this PR. Thanks.

NanlinXie (approved):

dbkinder:
How do we guarantee that the duplicated regex patterns in vue and xsd files are kept in sync?  Better is to have a single source for this regex pattern (and the tooltip documentation) in the xsd files.

junjiemao1:
> How do we guarantee that the duplicated regex patterns in vue and xsd files are kept in sync? Better is to have a single source for this regex pattern (and the tooltip documentation) in the xsd files.

This patch does not duplicate the regex pattern; it fetches the pattern from the json schema which is generated from the XSD files.

The error message (b-form-invalid-feedback) is duplicated, though.

Issue #7880, opened 2022-07-08
fail to run acrn-dm with wrong iasl path


shiqingg:
If user's host machine has a special iasl version at specific folder and could be indexed by "which"
e.g. $ which iasl
/home/xxx/bin/iasl

then acrn-dm will fail to run in the case no "/home/xxx/bin/iasl" is found in target machine.

shiqingg:
[External_System_ID] ACRN-9254

PR #7881, opened 2022-07-08
Refine the detection and the version verification of "iasl"

shiqingg:
ACRN has dependency on "iasl" compiler both at build time and at run time.
 - At build time (on the *dev* machine),
   config_tools depends on "iasl" to generate the binary of ACPI tables for
   pre-launched VMs.
 - At run time (on the *target* machine),
   acrn-dm depends on "iasl" to build the ACPI tables for post-launched VMs.

Current implementation may cause trouble when the "iasl" path is different
between the *dev* machine and the *target* machine. Because ASL_COMPILER
is determined on the *dev* machine and passed to acrn-dm, while
acrn-dm is running on the *target* machine.

This patch series refine the detection and the version verification of "iasl".
1. Detect the "iasl" compiler
   - At build time (on the *dev* machine),
     config_tools uses ASL_COMPILER as the path to the "iasl" compiler.
     By default, ASL_COMPILER is initialized by "which iasl".
     User could easily override it by specifying ASL_COMPILER as the
     build option, like below:
       make BOARD=xxxx SCENARIO=yyyy ASL_COMPILER=/usr/local/bin/iasl

   - At run time (on the *target* machine),
     if "--iasl <iasl_compiler_path>" is specified as the acrn-dm parameter,
     acrn-dm uses <iasl_compiler_path> as the path to the "iasl" compiler;
     otherwise, "which iasl" is used to detect the "iasl" compiler.

2. Verify the "iasl" version
   To avoid hardcoding the minimum "iasl" version in multiple places,
   define IASL_MIN_VER in the top-level Makefile, and pass it to
   config_tools and acrn-dm.
   Then, config_tools and acrn-dm verify whether the "iasl" version is newer
   than IASL_MIN_VER at build time and run time respectively.

v2 -> v3:
 - use "strlen" rather than "strncmp" to check whether asl_compiler
   has been set or not

v1 -> v2:
 - remove "iasl_param" and "with_iasl_param" to simplify the logic
 
Tracked-On: #7880


ywan170 (approved):

acrnsi:

dbkinder (approved):
Doc changes LGTM

doc/reference/hv-make-options.rst
```diff
@@ -34,8 +34,8 @@ The following table shows ACRN-specific command-line options:
        is ``n``.)
 
    * - ``ASL_COMPILER``
-     - Optional. Use an ``iasl`` compiler that is not in the default path
-       (``/usr/sbin``).
+     - Optional. Specify the path to ``iasl`` compiler on the ``dev`` machine.
```
gvancuts, line 37:
```suggestion
     - Optional. Specify the path to the ``iasl`` compiler on the development machine.
```

doc/reference/hv-make-options.rst
```diff
@@ -34,8 +34,8 @@ The following table shows ACRN-specific command-line options:
...
+       (Default value is detected by ``which iasl``.)
```
gvancuts, line 38:
```suggestion
       (If not provided, the default value is derived from ``which iasl``.)
```

doc/user-guides/acrn-dm-parameters.rst
```diff
@@ -325,6 +325,21 @@ Here are descriptions for each of these ``acrn-dm`` command line parameters:
 
       --ssram
 
+----
+
+``--iasl <iasl_compiler_path>``
+   Specify the path to ``iasl`` compiler on the ``target`` machine.
```
gvancuts, line 331:
```suggestion
   Specify the path to ``iasl`` compiler on the target machine.
```

doc/user-guides/acrn-dm-parameters.rst
```diff
@@ -325,6 +325,21 @@ Here are descriptions for each of these ``acrn-dm`` command line parameters:
...
+
+   If ``--iasl <iasl_compiler_path>`` is specified as the acrn-dm parameter,
```
gvancuts, line 333:
```suggestion
   If ``--iasl <iasl_compiler_path>`` is specified as the ``acrn-dm`` parameter,
```

doc/user-guides/acrn-dm-parameters.rst
```diff
@@ -325,6 +325,21 @@ Here are descriptions for each of these ``acrn-dm`` command line parameters:
...
+   acrn-dm uses <iasl_compiler_path> as the path to the ``iasl`` compiler;
+   otherwise, ``which iasl`` is used to detect the ``iasl`` compiler.
```
gvancuts, line 335:
```suggestion
   otherwise, ``which iasl`` is used to detect where the ``iasl`` compiler is located.
```

doc/user-guides/acrn-dm-parameters.rst
```diff
@@ -325,6 +325,21 @@ Here are descriptions for each of these ``acrn-dm`` command line parameters:
...
+
+   usage::
+
+      --iasl /usr/local/bin/iasl
+
+   uses ``/usr/local/bin/iasl`` as the path to ``iasl`` compiler
```
gvancuts, line 341:
```suggestion
   uses ``/usr/local/bin/iasl`` as the path to the ``iasl`` compiler
```

hypervisor/Makefile
```diff
@@ -11,6 +11,8 @@ GCC_MINOR=$(shell echo __GNUC_MINOR__ | $(CC) -E -x c - | tail -n 1)
 #enable stack overflow check
 STACK_PROTECTOR := 1
 
+ASL_COMPILER ?= $(shell which iasl)
```
gvancuts, line 14: We should add a check somewhere to make sure `ASL_COMPILER` is initialized to a value. In the event it is not (basically in case `iasl` is not on the dev machine) then the compilation will fail with the following error message which is confusing to the user:
```
...
VM configuration check pass!
generate the binary of ACPI tables for pre-launched VMs ...
python3 ../misc/config_tools/acpi_gen/bin_gen.py --board /home/gvancuts/work/acrn-hypervisor/build/hypervisor/.board.xml --scenario /home/gvancuts/work/acrn-hypervisor/build/hypervisor/.scenario.xml --asl /home/gvancuts/work/acrn-hypervisor/build/hypervisor/configs --out /home/gvancuts/work/acrn-hypervisor/build/hypervisor --iasl_path  --iasl_min_ver 20190703
usage: python3 bin_gen.py --board [board] --scenario [scenario] --iasl_path [the path to the iasl compiler] --iasl_min_ver [the minimum iasl version][ --out [output dir of acpi ASL code]]
bin_gen.py: error: argument --iasl_path: expected one argument
make[1]: *** [Makefile:424: pre_build] Error 2
make[1]: Leaving directory '/home/gvancuts/work/acrn-hypervisor/hypervisor'
make: *** [Makefile:117: hypervisor] Error 2
```

The user will see that he is missing the `--iasl_path` argument but that is an internal thing to the Makefile. What he really needs to do it install `iasl` and/or provide the path to it but using the `ASL_COMPILER` variable (not the `--iasl_path` argument).

devicemodel/core/main.c
```diff
@@ -1020,6 +1028,18 @@ main(int argc, char *argv[])
...
 	}
+
+	if (get_iasl_compiler() != 0) {
+		pr_err("Cannot find Intel ACPI ASL compiler tool \"iasl\".\n");
+		exit(1);
+	}
+
+	if (check_iasl_version() != 0) {
+		pr_err("Please install iasl tool with version >= %s from https://www.acpica.org/downloads \n",
+			IASL_MIN_VER);
```
gvancuts, line 1039: How about being more verbose here and also tell the user that he should install the min version and provide the path to it if it's not on the PATH?

We could also run into the situation whereby `iasl` is on the system (was installed via the `acpica-tools` package) and another version compiled by the user is also installed but in a different path (e.g. `/usr/sbin` vs. `/usr/bin`). In that case, the error message can be confusing because it will complain that the version is too old since it's only looking at one of them.

gvancuts (changes requested):

hypervisor/Makefile
shiqingg: I have added a check to make sure ASL_COMPILER is initialized to a value.

devicemodel/core/main.c
shiqingg: I have updated the message.

ywan170 (approved):

gvancuts (approved):

NanlinXie (approved):

lifeix (approved):

PR #7888, opened 2022-07-12
hv: Change sched_event back to boolean-based implementation

yfliuuu:
Commit d575edf79a9b8dcde16574e28c213be7013470c7 changes the internal
    implementation of wait_event and signal_event to use a counter instead
    of a boolean value.
    
    The background was:
    ACRN utilizes vcpu_make_request and signal_event pair to shoot down
    other vcpus and let them wait for signals. vcpu_make_request eventually
    leads to target vcpu calling wait_event.
    
    However vcpu_make_request/signal_event pair was not thread-safe,
    and concurrent calls of this pair of API could lead to problems.
    One such example is the concurrent wbinvd emulation, where vcpus may
    concurrently issue vcpu_make_request/signal_event to synchronize wbinvd
    emulation.
    
    d575edf commit uses a counter in internal implementation of
    wait_event/signal_event to avoid data races.
    
    However by using a counter, the wait/signal pair now carries semantics of
    semaphores instead of events. Semaphores require caller to carefully
    plan their calls instead of multiply signaling any number of times to the same
    event, which deviates from the original "event" semantics.
    
    This patch changes the API implementation back to boolean-based, and
    re-resolve the issue of concurrent wbinvd in next patch.
    
    This also partially reverts commit 10963b04d127c3a321b30467b986f13cbd09db66,
    which was introduced because of the d575edf.

### DIFF
--- a/hypervisor/arch/x86/guest/lock_instr_emul.c
+++ b/hypervisor/arch/x86/guest/lock_instr_emul.c
@@ -60,25 +60,7 @@ void vcpu_complete_lock_instr_emulation(struct acrn_vcpu *cur_vcpu)
 	if (cur_vcpu->vm->hw.created_vcpus > 1U) {
 		foreach_vcpu(i, cur_vcpu->vm, other) {
 			if ((other != cur_vcpu) && (other->state == VCPU_RUNNING)) {
-				/*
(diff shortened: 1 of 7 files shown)


acrnsi:

yfliuuu:
@dongyaozu Hi Eddie, could you please help approve this PR?

Issue #7876, opened 2022-07-06
ACRN shall report TSC frequency to vCPUs


chejianj:
ACRN shall report TSC frequency to vCPUs. The interfaces for reporting varies along with the guest kernels.

For VMs running Linux, the hypervisor shall report TSC frequency (in KHz) in CPUID leaf 40000010H.EAX and the Linux kernel shall use this value rather than trying to calibrate.
For VMs running Windows, the hypervisor shall report TSC frequency in MSR 40000022H, as is required by Microsoft Hypervisor TLFS specification.

chejianj:
[External_System_ID] ACRN-9244

PR #7890, opened 2022-07-14
hv: tsc: calibrate TSC by HPET

chejianj:
On some platforms CPUID.0x15:ECX is zero and CPUID.0x16 can
only return the TSC frequency in MHZ which is not accurate.
For example the TSC frequency obtained by CPUID.0x16 is 2300
MHZ and the TSC frequency calibrated by HPET is 2303.998 MHZ
which is much closer to the actual TSC frequency 2304.000 MHZ.
This patch adds the support of using HPET to calibrate TSC
when HPET is available and CPUID.0x15:ECX is zero.

v3->v4:
  - move calc_tsc_by_hpet into hpet_calibrate_tsc

v2->v3:
  - remove the NULL check in hpet_init
  - remove ""& 0xFFFFFFFFU" in tsc_read_hpet
  - add comment for the counter wrap in the low 32 bits in
    calc_tsc_by_hpet
  - use a dedicated function for hpet_calibrate_tsc

v1->v2:
  - change native_calibrate_tsc_cpuid_0x15/0x16 to
    native_calculate_tsc_cpuid_0x15/0x16
  - move hpet_init to BSP init
  - encapsulate both HPET and PIT calibration to one function
  - revise the commit message with an example"

Tracked-On: #7876

### DIFF
--- a/hypervisor/arch/x86/cpu.c
+++ b/hypervisor/arch/x86/cpu.c
@@ -235,6 +235,9 @@ void init_pcpu_post(uint16_t pcpu_id)
 		/* Print Hypervisor Banner */
 		print_hv_banner();
 
+		/* Initialie HPET */
+		hpet_init();
+
 		/* Calibrate TSC Frequency */
 		calibrate_tsc();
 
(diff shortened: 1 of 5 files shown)


acrnsi:

lifeix (approved):

Issue #7902, opened 2022-07-20
coding style cleanup in pci.c


yonghuah:
validate inputs explicitly, instead of using ASSERT.

yonghuah:
[External_System_ID] ACRN-9201

PR #7903, opened 2022-07-20
hv: validate inputs in vpci_mmio_cfg_access

yonghuah:
Validate inputs explicitly instead of using ASSERT, which
 works only for debug build.

Tracked-On: #7902

### DIFF
--- a/hypervisor/dm/vpci/vpci.c
+++ b/hypervisor/dm/vpci/vpci.c
@@ -192,9 +192,16 @@ static int32_t vpci_mmio_cfg_access(struct io_request *io_req, void *private_dat
 	bdf.value = (uint16_t)((address - pci_mmcofg_base) >> 12U);
 
 	if (mmio->direction == ACRN_IOREQ_DIR_READ) {
-		ret = vpci_read_cfg(vpci, bdf, reg_num, (uint32_t)mmio->size, (uint32_t *)&mmio->value);
+		uint32_t val = ~0U;
+
+		if (pci_is_valid_access(reg_num, (uint32_t)mmio->size)) {
+			ret = vpci_read_cfg(vpci, bdf, reg_num, (uint32_t)mmio->size, &val);
+		}
+		mmio->value = val;
 	} else {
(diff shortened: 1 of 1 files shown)


acrnsi:

hypervisor/hw/pci.c
```diff
@@ -177,23 +177,27 @@ static inline uint32_t mmcfg_off_to_address(union pci_bdf bdf, uint32_t offset)
...
+			break;
+		case 2U:
+			val = (uint32_t)mmio_read16(hva);
+			break;
+		default:
+			val = mmio_read32(hva);
+			break;
+		}
+	} else {
+		pr_err("the offset should be aligned with 2/4 byte\n");
```
dbkinder, line 200: Will this message make sense to the developer without any context, such as the name of the function being called?

hypervisor/hw/pci.c
```diff
@@ -207,20 +211,26 @@ static uint32_t pci_mmcfg_read_cfg(union pci_bdf bdf, uint32_t offset, uint32_t
...
+			break;
+		case 2U:
+			mmio_write16((uint16_t)val, hva);
+			break;
+		default:
+			mmio_write32(val, hva);
+			break;
+		}
+	} else {
+		pr_err("the offset should be aligned with 2/4 byte\n");
```
dbkinder, line 233: Same question.  Will this message make any sense without more context?

yonghuah: @dbkinder  thank you, after re-think this fix, seems need to double check and validate these inputs in the caller of these functions, if so, we don't need to break the @pre of them.

so, pls hold on this PR for now, sorry for this confusing.

hypervisor/hw/pci.c
dongyaozu, line 200: If this error really happens, pr_err is insufficient.  We need to let the guest/VCPU die rather than the HV/PCPU die.

yonghuah:
@dongyaozu  Unlock this PR and pls review, thanks.

hypervisor/hw/pci.c
yonghuah: will follow the same style in PIO PCI Config READ/Write functions.

lifeix (approved):

Issue #7820, opened 2022-06-24
Create an "ACRN Sample App"


gvancuts:
Create an "ACRN Sample App" as described on this [wiki page](https://github.com/projectacrn/acrn-hypervisor/wiki/ACRN-Sample-App).

NanlinXie:
[External_System_ID] ACRN-9246

PR #7936, opened 2022-07-29
SampleApplication: Add Makefile for the entire SampleApplication

liudlong:
Add the Makefile in the SampleApplication root directory. User
don't need to separate build the userapp and rtapp, just need make
once at the SampleApplication root directory.

Tracked-On #7820

### DIFF
--- a/misc/sample_application/Makefile
+++ b/misc/sample_application/Makefile
@@ -1,0 +1,18 @@
+T := $(CURDIR)
+OUT_DIR ?= $(shell mkdir -p $(T)/build;cd $(T)/build/;pwd)
+
+.PHONY: all userapp rtapp
+all: userapp histapp rtapp
+
+userapp:
+	$(MAKE) -C $(T)/uservm OUT_DIR=$(OUT_DIR)
+histapp:
+	cp $(T)/uservm/histapp.py $(OUT_DIR)
+rtapp:
+	$(MAKE) -C $(T)/rtvm OUT_DIR=$(OUT_DIR)
+
+.PHONY: clean
+
+clean:
+	rm -rf $(OUT_DIR)
+
--- a/misc/sample_application/rtvm/Makefile
+++ b/misc/sample_application/rtvm/Makefile
@@ -1,12 +1,14 @@
 CC = gcc
-CP = g++
+CXX ?= g++
+T := $(CURDIR)
+OUT_DIR ?= $(shell mkdir -p $(T)i/../build;cd $(T)/../build;pwd)
 
 CFLAGS = -Wall -Wextra -Wabi=11 -pedantic
 
 LDLIBS = -lrt -pthread
 
 all: rtApp.c
-	$(CC) $(CFLAGS) -o rtapp rtApp.c ivshmemlib.c
+	$(CXX) $(CFLAGS) -o $(OUT_DIR)/rtApp rtApp.c ivshmemlib.c
 
 clean:
-	rm rtapp
+	rm $(OUT_DIR)/rtApp
--- a/misc/sample_application/uservm/Makefile
+++ b/misc/sample_application/uservm/Makefile
@@ -1,12 +1,19 @@
-CC = gcc
-CP = g++
+CC ?= gcc
+CXX ?= g++
+T := $(CURDIR)
+OUT_DIR ?= $(shell mkdir -p $(T)i/../build;cd $(T)/../build;pwd)
 
 CFLAGS = -Wall -Wextra -Wabi=11 -pedantic
 
 LDLIBS = -lrt -pthread
 
-all: userApp.cpp
-	$(CP) $(CFLAGS) -o userapp userApp.cpp ivshmemlib.c $(LDLIBS)
+all: userApp histapp
+
+userApp:
+	$(CXX) $(CFLAGS) -o $(OUT_DIR)/userApp userApp.cpp ivshmemlib.c $(LDLIBS)
+histapp:
+	cp $(T)/histapp.py $(OUT_DIR)
 
 clean:
-	rm userapp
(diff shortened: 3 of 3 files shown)


misc/sample_application/Makefile
```diff
@@ -0,0 +1,18 @@
+T := $(CURDIR)
+OUT_DIR ?= $(shell mkdir -p $(T)/build;cd $(T)/build/;pwd)
+
+.PHONY: all userapp rtapp
+all: userapp rtapp
+
+userapp:
+	cp $(T)/uservm/histapp.py $(OUT_DIR)
+	$(MAKE) -C $(T)/uservm  -o=$(OUT_DIR)
```
junjiemao1, line 9: Do you want to pass `OUT_DIR` to the sub-makefiles? `-o` is not for that purpose.

You should use `OUT_DIR=$(OUT_DIR)` here and use `$(OUT_DIR)` rather than hardcoding `../build` in the sub-makefiles.

liudlong: Re-push the new modification about the OUT_DIR.

junjiemao1: You are still not using `$(OUT_DIR)` in the sub-makefiles, which means passing `OUT_DIR` to them has no effect at all.

misc/sample_application/rtvm/Makefile
```diff
@@ -6,7 +6,7 @@ CFLAGS = -Wall -Wextra -Wabi=11 -pedantic
 LDLIBS = -lrt -pthread
 
 all: rtApp.c
-	$(CC) $(CFLAGS) -o rtapp rtApp.c ivshmemlib.c
+	$(CC) $(CFLAGS) -o ../build/rtapp rtApp.c ivshmemlib.c
```
gvancuts, line 9: Should be using `$(OUT_DIR)/rtapp` here instead I guess?

misc/sample_application/uservm/Makefile
```diff
@@ -6,7 +6,7 @@ CFLAGS = -Wall -Wextra -Wabi=11 -pedantic
 LDLIBS = -lrt -pthread
 
 all: userApp.cpp
-	$(CP) $(CFLAGS) -o userapp userApp.cpp ivshmemlib.c $(LDLIBS)
+	$(CP) $(CFLAGS) -o ../build/userapp userApp.cpp ivshmemlib.c $(LDLIBS)
```
gvancuts, line 9: Ditto

gvancuts (changes requested):
For that `$(OUT_DIR)` thing to work, you'd also want a specific target added to the top-level Makefile (i.e. `acrn-hypervisor/Makefile`), so the user would run `make sample_application` from the top-level folder and build those.

misc/sample_application/Makefile
liudlong: 1 . Modify the Makefile in the parent directory pass the OUT_DIR to the sub-makefies.
2. Modify the Makefile in the sub-directory in case the user just run the make command in the sub-directory.
Fore update the commit in this PR

misc/sample_application/rtvm/Makefile
liudlong: 1 . Modify the Makefile in the parent directory pass the OUT_DIR to the sub-makefies.
2. Modify the Makefile in the sub-directory in case the user just run the make command in the sub-directory.
Fore update the commit in this PR

misc/sample_application/uservm/Makefile
liudlong: @gvancuts 
1 . Modify the Makefile in the parent directory pass the OUT_DIR to the sub-makefies.
2. Modify the Makefile in the sub-directory in case the user just run the make command in the sub-directory.
Fore update the commit in this PR

misc/sample_application/uservm/Makefile
```diff
@@ -1,12 +1,16 @@
-CC = gcc
-CP = g++
+CC ?= gcc
+CP ?= g++
```
junjiemao1, line 2: The typical name for g++ is `CXX`, not `CP`.

liudlong: > The typical name for g++ is `CXX`, not `CP`.

 Replace the "CP ?= g++" with "`CXX`?=g++" . thanks

misc/sample_application/uservm/Makefile
```diff
@@ -1,12 +1,16 @@
...
+OUT_DIR ?= $(shell mkdir -p $(T)i/../build;cd $(T)/../build;pwd)
 
 CFLAGS = -Wall -Wextra -Wabi=11 -pedantic
 
 LDLIBS = -lrt -pthread
 
 all: userApp.cpp
-	$(CP) $(CFLAGS) -o userapp userApp.cpp ivshmemlib.c $(LDLIBS)
+	$(CXX) $(CFLAGS) -o $(OUT_DIR)/userApp userApp.cpp ivshmemlib.c $(LDLIBS)
+	cp $(T)/histapp.py $(OUT_DIR)
```
junjiemao1, line 12: I don't remember you have this `cp histapp.py` command previously. Is it added intentionally? The Makefile you added under misc/sample_application does the same thing.

If it is by intention, please split the commands into two goals, one generating `$(OUT_DIR)/userApp` while the other `$(OUT_DIR)/histapp.py`. The goal `all` shall depend on those two output files rather than do everything by itself.

The idea is that if a Makefile should not rebuild something that has not changed since its last build.

junjiemao1 (changes requested):

junjiemao1 (approved):
Looks good to me. Thanks!

NanlinXie (approved):

NanlinXie:
@gvancuts , we will merge this patch now, please be free to let us know if you have further concern.

PR #7971, opened 2022-08-10
doc: update hypervisor shell usage

mgcao:
ACRN hypervisor shell has been improved to be bash-like shell, so update
its usage


dbkinder (commented):
LGTM, thanks.

Issue #7975, opened 2022-08-11
GSG python package xmlschema/elementpath incompatibility


zihengL1:
On a clean Ubuntu 20.04.2 installation, I followed the GSG, and while installing the ACRN pre-requisites I got this error:

david@X220-20:~$ sudo pip3 install "elementpath==2.5.0" lxml xmlschema defusedxml tqdm
Collecting elementpath==2.5.0
  Downloading elementpath-2.5.0-py3-none-any.whl (181 kB)
     |████████████████████████████████| 181 kB 1.2 MB/s 
Collecting lxml
  Downloading lxml-4.9.1-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.manylinux_2_24_x86_64.whl (6.9 MB)
     |████████████████████████████████| 6.9 MB 9.7 MB/s 
Collecting xmlschema
  Downloading xmlschema-2.0.1-py3-none-any.whl (354 kB)
     |████████████████████████████████| 354 kB 28.2 MB/s 
Collecting defusedxml
  Downloading defusedxml-0.7.1-py2.py3-none-any.whl (25 kB)
Collecting tqdm
  Downloading tqdm-4.64.0-py2.py3-none-any.whl (78 kB)
     |████████████████████████████████| 78 kB 6.5 MB/s 
ERROR: xmlschema 2.0.1 has requirement elementpath<4.0.0,>=3.0.0, but you'll have elementpath 2.5.0 which is incompatible.
Installing collected packages: elementpath, lxml, xmlschema, defusedxml, tqdm
Successfully installed defusedxml-0.7.1 elementpath-2.5.0 lxml-4.9.1 tqdm-4.64.0 xmlschema-2.0.1

 

On my other Ubuntu 20.04 system the xmlschema version is 1.9.2, leading me to think we have another version requirement missing in the GSG.  Do we need to require xmlschema==1.9.2?

zihengL1:
[External_System_ID] ACRN-9283

dbkinder:
In general we've not specified specific compatible versions of apt or python packages. The "latest" GSG was recently updated to say:

sudo pip3 install "elementpath==2.5.0" lxml "xmlschema==1.10.0" defusedxml tqdm

while a more complete solution is underway.

PR #7983, opened 2022-08-12
config-tools: modified requirements.txt and update GSG

zihengL1:
Moved requirements.txt form misc/config_tools/configurator
directory to misc/config_tools directory and modified
relevant python package installing steps in GSG.

Tracked-On: #7975

### DIFF
--- a/misc/config_tools/requirements.txt
+++ b/misc/config_tools/requirements.txt
@@ -1,9 +1,9 @@
 defusedxml
 lxml
-elementpath
-xmlschema
-build
+elementpath==2.5.3
+xmlschema==1.9.2
 tqdm
+build
 requests
 xmltodict
 sphinx
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -142,7 +142,8 @@ To set up the ACRN build environment on the development computer:
 
    .. code-block:: bash
 
-      sudo pip3 install "elementpath==2.5.0" lxml "xmlschema==1.10.0" defusedxml tqdm
+      cd ~/acrn-work/acrn-hypervisor/misc/config_tools
+      sudo pip3 install -r requirements.txt
 
 #. Build and install the iASL compiler/disassembler used for advanced power management,
    device discovery, and configuration (ACPI) within the host OS:
--- a/doc/tutorials/acrn_configurator_tool.rst
+++ b/doc/tutorials/acrn_configurator_tool.rst
@@ -503,7 +503,7 @@ how to build the Debian package from source code.
 
    .. code-block:: bash
 
-      cd ~/acrn-work/acrn-hypervisor/misc/config_tools/configurator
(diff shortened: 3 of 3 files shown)


misc/config_tools/requirements.txt
```diff
@@ -1,10 +1,10 @@
...
 lxml
-elementpath
-xmlschema
+elementpath==2.5.3
+xmlschema==1.9.2
 build
 tqdm
 requests
 xmltodict
 sphinx
```
dbkinder, line 9: The previous list (in the GSG) only had 5 python packages.  Why do we need the new ones added here: build, requests, xmltodict, and sphinx?  (If they're for the configurator, we don't have the user build the configurator any more so we don't need those in the general developer prerequisites.

We need an older version of sphinx for documentation and not just the latest available.

dbkinder (changes requested):
There seem to be too many python packages as prerequisites for the GSG...

dbkinder (approved):
That looks better, thanks.  This looks like the right list for the general developer working on ACRN. I'm a bit concerned now that we don't have the prerequisites for building the configurator recorded any more, but that's a different problem.

@NanlinXie @zihengL1This PR removes misc/config_tools/configurator/requirements.txt
so the instructions for [building the configurator from source](https://projectacrn.github.io/latest/tutorials/acrn_configurator_tool.html#build-acrn-configurator-from-source-code) will now fail.  As for the documentation, it's OK for this special instructions to have additional dependencies, but we need to be careful not to introduce incompatible dependencies among the different tools.

NanlinXie (approved):

dbkinder (commented):
@NanlinXie @zihengL1 We discussed not installing python packages that are not needed for developers building ACRN, yet some packages that are only needed for building the configurator got back on this list in the GSG and the PR was merged.  :(

junjiemao1:
Mmm... Sorry for not noticing the discussion in this PR before.

Dropping dependencies of the configurator may not be a good idea if we already have one specifying its dependencies. How about going back to the original approach that maintains two requirements.txt under `misc/config_tools` and `misc/config_tools/configurator` (which was what we want to get rid of at that time, unfortunately)?

dbkinder:
@junjiemao1 @NanlinXie Are we going to revert this change then, or leave it as merged for the 3.1 release?

dbkinder:
@junjiemao1 @NanlinXie When doing a full GSG walkthrough, the instructions in the GSG say to checkout the v3.0 tag so when you try to use the requirements.txt file it doesn't exist (it's only on the master branch).  I submitted PR #8015 to revert the GSG to not use requirements.txt while we figure things out.  (When we make the v3.1 release, we can update to use the v3.1 tag in the GSG, but that's not ready yet.)

junjiemao1:
> @junjiemao1 @NanlinXie Are we going to revert this change then, or leave it as merged for the 3.1 release?

I would suggest we keep it and add another PR to split it into two dependency lists under different directories if we are fine with keeping two of them.

PR #7993, opened 2022-08-16
Acrn waag multi

yakuizhao:
ACRN uses the DMABuf to render the framebuffer submitted from guest vm.
But the DMABuf doesn't work when guest_vm is booted with multi-display.
This patch set tries to fix the display issues when multi-display is used.

This issue is related with the  limitation in libSDL. One small texture(32x32) is added to WA the issue and
helps to assure that the libSDL can configure its internal state correctly.

### DIFF
--- a/devicemodel/hw/vdisplay_sdl.c
+++ b/devicemodel/hw/vdisplay_sdl.c
@@ -1254,7 +1254,7 @@ vdpy_init(int *num_vscreens)
 
 	vdpy.s.n_connect++;
 	if (num_vscreens)
-		*num_vscreens = 1;
+		*num_vscreens = vdpy.vscrs_num;
 	return vdpy.s.n_connect;
 }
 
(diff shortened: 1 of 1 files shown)


devicemodel/hw/vdisplay_sdl.c
```diff
@@ -611,6 +612,22 @@ sdl_gl_display_init(void)
...
+
+	if (vscr == NULL)
+		return;
+
+	bogus_rect.x = 0;
+	bogus_rect.y = 0;
+	bogus_rect.w = 32;
+	bogus_rect.h = 32;
+	SDL_RenderClear(vscr->renderer);
+	SDL_RenderCopy(vscr->renderer, vscr->bogus_tex, NULL, &bogus_rect);
```
ywan170, line 627: Please add some comments here or a TODO for monitoring SDL update.

yakuizhao: > Please add some comments here or a TODO for monitoring SDL update.
It is updated in the new version.

ywan170 (approved):

PR #8001, opened 2022-08-17
SampleApplication: Update the image create scripts

liudlong:
1. Delete the generated images if the create scripts failed.
2. Enable root user login on GUI.
3. Enable systemd service in case there are many error logs.

Tracked-On: #7820


NanlinXie (approved):

szhen11:
start to run premerge test

PR #8002, opened 2022-08-17
misc: add a uio init setting

weichenli-intel:
We need set uio when the hmi and rt user VM setup, so creat a rc.local
in the '/etc' and run it after system setup.

Tracked-On: #6690

### DIFF
--- a/misc/sample_application/image_builder/create_image.sh
+++ b/misc/sample_application/image_builder/create_image.sh
@@ -152,6 +152,25 @@ EOF
         sudo chown root:root /etc/schroot/chroot.d/acrn-guest
 }
 
+function create_uio_config() {
+    local mount_point=$1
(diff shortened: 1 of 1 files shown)


NanlinXie (approved):

szhen11:
start to run premerge test

misc/sample_application/image_builder/create_image.sh
```diff
@@ -152,6 +152,25 @@ EOF
...
+modprobe uio_pci_generic
+
+for i in {0..3}
+do
+bash -c 'echo "1af4 1110" > /sys/bus/pci/drivers/uio_pci_generic/new_id'
+sleep 1
+done
+EOF
+
+    sudo mv ${temp_file} /etc/rc.local && \
```
gvancuts, line 170: I'm assuming we are not overwriting any existing `/etc/rc.local` here, or are we?

PR #8014, opened 2022-08-17
SampleApplication: Add dhclient command

liudlong:
1.Add the dhclient for the VM images.
2.Disable the schroot data mapping.

Tracked-On: #7820

### DIFF
--- a/misc/sample_application/image_builder/create_image.sh
+++ b/misc/sample_application/image_builder/create_image.sh
@@ -146,7 +146,7 @@ profile=desktop
 personality=linux
 preserve-environment=true
 EOF
-
+    sudo sed -i -e '3,9 s/^/#/' /etc/schroot/default/nssdatabases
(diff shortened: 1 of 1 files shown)


misc/sample_application/image_builder/create_image.sh
```diff
@@ -156,6 +156,7 @@ function create_uio_config() {
     local temp_file=$(mktemp /tmp/rc.local.XXXX)
 
     cat << EOF > ${temp_file}
+dhclient -q
```
gvancuts, line 159: Do we need this because DHCP is not enabled by default in the image?

NanlinXie (approved):

misc/sample_application/image_builder/create_image.sh
gvancuts: Can you explain why we need this?

PR #8020, opened 2022-08-18
doc: add sample application user guide

dbkinder:


dbkinder:
@NanlinXie I've gone ahead and merged this so we can use it for the DX study.  We can make further edits on GitHub as needed.

PR #8022, opened 2022-08-19
doc: update sample app user guide

dbkinder:
After review with engineering, we added material for showing what a
successful hmi_vm.img build ending looks like (and a hint of what to do
if it looks like the build failed.  Also added using a dhclient commmand
to the instructions in case the HMI_VM didn't get an IP address.


NanlinXie (commented):
approved.

Issue #8029, opened 2022-08-22
Add a python package in GSG doc.


6ZhangWei:
The [Getting Started Guide](https://projectacrn.github.io/latest/getting-started/getting-started.html#prepare-the-development-computer) specifies that we need to install the Python packages lxml, xmlschema, and defusedxml.

However, a make clean involves the execution of the Python script [manager.py](https://github.com/projectacrn/acrn-hypervisor/blob/master/misc/config_tools/configurator/packages/configurator/thirdLib/manager.py). This script imports the package requests.py, which might not be installed if the developer follows the Getting Started Guide.

Please update the Getting Started Guide such that it includes the command

sudo pip3 install "elementpath==2.5.0" lxml "xmlschema==1.9.2" defusedxml tqdm requests

6ZhangWei:
[External_System_ID] ACRN-9144

PR #8030, opened 2022-08-22
doc: Add a python package in GSG

6ZhangWei:
The Getting Started Guide specifies that we need to install the Python packages lxml, xmlschema, and defusedxml.
However, a make clean involves the execution of the Python script manager.py. This script imports the package requests.py, which might not be installed if the developer follows the Getting Started Guide.

Add requests follow this command in GSG.
sudo pip3 install "elementpath==2.5.0" lxml "xmlschema==1.9.2" defusedxml tqdm requests

[External_System_ID] ACRN-9144
Tracked-On: #8029
Signed-off-by: Zhang Wei <wei6.zhang@intel.com>

### DIFF
--- a/doc/getting-started/getting-started.rst
+++ b/doc/getting-started/getting-started.rst
@@ -142,7 +142,7 @@ To set up the ACRN build environment on the development computer:
 
    .. code-block:: bash
 
-      sudo pip3 install "elementpath==2.5.0" lxml "xmlschema==1.9.2" defusedxml tqdm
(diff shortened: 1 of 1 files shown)


dbkinder (approved):
I'm going to approve this change, but for v3.2 we really need to isolate what's needed for building the configurator out of the normal ACRN user prerequisites since we don't ask the user to build the configurator any more (we provide a pre-built package.)

PR #8058, opened 2022-08-25
misc: add assert to check the total memory size of User VMs

weichenli-intel:
The current code does not detect the memory applied by all VMs, if it
exceeds the native memory, there will be an allocate failed and these
VMs can't run.

Tracked-On: #6690
Reviewed-by: Junjie Mao junjie.mao@intel.com


NanlinXie (approved):

szhen11:
start to run premerge test

szhen11:
start to run premerge test

PR #8069, opened 2022-08-26
config_tools/upgrader: keep cache allocation policies properly

junjiemao1:
The current generic data moving policy in the upgrader do not work well
with complex nodes having multiple-occurred descendants. That causes the
lost of cache allocation policies in scenario XMLs when upgrading a v3.0
scenario XML to the coming v3.1.

This patch adds another generic-purpose mover that simply copies a whole
subtree unchanged.

Most nodes of this kind are handled by special movers as their formats
changed dramatically in v3.0, which is why that issue is not identified
earlier.

Tracked-On: #6690


NanlinXie (approved):

szhen11:
start to run premerge test

Issue #8068, opened 2022-08-26
[config_tool] Imported scenarios not populating CAT widget


chuangxke:
**Describe the bug**
CAT widget data doesn't reflash after import another scenario

**To Reproduce**
Open configurator UI
Import a scenario file with CAT configured
View CAT widget

**Expected behavior**
Be able to see the previously saved CAT configuration in the CAT widget

chuangxke:
[External_System_ID] ACRN-9275

PR #8074, opened 2022-08-26
[config_tool] Imported scenarios not populating CAT widget

chuangxke:
fix CAT widget data doesn't refresh after import another scenario

the focus will go to "Basic Parameters" tab after import a scenario everytime. so that data could be refresh.

Tracked-On: #8068
Signed-off-by: Chuang-Ke <chuangx.ke@intel.com>


NanlinXie (approved):

szhen11:
start to run premerge test

PR #8107, opened 2022-09-01
doc: Update references to scenario options doc

amyreye:

### DIFF
--- a/doc/tutorials/rtvm_performance_tips.rst
+++ b/doc/tutorials/rtvm_performance_tips.rst
@@ -189,8 +189,9 @@ Tip: Disable the software workaround for Machine Check Error on Page Size Change
    By default, the software workaround for Machine Check Error on Page Size
    Change is conditionally applied to the models that may be affected by the
    issue. However, the software workaround has a negative impact on
-   performance. If all guest OS kernels are trusted, the
-   :option:`hv.FEATURES.MCE_ON_PSC_DISABLED` option could be set for performance.
+   performance. If all guest OS kernels are trusted, you can disable the
+   software workaround (by deselecting the :term:`Enable MCE workaround` option
+   in the ACRN Configurator tool) for performance.
 
 .. note::
    The tips for preempt-RT Linux are mostly applicable to the Linux-based RTOS as well, such as Xenomai.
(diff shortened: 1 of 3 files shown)


doc/tutorials/rtvm_performance_tips.rst
```diff
@@ -189,8 +189,9 @@ Tip: Disable the software workaround for Machine Check Error on Page Size Change
    By default, the software workaround for Machine Check Error on Page Size
    Change is conditionally applied to the models that may be affected by the
    issue. However, the software workaround has a negative impact on
-   performance. If all guest OS kernels are trusted, the
-   :option:`hv.FEATURES.MCE_ON_PSC_DISABLED` option could be set for performance.
+   performance. If all guest OS kernels are trusted, you can disable the
```
dbkinder, line 192: Looks like the default changed to NOT enable this work around, so we need to change the text on the lines above and here.  @NanlinXie can you confirm?

dbkinder (changes requested):
I think the default for 'Enable MCE workaround" changed

doc/tutorials/rtvm_performance_tips.rst
NanlinXie: Yes , confirmed. Here we need uncheck "Enable MCE workaround" this option.

amyreye: @NanlinXie Not sure I understand your answer. Is the following text accurate?

> Tip: Disable the software workaround for Machine Check Error on Page Size Change.
> By default, the software workaround for Machine Check Error on Page Size Change is conditionally applied to the models that may be affected by the issue. However, the software workaround has a negative impact on performance. If all guest OS kernels are trusted, you can disable the software workaround (by changing the [:term:`Enable MCE workaround`](https://github.com/projectacrn/acrn-hypervisor/blob/17e7d00eb67cdbc38b665cb9d30104777e75d5a1/doc/tutorials/rtvm_performance_tips.rst#id3) option in the ACRN Configurator tool) for performance.

NanlinXie: The last sentence could be modified to "the "Enable MCE workaround" option could be unset for performance."

Issue #8111, opened 2022-09-05
[life_mngr] Sync between SOS and RTVM failed when startup hence life_mngr cannot work


chejianj:
1) deploy life_mngr on Service VM and RTVM
2) Sync between Service VM and RTVM failed

The reason is that RTVM will write 1 to fifo and the string received by Service VM is preceded by -1

chejianj:
[External_System_ID] ACRN-9352

gvancuts:
Updated `SOS` to `Service VM`

PR #8112, opened 2022-09-05
misc: life_mngr: revise try_receive_message_by_uart

chejianj:
Revise try_receive_message_by_uart to read one char from uart one time.
With this implementation each char can be checked. This can be used to
address the following 2 problems:
1) nosie data: it is found that there is noise data in the uart from
   guest VM when guest startup.
2) split multiple commands

Tracked-On: #8111


yakuizhao (approved):
LGTM

PR #8142, opened 2022-09-14
misc: add assertion to check the BDF value

weichenli-intel:
The BDF of user setting should skip the slot '00', '01', '02', '1f', all these slots should select one of the 03~1e.

This patch add an assertion to check the above policy.

Tracked-On: #6690

### DIFF
--- a/misc/config_tools/schema/checks/vbdf_assignment.xsd
+++ b/misc/config_tools/schema/checks/vbdf_assignment.xsd
@@ -31,4 +31,22 @@
     </xs:annotation>
   </xs:assert>
 
+  <xs:assert test="every $ivshmem in /acrn-config//IVSHMEM_VM satisfies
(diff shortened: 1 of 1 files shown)


junjiemao1:
Have you tried those assertions with the elementpath & xmlschema versions we use in the GSG?

weichenli-intel:
> Have you tried those assertions with the elementpath & xmlschema versions we use in the GSG?

My native version is the same with GSG:
xmlschema                     1.9.2
elementpath                   2.5.0

junjiemao1:
> > Have you tried those assertions with the elementpath & xmlschema versions we use in the GSG?
> 
> My native version is the same with GSG: xmlschema 1.9.2 elementpath 2.5.0

Fine with me then.

NanlinXie (approved):

PR #8161, opened 2022-09-19
dm: virtio-blk: fix parameter err

conghuic23:
Fix the truncate issue for virtio block parameter.

### DIFF
--- a/devicemodel/hw/pci/virtio/virtio_block.c
+++ b/devicemodel/hw/pci/virtio/virtio_block.c
@@ -506,7 +506,11 @@ virtio_blk_init(struct vmctx *ctx, struct pci_vdev *dev, char *opts)
 		if (strcmp("iothread", opt) == 0) {
 			use_iothread = true;
 		} else {
-			opts_tmp = opts_start;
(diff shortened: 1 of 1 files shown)


devicemodel/hw/pci/virtio/virtio_block.c
```diff
@@ -506,7 +506,7 @@ virtio_blk_init(struct vmctx *ctx, struct pci_vdev *dev, char *opts)
 		if (strcmp("iothread", opt) == 0) {
 			use_iothread = true;
 		} else {
-			opts_tmp = opts_start;
+			opts_tmp = opts;
```
yakuizhao, line 509: Can some comments be added?
It seems that the opts_start is assigned by using strdup(opts). 
This seems quite strange. It will be better to add some comments about the buffer pointed by opts_start is already changed.

yakuizhao (approved):
LGTM.

szhen11:
start to run premerge test

gvancuts:
Shouldn't we also add a new copyright info to this `virtio_block.c` file since there are modifications made to it from Intel engineers?

/cc @NanlinXie

Issue #8168, opened 2022-09-19
Add Hypervisor-Managed Processor Performance Controls


izhouwu:

izhouwu:
[External_System_ID] [ACRN-9359]

PR #8174, opened 2022-09-20
add ACRN CPU performance management

izhouwu:
1. Background:

Customers of the industry segment are asking for more computation power
in WaaG. Currently there are two ways to manage CPU frequency when
running ACRN:
  - Pass through HWP/p-state to guest. This doesn't work for Windows
    because Windows running in a VM relies on Hyper-V to manage the
    frequency. So it does not load frequency driver.
  - Or rely on service VM's Linux HWP driver, which can set the
    frequency policy to hardware autonomous selection. But it does not
    work either because when a CPU is set to offline, that driver
    sets HWP autonomous frequency selection range to minimal. Also we
    don't want the RTVM's frequency been changed by autonomous selection.

So we have to let ACRN manage CPU frequency.

2. The design:

The design of ACRN CPU performance management is to let hardware do the
autonomous frequency selection(or set to a fixed value), and remove all
guests' capability to control CPU frequency.

If the platform supports HWP, we can use HWP to do the hardware
autonomous selection. The hypervisor just need to setup HWP_REQUEST at
start up, and no need to be governing the frequency after this. But for
ACPI p-state, the CPU frequency can only be fixed.

Cores running RTVMs are always fixed to nominal/guaranteed frequency, to
get more certainty in latency.

Users may have preference on how the CPU frequency is managed. So we
provide two performance policy types for users to choose from:
  - 'Performance': CPU runs at its CPU runs at its maximum
    frequency. Enable hardware autonomous frequency selection if HWP is
    presented.
  - 'Nominal': CPU runs at its guaranteed frequency.
(*Note: The performance 'policy' was called 'governor' in the first
edition of this patch series. But ACRN don't actually 'govern' the
frequency. It just apply the policy at start up. So the word 'policy'
might be more appropriate.)

The policy type is passed to hypervisor through boot parameter, as
either 'cpu_perf_policy=Nominal' or 'cpu_perf_policy=Performance'.
The default type is 'Performance'.

An configurate item is provided to user to choose the policy.
Config-tool will then generate the boot parameter automatically.
But it can always be changed by editting the grub cfg.
3. The implementation:
  - Add a config item and automatically generate policy parameter.
    (1/6, 2/6)
  - Extract CPU frequency info and generate frequency limits from
    config-tools. This could save hypervisor code. (3/6, 4/6)
  - Implement CPU frequency initializer. (5/6)
  - Remove guest's ability to control CPU frequency by removing the
    guests' HWP/EIST CPUIDs and blocking the related MSR accesses.
    Also removed hypercall cmds for getting px info, so that DM do
    not generate _PSS/_PPC for post-launched VMs anymore. (6/6)

Tracked-On: #8168

### DIFF
--- a/hypervisor/arch/x86/pm.c
+++ b/hypervisor/arch/x86/pm.c
@@ -19,6 +19,8 @@
 #include <asm/lapic.h>
 #include <asm/tsc.h>
 #include <delay.h>
+#include <asm/board.h>
+#include <asm/cpuid.h>
 
 struct cpu_context cpu_ctx;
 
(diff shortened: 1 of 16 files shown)


izhouwu:
The origin patch series contains 6 patches. To pass the LOC check, patch 6/6 will be later uploaded in another PR.

hypervisor/arch/x86/pm.c
```diff
@@ -271,3 +273,81 @@ void reset_host(void)
...
+	uint32_t cpuid_06_eax, unused;
+	struct acrn_boot_info *abi = get_acrn_boot_info();
+	const char *cmd_src = abi->cmdline;
+
+	/*
+	 * Parse cmdline, decide which policy type to use.
+	 * User can either specify cpu_perf_policy=Nominal or cpu_perf_policy=Performance
+	 * The default type is 'Performance'
+	 */
+	if(strstr_s(cmd_src, MAX_BOOTARGS_SIZE, "cpu_perf_policy=Nominal", 24U) != NULL) {
```
gvancuts, line 290: Have we verified that this works correctly also when a command-line `uart=` parameter is present?

gvancuts (commented):
Do not forget to also add documentation about this ;-)

misc/config_tools/schema/config.xsd
```diff
@@ -129,6 +129,14 @@ If your VM is not a security VM, leave this option unchecked. </xs:documentation
         <xs:documentation>Configure Software SRAM. This feature reserves memory buffers as always-cached memory to improve an application's real-time performance.</xs:documentation>
       </xs:annotation>
     </xs:element>
+    <xs:element name="CPU_PERFORMANCE_POLICY" type="policyType" default="Performance">
+      <xs:annotation acrn:title="CPU performance policy type" acrn:views="advanced">
+        <xs:documentation>Select the default performance policy type for all CPUs. The hypervisor will setup CPU frequency based on the policy type.
+
+- ``Performance``: CPU runs at its maximum frequency. Enable hardware autonomous frequency selection if HWP is presented.
```
amyreye, line 136: Change "if HWP is presented" to "if the system supports [full term] (HWP)". Replace [full term] with the term that HWP stands for (is it Hardware P-state?).

misc/config_tools/schema/config.xsd
```diff
@@ -129,6 +129,14 @@ If your VM is not a security VM, leave this option unchecked. </xs:documentation
         <xs:documentation>Configure Software SRAM. This feature reserves memory buffers as always-cached memory to improve an application's real-time performance.</xs:documentation>
       </xs:annotation>
     </xs:element>
+    <xs:element name="CPU_PERFORMANCE_POLICY" type="policyType" default="Performance">
+      <xs:annotation acrn:title="CPU performance policy type" acrn:views="advanced">
+        <xs:documentation>Select the default performance policy type for all CPUs. The hypervisor will setup CPU frequency based on the policy type.
```
amyreye, line 134:
```suggestion
        <xs:documentation>Select the default performance policy type for all CPUs. The hypervisor will set up the CPU frequency based on the policy type.
```

misc/config_tools/schema/types.xsd
```diff
@@ -147,6 +147,24 @@ higher value (lower severity) are discarded.</xs:documentation>
   </xs:restriction>
 </xs:simpleType>
 
+<xs:simpleType name="policyType">
+  <xs:annotation>
+    <xs:documentation>A string specifying the CPU frequency policy type:
+
+- ``Performance``: CPU runs at its maximum frequency. Enable hardware autonomous frequency selection if HWP is presented.
```
amyreye, line 154: Change "if HWP is presented" to "if the system supports [full term] (HWP)". Replace [full term] with the term that HWP stands for (is it Hardware P-state?).

amyreye:
Is this change for v3.1 or v3.2? If v3.2, can you postpone merging the PR until after the v3.1 release (or at least until after documentation on master is merged in the v3.1 branch)? I am concerned that the new element added to the XSD file will appear in the v3.1 [scenario options documentation](https://projectacrn.github.io/latest/reference/config-options.html) if we merge the PR before I have cherry picked the v3.1 docs. What do you think, @NanlinXie ?

hypervisor/arch/x86/pm.c
izhouwu: I checked the code and did some extra test about this. And it works fine with 'uart=' parameter.

misc/config_tools/schema/types.xsd
izhouwu: The full term is ‘Hardware-Controlled Performance States’. I have changed it and force pushed.

misc/config_tools/schema/config.xsd
izhouwu: This was also changed in the last push

NanlinXie:
> Is this change for v3.1 or v3.2? If v3.2, can you postpone merging the PR until after the v3.1 release (or at least until after documentation on master is merged in the v3.1 branch)? I am concerned that the new element added to the XSD file will appear in the v3.1 [scenario options documentation](https://projectacrn.github.io/latest/reference/config-options.html) if we merge the PR before I have cherry picked the v3.1 docs. What do you think, @NanlinXie ?

Agreed.

NanlinXie (approved):
Approved for configtool part.

amyreye:
> > Is this change for v3.1 or v3.2? If v3.2, can you postpone merging the PR until after the v3.1 release (or at least until after documentation on master is merged in the v3.1 branch)? I am concerned that the new element added to the XSD file will appear in the v3.1 [scenario options documentation](https://projectacrn.github.io/latest/reference/config-options.html) if we merge the PR before I have cherry picked the v3.1 docs. What do you think, @NanlinXie ?
> 
> Agreed.

This PR will not affect v3.1 docs, feel free to merge.

szhen11:
@dongyaozu Please help to review and approve this PR, thanks.

dongyaozu (approved):

PR #8176, opened 2022-09-20
doc: add hypercall ABI in the HLD

junjiemao1:
The application binary interface (ABI) is an important aspect of hypercalls. But unfortunately it is not yet defined anywhere in the current HLD.

This patch adds the current ABI specification to the hypercall section of HLD.


gvancuts:
Thanks for adding this @junjiemao1 . Do we have any policy in terms of ABI compatibility between versions (minor, major, etc.)?

I also think it would be good to do something similar for `acrn-dm` and provide a list of "official" APIs that an application can use to interact with it.

doc/developer-guides/hld/hv-hypercall.rst
```diff
@@ -9,6 +9,28 @@ is from the hypervisor to the guest VM. The hypervisor supports
...
 PCI assignment, guest memory mapping, power management, and secure world switch.
 
+The application binary interface (ABI) of ACRN hypercalls is defined as follows.
+
+- A guest VM executes the ``vmcall`` instruction to trigger a hypercall.
+
+- Input parameters of a hypercall includes:
+
+  - An hypercall ID in ``R8``, which specifies the kind of service requested by
+    the guest VM.
```
amyreye, line 19: What is R8, RDI, RSI? Are they registers? If so, should we say that they are registers for the sake of clarity? (Or would the audience know that?)

doc/developer-guides/hld/hv-hypercall.rst
```diff
@@ -9,6 +9,28 @@ is from the hypervisor to the guest VM. The hypervisor supports
...
+
+  - The first parameter in ``RDI`` and the second in ``RSI``. The semantics of
+    those two parameters vary among different kinds of hypercalls and are
+    defined in the hypercall APIs reference. For hypercalls targeting a specific
+    VM, the first parameter is typically the ID of that VM.
```
amyreye, line 24: Why would a hypercall target a specific VM? The introduction above says a hypercall is from a VM to the hypervisor.

junjiemao1:
> Thanks for adding this @junjiemao1 . Do we have any policy in terms of ABI compatibility between versions (minor, major, etc.)?

The ABI is intended to be stable and has been so since the launch of the project. I don't foresee any change to it in the future.

> I also think it would be good to do something similar for `acrn-dm` and provide a list of "official" APIs that an application can use to interact with it.

There is one [page](https://projectacrn.github.io/latest/api/devicemodel_api.html) on that topic, but I'm not sure to whom that spec is designed to help.

Typically the device model does not interact with other applications via APIs, but rather command line parameters, sockets and events. I remember we have a list of the DM parameters in the doc, but probably none for the latter two kinds.

doc/developer-guides/hld/hv-hypercall.rst
junjiemao1: Yes. They are registers. Will explicitly state so in the next revision.

doc/developer-guides/hld/hv-hypercall.rst
junjiemao1: Using ``target`` here could be ambiguous. I rephrased it to ``For hypercalls requesting operations on a specific VM``.

amyreye (approved):

PR #8184, opened 2022-09-21
Add the support of PCI rom for passthru GPU device

yakuizhao:
PCI ROM is the firmware specific to PCI device and it is provided by
the device vendor. The PCI rom resides in 0x30 offset of PCI config space.
This can be used to check whether the PCI rom exists. And when it exists,
it can load the firmware from the addr that is obtained from ROM bar addr.

   This patch set tries to add the support of PCI rom bar so that the rom file
can be loaded for the passthrough device. Based on the PCI rom, it is convenient
to add the GPU GOP for OVMF firmward instead of integrating the GPU Gop into
the OVMF.
   The PCI rom bar is handled in DM. The minor change is in HV so that
it can forward the access of ROM_bar reg to DM. Now this is handled only
when the ACRN_PTDEV_QUIRK_ASSIGN is specified for the passthru device.

   It is noted that the rom_file has no dependency on the ROM bar of physical
PCI device. It can be provided by the device vendor. It can be dumped from the
ROM bar of PCI device. It can also be converted from EFI image by using EfiRom too

### DIFF
--- a/hypervisor/dm/vpci/vpci.c
+++ b/hypervisor/dm/vpci/vpci.c
@@ -436,10 +436,15 @@ static const struct cfg_header_perm cfg_hdr_perm = {
 /*
  * @pre offset + bytes < PCI_CFG_HEADER_LENGTH
  */
-static void read_cfg_header(const struct pci_vdev *vdev,
+static int32_t read_cfg_header(const struct pci_vdev *vdev,
 		uint32_t offset, uint32_t bytes, uint32_t *val)
 {
-	if (vbar_access(vdev, offset)) {
+	int32_t ret = 0;
+
+	if ((offset == PCIR_BIOS) && is_quirk_ptdev(vdev)) {
+		/* the access of PCIR_BIOS is emulated for quirk_ptdev */
+		ret = -ENODEV;
+	} else if (vbar_access(vdev, offset)) {
 		/* bar access must be 4 bytes and offset must also be 4 bytes aligned */
 		if ((bytes == 4U) && ((offset & 0x3U) == 0U)) {
 			*val = pci_vdev_read_vcfg(vdev, offset, bytes);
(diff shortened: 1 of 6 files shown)


ywan170 (approved):

hypervisor/dm/vpci/vpci.c
```diff
@@ -436,10 +436,16 @@ static const struct cfg_header_perm cfg_hdr_perm = {
 /*
  * @pre offset + bytes < PCI_CFG_HEADER_LENGTH
  */
-static void read_cfg_header(const struct pci_vdev *vdev,
+static int32_t read_cfg_header(const struct pci_vdev *vdev,
 		uint32_t offset, uint32_t bytes, uint32_t *val)
 {
-	if (vbar_access(vdev, offset)) {
+	int32_t ret;
```
lifeix, line 442: I prefer using "int32_t ret = 0" here.

hypervisor/dm/vpci/vpci.c
```diff
@@ -460,15 +466,22 @@ static void read_cfg_header(const struct pci_vdev *vdev,
...
 
 /*
  * @pre offset + bytes < PCI_CFG_HEADER_LENGTH
  */
-static void write_cfg_header(struct pci_vdev *vdev,
+static int32_t write_cfg_header(struct pci_vdev *vdev,
 		uint32_t offset, uint32_t bytes, uint32_t val)
 {
-	if (vbar_access(vdev, offset)) {
+	int32_t ret;
```
lifeix, line 478: ditto

lifeix (changes requested):

hypervisor/dm/vpci/vpci.c
yakuizhao: Thanks for pointing out. 
It is changed.

lifeix (approved):

Issue #8051, opened 2022-08-25
ADL-S][S5]Can't shutdown WaaG with user VM name same as in launch script


lifeix:
[External_System_ID] ACRN-7654

PR #8218, opened 2022-09-28
release_3.1_hotfix: life_mngr: enlarge windowns VM name buffer length

lifeix:
Enlarge windowns oVM name buffer length from 12 to 32.

Tracked-On: #8051


yakuizhao (approved):
LGTM

PR #8226, opened 2022-10-08
Refine the logic of allocate sworld memory

weichenli-intel:
This series try to refine the allocate sworld memory logic, the current
code allocate for all post launched VM, it's a waste of memory.

The new proposal try to count trusty VM number by offline tool and
allocate sworld memory only for trusty VM, then add a new interface to
get the right page index.


NanlinXie (approved):
Approved for configtool part.

dongyaozu (approved):

PR #8251, opened 2022-10-22
Issue8246 debianization improvement

tttech-industrial-buchsbaum:
This patch series fixes issues with configuration file detection when using ACRN_BOARDLIST, ACRN_SCENARIOLIST and CONFIGDIRS other than the default settings in debian/rules, e.g. using

`debuild -eACRN_BOARDLIST= -eACRN_SCENARIOLIST= -eCONFIGDIRS=<custom config path> -- binary`

to build all custom board/scenario configurations from  <custom config path>.
Remember: `-eACRN_BOARDLIST=` means detecting all available boards, whereas not setting `ACRN_BOARDLIST` means using the debian/rules default settings. The same applies to `ACRN_SCENARIOLIST`.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -32,8 +32,8 @@ rwildcard=$(foreach d,$(wildcard $(1:=/*)),$(call rwildcard,$d,$2) $(filter $(su
 unquote = $(subst $\",,$1)
 
 # set these variables to define build of certain boards/scenarios, e.g.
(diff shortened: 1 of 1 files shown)


junjiemao1 (approved):
LGTM, thanks!

NanlinXie (approved):

PR #8266, opened 2022-10-27
doc: Create a requirements.txt for python package on GSG

zhangrouyu:
Signed-off-by: zhangrouyu <rouyu.zhang@intel.com>

### DIFF
--- a/misc/config_tools/requirements.txt
+++ b/misc/config_tools/requirements.txt
@@ -1,5 +1,5 @@
 defusedxml
 lxml
-elementpath==2.5.3
-xmlschema==1.9.2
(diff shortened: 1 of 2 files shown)


misc/config_tools/requirements.txt
```diff
@@ -1,5 +1,5 @@
 defusedxml
 lxml
-elementpath==2.5.3
-xmlschema==1.9.2
-tqdm
\ No newline at end of file
+elementpath>=2.5.0
+xmlschema
```
dbkinder, line 4: Since you removed a version requirement, can you confirm that ANY version of xmlschema is OK?

PR #8272, opened 2022-10-28
doc: handle anonymous complex types in XML schema

junjiemao1:
The schema of scenario XMLs uses both named and anonymous complex types, but the configdoc.xsl today only works for named complex types. That causes improper rendering of config items related to virtio devices (whose schema uses anonymous types) as well as wrong applicable VM icons which states that virtio devices apply to not only post-launched VMs but also pre-launched and service VM, which does not make sense.

This patch improves configdoc.xsl by adding support of anonymous complex types so that they are walked through in the same way as named ones.


dbkinder (approved):
Yep, looks better with this fix.

PR #8273, opened 2022-10-28
update ubuntu version from 20.04.5 to 22.04.1 in GSG document

zhangrouyu:
update ubuntu version and remove the python3.8-venv.
ubuntu22.04 installs python3.10 by default, and an error will be reported when installing python3.8-venv.

Signed-off-by: zhangrouyu <rouyu.zhang@intel.com>


doc/getting-started/getting-started.rst
```diff
@@ -119,7 +119,7 @@ To set up the ACRN build environment on the development computer:
 
       sudo apt install -y gcc git make vim libssl-dev libpciaccess-dev uuid-dev \
            libsystemd-dev libevent-dev libxml2-dev libxml2-utils libusb-1.0-0-dev \
-           python3 python3-pip python3.8-venv libblkid-dev e2fslibs-dev \
+           python3 python3-pip libblkid-dev e2fslibs-dev \
```
gvancuts, line 122: Does that mean to Python venv is already being installed although not explicitly mentionned in the list?

dbkinder: On a clean Ubuntu 22.04 system, Python venv is not installed and won't be after this list of packages is installed.  Python venv was only used for building the configurator, which we don't require the user to do any more.

dbkinder (approved):
LGTM, thanks.

PR #8281, opened 2022-11-02
Disable PMU passthrough in release mode

liuhang-bit:
ACRN v2.6 introduces PMU passthrough to RT VMs which have 
LAPIC passthrough enabled, which is mainly for performance 
profiling purposes at development time only. As of now there
 is no foreseeable use case that require PMU in production.

This patch series makes that mechanism effective only in debug
builds to avoid potential side effects in production environments.


NanlinXie (approved):
Approved for configtool part

lifeix (approved):

szhen11:
start to run premerge test

PR #8283, opened 2022-11-02
misc: rename common.py to avoid private library conflict

weichenli-intel:
The offline tool use a utility libary and use "common" to named it, this name conflict with some customer's local library, so we rename it to a better and clear name.

Tracked-On: #6690


misc/config_tools/acpi_gen/asl_gen.py
```diff
@@ -669,8 +669,8 @@ def gen_dsdt(board_etree, scenario_etree, allocation_etree, vm_id, dest_path):
             bus_number = int(m.group(1), 16)
             bdf = f"{bus_number:02x}:{device_number:02x}.{function_number}"
             address = hex((device_number << 16) | (function_number))
-            device_node = common.get_node(f"//bus[@address='{hex(bus_number)}']/device[@address='{address}']", board_etree)
-            alloc_node = common.get_node(f"/acrn-config/vm[@id='{vm_id}']/device[@name='PTDEV_{bdf}']", allocation_etree)
+            device_node = acrn_config_utilities.get_node(f"//bus[@address='{hex(bus_number)}']/device[@address='{address}']", board_etree)
+            alloc_node = acrn_config_utilities.get_node(f"/acrn-config/vm[@id='{vm_id}']/device[@name='PTDEV_{bdf}']", allocation_etree)
```
junjiemao1, line 673: For methods like `get_node` which are commonly used throughout a script, how about using `from acrn_config_utilities import get_node` instead so that we can invoke it without repeating `acrn_config_utilities.` every time we use it?

weichenli-intel: I have update.

szhen11:
start to run premerge test

junjiemao1 (approved):

NanlinXie (approved):

PR #8285, opened 2022-11-02
doc: Upgrade RT kernel to 5.15 in sample app guide

NanlinXie:
doc: Upgrade RT kernel to 5.15 in sample app guide

Signed-off-by: Xie, nanlin <nanlin.xie@intel.com>


dbkinder (approved):
OK, but we'll need to update these checkout tags for the actual v3.2 release

Issue #8303, opened 2022-11-10
Confusion about the daily tag in ACRN version


6ZhangWei:
I have build the ACRN Hypervisor based on the Release v3.0.1. I have used this command to clone the ACRN repo:
git clone https://github.com/projectacrn/acrn-hypervisor.git -b v3.0.1

However, after installing the Debian package, I can see this output in the HV console:
ACRN:\>version
HV 3.01-40-2022-08-05 09:21:56-7112452d5-dirty DBG (daily tag: acrn-2022w27.1-180000p) shared@tgl-h-rvp build by 1000

The version information starts with 3.01-40, which is good. But I don’t understand the daily tag: acrn-2022w27.1-180000p. Wasn’t the v3.0.1 Release created this week (ww32)?

6ZhangWei:
[External_System_ID] ACRN-9315

PR #8304, opened 2022-11-10
 hv: change the version format

6ZhangWei:
hv: modify the version info

The daily tag should be the latest git-tag. And this tag name is also showed in the head of version-string. The content of daily tag is redundant, also a wrong daily tag name was fetched in release3.0. This patch won't show the daily tag.

Tracked-On #8303
Signed-off-by: Zhang Wei <wei6.zhang@intel.com>

### DIFF
--- a/hypervisor/debug/shell.c
+++ b/hypervisor/debug/shell.c
@@ -738,8 +738,8 @@ static int32_t shell_version(__unused int32_t argc, __unused char **argv)
 {
 	char temp_str[MAX_STR_SIZE];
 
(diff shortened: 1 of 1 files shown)


dongyaozu:
@lifeix can you review and ack?

lifeix:
> unstable-acrn

lifeix:
@6ZhangWei  Please fix this JIRA bug first. About the version format, you could send a RFC patch to acrn mail-list to discuss.

PR #8306, opened 2022-11-10
Issue8262 some kernel cmdline update is missing in board inspector

tttech-industrial-buchsbaum:
Installing python3-acrn-board-inspector should add a separate grub menu entry (which is also set as default) to ensure the kernel is parameterized correctly for the use with ACRN board inspector. So, rebooting after install will ensure the ACRN board inspector can run within the required environment. Deinstallation completely removes the this temporary entry again.


debian/grub/20_acrn-board-inspector
```diff
@@ -0,0 +1,269 @@
...
+EOF
+}
+
+machine=`uname -m`
+case "x$machine" in
+    xx86_64)
+	GENKERNEL_ARCH="x86"
+	list=
+	for i in /boot/vmlinuz-* /vmlinuz-* /boot/kernel-* ; do
+	    if grub_file_is_not_garbage "$i" ; then list="$list $i" ; fi
```
junjiemao1, line 172: Will this generate one top-level menu entry for each kernel found under /boot? How about adding an upper limit of the numbers of entries to generate?

tttech-industrial-buchsbaum: Yes it will. This is a more or less a copy of what the 10_linux grub-mkconfig helper script does. They also do not limit the number of entries. Since the usual scenario is installing the ACRN board inspector on a fresh install, this will only lead to only one additional top-level menu entry. It virtually doubles the top-level boot entries you had before installing ACRN board inspector. Since this is the way it is also done in other grub helper scripts and considering the main use case I do not think we need any limitation. BTW, after obtaining the board configuration, the board inspector should be deinstalled again before continuing to prepare the system for ACRN!

junjiemao1: One difference between 10_linux/25_linux_acrn and this script is that the later does not create any submenu.

But your point also makes sense: if it is a freshly installed distribution, doubling the top-level entries is not a major concern.

May I know why the board inspector shall be removed?

tttech-industrial-buchsbaum: Does it make sense to have the board inspector on a ACRN enabled system?

tttech-industrial-buchsbaum:
Update PR to remove default boot behavior.

junjiemao1:
> Update PR to remove default boot behavior.

Thanks for updating! Can we merge the second and third commits so that we do not introduce some files only to remove them in the next commit in the git history?

tttech-industrial-buchsbaum:
Finally updated the PR commits correctly by rebasing them. Also folded the former last two commits. This should now be applicable directly without hassles.

junjiemao1 (approved):
LGTM. Thanks!

NanlinXie (approved):

PR #8307, opened 2022-11-11
misc: fix the vCPU number issue of document

weichenli-intel:
Current code use a invalid parameter to print the vCPU number, this patch use the length of vCPU list to fix the above issue.

Tracked-On: #6690

### DIFF
--- a/misc/config_tools/scenario_config/config_summary.py
+++ b/misc/config_tools/scenario_config/config_summary.py
@@ -16,6 +16,7 @@ class GenerateRst:
     pci_vuart = {}
     pci_ivshmem = {}
     amount_l3_cache = {}
+    service_vm_used_pcpu_list = []
 
     # Class initialization
     def __init__(self, board_file_name, scenario_file_name, rst_file_name) -> None:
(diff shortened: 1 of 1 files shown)


junjiemao1 (approved):

misc/config_tools/scenario_config/config_summary.py
```diff
@@ -302,7 +302,10 @@ def get_basic_information_table(self, vm_node):
                 vcpu_num.add(item)
         amount_vm_l3_cache = self.get_amount_l3_cache(vm_node)
         parameter_dict["Load Order"] = load_order
-        parameter_dict["Number of vCPUs"] = len(vcpu_num)
```
junjiemao1, line 305: Since `vcpu_num` is no longer used, remove line 281 and line 297-302 as well.

junjiemao1 (changes requested):

misc/config_tools/scenario_config/config_summary.py
weichenli-intel: Done

junjiemao1 (approved):

NanlinXie (approved):

Issue #8344, opened 2022-11-28
debian/debian_build.sh fails when a work folder contains files other than XML


junjiemao1:
**Describe the bug**
debian/debian_build.sh fails when a work folder contains files other than XML.

**Platform**
Any

**Codebase**
Main branch

**Scenario**
Any

**To Reproduce**
1. Create a work folder that contains a board XML, a scenario XML and some arbitrary text files.
2. Build debian packages using `debuild -eACRN_BOARDLIST= -eACRN_SCENARIOLIST= -eCONFIGDIRS=test_scenario -- binary`

**Expected behavior**
Build succeeds.

**Error log**
Traceback (most recent call last):
  File "/usr/lib/python3.10/xml/etree/ElementTree.py", line 1713, in feed
    self.parser.Parse(data, False)
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 1, column 0
 
During handling of the above exception, another exception occurred:
 
Traceback (most recent call last):
  File "/home/david/acrn-work/acrn-hypervisor/misc/config_tools/scenario_config/default_populator.py", line 102, in <module>
    main(args)
  File "/home/david/acrn-work/acrn-hypervisor/misc/config_tools/scenario_config/default_populator.py", line 89, in main
    pipeline.run(obj)
  File "/home/david/acrn-work/acrn-hypervisor/misc/config_tools/scenario_config/pipeline.py", line 68, in run
    stage.run(obj)
  File "/home/david/acrn-work/acrn-hypervisor/misc/config_tools/scenario_config/xml_loader.py", line 18, in run
    etree = parse(xml_path)
  File "/usr/lib/python3/dist-packages/defusedxml/common.py", line 100, in parse
    return _parse(source, parser)
  File "/usr/lib/python3.10/xml/etree/ElementTree.py", line 1222, in parse
    tree.parse(source, parser)
  File "/usr/lib/python3.10/xml/etree/ElementTree.py", line 586, in parse
    parser.feed(data)
  File "/usr/lib/python3.10/xml/etree/ElementTree.py", line 1715, in feed
    self._raiseerror(v)
  File "/usr/lib/python3.10/xml/etree/ElementTree.py", line 1622, in _raiseerror
    raise err
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 1, column 0

NanlinXie:
[External_System_ID] ACRN-9524

PR #8345, opened 2022-11-28
debian/rules: search for XML files only

junjiemao1:
When searching for scenario XMLs that are saved under the same directory as a board XML, debian/rules uses the wildcard `*` which includes other non-XML files. That causes some non-XML files to be considered as scenario XMLs as well and will cause build-time errors when the build system attempts to parse them as XMLs.

Change the wildcard expression to `*.xml` to restrict the found files to be XML.

Tracked-On: #8344


junjiemao1:
@tttech-industrial-buchsbaum Please review this fix. Thanks!

szhen11:
start to run premerge test

tttech-industrial-buchsbaum:
This is an obvious problem and we really need this fix!

NanlinXie (approved):

Issue #8246, opened 2022-10-20
Debianization improvement


NanlinXie:
After we have new debianization solution in v3.0, we have more request to improve, create this issue to track the improvement for debianization.

PR #8346, opened 2022-11-29
debian/rules: change default BOARDLIST and SCENARIOLIST to empty

junjiemao1:
The variables BOARDLIST and SCENARIOLIST serve as filters of XMLs that are found under the user-given directories, and there is no need to assume any filter if a user does not specify that explicitly.

Update the default filters to none so that all found XMLs will be used if a user does not state otherwise.

Tracked-On: #8246


junjiemao1:
@tttech-industrial-buchsbaum Here's the default-value-cleanup change we just talked about :)

tttech-industrial-buchsbaum:
This absolutely makes sense and looks fine to me.

NanlinXie (approved):

PR #8355, opened 2022-12-02
doc: update sample-app version &command to match release_3.2

nlesxmz:
1. update hypervisor and kernel version to release_3.2
2. modify debian package (of ACRN hypervisor and tools) build command
3. delete iasl command lines (substitute: acpica-tools generate iasl in GSG)
4. remove use of USB stick to copy files from development computer to target (only use scp)
5. remove grub menu after reboot (it will boot automatically.)

Signed-off-by: Min Yang <minx.yang@intel.com>


nlesxmz:
filed by mistake

PR #8365, opened 2023-01-20
doc: fix missing numbered-step icon

dbkinder:
Sphinx version 5 generates &lt;section&gt; instead of &lt;div&gt; for section headings.  This throws off the custom CSS style for numbered steps so the number icon wasn't showing up.  Tweak the custom CSS to look for both &lt;section&gt; and &lt;div&gt; with class=numbered-step to handle Sphinx 4 and 5-generated content.

### DIFF
--- a/doc/static/acrn-custom.css
+++ b/doc/static/acrn-custom.css
@@ -291,7 +291,8 @@ body {
    counter-reset: step-count;
 }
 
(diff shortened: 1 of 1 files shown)


NanlinXie:
Thanks @dbkinder

Issue #8382, opened 2023-04-21
Failed to build with gcc 12


jiaqingz-intel:
Building ACRN fails with gcc-12 (Ubuntu 12.1.0-2ubuntu1~22.04) 12.1.0

```
In file included from uart_channel.c:15:
In function ‘destroy_uart_channel_devs’,
    inlined from ‘deinit_uart_channel’ at uart_channel.c:354:3:
uart_channel.c:313:9: error: pointer ‘c_dev’ used after ‘free’ [-Werror=use-after-free]
  313 |         LIST_FOREACH(c_dev, &c->tty_open_head, open_list) {
      |         ^~~~~~~~~~~~
uart_channel.c:319:17: note: call to ‘free’ here
  319 |                 free(c_dev);
      |                 ^~~~~~~~~~~ 
```

```
hw/pci/gvt.c: In function ‘gvt_init_config.constprop’:
hw/pci/gvt.c:263:41: error: array subscript 257 is above array bounds of ‘uint8_t[256]’ {aka ‘unsigned char[256]’} [-Werror=array-bounds]
  263 |                         gvt->host_config[cap_ptr + 4]);
      |                         ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
hw/pci/gvt.c:48:17: note: while referencing ‘host_config’
   48 |         uint8_t host_config[PCI_REGMAX+1];
      |                 ^~~~~~~~~~~
```

jiaqingz-intel:
[External_System_ID] ACRN-9898

jiaqingz-intel:
Fixed

PR #8383, opened 2023-04-21
misc: life_mngr: fix use-after-free in uart channel

jiaqingz-intel:
LIST_FOREACH() doesn't allow var to be removed or freed within the loop, but c_dev is freed inside the loop here. gcc 12 also reports error on it. This patch uses list_foreach_safe() macro instead for freeing var within the loop safely.

Tracked-On: #8382


ywan170 (approved):
Thanks, @jiaqingz-intel !

misc/services/life_mngr/list.h
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright (C) 2023 Intel Corporation.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <sys/queue.h>
+
+#define list_foreach_safe(var, head, field, tvar)	\
+for ((var) = LIST_FIRST((head));			\
+	(var) && ((tvar) = LIST_NEXT((var), field), 1); \
+	(var) = (tvar))
```
ywan170, line 11: Can we use LIST_FOREACH_SAFE directly?

jiaqingz-intel: No, glibc does not provide `LIST_FOREACH_SAFE` macro.

> LIST_FOREACH_SAFE(), which is present on the BSDs but is not
> present in glibc, fixes this limitation by allowing var to safely
> be removed from the list and freed from within the loop without
> interfering with the traversal.
> https://man7.org/linux/man-pages/man3/list.3.html

Issue #8368, opened 2023-02-13
AttributeError: type object 'XPath2Parser' has no attribute 'SYMBOLS'


saininav:
Latest elementpath version 4.0.1 has removed 'SYMBOLS' attribute, which causing failure in ACRN 3.1.0
https://github.com/sissaschool/elementpath/commit/a76aeb88b0d103420239df1d5d001f7d7f56b5c9

Error log:

misc/config_tools/scenario_config/elementpath_overlay.py", line 24, in CustomParser
|     SYMBOLS = BaseParser.SYMBOLS | {
|               ^^^^^^^^^^^^^^^^^^
| AttributeError: type object 'XPath2Parser' has no attribute 'SYMBOLS'

NanlinXie:
[External_System_ID] ACRN-9902

PR #8384, opened 2023-04-23
config_tools: adapt to elementpath >= 4.0.0

junjiemao1:
Starting from elementpath 4.0.0, the XPath2Parser class no longer has a SYMBOLS member and completeness checking based on that are removed as well. This patch updates the elementpath_overlay in the config tools to fix the compilation issue when using recent elementpath versions.

Tracked-On: #8368


NanlinXie (approved):

wenlingz:
start to run premerge test

PR #8416, opened 2023-05-30
enable VHWP for guests who owns pCPUs

izhouwu:
Background:

Currently CPU frequency control is hidden to guests, and controlled by
hypervisor. While it is sufficient in most cases, some guest OS may
still need CPU performance info to make multi-core scheduling decisions.

This is seen on Linux kernel, which uses HWP highest performance level
as CPU core's priority in multi-core scheduling (CONFIG_SCHED_MC_PRIO).
Enabling this kernel feature could improve performance as single thread
workloads are scheduled on the highest performance cores. This is
significantly useful for guests with hybrid cores.

Design:

The concept is to create a virtual HWP(VHWP) interface for guests, so
that guest can have some CPU performance managing and prioritized MC
scheduling abilities.

For guests owning its pCPUs, we can just allow vCPU read
IA32_HWP_CAPABLITIES and write certain bit fields in IA32_HWP_REQUEST.
But for CPU sharing guests, vCPU’s performance capability involves how
they share CPU time slice, and it is much more complicated. Currently we
only support guest who owns its pCPUs. Windows guests are also excluded
for this feature, as it does not load frequency driver in VM.

Intel_pstate driver also relies on CONFIG_ACPI_CPPC_LIB to implement
prioritized MC scheduling, this means we also need to provide ACPI _CPC
in DM.

Another thing is thermal. Thermal events are delivered through lapic
thermal LVT. Currently ACRN does not support delivering those interrupts
to guests by virtual lapic. They need to be virtualized to provide
guests some thermal management abilities in the future. Currently we
just hide those thermal program interfaces from guests.


izhouwu:
The patch series contains 6 patches. However LOC exceeds CI limit.

The 6/6 [hv: hide thermal interface from guests] will be later opened in another PR.

wenlingz:
@izhouwu  please rebase your code, thanks

izhouwu:
> @izhouwu please rebase your code, thanks

done

NanlinXie (approved):
Approved for configtool part.

chejianj (approved):

hypervisor/arch/x86/guest/vmsr.c
```diff
@@ -672,14 +670,32 @@ int32_t rdmsr_vmexit_handler(struct acrn_vcpu *vcpu)
...
+	case MSR_IA32_PM_ENABLE:
+	case MSR_IA32_HWP_CAPABILITIES:
+	case MSR_IA32_HWP_REQUEST:
+	case MSR_IA32_HWP_STATUS:
+	case MSR_IA32_MPERF:
+	case MSR_IA32_APERF:
+	{
+		if (is_vhwp_configured(vcpu->vm)) {
+			v = msr_read(msr);
+		} else {
```
lifeix, line 694: Could we just return a errno to let the Hypervisor to inject a GP into the guest ?

izhouwu: > Could we just return a errno to let the Hypervisor to inject a GP into the guest ?

I have changed it to returning errno. The new version is pushed.

lifeix (approved):

wenlingz:
start to run premerge test

PR #8421, opened 2023-06-06
doc: add mirror for v2.7 intersphinx inventory file

jiaqingz-intel:
This patch enables intersphinx to pull the objects.inv file directly from projectarcn/projectacrn.github.io repo when github pages was experiencing downtime.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -53,7 +53,8 @@ extlinks = {
 # (We only need to link to an archive that has the older release notes)
 
 intersphinx_mapping = {
-        'v2.7': ('https://projectacrn.github.io/2.7', None),
+        'v2.7': ('https://projectacrn.github.io/2.7',
+                 (None, 'https://raw.githubusercontent.com/projectacrn/projectacrn.github.io/master/2.7/objects.inv')),
         }
 
 graphviz_output_format='png'


NanlinXie:
@dbkinder , Could you help to reivew this PR?

doc/conf.py
```diff
@@ -53,7 +53,8 @@
 # (We only need to link to an archive that has the older release notes)
 
 intersphinx_mapping = {
-        'v2.7': ('https://projectacrn.github.io/2.7', None),
+        'v2.7': ('https://projectacrn.github.io/2.7',
+                 (None, 'https://raw.githubusercontent.com/projectacrn/projectacrn.github.io/master/2.7/objects.inv')),
```
dbkinder, line 57: While this change shouldn't cause any issues (the inventory files should be identical in the two locations), I presume you've had issues with projectacrn.github.io not being available when you tried to generate documentation.  (This intersphinx mapping is used only during doc generation and not while browsing content on the site.)  I've never had an issue with the projectacrn.github.io site being down while generating docs. I'm assuming you have seen this issue enough to introduce this secondary source for the objects.inv file.

BTW, this intersphinx mapping was introduced so we don't have to generate release notes for old releases.  The old release notes referenced documents that no longer existed in the current release, which caused doc build errors.  The v2.7 release was the last one where all releases notes from v2.7 back to v0.1 were generated.

dbkinder (approved):
I'm not sure this is really needed, but it doesn't cause any harm.  (I've never had an issue where the projectacrn.github.io site wasn't available during doc generation, but the github.com site was.

doc/conf.py
jiaqingz-intel: > While this change shouldn't cause any issues (the inventory files should be identical in the two locations), I presume you've had issues with projectacrn.github.io not being available when you tried to generate documentation. (This intersphinx mapping is used only during doc generation and not while browsing content on the site.) I've never had an issue with the projectacrn.github.io site being down while generating docs. I'm assuming you have seen this issue enough to introduce this secondary source for the objects.inv file.

Yesterday github.io was down in our internal network for several hours, that's why I submitted this PR.

> BTW, this intersphinx mapping was introduced so we don't have to generate release notes for old releases. The old release notes referenced documents that no longer existed in the current release, which caused doc build errors. The v2.7 release was the last one where all releases notes from v2.7 back to v0.1 were generated.

Yes I checked git log and was aware of that. I'd like to ask when there is other API changes, or deprecated options, how to generate the link in a clean way? For example, if there are references in v3.2 release note that not available in v3.3, what should be done in such case?

jiaqingz-intel:
Closed as the network issue has been solved

Issue #8425, opened 2023-06-09
Cannot build doc with python 3.11 and doxygen 1.9.4


jiaqingz-intel:
build doc fails on Ubuntu 23.04 (python 3.11.2 + doxygen 1.9.4)

```
acrn-hypervisor/doc/scripts/filter-known-issues.py:39: DeprecationWarning: module 'sre_constants' is deprecated
  import sre_constants
warning: Tag 'LATEX_SOURCE_CODE' at line 1812 of file '-' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'RTF_SOURCE_CODE' at line 1894 of file '-' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'DOCBOOK_PROGRAMLISTING' at line 1992 of file '-' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'CLASS_DIAGRAMS' at line 2190 of file '-' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
acrn-hypervisor/devicemodel/include/pci_core.h:351: warning: found documented return type for pci_generate_msi that does not return anything
acrn-hypervisor/devicemodel/include/pci_core.h:361: warning: found documented return type for pci_generate_msix that does not return anything
acrn-hypervisor/devicemodel/include/pci_core.h:371: warning: found documented return type for pci_lintr_assert that does not return anything
acrn-hypervisor/devicemodel/include/pci_core.h:380: warning: found documented return type for pci_lintr_deassert that does not return anything
acrn-hypervisor/devicemodel/include/pci_core.h:436: warning: found documented return type for pci_set_cfgdata16 that does not return anything
acrn-hypervisor/devicemodel/include/pci_core.h:455: warning: found documented return type for pci_set_cfgdata32 that does not return anything
acrn-hypervisor/devicemodel/include/pci_core.h:417: warning: found documented return type for pci_set_cfgdata8 that does not return anything
acrn-hypervisor/devicemodel/include/virtio.h:533: warning: found documented return type for virtio_config_changed that does not return anything
acrn-hypervisor/devicemodel/include/virtio.h:564: warning: found documented return type for virtio_linkup that does not return anything
[...]

PR #8426, opened 2023-06-09
doc: building doc with python 3.11 + doxygen 1.9.4

jiaqingz-intel:
This patchset enables building document with python 3.11 and doxygen 1.9.4. A major change is that all '@return None' are removed as void-returning functions shall not have any return value documented as required by doxygen 1.9.4 or later versions.

Verified doc can be built successfully on Ubuntu 23.04 (python 3.11.2 + doxygen 1.9.4) and Ubuntu 22.04 (python 3.10.6 + doxygen 1.9.1)

Tracked-On: #8425

### DIFF
--- a/devicemodel/hw/pci/core.c
+++ b/devicemodel/hw/pci/core.c
@@ -2012,8 +2012,6 @@ pci_msix_enabled(struct pci_vdev *dev)
  *
  * @param dev Pointer to struct pci_vdev representing virtual PCI device.
  * @param index MSIx table entry index.
- *
- * @return None
  */
 void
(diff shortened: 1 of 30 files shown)


NanlinXie:
@dbkinder , please help to review, thanks!

dbkinder (approved):
Thanks @jiaqingz-intel 
Looks OK to me, though I haven't tried using Ubuntu 23.04 or the later versions of Python.  While not critical, you might want to update tutorials/docbuild.rst with any more information about versions we've tested to work.

NanlinXie (approved):

PR #8427, opened 2023-06-09
enable VHWP for guests who owns pCPUs(part 2)

izhouwu:
Background:

Currently CPU frequency control is hidden to guests, and controlled by
hypervisor. While it is sufficient in most cases, some guest OS may
still need CPU performance info to make multi-core scheduling decisions.

This is seen on Linux kernel, which uses HWP highest performance level
as CPU core's priority in multi-core scheduling (CONFIG_SCHED_MC_PRIO).
Enabling this kernel feature could improve performance as single thread
workloads are scheduled on the highest performance cores. This is
significantly useful for guests with hybrid cores.

Design:

The concept is to create a virtual HWP(VHWP) interface for guests, so
that guest can have some CPU performance managing and prioritized MC
scheduling abilities.

For guests owning its pCPUs, we can just allow vCPU read
IA32_HWP_CAPABLITIES and write certain bit fields in IA32_HWP_REQUEST.
But for CPU sharing guests, vCPU’s performance capability involves how
they share CPU time slice, and it is much more complicated. Currently we
only support guest who owns its pCPUs. Windows guests are also excluded
for this feature, as it does not load frequency driver in VM.

Intel_pstate driver also relies on CONFIG_ACPI_CPPC_LIB to implement
prioritized MC scheduling, this means we also need to provide ACPI _CPC
in DM.

Another thing is thermal. Thermal events are delivered through lapic
thermal LVT. Currently ACRN does not support delivering those interrupts
to guests by virtual lapic. They need to be virtualized to provide
guests some thermal management abilities in the future. Currently we
just hide those thermal program interfaces from guests.

### DIFF
--- a/hypervisor/arch/x86/guest/vlapic.c
+++ b/hypervisor/arch/x86/guest/vlapic.c
@@ -862,9 +862,9 @@ vlapic_trigger_lvt(struct acrn_vlapic *vlapic, uint32_t lvt_index)
 {
 	uint32_t lvt;
 	int32_t ret = 0;
+	struct acrn_vcpu *vcpu = vlapic2vcpu(vlapic);
 
 	if (!vlapic_enabled(vlapic)) {
-		struct acrn_vcpu *vcpu = vlapic2vcpu(vlapic);
 		/*
 		 * When the local APIC is global/hardware disabled,
 		 * LINT[1:0] pins are configured as INTR and NMI pins,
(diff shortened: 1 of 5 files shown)


izhouwu:
The patch series contains 6 patches. Due to CI LOC limit, the PR is split into 2 parts

This is part2 6/6 [hv: hide thermal interface from guests]

PR #8430, opened 2023-06-16
hv: vpci: ignore PCI I/O BAR with non-zero upper 16 bits

jiaqingz-intel:
On x86 platform, the upper 16 bit of I/O BAR should be initialized to
zero by BIOS. Howerever, some buggy BIOS still programs the upper 16
bits to non-zero, which causes error in check_pt_dev_pio_bars(). Since
I/O BAR reprogramming by VM is currently unsupported, this patch
ignores such I/O BARs when creating vpci devices to make VM boot.

Tracked-On: #8373


dongyaozu:
My comments are ignored?

dongyaozu (approved):

Issue #8432, opened 2023-06-20
Flickering screen when passing ADL-N and RPL-P platforms


jiaqingz-intel:
Black screen or flickering screen when passing ADL-N and RPL-P platforms

And kernel i915 driver reports error with `ERROR Fault errors on pipe A: 0x00000080`

jiaqingz-intel:
[External_System_ID] ACRN-10146

jiaqingz-intel:
Reopen due to the solution/workaround is not merged in master

PR #8433, opened 2023-06-21
dm: passthrough DSM region for ADL-N and RPL

jiaqingz-intel:
The Data of Stolen Memory (DSM) region on Alder Lake-N and Raptor Lake platform is indicated by the BDSM register (0xC0 and 0xC4 in PCI config space), which is the same as Gen 11 (Tiger Lake) iGPU. This patch adds ADL-N and RPL iGPU device id to passthrough the DSM region properly when using GVT-d.

The PCI device ids are taken from i915 kernel driver.

Tracked-On: #8432


chejianj (approved):

wenlingz:
start to run premerge test

Issue #8448, opened 2023-07-10
[doc] The script in Sample Application Guide is not working


jiaqingz-intel:
1. Cannot install `linux-modules-extra-$(uname -r)` as it runs in chroot environment, `uname` reads host kernel version
2. Need to install new kernel to support GVT-d on 13th Gen Intel Processors
3. The documentation need to be updated about how to enable acrn-br0 for devicemodel.

jiaqingz-intel:
Fix merged

PR #8449, opened 2023-07-10
doc: replace qemu-kvm with qemu-utils in sample application guide

jiaqingz-intel:
The create_image.sh script uses qemu-img to create image files. Users only need to install qemu-image in qemu-utils instead of the whole qemu package.

Tracked-On: #8448


NanlinXie (approved):

dbkinder (approved):
Changes look OK to me

PR #8451, opened 2023-07-11
doc: building doc with python 3.11 + doxygen 1.9.4

jiaqingz-intel:
This patchset enables building document with python 3.11 and doxygen 1.9.4. A major change is that all '@return None' are removed as void-returning functions shall not have any return value documented as required by doxygen 1.9.4 or later versions.

Verified doc can be built successfully on Ubuntu 23.04 (python 3.11.2 + doxygen 1.9.4) and Ubuntu 22.04 (python 3.10.6 + doxygen 1.9.1)

Tracked-On: #8425

### DIFF
--- a/devicemodel/hw/pci/core.c
+++ b/devicemodel/hw/pci/core.c
@@ -2012,8 +2012,6 @@ pci_msix_enabled(struct pci_vdev *dev)
  *
  * @param dev Pointer to struct pci_vdev representing virtual PCI device.
  * @param index MSIx table entry index.
- *
- * @return None
  */
 void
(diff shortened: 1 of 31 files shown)


jiaqingz-intel:
This reopens #8426, the original one was closed by accident

dbkinder (approved):
I think this change is OK.  Have you tested it with Doxygen 1.9.1 as well?

jiaqingz-intel:
> I think this change is OK. Have you tested it with Doxygen 1.9.1 as well?

It works with 1.9.1

Issue #8507, opened 2023-09-11
Halting Physical CPUs on Idle


izhouwu:
When there is no active vCPU on a physical core, the hypervisor shall trigger entering C-states (at least C1 but can be deeper) on that physical core.
This is to reduce CPU idle power.

izhouwu:
[External_System_ID] ACRN-9962

PR #8508, opened 2023-09-11
hv: switch to dynamic timer in bvt scheduler

izhouwu:
When bvt scheduler picks up a thread to run, it sets up a counter ‘run_countdown’ to determine how many ticks it should remain running. Then the timer will decrease run_countdown by 1 on every 1000Hz tick interrupt, until it reaches 0. The tick interrupt consumes a lot of power during idle (if we are using HLT in idle thread).

This patch is to switch the 1000 HZ timer to a dynamic one, which only interrupt on run_countdown expires.

Tracked-On: #8507

### DIFF
--- a/hypervisor/common/sched_bvt.c
+++ b/hypervisor/common/sched_bvt.c
@@ -21,8 +21,6 @@ struct sched_bvt_data {
 	uint16_t weight;
 	/* virtual time advance variable, proportional to 1 / weight */
 	uint64_t vt_ratio;
-	/* the count down number of mcu until reschedule should take place */
-	uint64_t run_countdown;
 	/* actual virtual time in units of mcu */
 	int64_t avt;
 	/* effective virtual time in units of mcu */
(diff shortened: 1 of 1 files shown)


hypervisor/common/sched_bvt.c
```diff
@@ -166,12 +158,7 @@ static int sched_bvt_init(struct sched_control *ctl)
 
 	/* The tick_timer is periodically */
 	initialize_timer(&bvt_ctl->tick_timer, sched_tick_handler, ctl,
-			cpu_ticks() + tick_period, tick_period);
-
-	if (add_timer(&bvt_ctl->tick_timer) < 0) {
-		pr_err("Failed to add schedule tick timer!");
-		ret = -1;
-	}
+			0, 0);
```
junjiemao1, line 161: Just move the two parameters to the same line as `initialize_timer`.

junjiemao1 (approved):

PR #8527, opened 2023-10-18
doc: specify the python pkg version

Kunhui-Li:
specify the python pkg version using the command
'sudo pip3 install "elementpath==2.5.0" lxml xmlschema defusedxml tqdm' because fix PR #8384 is not in v3.2 branch.


NanlinXie (approved):

NanlinXie:
Close this since there's no need for this.

PR #8529, opened 2023-10-19
Fix Debian packaging postinst partition finding

xzhangxa:
If the root partition is bind mounted with / and another, the current postinst script (using command lsblk) will fail to find the partition: $type will be "/" only and cause the following command may find the wrong partition.

And current command forces the root partition to be ext4.

This patch fixes the two issues.

Two setups with bind mounted root partition;

Ubuntu 22.04 desktop, bind mount firefox snap:
```
> lsblk
nvme0n1      259:19   0 931.5G  0 disk
├─nvme0n1p1  259:20   0   243M  0 part /boot/efi
├─nvme0n1p2  259:21   0 927.5G  0 part /var/snap/firefox/common/host-hunspell
│                                      /
```

Windows WSL2, bind mount WSLg:
```
> lsblk
sdc    8:32   0     1T  0 disk /mnt/wslg/distro
                               /
```

Tracked-On: #8532

### DIFF
--- a/misc/packaging/acrn-hypervisor.postinst
+++ b/misc/packaging/acrn-hypervisor.postinst
@@ -21,7 +21,7 @@ GENERATED_PARAMS=(cpu_perf_policy=Performance)
 #ACRN parameters End
 
 ACRNBIN="/boot/acrn.${SCENARIO}.${BOARD}.bin"
-type=$(lsblk -l |awk '$NF == "/" {print $1}')
+type=$(findmnt / |awk '$1 == "/" {print $2}')
 
 pattern='^/dev/.* UUID="([^"]+)".* PARTUUID="([^"]+)"'
 while IFS= read -r line; do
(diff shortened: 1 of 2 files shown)


shuaizhx:
@xzhangxa please add Tracked-On

xzhangxa:
> @xzhangxa please add Tracked-On

Tracked-On: https://github.com/projectacrn/acrn-hypervisor/issues/8532

junjiemao1 (approved):

xzhangxa:
I'd like to change the code a little from using `df` to `findmnt /`. When using `df` if the kernel cmdline has `root=/dev/xxx` the `df` command will show "/" on partition "/dev/root", but we need the real hardware partition. So it won't work in this case (it's also the case that if you install a new acrn-hypervisor deb under the previous acrn hypervisor+SOS).

So `lsblk` fails when bind mount root partition and `df` fails when kernel cmdline has explicit "root=/dev/xxx".

Change to use `findmnt /` instead, it will always report the real hardware partition.

yuhuanX:
start to run premerge test

yuhuanX:
Please 'terryzouhao' or 'NanlinXie' approve this patch.

NanlinXie (approved):

yuhuanX:
start to run premerge test

yuhuanX:
start to run premerge test

yuhuanX:
start to run premerge test

Issue #8547, opened 2024-01-19
Notification of VM Events


izhouwu:
When any of the following VM event occur, the hypervisor and device model shall send an event in JSON format through the device model command monitor. Events shall encode the event type and the information mentioned below.

Changes to RTC (offset of new vRTC value to the old value is required).
Crashes due to triple fault.
Watchdog timeout.
Shutdown.
Reboot.

izhouwu:
[External_System_ID] ACRN-9966

PR #8556, opened 2024-02-04
hv: add max_len for sbuf_put param

izhouwu:
sbuf_put copies sbuf->ele_size of data, and puts into ring. Currently this function assumes that data size from caller is no less than sbuf->ele_size.

But as sbuf->ele_size is usually setup by some sources outside of the HV (e.g., the service VM), it is not meant to be trusted. So caller should provide the max length of the data for safety reason. sbuf_put() will return UINT32_MAX if max_len of data is less than element size.

Additionally, a helper function sbuf_put_many() is added for putting multiple entries.

Tracked-On: #8547

### DIFF
--- a/hypervisor/common/sbuf.c
+++ b/hypervisor/common/sbuf.c
@@ -27,8 +27,13 @@ uint32_t sbuf_next_ptr(uint32_t pos_arg,
 
 /**
  * The high caller should guarantee each time there must have
- * sbuf->ele_size data can be write form data and this function
- * should guarantee execution atomically.
+ * sbuf->ele_size data can be write form data.
+ *
+ * As sbuf->ele_size is possibly setup by some sources outside of the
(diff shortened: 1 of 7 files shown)


hypervisor/common/sbuf.c
```diff
@@ -104,3 +113,29 @@ int32_t sbuf_setup_common(struct acrn_vm *vm, uint16_t cpu_id, uint32_t sbuf_id,
 
 	return ret;
 }
+
+/* try put a batch of elememts from data to sbuf
+ * data_size should be equel to n*elem_size, data not enough to fill the elem_size will be ignored.
```
junjiemao1, line 118: It seems that constraint does not hold for the callsites in profiling.c. Shall we properly handle the case where `data_size` is not a multiple of `elem_size`?

junjiemao1 (approved):

PR #8561, opened 2024-02-27
dm: fix LPC UART no output issue after reboot

zhangqiang4:
The meaning of lpc_uart_vdev::enabled was changed from runtime enablement status of UARTs to user configuration in commit 852f10cc3. So it's incorrect to clear it in lpc_deinit() which will cause UARTs disabled after reboot.

Tracked-On: #8537
Fixes: 852f10cc3 ("dm: lpc: only emulate COM ports specified in command line")


zhangqiang4:
@chejianj Please help approve this PR. This commit is simple and got ACKed by Jiaqing.

yuhuanX:
start to run premerge test

Issue #8564, opened 2024-03-08
Cannot send control keys such as up/down arrow to VM virtual console


izhouwu:
Use up/down arrow in guest VM virtual console does not get expected result:

 

ACRN:\>
ACRN:\>vm_console 0

----- Entering VM 0 Shell -----

root@intel-corei7-64:~# ls

(press up arrow)
root@intel-corei7-64:~#]A

izhouwu:
[External_System_ID] ACRN-10290

PR #8565, opened 2024-03-08
hv: console reads all input chars in one poll

izhouwu:
For uart console, some control keys are defined as byte sequences, such as:
 * up arrow - 0x1b/0x5b/0x41
 * F8 - 0x1b/0x5b/0x31/0x39/0x7e

Currently hv console only read one char per poll.
When guest vuart console is active, those byte sequences may not be sent to guest vuart in good timing due to the poll interval. Thus control keys such as up/down can not be used in shell or vim.

The solution is to read all input chars in one poll, so that control keys can be received by guest OS properly.

Tracked-On: #8564


junjiemao1 (approved):

yuhuanX:
start to run premerge test

Issue #8580, opened 2024-04-17
Return value for default case in hcall_vm_intr_monitor() is not right


yisun-git:
In hcall_vm_intr_monitor(), the default case for intr_hdr->cmd is a wrong case. So, it should return error code back. But it returns success code 0 in current codes.

PR #8581, opened 2024-04-18
hv: return error code for default case in hcall_vm_intr_monitor

yisun-git:
In hcall_vm_intr_monitor(), the default case for intr_hdr->cmd is a wrong case. So, it should return error code back. But it returns success code 0 in current codes.

Tracked-On: #8580


yuhuanX:
start to run premerge test

PR #8597, opened 2024-05-13
misc: fix v-model used on props for vue3 make strictly checking

dongpingx:
I make this change during recent building configurator but got a failure. The change is to fix the problem caused by vue3's strictly syntax chacking. It is reported for newly releases only and before v3.2.25 it is a warning. I follow the error message and suggestion to change the code. Error message said "v-model cannot be used on a prop, because local prop bindings are not writable".
Suggestion said "Use v-bind binding combined with v-on listener to emit update" I ran the build & install procedure to verify the problem was solved.

Tracked-on: #8596

### DIFF
--- a/misc/config_tools/configurator/packages/configurator/src-tauri/Cargo.lock
+++ b/misc/config_tools/configurator/packages/configurator/src-tauri/Cargo.lock
@@ -1362,7 +1362,7 @@ dependencies = [
  "futures-sink",
  "futures-util",
  "http",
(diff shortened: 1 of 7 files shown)


yuhuanX:
@dongpingx , please use Tracked-On, instead of Tracked-on

NanlinXie (approved):

yuhuanX:
start to run premerge test

PR #8603, opened 2024-05-27
hv: vpci: clear vdev structure on device deassign

jiaqingz-intel:
In devicemodel, a passthrough device is deassigned and then assigned to guest on guest reboot. Each time hypervisor allocates a new pci_vdev structure to keep its info. As it was stored in a statically-allocated array, it will eventually use up all slots, resulting both resource leak and out-of-bounds access.

Fix it by clearing the corresponding vdev structure on device deassign, thus a bitmap is introduced to track the usage, replacing the existing array count.

Tracked-On: #8590

### DIFF
--- a/hypervisor/dm/vpci/vpci.c
+++ b/hypervisor/dm/vpci/vpci.c
@@ -278,7 +278,7 @@ void deinit_vpci(struct acrn_vm *vm)
 	struct pci_vdev *vdev, *parent_vdev;
 	uint32_t i;
 
-	for (i = 0U; i < vm->vpci.pci_vdev_cnt; i++) {
+	for (i = 0U; i < CONFIG_MAX_PCI_DEV_NUM; i++) {
 		vdev = (struct pci_vdev *) &(vm->vpci.pci_vdevs[i]);
 
 		/* Only deinit the VM's own devices */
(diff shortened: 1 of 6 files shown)


yuhuanX:
@jiaqingz-intel ，code scan with clang-tidy  fail.

acrn-hypervisor/hypervisor/dm/vpci/vpci.c:669:3: warning: vpci_init_vdev() has multiple return statements [acrn-c-cu-02]
                return NULL;
                ^
acrn-hypervisor/hypervisor/dm/vpci/vpci.c:694:2: note: first return statement in the same function
        return vdev;
        ^
acrn-hypervisor/hypervisor/dm/vpci/vroot_port.c:145:5: warning: create_vrp() has multiple return statements [acrn-c-cu-02]
                                return -EFAULT;
                                ^
acrn-hypervisor/hypervisor/dm/vpci/vroot_port.c:154:2: note: first return statement in the same function
        return 0;
        ^

hypervisor/dm/vpci/vpci.c
```diff
@@ -730,62 +752,73 @@ int32_t vpci_assign_pcidev(struct acrn_vm *tgt_vm, struct acrn_pcidev *pcidev)
...
-		}
+	if ((vdev_in_service_vm == NULL) || (vdev_in_service_vm->user != vdev_in_service_vm) ||
+			(vdev_in_service_vm->pdev == NULL) ||
+			is_host_bridge(vdev_in_service_vm->pdev) || is_bridge(vdev_in_service_vm->pdev)) {
+		pr_fatal("%s, can't find PCI device %x:%x.%x for vm[%d] %x:%x.%x\n", __func__,
+			pcidev->phys_bdf >> 8U, (pcidev->phys_bdf >> 3U) & 0x1fU, pcidev->phys_bdf & 0x7U,
+			tgt_vm->vm_id,
+			pcidev->virt_bdf >> 8U, (pcidev->virt_bdf >> 3U) & 0x1fU, pcidev->virt_bdf & 0x7U);
+		ret = -ENODEV;
+		goto sos_out;
```
lifeix, line 763: Please don't use goto.

lifeix (changes requested):

hypervisor/dm/vpci/vpci.c
jiaqingz-intel: Done

junjiemao1:
I think we should also populate the error at the caller of vpci_init_vdev() in create_ivshmem_vdev() and create_vmcs9900_vdev(). They can be called via hypercalls and the service VM needs to know when it fails to create a virtual device.

As for vpci_init_vdevs(), you may want to leave a comment why the error can be silently ignored (probably let the VM initialize with the restricted number of virtual devices).

hypervisor/dm/vpci/vroot_port.c
```diff
@@ -140,14 +141,19 @@ int32_t create_vrp(struct acrn_vm *vm, struct acrn_vdev *dev)
 			dev_config->vdev_ops = &vrp_ops;
 
 			vdev = vpci_init_vdev(&vm->vpci, dev_config, NULL);
+			if (vdev == NULL) {
+				pr_err("%s: failed to create virtual root port\n", __func__);
+				ret = -EFAULT;
+				break;
+			}
```
junjiemao1, line 148: There are two other calls to vpci_init_vdev() without checks on the returned pointer, one in create_vmcs9900_vdev() and the other in create_ivshmem_vdev(). Both can be called via hypercalls, and thus require the errors to be reported.

hypervisor/dm/vpci/vroot_port.c
```diff
@@ -140,14 +141,19 @@ int32_t create_vrp(struct acrn_vm *vm, struct acrn_vdev *dev)
...
 
 			break;
 		}
 	}
 
-	return 0;
+	return ret;
 }
 
 int32_t destroy_vrp(__unused struct pci_vdev *vdev)
```
junjiemao1: Need call vpci_deinit_vdev() in destroy_vrp().

junjiemao1 (changes requested):

hypervisor/dm/vpci/vroot_port.c
jiaqingz-intel: Fixed in 2nd commit

junjiemao1 (approved):

yuhuanX:
start to run premerge test

yuhuanX:
start to run premerge test

Issue #8617, opened 2024-05-30
0-sized e820 entry after e820_alloc_region()


jiaqingz-intel:
```
[0us][cpu=0][(null)][sev=1][seq=173]:e820_alloc_region: after hv_e820:
[0us][cpu=0][(null)][sev=1][seq=174]:hv_e820[0]:type: 0x1 Base: 0x0000000000000000 length: 0x000000000009e000
[0us][cpu=0][(null)][sev=1][seq=175]:hv_e820[1]:type: 0x2 Base: 0x000000000009e000 length: 0x0000000000062000
[0us][cpu=0][(null)][sev=1][seq=176]:hv_e820[2]:type: 0x1 Base: 0x0000000000100000 length: 0x0000000002100000
[0us][cpu=0][(null)][sev=1][seq=177]:hv_e820[3]:type: 0x1 Base: 0x0000000002e1e000 length: 0x0000000000000000
[0us][cpu=0][(null)][sev=1][seq=178]:hv_e820[4]:type: 0x1 Base: 0x0000000006e17000 length: 0x000000001e53c000
[0us][cpu=0][(null)][sev=1][seq=179]:hv_e820[5]:type: 0x2 Base: 0x0000000025353000 length: 0x0000000000080000
[0us][cpu=0][(null)][sev=1][seq=180]:hv_e820[6]:type: 0x1 Base: 0x00000000253d3000 length: 0x0000000006a2d000
[0us][cpu=0][(null)][sev=1][seq=181]:hv_e820[7]:type: 0x2 Base: 0x000000002be00000 length: 0x0000000000400000
[0us][cpu=0][(null)][sev=1][seq=182]:hv_e820[8]:type: 0x1 Base: 0x000000002c200000 length: 0x00000000001ed000
[0us][cpu=0][(null)][sev=1][seq=183]:hv_e820[9]:type: 0x2 Base: 0x000000002c3ed000 length: 0x00000000002fe000
[0us][cpu=0][(null)][sev=1][seq=184]:hv_e820[10]:type: 0x1 Base: 0x000000002c6eb000 length: 0x0000000000001000
[0us][cpu=0][(null)][sev=1][seq=185]:hv_e820[11]:type: 0x2 Base: 0x000000002c6ec000 length: 0x0000000001c57000
[0us][cpu=0][(null)][sev=1][seq=186]:hv_e820[12]:type: 0x1 Base: 0x000000002e343000 length: 0x000000000566a000
[...]

PR #8628, opened 2024-06-17
e820: properly reserve memory for multiboot modules

jiaqingz-intel:
In current implementation, if there are multiple continous 4k-aligned modules, 0-sized e820 entries will be created between these regions. And for non-4k-aligned modules, when two of them are located in one page, the second memory range will not be reserved as it was not in one e820 entry after the first is reserved, making it vulnerable.

This patch fixes it by marking the exact memory range of multiboot modules as unusable first, then shrinking the e820 entries to page boundary. If the module crosses multiple e820 entries, possibly due to a buggy bootloader, hypervisor will panic immediately to prevent modules getting corrupted.

Tracked-On: #8617

### DIFF
--- a/hypervisor/arch/x86/e820.c
+++ b/hypervisor/arch/x86/e820.c
@@ -119,43 +119,6 @@ static void insert_e820_entry(uint32_t index, uint64_t addr, uint64_t length, ui
 	hv_e820[index].type = type;
 }
 
-static uint64_t e820_alloc_region(uint64_t addr, uint64_t size)
-{
-	uint32_t i;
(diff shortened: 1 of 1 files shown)


wenlingz:
@lifeix @junjiemao1 please help approve this PR, thanks

junjiemao1 (approved):

yuhuanX:
start to run premerge test

yuhuanX:
start to run premerge test

yuhuanX:
start to run premerge test

Issue #8652, opened 2024-07-11
HV: remove Service VM delayed loading


haiwei-li:
Now multiboot modules memory is already reserved from e820 in function `alloc_mods_memory()` and Service VM will not corrupt pre-launched VM modules.

So remove the code of Service VM delayed loading.

PR #8655, opened 2024-07-17
hv: remove Service VM delayed loading

haiwei-li:
Now multiboot modules memory is already reserved from e820 in function `alloc_mods_memory()` and Service VM will not corrupt pre-launched VM modules.

So remove the code of Service VM delayed loading.

Tracked-On: #8652


lifeix (approved):

yuhuanX:
start to run premerge test

Issue #8657, opened 2024-07-17
Configurator Main VM Support


dongpingx:
Support Main VM, i.e., the Main VM scenario is similar to partitioned scenario, which select one VM as Main VM and owns the physical resources.

PR #8658, opened 2024-07-17
misc: Configurator Main VM Support

dongpingx:
This patch is to support Main VM, i.e., the Main VM scenario is similar to paritioned scenario, which select one VM as Main VM and owns the physical resources.

I extracted all pcis from the board, substracted them with dispatched ones and then make an assignment to replace the old enums.

This is implemented through two methods called updateSchema which is to update schema on the current VM and updateLoadSchema which is to maintain pcis, i.e. removed the pcis listed on the pre-launched VM, while loading scenario.xml.

I tested locally and confirmed the functionalties above are implemented.

Tracked-On: #8657


NanlinXie (approved):

yuhuanX:
start to run premerge test

Issue #8659, opened 2024-07-17
update libwebkit2gtk install method & remove unnecessary package


dongpingx:
1. update libwebkit2gtk install method
2. remove unnecessary package: libappindicator3-dev

PR #8660, opened 2024-07-17
doc: update libwebkit2gtk install method & remove unnecessary package

dongpingx:
1.We updated libwebkit2gtk-4.0-dev install method for Canonical had removed that package already, but Tauri need it, so we advice the user to download the dependencies through Ubuntu22.04's repo or adopt our links directly.

2.remove unnecessary package: libappindicator3-dev, we found it is possible to build acrn without that dependency, so we remove it now.

Tracked-On: #8659


yuhuanX:
start to build

NanlinXie (approved):

PR #8676, opened 2024-07-30
[release_3.3] Remove debconf call in grub script

jiaqingz-intel:
Calling debconf in grub script may fail when installing with dpkg as
debconf database might be occupied by dpkg, causing update-grub fail
with message below

```
debconf: DbDriver "config": /var/cache/debconf/config.dat is locked by another process: Resource temporarily unavailable
```

This patchset fixes issue above by creating a symlink to scenario file
at fixed path /usr/share/acrn/scenario.xml and let grub-acrn script
read scenario here so that debconf is no longer called.


NanlinXie (approved):

yuhuanX:
start to run premerge test

Issue #8621, opened 2024-06-14
trace: hv shall provide a tool that can trace vcpu usage in pcpu sharing case


nacui-intel:
Show cpu usage by acrntrace when different vms share the same physical cpu.

PR #8686, opened 2024-08-07
doc: add cpu usage in acrntrace

nacui-intel:
This patch adds the description of newly added parameter of acrntrace analyzer, as well as the user guide that how to analyze the cpu usage of VMs in cpu sharing case.

Tracked-On: #8621


yuhuanX:
start to build

yuhuanX:

PR #8703, opened 2024-08-19
doc: Document requirements for pre-launched VM

daichifukui:
Hi all,

For a pre-launched VM, it is essential to include 'reboot=acpi' in the 'bootargs' of the scenario.xml file. Additionally, if the VM is named VM0, the grub.cfg file must contain
'module2 /boot/ACPI_VM0.bin ACPI_VM0'.

Currently, several scenario files located in misc/config_tools/data already support 'reboot=acpi'. Furthermore, the postinst Debian script automatically adds the 'module2' command to grub.cfg if the scenario is 'partitioned'. Therefore, by using the provided scenarios, the pre-launched VM can be used without any additional configuration.

However, when using a custom scenario file, there is a risk of omitting the necessary parameter and command for a pre-launched VM. To avoid such errors, it is useful to document these requirements.

Hence this pull request.  
I'd appreciate it if you could review this request and merge it with the upstream.

### DIFF
--- a/doc/introduction/index.rst
+++ b/doc/introduction/index.rst
@@ -272,6 +272,10 @@ can define your own configuration scenarios.
   are automatically launched at boot time by the hypervisor.  The User VMs can
   be Real-Time VMs (RTVMs), Safety VMs, or standard User VMs.
 
+  When selecting this scenario, ensure that ``reboot=acpi`` is included in the
+  ``bootargs`` of ``scenario.xml``. Additionally, if the target VM is named VM0,
+  ``grub.cfg`` must contain a menuentry with ``module2 /boot/ACPI_VM0.bin ACPI_VM0``.
+
 * **Hybrid** scenario simultaneously supports both sharing and partitioning on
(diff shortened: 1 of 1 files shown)


yuhuanX:

yuhuanX:
start to build

yuhuanX:
start to run premerge test

doc/introduction/index.rst
```diff
@@ -272,6 +272,10 @@ can define your own configuration scenarios.
   are automatically launched at boot time by the hypervisor.  The User VMs can
   be Real-Time VMs (RTVMs), Safety VMs, or standard User VMs.
 
+  When selecting this scenario, ensure that ``reboot=acpi`` is included in the
```
jsun26intel, line 275: The document here is for ACRN High-Level Architecture. The scenario doesn't care the RTVM OS is Linux or others. But your patch specifies the "reboot=acpi" cmdline which works for Linux only.

Issue #8598, opened 2024-05-17
ENODEV should be able to be set into RAX as hypercall return value


yisun-git:
Some hypercalls return -ENODEV which should be set into RAX as return value, e.g. HC_ASSIGN_PCIDEV. So, remove the check in vmcall_vmexit_handler() and change return value to -EACCESS if the hypercall is not sent from Service VM or allowed VM.

PR #8705, opened 2024-08-22
hv: ENODEV should be able to be set into RAX as hypercall return value

yisun-git:
Some hypercalls return -ENODEV which should be set into RAX as return value, e.g. HC_ASSIGN_PCIDEV. So, remove the check in vmcall_vmexit_handler() and change return value to -EACCESS if the hypercall is not sent from Service VM or allowed VM.

Tracked-On: #8598


junjiemao1 (approved):

yuhuanX:
start to run premerge test

PR #8706, opened 2024-08-26
doc: add IVSHMEM region ID support

yonghuah:
Add guide to configure IVSHMEM Region ID.


jsun26intel (changes requested):
we don't need to expose internal usage to public. Preserve 0~0xfff is enough.

PR #8720, opened 2024-08-30
misc: fail to save config_summary.rst while changing board.xml

dongpingx:
This patch is to fix the bug which will report while changing board.xml.

The former coder added data into the list without removing old ones, and the file config_summary.rst won't be dropped before generating a new one.

I removed the old data when I want to close the file config_summary.rst.

I tested by change board.xml with different pcpus and make confirmation the result is correct now.

Tracked-On: #8718

### DIFF
--- a/misc/config_tools/scenario_config/config_summary.py
+++ b/misc/config_tools/scenario_config/config_summary.py
@@ -448,7 +448,9 @@ class GenerateRst:
 
     # Close the Rst file after all information is written.
     def close_file(self):
-        self.file.close()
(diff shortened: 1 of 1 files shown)


yuhuanX:

6ZhangWei:
Can close this PR. Redo this patch in https://github.com/projectacrn/acrn-hypervisor/pull/8780

yuhuanX:
start to run premerge test

yuhuanX:
start to build

PR #8722, opened 2024-08-30
misc: configurator need to support configurable reg width for UART

dongpingx:
We found sometime we need to configure reg width for mmio and we change the width by hand for workaround, now we can configure the width by configurator.

I defined a macro named CONFIG_SERIAL_MMIO_REG_WIDTH in uart16550.c, collected mmio ports, loaded them and made a judgement whether the current serial_console was a mmio serial, it will show an option for mmio and hide for other serial ports.

I tested three sorts of serials: portio, pci and mmio, built acrn and checked the content generated in build/hypervisor/include/config.h, I confirmed the results are correct.

Tracked-On: #8721


yuhuanX:

yuhuanX:
start to run premerge test

Issue #8566, opened 2024-03-17
Add region ID support for IVSHMEM Devices


yonghuah:
1. Use “Subsystem ID” in PCI configuration space to indicate Link ID of IVSHMEM devices
2. “Subsystem Vendor ID” will be hardcoded to “0x8086”
3. Format to define IVSHMEM regions will be revised:

Link ID is required when configuring IVHSMEM instance:
Applicable to both DM-land and HV-land IVSHMEM devices.
4. Impacted components: ACRN Hypervisor/ Device Model and ACRN config tool.

yonghuah:
[External_System_ID] ACRN-10265

PR #8646, opened 2024-07-03
config_tools: Add ivshmem region ID to launch script

zhangckid:
Add missed ivshmem region ID.

Tracked-On: #8645

Reviewed-by: Junjie Mao <junjie.mao@intel.com>

PR #8736, opened 2024-09-20
config-tools: add IVSHMEM Region ID in guest launch script

yonghuah:
Add region ID when generating guest launch script file if IVSHMEM device is configured.

Tracked-On: #8566


yonghuah:
duplicated with below PR#
https://github.com/projectacrn/acrn-hypervisor/pull/8646

Issue #8761, opened 2025-02-24
improper usage of stac() and clac() in get_initrd_load_addr()


haoyu-tang:
**Describe the bug**
get_initrd_load_addr() called stac() twice, but doesn't call clac().

**Platform**
All

**Codebase**
Master

**Scenario**
All

**To Reproduce**
By code review.

**Expected behavior**
The second call of stac() should be clac().

**Additional context**
N/A

PR #8762, opened 2025-02-24
hv: fix stac/clac context in get_initrd_load_addr()

haoyu-tang:
Tracked-On: #8761


yuhuanX:
start to run premerge test

Issue #8771, opened 2025-04-25
ACPI PM register emulation is not working on QEMU


jiaqingz-intel:

jiaqingz-intel:
[External_System_ID] ACRN-12151

PR #8772, opened 2025-04-25
hv: pm: fix acpi register size calculation

jiaqingz-intel:
The Access Size field in ACPI GAS was not introduced before ACPI 2.0, Errata C. It is not guaranteed to be a non zero value, like QEMU programs it to 0. As it only indicates how many bytes it can be accessed at once, the register size should be determined by Bit Width and Bit Offset. In IO space, Bit Offset is always 0, the size is (Bit Width / 8).

Tracked-On: #8771

Reviewed-by: Li Fei <fei1.li@intel.com>


yuhuanX:
start to build

yuhuanX:
start to run premerge test

Issue #8737, opened 2024-09-23
RTVM: using io_uring can not bringup


476997779:
**Describe the bug**

**Platform**
Describe the platform information, 

**Codebase**
Describe which codebase is used.

**Scenario**
Describe the scenario: Hybrid, Industry or logical_partition

**To Reproduce**

**Expected behavior**

**Additional context**
Attach the following logs if available:

- native Linux boot kernel message log
- ACRN serial port log (when issues happens)
- Service VM kernel message when issue happens
- acrn-dm log: all files in /var/log/acrn-tm
- console log when running script to launch guest VM

PR #8775, opened 2025-05-20
dm: fix the issue that RTVM cannot bring up with io uring

YuanXin-Intel:
Device model uses polling mode to fetch the virtio blk request in RTVM. When RTVM brings up with io uring, the threads handling io uring and vq are not same, which would cause competition. To fix this issue, device model should handle vq and io uring in the same thread to avoid conflict.

Tracked-On: #8737

Signed-off-by: YuanXin-Intel <xin.yuan@intel.com>
Reviewed-by: Jian Jun Chen <jian.jun.chen@intel.com>

### DIFF
--- a/devicemodel/hw/pci/virtio/virtio.c
+++ b/devicemodel/hw/pci/virtio/virtio.c
@@ -84,7 +84,7 @@ void iothread_handler(void *arg)
 
 void
 virtio_set_iothread(struct virtio_base *base,
-			  bool is_register)
+			  bool is_register, bool ioevent_poll)
 {
 	struct virtio_vq_info *vq;
 	struct virtio_ops *vops;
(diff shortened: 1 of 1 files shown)


chejianj (approved):
LGTM

yuhuanX:
start to run premerge test

yuhuanX:
start to run premerge test

wistal:
This patch seems have impact/overhead on virtio-input, without this patch, virtio input works normally. with this patch, the virtio-input will got the virtio_input_notify_event_vq and virtio_input_notify_status_vq polled by the polling timer.
without this patch the above two callback function will not be polled by timer

to verify this issue, set dm  console=5 and enable the following debug code.
--- a/devicemodel/hw/pci/virtio/virtio_input.c
+++ b/devicemodel/hw/pci/virtio/virtio_input.c
@@ -24,7 +24,7 @@
 #include <sys/stat.h>
 #include <sys/eventfd.h>
 
-static int virtio_input_debug;
+static int virtio_input_debug=1;
 #define DPRINTF(params) do { if (virtio_input_debug) pr_dbg params; } while (0)
 #define WPRINTF(params) (pr_err params)
 
@@ -256,7 +256,7 @@ virtio_input_notify_status_vq(void *vdev, struct virtio_vq_info *vq)
                WPRINTF(("virtio_input: input_is_fifo or input_is_eventfd, doesnot support notify host\n"));
                //just warning, do nothing
        }
-
+       DPRINTF(("%s\n", __func__));
        while (vq_has_descs(vq)) {
                n = vq_getchain(vq, &idx, &iov, 1, NULL);
                if (n < 0) {

we can see the following log printed very quickly
virtio_input_notify_status_vq
virtio_input_notify_event_vq
...
virtio_input_notify_status_vq
virtio_input_notify_event_vq

PR #8780, opened 2025-08-19
misc: fail to save config_summary.rst while changing board.xml

6ZhangWei:
This patch is to fix the bug which will report while changing board.xml.

The issue was that all data structures were class variables, Data will be accumulated when calling
main(board_xml, scenario_xml, config_summary).

This patch moved all shared data to instance variables in __init__(). self.io_port = {}
self.io_description = []
self.pci_vuart = {}
self.pci_ivshmem = {}
self.amount_l3_cache = {}
self.service_vm_used_pcpu_list = []

Tracked-On: #8718

### DIFF
--- a/misc/config_tools/scenario_config/config_summary.py
+++ b/misc/config_tools/scenario_config/config_summary.py
@@ -91,19 +91,20 @@ class Doc:
 
 
 class GenerateRst:
-    io_port = {}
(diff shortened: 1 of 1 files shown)


6ZhangWei:
redo the PR 8720

Issue #8782, opened 2025-09-02
Rework Makefile to support multiple architecture and add RISC-V hello world skeleton


yfliuuu:
Rework Makefile to support multiple architecture and add RISC-V hello world skeleton

yfliuuu:
[External_System_ID] ACRN-13070

PR #8789, opened 2025-09-08
[multi-arch-dev] doc: Add initial multi-arch-coding-guideline

yfliuuu:
Tracked-On: #8782

### DIFF
--- a/multi-arch-coding-guideline.rst
+++ b/multi-arch-coding-guideline.rst
@@ -1,0 +1,341 @@
+Multi-architecture development coding guideline
+###############################################
+
+This document is intended for guiding initial development of
+ACRN multi-architecture.
+
+This document is NOT a coding style guideline (where one normally expects
+that the document explains things such as indentation, naming convention,
+typedef rules, etc.)
+
+During early development of multi-architecture project, the APIs
+and data structures should be organized in a consistent approach for
+better readability and maintainability. There are multiple ways to organize
+these factors and this documentation lays out a guideline to assist future
+development.
+
+General Assumptions and Practices
+*********************************
+
+Due to FuSa constraints, we have several compiler features or coding styles that
+are NOT encouraged in ``hypervisor/`` code:
+
+* Weak link (``__weak``): There are several MISRA rules that discourage uses of weak
+  link (Rule 1.2, 5.3, 5.8, 5.9). See `coding-guidelines`_ for more information.
+
+* Function-like macro: Use inline function whenever possible. Function-like macros
+  should be avoided unless absolutely necessary. This also implies that section-based
+  variable placement is also discouraged (utilizes linker to group symbols in certain
+  binary sections. See ``devicemodel/include/types.h``, macro ``SET_*``).
+
+Practices and general rule of thumb:
+
+* Enable compiler ``-O2`` as default.
+
+* APIs that are public (globally visible) but implemented in arch domains should start
+  their name with ``arch_``. If this API is optional, a default/empty ``arch_`` API
+  can be implemented in common C/header file.
+
+* Architecture headers should be exposure-controlled: If a structure or an API is used/shared
+  only within the module, then do not expose it to public headers, put it to private headers.
+  There is also a reverse statement: all arch symbols exposed to common scope should be public.
+  This is nearly impossible in practice. So we follow only the former part, and leave the latter
+  as future optimization.
+
+.. _coding-guidelines: https://projectacrn.github.io/latest/developer-guides/coding_guidelines.html
+
+Data Structures
+***************
+
+In multi-architecture project we might have a common data structure that needs
+to carry some architecture specific information (for example, a vCPU object).
+
+For this type of data structures we prefer the following style:
+
+In common header:
+
+.. code-block:: c
+
+   /* File: common.h */
+   #include <arch-public.h>
+
+   struct foo {
+       <common member1>;
+       <common member2>;
+       <common member3>;
+       struct arch_foo arch;
+   };
+
+   struct bar {
+       <common member1>;
+       <common member2>;
+       <common member3>;
+   }
+
+And in architecture specific header:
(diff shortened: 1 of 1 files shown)


multi-arch-coding-guideline.rst
```diff
@@ -0,0 +1,314 @@
...
+
+   /* Forward declaration when the reference of struct is opaque */
+   struct foo;
+   struct bar;
+
+   struct foo_arch {
+       <arch member1>;
+       <arch member2>;
+       <arch member3>;
+       struct foo *opaque_foo;
```
ywan170: Can we leverage container_of to avoid this pointer?

multi-arch-coding-guideline.rst
```diff
@@ -0,0 +1,314 @@
...
+
+And in architecture specific header:
+
+.. code-block:: c
+
+   /* File: arch-public.h */
+
+   /* Forward declaration when the reference of struct is opaque */
+   struct foo;
+   struct bar;
```
ywan170: I haven't got the point here, why can not include the common.h directly?

multi-arch-coding-guideline.rst
```diff
@@ -0,0 +1,314 @@
...
+   #endif
+
+   /*
+    * Architecture MAY implement this API.
+    * If it does not, a common default is provided in common C file.
+    * Due to FuSa constraint described above, __weak is discouraged.
+    * Define __ARCH_HAS_OPTIONAL_WITH_DEFAULT in arch header to
+    * indicate implementation instead.
+    */
+   void arch_api_optional_with_default(void);
```
ywan170, line 185: The arch_ is a prefix for arch depended implementation. If there's a default generic implementation that should be without arch_ prefix. You can refer to the patch I wrote for Haoyu.

multi-arch-coding-guideline.rst
```diff
@@ -0,0 +1,314 @@
...
+Here we consider only common-arch interfaces. Interfaces within each architecture
+are out of scope of this section.
+
+There are several form of interfaces used in different use cases.
+
+Function prototypes
+===================
+
+This is the most common form of interface where a function is referenced in common scope
+and implemented in arch-specific scope.
```
ywan170, line 132: We need to empower the principle is that arch/x86/* should only implement common API with arch_ prefix, never implement a common API directly. If a common API doesn't have any generic code logic, then wrap the arch_xxx  as a simple function in common.h like your helper case.

multi-arch-coding-guideline.rst
yfliuuu: The example here was not very appropriate. I'll add another.

Yes we can use container_of when we tried to get outer struct with inner struct pointer.
My original point is that, if a struct is referenced as opaque pointer, we need a forward declaration.

multi-arch-coding-guideline.rst
yfliuuu: Common header includes arch-public header first, which means the arch-public content expands before common header. If arch-public again includes common, that part simply becomes empty because of the outer header guard from common header.

If arch-public header needs to reference contents in common (opaque reference), a forward declaration is needed.

A simple example is, a common vcpu header defines struct vcpu, and arch-public vcpu header has API that needs to take struct vcpu * as input.

However if arch header needs to explicitly reference members inside struct vcpu (maybe through inline function), then this creates circular dependency. There are usually two solutions: 1, do not inline this function and put implementation to C file, 2, that function is NOT an arch API and it should not be placed in arch-public header.

multi-arch-coding-guideline.rst
yfliuuu: My understanding is: arch_ functions are functions that are supposed to be implemented by arch domain. However some arch may choose to do it, some arch don't. If they choose to do it, the implementation is in arch, and it is correct to start function with arch_.

Having a default in common domain doesn't mean that it is not an arch_ function.

KVM has the same logic. KVM has some arch_ default API implemented in common C file (for example kvm_arch_create_vm_debugfs).

multi-arch-coding-guideline.rst
yfliuuu: Sure, will emphasize this.

yuhuanX:
start to build

yuhuanX:
start to build

multi-arch-coding-guideline.rst
shiqingg, line 1: We'd better place `multi-arch-coding-guideline.rst` under `doc/developer-guide/` and update `coding_guidelines.rst` to include it, ensuring it is built as part of the documentation.

multi-arch-coding-guideline.rst
```diff
@@ -0,0 +1,341 @@
...
+development.
+
+General Assumptions and Practices
+*********************************
+
+Due to FuSa constraints, we have several compiler features or coding styles that
+are NOT encouraged in ``hypervisor/`` code:
+
+* Weak link (``__weak``): There are several MISRA rules that discourage uses of weak
+  link (Rule 1.2, 5.3, 5.8, 5.9). See `coding-guidelines`_ for more information.
```
shiqingg, line 24: `Rule 1.2, 5.3, 5.8, and 5.9` are MISRA-C rule IDs, while the references point to the ACRN coding guidelines.
To maintain consistency, we can replace these MISRA-C rule IDs with the corresponding ACRN coding guideline rule IDs: `Rule 1.2, 5.3, 5.8, 5.9` -> `ACRN Rule C-LE-*`

multi-arch-coding-guideline.rst
```diff
@@ -0,0 +1,341 @@
...
+       /* short implementation */
+   }
+
+In architecture-specific C file:
+
+.. code-block:: c
+
+   /* File: arch.c */
+   #include <common.h>
+   #include <arch-priate.h>
```
shiqingg, line 243: priate -> private?

Issue #8792, opened 2025-09-12
hv: timer: support multi-arch and add RISC-V timer


yisun-git:
[External_System_ID] ACRN-13077

PR #8787, opened 2025-09-05
hv: ipi/smpcall: add multi-arch framework and RISC-V support

shiqingg:
hv: ipi/smpcall: add multi-arch framework and RISC-V support

This patchset adds multi-arch support, focusing on IPI and SMP call features.
It:
  - introduces common SMP call frameworks
  - adapts x86 code to the new SMP call framework
  - implements the IPI and SMP call for RISC-V

Changelog:
v3 -> v4:
 * Renamed SBI_EXPERIMENTAL_x and SBI_VENDOR_x.

v2 -> v3:
 * Added description for sbi_ecall().
 * Renamed SBI-related enums, macros, and data structures to align with
   the SBI specification.
 * Updated commit message and code comments to state explicitly that
   legacy SBI extensions are not supported in ACRN.
 * Refined the prototype of sbi_send_ipi() to align with the SBI spec:
     From: int64_t sbi_send_ipi(uint64_t mask)
     To:   int64_t sbi_send_ipi(uint64_t mask, uint64_t mask_base)
   In ACRN it is invoked as sbi_send_ipi(dest_mask, 0UL), with mask_base
   set to 0UL.
 * Renamed send_single_ipi() and send_dest_ipi_mask() to
   arch_send_single_ipi() and arch_send_dest_ipi_mask() respectively.
 * Merged the following two patches into one:
     [RFC PATCH v2 4/7] hv: introduce common/smp.{c,h}
     [RFC PATCH v2 5/7] hv: smpcall: x86: adapt to common SMP call
 * Split per_cpu.h implementation into a separate commit.

v1 -> v2:
 * Dropped the ops way per discussion in
   https://lists.projectacrn.org/g/acrn-dev/topic/rfc_patch_3_6_hv_ipi/114877139

Tracked-On: #8786

PR #8793, opened 2025-09-12
hv: timer: support multi-arch and add RISC-V timer

yisun-git:
This patch-set adjusts common timer codes to support multi-arch by moving
arch specific codes from common to x86. Then, it adds RISC-V timer codes.

Change Log:
Formal v1 -> Formal v2
1. Remove hw_timer.h because set_hw_timeout/init_hw_timer are arch
   specific functions. So rename them to arch_set_timer_count/
   arch_init_timer and move the declaration into common timer.h.
2. Add "FIXME" in trap.c because the timer irq handling mechanism may
   need be changed to align with x86.

RFC v2 -> Formal v1
1. Rename set_hw_timeout() to arch_set_timer_count(). Then, Refine
   update_physical_timer() to call arch_set_timer_count() to set arch
   specific timer.
2. Rename init_hw_timer() to arch_init_hw_timer().
3. Add SBI timer interface on Shiqing's PR #8787.
4. Remove sstc.c/sstc.h and move the codes into risc-v timer.c.
5. Remove set_deadline() and call SBI or SSTC interface directly in
   arch_set_timer_count() to simplify the codes. Furthermore, remove the not
   necessary check for timeout value in original set_deadline().
6. Remove get_tick() and move its codes into arch_cpu_ticks() to reduce
   function layer.
7. Remove get_cpu_khz() and move its codes into arch_cpu_tickrate() to
   reduce function layer.
8. Add sexpt_handler() and modify flow of sint_handler(0 in trap.c to better
   show the work flow.

RFC v1 -> RFC v2:
1. Refine change about update_physical_timer() to abstract arch specific
   codes out as arch_set_timer_count().
2. Rename cpu_ticks/cpu_tickrate to arch_cpu_ticks/arch_cpu_tickrate to
   follow naming convention. Furthermore, add cpu_ticks() in common
   timer.c to facilitate external callers.
3. Refine codes in sbi.c to fix review comments in https://lists.projectacrn.org/g/acrn-dev/topic/rfc_2_2_risc_v_initial/115003295
4. Split SSTC codes out from sbi.c to sstc.c.
5. In risc-v timer.c and timer.h, Remove preinit_timer(), cpu_khz, boot_count
   and get_cycles() which are not used. Remove unused header file including.
   And use MIN_TIMER_PERIOD_US which is declared in common timer.h.
6. Refine codes in trap.c to fix review comments in https://lists.projectacrn.org/g/acrn-dev/topic/rfc_2_2_risc_v_initial/1150032

Tracked-On: #8792


yisun-git:
@ywan170, please help to review. Thanks!

hypervisor/include/arch/riscv/asm/timer.h
```diff
@@ -0,0 +1,17 @@
...
+ * Authors:
+ *   Haicheng Li <haicheng.li@intel.com>
+ */
+
+#ifndef __RISCV_TIMER_H__
+#define __RISCV_TIMER_H__
+
+#include <timer.h>
+
+void timer_irq_handler(void);
```
ywan170, line 20: Add a FIXME too, this should be registered through request_irq instead of an arch private API

ywan170 (approved):

yisun-git:
Updated hv-timer.rst to remove cpu_tickrate which has been replaced by arch specific interface arch_cpu_tickrate. Otherwise, build doc fails. Please review again. Thanks!

shiqingg (approved):
Fine with me on the doc changes.

PR #8808, opened 2025-09-22
multi-arch add risc-v io interface

haoyu-tang:
MMIO read/write without memory order should be common. 
ARCH without PIO support shouldn't use PIO APIs, so implement them as empty. 
Add RISC-V MMIO APIs with memory order.

### DIFF
--- a/hypervisor/include/common/io.h
+++ b/hypervisor/include/common/io.h
@@ -1,0 +1,261 @@
+/*
+ * Copyright (C) 2025 Intel Corporation.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef IO_H
+#define IO_H
+
+#include <types.h>
+#include <asm/io.h>
+
+#ifndef HAS_ARCH_PIO
+/* Write 1 byte to specified I/O port */
+static inline void arch_pio_write8(__unused uint8_t value, __unused uint16_t port){}
+
+/* Read 1 byte from specified I/O port */
+static inline uint8_t arch_pio_read8(__unused uint16_t port) { return 0xffU;}
+
+/* Write 2 bytes to specified I/O port */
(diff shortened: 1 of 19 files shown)


lifeix:
@ywan170  @chejianj Please help to review

yfliuuu:
Any reviewers?

hypervisor/include/common/io.h
```diff
@@ -0,0 +1,261 @@
...
+
+/* Read 4 bytes to specified I/O port */
+static inline uint32_t arch_pio_read32(__unused uint16_t port) { return 0xffffffffU;}
+
+/* Write 8 bytes to specified I/O port */
+static inline void arch_pio_write(__unused uint64_t v, __unused uint16_t addr, __unused size_t sz) {}
+
+/* Read 8 bytes to specified I/O port */
+static inline uint64_t arch_pio_read(__unused uint16_t addr, __unused size_t sz) { return 0xffffffffU;}
+#endif /* HAS_ARCH_PIO */
```
ywan170, line 37: PIO is X86 specific interface, they should be not common.

ywan170:
> Any reviewers?

Just give a comment.

hypervisor/include/common/io.h
lifeix: Yes, only x86 has PIO. On x86 platform, a uart may use PIO BAR; on risc-v platform, a uart may use MMIO BAR. I suppose we would implement the uart as common. So for risc-v, this PIO APIs would not used, so just implement them empty here.
But there may have execption: in some arm/risc-v SOC, there still have some PCI devices which has PIO. In this case, may need to implement risc-v arch special PIO APIs to access PIO BAR.

ywan170: OK to me

ywan170 (approved):

Issue #8816, opened 2025-09-28
hv: riscv: Initialize timer when boot


yisun-git:
[External_System_ID] ACRN-13090

Call timer_init() to initialize timer when boot and moves timer.c and softirq.c out from x86 to make them be built for risc-v too.

PR #8806, opened 2025-09-22
refactor acrnlog and acrntrace to common

lifeix:
This patch sereis also decocuple hvlog, console and npk_log.

Jiaqing Zhao (3):
  hv: make acrntrace a common feature
  hv: move sbuf_share_setup() and sbuf_reset() to common
  hv: enable acrnlog in release builds

Tracked-On: #8805
    Signed-off-by: Jiaqing Zhao <jiaqing.zhao@linux.intel.com>
    Reviewed-by: Yifan Liu  <yifan1.liu@intel.com>

PR #8817, opened 2025-09-28
hv: riscv: Initialize timer when boot

yisun-git:
This patch calls timer_init() to initialize timer when boot and moves timer.c and softirq.c out from x86 to make them be built for risc-v too.

To pass the compilation, it has some dependencies.
1. In softirq.c, it calls CPU_IRQ_DISABLE_ON_CONFIG/CPU_IRQ_ENABLE_ON_CONFIG which will be implemented by https://lists.projectacrn.org/g/acrn-dev/topic/patch_v2_3_5_hv_riscv/115408231
2. In timer.c, it calls TRACE_2L which will be implemented by https://github.com/projectacrn/acrn-hypervisor/pull/8806
3. In timer.c, it calls ASSERT()->asm_assert() which has not been available for debug build.
4. In local_irq_save(), "*(uint32_t *)(x) = val" cannot pass compilation. Shiqing will fix it in https://lists.projectacrn.org/g/acrn-dev/topic/patch_v2_3_5_hv_riscv/115408231

Tracked-On: #8816

### DIFF
--- a/hypervisor/Makefile
+++ b/hypervisor/Makefile
@@ -155,6 +155,8 @@ COMMON_C_SRCS += lib/memory.c
 COMMON_C_SRCS += lib/bits.c
 COMMON_C_SRCS += common/ticks.c
 COMMON_C_SRCS += common/delay.c
+COMMON_C_SRCS += common/timer.c
+COMMON_C_SRCS += common/softirq.c
 
 # library componment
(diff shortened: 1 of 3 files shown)


yuhuanX:
start to build

ywan170 (approved):

Issue #8831, opened 2025-10-20
memory multiarch abstraction and riscv support


liuhang-bit:
[External_System_ID] ACRN-13101

PR #8832, opened 2025-10-20
 memory multiarch abstraction and riscv support

liuhang-bit:
This patch-set seperate common memory related data structure and interfaces from
x86 and add riscv memory code, details as follows:

1. struct pgtable has been modified and put into the common code, callbacks tweak_exe_right/recover_exe_right has been removed because of x86 specific implementation, and clflush_page_walk is changed to flush_cache_pagewalk, new callback set_pgentry has been added to wrap per memory translation page table levels arch specific implementation, and get_default_right callback has been removed and the logic has been integrated into set_pgentry.
2. The common memory related data structure also includes structure page_pool, the related interfaces includes init_page_pool/alloc_page/free_page
3. pte sanitize operation related function has become common code, since x86 need to sanitize pte with specific hpa, since the operation is duplicated with memset when alloc_page for other arch, so replace it with sanitize_pte.
4. the common function and variables naming style has been changed to pgtln, which is independent with any arch, to reserve arch specific naming , introduce mm_common.h for each arch to map common to arch.
5. move some functions such as hpa2hva to common.
6. riscv only need to map hv runtime memory and MMIO address space, donot map guest memory
7. provide common interfaces pre_user_access() and post_user_access() to allow access guest related address, if SMAP is available, by default, PAGE_ATTR_IO/PAGE_ATTR_PMA attributes are not used, since they need detection for svpbmt extension. 

Tracked-On: #8831
Signed-off-by: hangliu1 <hang1.liu@intel.com>


ywan170 (approved):
Please fix all my new comments based on v9.

yuhuanX:
start to build

shiqingg (approved):

Issue #8805, opened 2025-09-22
Enable Console for risc-v


lifeix:

lifeix:
[External_System_ID] ACRN-13096

PR #8835, opened 2025-10-23
dump: dump hv image delta for risc-v

lifeix:
dump hypervisor image load address delta for risc-v.

Note: using the riscv64-linux-gnu-addr2line to parse the `Host Call Trace` may not accurate when the complier using `O2` to build the hypervisor. In this case, it'd better to use riscv64-linux-gnu-objdump to `disassemble -D acrn.out` to parse the `Host Call Trace` manually to change the complier to use `O1` to build the hypervisor.

Tracked-On: #8805

Reviewed-by: Jian Jun Chen <jian.jun.chen@intel.com>

### DIFF
--- a/hypervisor/debug/riscv/dump.c
+++ b/hypervisor/debug/riscv/dump.c
@@ -5,6 +5,7 @@
  */
 
 #include <types.h>
+#include <reloc.h>
 #include <asm/irq.h>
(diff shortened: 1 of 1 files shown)


lifeix:
@ywan170  Please help to review

yuhuanX:
start to build

PR #8837, opened 2025-10-30
vcpu/vm/vm_config multi-architecture

yfliuuu:
This patch series focus on multi-architecture of ACRN vcpu/vm/vm_config.

This series is part 1 of previous Service-VM/Pre-launched VM enabling.
More parts will follow in separate series.

Note that this patch series only guarantee that each commit compiles for
x86 architecture. It does not guarantee compilation for RISC-V.

Due to the complexity of the dependency relationship, making each commit
compiles for RISC-V will result in basically every change to be put inside
an extremely large commit, making it nearly un-reviewable.


yfliuuu:
@ywan170 Pls help approve.
@shiqingg Pls help review the last doc commit.

ywan170 (approved):

shiqingg (approved):

jsun26intel (approved):

Issue #8849, opened 2025-11-25
No IP in SOS hybrid mode on RPL NUC I5


yisun-git:
How Found

Function test

 

HW/Board

RPL NUC I5, cpu: i5-13500H

 

Build link

build acrn from [https://github.com/projectacrn/acrn-hypervisor.git]

commit: bc90c7bc418f52682b3e753356beb6fbbbf5b602  branch: master

build kernel from [https://github.com/projectacrn/acrn-kernel.git]

commit: a3634b1a933e80ff672cc47bd32d2f036cfbe755   branch: 6.12/linux

 

Steps
 # install ubuntu 24.04 on RPL NUC I5
 # build acrn in hybrid mode:
configure1(Hybrid Mode)
| Component | CPU | Memory | Graphic | USB input | USB disk | Ethernet(pcie) | storage(pcie) | vUART | Inter-UART |
|--------------------|-------|---------|---------------------|-------------|----------|-----------------|-------------------|-------|--------------------------------|
| SOS | 7P+8E | 15G | Headless | | | | | Y | SOS<-> UOS, SOS<->Tee |
| UOS(UAAG) | 4P | 16384M | virtio-gpu SRIOV | Mdediator/pT| PT | virtio-net | nvme-partition PT | Y | SOS<-> UOS |
| Pre-launchVM (TEE) | 1 | 256M | / | / | / | / | / | Y | SOS<->TEE |
 # install acrn/kernel following [https://projectacrn.github.io/latest/getting-started/getting-started.html] on RPL NUC I5
 # launch SOS

 

Expected result

1. sos got ip successfully

 

Actual result

1. sos failed to got ip:

root@acrn-MINIPC-PN64-E1:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host noprefixroute
       valid_lft forever preferred_lft forever
2: sit0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN group default qlen 1000
    link/sit 0.0.0.0 brd 0.0.0.0
[...]

PR #8850, opened 2025-12-01
hv: pci: refine codes to make sos can manage bridge

yisun-git:
In previous codes, the bridge is owned by hypervsior if there is pre-launched VM configured. This can cause bug, e.g. the net card behind the bridge cannot get IP because the write to bridge's command register to set BusMaster bit is ignored.

This patch is to refine the codes to make the bridge be managed by sos if the device behind the bridge is not assigned to pre- launched VM. Furthermore, it adds the parent device into struct pci_pdev to construct PCI hierarchy.

Tracked-On: #8849

### DIFF
--- a/hypervisor/hw/pci.c
+++ b/hypervisor/hw/pci.c
@@ -384,6 +384,7 @@ struct pci_bdf_mapping_group {
 struct pci_bus_num_to_drhd_index_mapping {
 	uint8_t bus_under_scan;
 	uint32_t bus_drhd_index;
+	struct pci_pdev *parent_pdev;
 };
 
 static uint32_t pci_check_override_drhd_index(union pci_bdf pbdf,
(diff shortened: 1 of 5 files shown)


lifeix (approved):

mz-intel:
start to run premerge test

szhen11:
start to run premerge test

szhen11:
start to run premerge test

hypervisor/hw/pci.c
```diff
@@ -864,17 +869,17 @@ struct pci_pdev *pci_init_pdev(union pci_bdf bdf, uint32_t drhd_index)
...
+			/* PCI debug uart and RP for EP which has been assigned to pre-launched VM are owned by HV */
+			if (is_pci_dbg_uart(bdf)) {
 				hv_owned_pci_pdevs[num_hv_owned_pci_pdev] = pdev;
 				num_hv_owned_pci_pdev++;
 			}
+
+			if (allocate_to_prelaunched_vm(pdev) && parent) {
+				hv_owned_pci_pdevs[num_hv_owned_pci_pdev] = parent;
+				num_hv_owned_pci_pdev++;
+			}
```
chejianj, line 881: If we passed through 2 pci devices under the same bridge to pre-launched VM, the parent bridge will be added to the hv_owned_pci_pdevs twice, although it is harmless in the current code since hv_owned_pci_pdevs is only used to check whether a device is owned by hypervisor.

another question is that we do not need to consider parent bridge of the bridge?

